PR #36, opened 2026-04-01
Fix image field removal not persisting after save

scottbuscemi:
## Summary

- Change image field removal to send `null` instead of `undefined` so the value survives `JSON.stringify` and reaches the server
- `undefined` values are stripped by `JSON.stringify`, so the server never saw the field in the update payload and kept the old value

## Test plan
- [ ] Add a featured image to a post/page and save
- [ ] Remove the featured image and save
- [ ] Refresh the page — the image should remain removed
- [ ] Re-add an image — should still work as before


travisbreaks (commented):
The undefined-to-null fix is correct for JSON serialization. A few things:

1. **Scope check**: Are there other field types beyond images that could have the same undefined vs null serialization issue? If custom fields (file uploads, relations, etc.) follow the same pattern, this fix might need to be broader.

2. **Server-side interpretation**: Does the server handler correctly distinguish `null` (clear the field) from a missing key (no change)? If the server treats both as "no change," then sending `null` won't actually clear the image. Worth verifying the PATCH/PUT handler checks for explicit `null`.

3. **Type safety**: A type guard or utility like `valueOrNull(field)` at the serialization boundary could prevent recurrence for future field types.

Small fix, correct direction.

BenjaminPrice (changes requested):
Solid fix for image field removal persistence; one visual bug noted during testing.

### Tested scenarios

- Removed a featured image and saved; confirmed the removal persisted correctly after save and page reload.
- Verified the JSON payload in the network tab contains the expected data when saving after image removal.

### Bugs

- After removing an image and letting auto-save persist the removal, adding a new image causes the screen to refresh on auto-save, and the newly added image disappears from the UI. The image is actually saved correctly (a manual page reload shows it), so this is a visual/rendering bug rather than a data-loss issue.

ascorbic (approved):
Thanks!

PR #51, opened 2026-04-02
Add note about re-linking CLI bins after build

thejackshelton:
After `pnpm build`, the emdash CLI bin needs a second pnpm install to be linked. Added a note to `CONTRIBUTING.md` so new contributors don't hit command not found when running `emdash seed`.


BenjaminPrice:
This becomes unnecessary after #75 merges

PR #57, opened 2026-04-02
Fix: close "Create byline" dialog when Cancel is clicked

jakaria-istauk:
## Problem

Clicking the **Cancel** button in the "Quick create byline" popup did not close the dialog.

## Root Cause

The `Dialog.Close` component passes its close handler via the render prop spread (`{...p}`). In the Cancel button, `onClick={resetQuickCreate}` was placed **after** `{...p}`, silently overriding the close handler:

```tsx
// Before — onClick overrides p.onClick (the close handler)
<Button {...p} variant="secondary" onClick={resetQuickCreate}>
```
##Fix
Compose both handlers so the form state is reset and the dialog closes:

```tsx
// After — both handlers run
onClick={(e) => {
    resetQuickCreate();
    p.onClick?.(e);
}}
```
## Testing
1. Open a content item in the editor
2. Scroll to the Bylines section
3. Click Quick create byline
4. Click Cancel — dialog should close immediately


ascorbic (approved):
Thanks

PR #58, opened 2026-04-02
Default ownership to current user when no author is assigned

jakaria-istauk:
### Problem
 
When editing a content item with no author assigned, the Ownership selector showed **"Unassigned"** instead of defaulting to the currently logged-in user. This meant editors had to manually set themselves as author every time.
 
---
 
### Changes
 
- When a content item has no `authorId`, the Ownership selector now displays the current user as the default selection.
- On load, `onAuthorChange` is fired once with the current user's ID to persist the default to the backend — no manual interaction required.
 
---
 
### Behaviour
 
| Scenario | Before | After |
|---|---|---|
| Item has no author | Shows "Unassigned" | Shows current user |
| Item has an author | Shows that author | Unchanged |
 
---
 
### Testing
 
1. Open a content item that has no author assigned.
2. Check the Ownership section in the sidebar — it should show the current logged-in user.
3. Save the item — the author should be persisted without needing to manually select it.


jakaria-istauk:

ascorbic:
This would assign every unassigned page to you just by viewing it. That's not expected behaviour. It's good to default *new* entries to the current user, but for existing pages that's not what we want.

PR #72, opened 2026-04-02
fix(cli): unwrap API response envelope in device flow login

### LINKED ISSUE #54: CLI `emdash login` fails against remote instances
The CLI `emdash login` command fails against remote EmDash instances due to two distinct issues.

### Issue 1: Response envelope mismatch (device flow)

The server's OAuth device flow endpoints wrap responses in `{"data": {...}}`, but the CLI expects the fields at the top level.

**Affected endpoints:**
- `POST /_emdash/api/oauth/device/code` — returns `{"data": {"device_code": ..., "verification_uri": ..., "user_code": ...}}`
- `POST /_emdash/api/oauth/device/token` — returns `{"data": {"access_token": ..., "expires_in": ...}}`

**Symptoms:**
- `verification_uri` and `user_code` display as `undefined` during login
- After authorizing in the browser, login fails with `ERROR Invalid time value` (because `expires_in` is `undefined`)

These were verified independently — patching only the device code response allowed login to proceed to the token exchange, which then failed with the same wrapping issue.

**Fix:** Unwrap the `.data` property when parsing these responses in the CLI

---

### Issue 2: Device flow token lacks admin scope

Separately from the envelope issue, the CLI's device code request does not include a `scope` parameter:

```json
{"client_id": "emdash-cli"}
```

The resulting token is issued without admin scope, so all subsequent CLI commands fail with `INSUFFICIENT_SCOPE` / `Token lacks required scope: admin` — even when the authorizing user is an admin.

This was verified by patching issue 1, completing a successful login, then running `emdash whoami` which returned a 403.

**Fix:** Adding `"scope": "admin"` to the device code request resolved this for me.

--- 

### Steps to reproduce

1. Deploy an EmDash site to Cloudflare (with or without Cloudflare Access)
2. Run `npx emdash login --url https://<your-site>`
3. Observe `undefined` for the verification URL and user code (issue 1)
4. If issue 1 is patched locally, complete the browser authorization
5. Observe `ERROR Invalid time value` (issue 1, token endpoint)
6. If both envelope issues are patched, complete login and run `npx emdash whoami`
7. Observe `Failed to fetch user info: 403` / `INSUFFICIENT_SCOPE` (issue 2)

### Environment

- EmDash v0.1.0 (both server and CLI)
- Deployed to Cloudflare Workers with D1
- Tested with and without Cloudflare Access (Access auth itself works correctly; the issues are in the subsequent EmDash device flow)

### Workaround

I patched`node_modules/emdash/dist/cli/index.mjs` with the unwrap fix and scope addition described above. Patches need to be re-applied after any `npm install`.

travisbreaks:
## Summary
Fixes #54. Two bugs in `emdash login` against remote (Cloudflare-deployed) instances:

- **Response envelope mismatch**: Server-side API wraps successful responses in `{ data: { ... } }` via `apiSuccess()`, but the CLI casts the raw response body directly as `DeviceCodeResponse` / `TokenResponse`. All fields resolve to `undefined`.
- **Missing scope**: Device code request omits `scope: "admin"`, causing issued tokens to have no scopes. Subsequent CLI commands fail with `INSUFFICIENT_SCOPE`.

## Changes
- Unwrap `.data` from device code response
- Unwrap `.data` from token polling success response (error responses are already flat, no change needed)
- Add `scope: "admin"` to device code request body

## Test plan
- [ ] `emdash login` against a remote Cloudflare-deployed instance completes successfully
- [ ] Token includes admin scope (or clamped scope for non-admin users)
- [ ] `emdash whoami` succeeds after login

Co-Authored-By: Claude Opus 4.6 (1M context) <tadao@travisfixes.com>


travisbreaks:

BenjaminPrice:
@travisbreaks I tested on DashHost.

As the creator of DashHost, I can attest that there's no middleware or anything that strips away anything from the request/response. The infrastructure deploys via wrangler using Workers for Platforms for each tenant.

ascorbic (commented):
Thanks. It looks like the whoami command has the same issue, so that could be fixed ina  followup.

PR #81, opened 2026-04-02
fix: repair invite flow — URL prefix + registration UI

### LINKED ISSUE #67: User invite system: wrong URL (missing /_emdash/ prefix) + no UI to complete registration
## Environment

- EmDash version: `0.0.3`
- `@emdash-cms/auth` version: `0.0.1`
- Adapter: Node.js + SQLite (Hetzner VPS)

---

## Bug 1: Invite URL missing `/_emdash/` prefix → 404

**File:** `packages/auth/src/invite.ts` (approx. line 65)

```ts
const url = new URL("/api/auth/invite/accept", config.baseUrl);
```

`config.baseUrl` is passed in as `https://example.com/_emdash` (from `getSiteBaseUrl()`). When constructing `new URL("/api/auth/invite/accept", "https://example.com/_emdash")`, the leading slash on the path is treated as an absolute path, which **replaces** the entire path of the base URL — discarding `/_emdash`.

**Resulting URL (wrong):**
```
https://example.com/api/auth/invite/accept?token=...
```

**Expected URL:**
```
https://example.com/_emdash/api/auth/invite/accept?token=...
```

Since the route is registered at `/_emdash/api/auth/invite/accept`, clicking the invite link returns a **404**.

**Suggested fix** — use a relative path or string concatenation instead:
```ts
const url = new URL(`${config.baseUrl}/api/auth/invite/accept`);
```

---

## Bug 2: `GET /invite/accept` returns raw JSON — no UI to complete registration

**File:** `src/astro/routes/api/auth/invite/accept.ts`

When the invitee visits the invite link (after fixing Bug 1), the `GET` handler validates the token and returns raw JSON:

```json
{"success":true,"email":"user@example.com","role":30,"roleName":"AUTHOR"}
```

There is no redirect to an admin UI page where the user can complete registration. The `POST /invite/complete` endpoint expects a WebAuthn passkey credential — but there is no frontend page to drive that flow.

**Result:** The user is never actually created. Checking the Users section in the admin dashboard confirms no new user appears despite the `success: true` response. The invite flow is effectively non-functional end-to-end.

**Expected behavior:** `GET /invite/accept` should redirect to an admin UI page (e.g. `/_emdash/admin/invite/accept?token=...`) that renders a form for the user to complete registration (set name, register passkey or set password).

---

## Summary

The invite flow has two compounding bugs:
1. The link given to admins to share points to a 404
2. Even after manually correcting the URL, visiting it returns JSON with no UI — and no user is actually created in the database

Both issues make the user invite system completely non-functional on the Node.js adapter.

SrikanthAlva:
## Summary

Fixes the two compounding bugs in the user invite system that made it completely non-functional (Issue #67):

- **Bug 1 — Invite URL drops `/_emdash/` prefix:** `new URL("/api/auth/invite/accept", baseUrl)` treats the leading `/` as an absolute path, discarding the `/_emdash` path from `baseUrl`. Fixed by using string concatenation and pointing the invite URL directly to the admin UI page.
- **Bug 2 — No UI to complete registration:** Added an `InviteAcceptPage` in the admin SPA that validates the invite token, displays the user's email and assigned role, and drives passkey registration via the existing `/auth/invite/complete` endpoint.

### Files changed

| File | Change |
|------|--------|
| `packages/auth/src/invite.ts` | Fix URL construction to preserve baseUrl path prefix; point invite URL to admin UI page |
| `packages/core/tests/unit/auth/invite.test.ts` | Add regression test for baseUrl with path prefix (`/_emdash`) |
| `packages/admin/src/components/InviteAcceptPage.tsx` | New invite acceptance page (modeled after `SignupPage.tsx`) |
| `packages/admin/src/lib/api/users.ts` | Add `validateInviteToken` and `completeInviteSignup` client functions |
| `packages/admin/src/lib/api/index.ts` | Re-export new functions and type |
| `packages/admin/src/router.tsx` | Register `/invite/accept` route |

## Test plan

- [x] New unit test `should preserve baseUrl path prefix in invite URL` passes
- [x] All 17 invite unit tests pass
- [x] All 2096 core package tests pass
- [x] All 41 auth package tests pass
- [x] `pnpm typecheck` passes (all packages)
- [x] `pnpm --silent lint:quick` returns 0 diagnostics
- [x] `pnpm format` clean
- [x] Manual: create an invite in the admin UI → verify the generated URL includes `/_emdash/admin/invite/accept`
- [x] Manual: visit the invite URL → verify the InviteAcceptPage loads with email/role displayed
- [x] Manual: complete passkey registration → verify user is created and redirected to admin dashboard

Closes #67

Made with [Cursor](https://cursor.com)


packages/core/src/astro/middleware/auth.ts
```diff
@@ -268,6 +267,10 @@ async function handleEmDashAuth(
 	const { emdash } = locals;
 
 	const isLoginRoute = url.pathname.startsWith("/_emdash/admin/login");
+	const isPublicAdminRoute =
+		isLoginRoute ||
```
BenjaminPrice: As `isLoginRoute` is no longer used anywhere but here, the above line can be consolidated into here and reduce the number of variables.

packages/admin/src/components/InviteAcceptPage.tsx
```diff
@@ -0,0 +1,243 @@
...
+			</div>
+		);
+	}
+
+	return (
+		<div className="min-h-screen flex items-center justify-center bg-kumo-base p-4">
+			<div className="w-full max-w-md">
+				{/* Header */}
+				<div className="text-center mb-8">
+					<div className="text-4xl font-bold mb-2">— EmDash</div>
```
BenjaminPrice: This is visually different from the Login page (see: https://github.com/emdash-cms/emdash/blob/5beb0ddc334deb862ba90cedbf03f052b58e4974/packages/admin/src/components/LoginPage.tsx#L272 ) and the Setup Wizard (see: https://github.com/emdash-cms/emdash/blob/5beb0ddc334deb862ba90cedbf03f052b58e4974/packages/admin/src/components/SetupWizard.tsx#L485 )

Could this structure be separated into a shared component?

BenjaminPrice (changes requested):

BenjaminPrice (commented):
### Bugs

- From the PR Description:
> (1) new URL("/api/auth/invite/accept", baseUrl) discarded the /_emdash path prefix because the leading / made it absolute

I'm unable to reproduce this locally. When I generate an invite locally it has the `/_emdash` path prefix.

### What I liked

- Generally looks good. I was able to reproduce the new registration page after accepting an invite. A few other small comments have been added.

BenjaminPrice:
Sorry, one more thing. This new Invitation flow could really benefit from some E2E tests being added.

packages/core/src/astro/middleware/auth.ts
SrikanthAlva: Updated this change

packages/admin/src/components/InviteAcceptPage.tsx
SrikanthAlva: Included the LogoLockup component in the InviteAcceptPage

SrikanthAlva:

BenjaminPrice (changes requested):
### Bugs

- The unit test `createInviteToken > should create a token and return url + email` fails because the generated invite URL now uses `/admin/invite/accept?token=` instead of the expected `/_emdash/api/auth/invite/accept?token=`. Either the test assertion or the URL construction needs to be updated so they agree on the correct path.

BenjaminPrice (changes requested):
See comment about failing test

jcheese1:
Need this in

ascorbic:
@SrikanthAlva it would be great if we could get this in. Are you able to get the tests passing?

ascorbic:
Superseded by #542

PR #82, opened 2026-04-02
Foundation + Instant Shell for IndexedDB Admin Caching

bhubbard:
**What this achieves:**

- On repeat visits, the admin shell renders instantly (sidebar, header, navigation) from cached manifest data
- The currentUser is also cached, so permission-dependent UI renders immediately
- Background fetches update the cache transparently
- If IndexedDB is unavailable (private browsing), everything degrades gracefully to standard useQuery

**Next phases (not yet implemented):**

- Phase 2: Cache content lists, media, users, bylines, taxonomies, menus, sections
- Phase 3: Optimistic mutations with conflict resolution
- Phase 4: Delta sync, sync status indicator, "Clear cache" button


ascorbic:
This is a useful addition, but can you rebase it on the latest main and use the PR template. If you've used an agent you can just ask it to do that.

PR #83, opened 2026-04-02
fix: allow hyphens in collection, field, and taxonomy slugs

### LINKED ISSUE #79: WordPress import crashes on collections with hyphens in slug (e.g. Elementor `elementor-hf`)
## Description

The WordPress import fails during the "prepare" step when the exported XML contains custom post types whose slugs include hyphens. EmDash's `SchemaRegistry.validateSlug` requires slugs to match `[a-z][a-z0-9_]*`, but WordPress post type slugs commonly use hyphens (e.g. `elementor-hf` from the Elementor page builder).

## Steps to reproduce

1. Export a WordPress site that uses **Elementor** (which registers the `elementor-hf` custom post type)
2. In the EmDash admin, start a WordPress import with the exported XML
3. The prepare step crashes immediately

## Error

```
SchemaError: collection slug must start with a letter and contain only lowercase letters, numbers, and underscores
    at SchemaRegistry.validateSlug (emdash/src/schema/registry.ts:705:10)
    at SchemaRegistry.createCollection (emdash/src/schema/registry.ts:121:8)
    at prepareImport (emdash/src/astro/routes/api/import/wordpress/prepare.ts:108:33)
```

## Expected behavior

The importer should either:
- Sanitize incoming WordPress post type slugs by replacing hyphens with underscores (e.g. `elementor-hf` → `elementor_hf`), or
- Skip non-content post types like `elementor-hf` that are plugin-internal, or
- Let the user choose which post types to import before validation runs

## Environment

- `emdash`: 0.0.3
- WordPress source site uses Elementor

antoineVIVIES:
Fixes #79

## Problem

WordPress plugins (Elementor, WooCommerce, ACF, etc.) register custom post types with hyphens in their slugs (e.g. `elementor-hf`, `shop-order`, `acf-field-group`). EmDash collection slugs previously required `[a-z][a-z0-9_]*`, so the import crashed with `SchemaError: INVALID_SLUG`.

## Fix

Instead of sanitizing hyphens away during import, allow hyphens natively in EmDash slugs. The slug validation pattern is updated from `/^[a-z][a-z0-9_]*$/` to `/^[a-z][a-z0-9_-]*$/` across all validation layers:

- **`database/validate.ts`** — SQL identifier validation
- **`schema/registry.ts`** — collection/field slug validation
- **`api/schemas/common.ts`** — shared Zod slug pattern
- **`api/schemas/schema.ts`** — collection/field API schemas
- **`api/schemas/taxonomies.ts`** — taxonomy API schemas
- **`api/handlers/taxonomies.ts`** — taxonomy name validation
- **`mcp/server.ts`** — MCP tool schemas
- **`seed/validate.ts`** — seed file validation
- **`admin/TaxonomyManager.tsx`** — admin UI validation + HTML pattern
- **`cloudflare/sandbox/bridge.ts`** — sandbox collection name validation
- **WordPress import `analyze.ts`** — `INVALID_SLUG_CHARS` updated to allow hyphens

## Test plan

- [x] Updated `wordpress-slug-sanitization.test.ts` — 15 tests pass, verifying hyphens are preserved
- [x] Lint clean (0 diagnostics)
- [x] Typecheck passes


antoineVIVIES:

antoineVIVIES:
Updated per @ascorbic's feedback — instead of sanitizing hyphens to underscores during WordPress import, EmDash now allows hyphens natively in collection, field, and taxonomy slugs.

**What changed:** slug validation updated from `[a-z][a-z0-9_]*` to `[a-z][a-z0-9_-]*` across all 10 validation points (schema registry, API schemas, MCP server, seed validation, SQL identifier validation, admin UI, Cloudflare sandbox). WordPress post types like `elementor-hf` and `shop-order` now pass through as-is.

antoineVIVIES:
@ascorbic sure Will go back to previous fix asap

packages/core/src/astro/routes/api/import/wordpress/analyze.ts
```diff
@@ -445,6 +448,16 @@ function isInternalMetaKey(key: string): boolean {
 	return false;
 }
 
+function sanitizeSlug(slug: string): string {
+	const sanitized = slug
+		.toLowerCase()
+		.replace(INVALID_SLUG_CHARS, "_")
+		.replace(LEADING_NON_ALPHA, "");
+	if (!sanitized) return "imported";
```
BenjaminPrice, line 464: The `INVALID_SLUG_CHARS` regex `/[^a-z0-9_]/g` is applied after `.toLowerCase()`, so it correctly handles uppercase. However, a slug consisting entirely of digits (e.g. 123) or underscores/hyphens produces "imported". If two different post types both degenerate to "imported", their content would merge into a single collection. This is an edge case but could cause data loss during import of unusual WP exports.

Perhaps we need to add some randomization to the slug, or we can get more sophisticated and have a map for all santized slugs and add a numerical incrementation.

PR #99, opened 2026-04-02
fix(admin): prevent dialogs from overflowing viewport on mobile

### LINKED ISSUE #92: Mobile UI modals don't fit
See https://x.com/fishcake2026/status/2039602523968041388

all3f0r1:
## Summary

Fixes #92

On mobile, admin dialogs overflow the viewport horizontally — content is clipped on the left side because `max-w-*` and `w-*` classes exceed the screen width.

- Add a global `max-width: calc(100vw - 2rem)` on `[role="dialog"]` in `styles.css` so all dialogs (current and future) stay within the viewport
- Fix `ContentPickerModal` which used a fixed `w-2xl` instead of `max-w-2xl`

## Test plan

- [ ] Open admin on a mobile viewport (≤ 400px)
- [ ] Trigger the Welcome modal → should fit within the screen
- [ ] Open a content picker dialog → should not overflow
- [ ] Open the media picker dialog → should not overflow
- [ ] On desktop, dialogs should render unchanged (their `max-w-*` classes are narrower than `100vw - 2rem`)


all3f0r1:
In styles.css, I wrapped the dialog override in a `@media (max-width: 639px)` query and added `min-width: 0 !important`. The previous max-width rule alone was ineffective because kumo's `size="lg"` sets `min-w-[32rem]` (512px), and CSS `min-width` beats `max-width` when they conflict. Now on mobile, `min-width` is reset so `max-width` can actually constrain the dialog. The media query also avoids affecting desktop where the min-width variants are appropriate.

Alongside, in `WelcomeModal.tsx`, I removed `size="lg"` — it was setting `min-w-[32rem]` (512px) on a dialog constrained to `sm:max-w-md` (448px), which makes no sense. The default "base" size (`sm:min-w-96` = 384px) is more appropriate for this modal.

@BenjaminPrice, can you give it another try, please?

BenjaminPrice (changes requested):
Content Modal width is correct now but the height doesn't scroll and the button to save is cut off

image

all3f0r1:
@BenjaminPrice Whoops indeed my test case didn't include that... My bad.

Can you give it another try, please?

Issue #110, opened 2026-04-02
@emdash-cms/x402: dist/middleware.mjs missing from published package


natsukingly:
## Bug

The `@emdash-cms/x402@0.1.0` package on npm is missing `dist/middleware.mjs` (and its `.d.mts` type file). The `package.json` exports map references these files:

```json
"./middleware": {
  "types": "./dist/middleware.d.mts",
  "default": "./dist/middleware.mjs"
}
```

But only `dist/index.mjs` and `dist/index.d.mts` are included in the published tarball.

## Impact

The Astro integration registers the middleware via `addMiddleware({ entrypoint: "@emdash-cms/x402/middleware" })`, which fails at runtime:

```
Error: Cannot find module '@emdash-cms/x402/middleware'
```

This makes the package **completely unusable** in any Astro project.

## Root Cause

The build tool (`tsdown`) has no config file specifying entry points. By default it only builds `src/index.ts`, so `src/middleware.ts` (and its dependency `src/enforcer.ts`) are never emitted to `dist/`.

## Fix

Add `src/middleware.ts` as an additional entry point in the tsdown build configuration so that `dist/middleware.mjs` and `dist/middleware.d.mts` are generated.

## Environment

- `@emdash-cms/x402`: 0.1.0
- `astro`: 6.1.3
- `@astrojs/cloudflare`: 13.1.7
- Platform: macOS (Cloudflare Workers dev via miniflare)

PR #112, opened 2026-04-02
fix(x402): add exports verification test to prevent missing dist artifacts

natsukingly:
## Summary

- Adds a test that verifies every path in `package.json` `"exports"` resolves to a file that exists on disk after build
- This test would have caught the missing `dist/middleware.mjs` that shipped in 0.1.0 (#110)

## Details

The test dynamically reads `package.json` exports and checks `existsSync()` for each declared path. If a new export is added but the build config doesn't include it as an entry point, this test will fail.

**Note:** This test depends on #48 (tsdown config) being merged first — without a build producing `dist/`, the test correctly fails.

## Test plan

- [ ] Merge #48 first, then run `pnpm build && pnpm test` in `packages/x402`
- [ ] Verify all 5 export paths pass


ascorbic:
Hey! If you sign the CLA we can get this in

PR #169, opened 2026-04-03
Enable page:metadata hooks for anonymous visitors on public pages

### LINKED ISSUE #166: page:metadata plugin hooks don't run for anonymous visitors on public pages
## Problem

The middleware (`src/astro/middleware.ts` lines 200–228) skips runtime initialization for anonymous visitors on public pages. This means `page:metadata` plugin hooks never fire for the audience that matters most — search engine crawlers and regular visitors.

`EmDashHead.astro` checks for the runtime via `getPageRuntime(Astro.locals)` and falls back to base-only SEO contributions when it's absent. Any plugin using the `page:metadata` hook to contribute meta tags, Open Graph, canonical URLs, robots directives, or JSON-LD schema markup only works for logged-in editors.

## Current behavior

For anonymous visitors on public pages:
```
middleware → early return (line 200-228) → locals.emdash not set
→ EmDashHead getPageRuntime() returns undefined → base contributions only
```

Plugin `page:metadata` hooks are never called. SEO plugins are effectively broken for the visitors that actually matter.

## Expected behavior

`page:metadata` hooks should run for **all** visitors so plugins can contribute SEO tags that search engines see.

## Why this is low-risk

The runtime is a **singleton** (`runtimeInstance` in middleware.ts). After the first request initializes it, `getRuntime()` is just a null-check. The per-request cost of calling `runtime.collectPageMetadata(page)` for anonymous visitors is negligible — it's just iterating registered hooks.

The early return at line 200 was likely an optimization to avoid session checks and full `locals.emdash` construction. But we don't need the full object — just the `collectPageMetadata` and `collectPageFragments` methods need to be available for `EmDashHead`.

## Suggested fix

After the early-return block for anonymous visitors, still initialize the runtime and attach the page contribution methods to locals:

```typescript
    const sessionUser = await context.session?.get("user");
        // ... existing setup check ...

        // Initialize runtime for page:metadata hooks (cached singleton, negligible cost)
        try {
            const config = getConfig();
            if (config) {
                const runtime = await getRuntime(config);
                locals.emdash = {
                    collectPageMetadata: runtime.collectPageMetadata.bind(runtime),
                    collectPageFragments: runtime.collectPageFragments.bind(runtime),
                };
            }
        } catch {
            // Non-fatal — EmDashHead falls back to base contributions
        }

        const response = await next();
        setBaselineSecurityHeaders(response);
        return response;
    }
}
```

This keeps the optimization of not constructing the full `locals.emdash` admin object for anonymous visitors, while still enabling plugin hooks for SEO.

## Use case

I'm building an SEO plugin (`page:metadata` hook) that generates:
- Proper `og:title` (without site name suffix)
- Meta robots with `max-snippet:-1, max-image-preview:large, max-video-preview:-1`
- Canonical URLs with trailing slash normalization
- Full JSON-LD schema graph (Person/Organization, WebSite, WebPage, Article)
- Configurable Person/Organization settings via `admin.settingsSchema`

All of this works perfectly for logged-in editors but is invisible to search engines.

jdevalk:
## Summary

- Initialize the runtime for anonymous visitors on public pages so `page:metadata` and `page:fragments` plugin hooks fire for all visitors
- Only attaches the two page contribution methods to `locals.emdash` (not the full admin object), preserving the existing performance optimization
- The runtime is a cached singleton — after first init, `getRuntime()` is just a null-check, so the per-request cost is negligible

## Problem

The middleware skips runtime initialization for anonymous visitors on public pages (lines 200–228). This means `EmDashHead`'s `getPageRuntime()` returns `undefined`, and it falls back to base-only SEO contributions. Plugin `page:metadata` hooks never fire for search engine crawlers or regular visitors.

## Fix

After the existing setup check, initialize the cached runtime and attach just `collectPageMetadata` and `collectPageFragments` to `locals.emdash`. This satisfies the structural check in `getPageRuntime()` so `EmDashHead` runs plugin hooks for all visitors.

## Test plan

- [ ] Verify anonymous visitors see plugin-contributed meta tags in `<head>`
- [ ] Verify logged-in editors still see the full admin toolbar and all existing functionality
- [ ] Verify performance is not impacted (runtime is cached singleton after first request)
- [ ] Verify `EmDashBodyStart`/`EmDashBodyEnd` work correctly with the minimal locals object

Fixes #166

### DIFF
--- a/packages/core/src/astro/middleware.ts
+++ b/packages/core/src/astro/middleware.ts
@@ -222,6 +222,24 @@ export const onRequest = defineMiddleware(async (context, next) => {
 				}
 			}
 
+			// Initialize the runtime for page:metadata and page:fragments hooks.
+			// The runtime is a cached singleton — after the first request, getRuntime()
+			// is just a null-check. This enables SEO plugins to contribute meta tags,
+			// canonical URLs, robots directives, and JSON-LD schema for all visitors,
+			// not just logged-in editors.
+			const config = getConfig();
+			if (config) {
+				try {
+					const runtime = await getRuntime(config);
+					locals.emdash = {
+						collectPageMetadata: runtime.collectPageMetadata.bind(runtime),
+						collectPageFragments: runtime.collectPageFragments.bind(runtime),
+					};
+				} catch {
+					// Non-fatal — EmDashHead will fall back to base SEO contributions
+				}
+			}
+
 			const response = await next();
 			setBaselineSecurityHeaders(response);
 			return response;


ascorbic:
This has already been fixed in #119. Thanks!

PR #183, opened 2026-04-03
fix: limit admin source aliasing to repo dev

masonjames:
## What does this PR do?

Fixes Astro dev so the raw `@emdash-cms/admin` source alias only applies when running an EmDash app inside this repo. External app installs now use the built admin package surface, while local monorepo/admin development keeps source aliasing for HMR.

The tests cover:

- local monorepo dev uses `packages/admin/src`
- external app dev uses `packages/admin/dist`
- sibling paths with matching prefixes do not count as inside this repo
- non-dev commands use `dist`

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Performance improvement
- [x] Tests

## Checklist

- [x] `pnpm typecheck` passes

## Screenshots / test output

- `pnpm --silent lint:quick`
- `pnpm --filter emdash exec vitest run tests/unit/astro/vite-config.test.ts`
- `pnpm typecheck`


ascorbic:
Can you explain a bit about why you're suggesting this change? I would rather just do a fix to ensure it doesn't apply when installed for end users.

masonjames:
Thanks — agreed. I was trying to keep the dev-only admin source alias from leaking into installed apps, but the first pass was too broad. The main use case is external Astro apps consuming EmDash — including linked/local installs — where the host app should see the same built admin package surface it would get from npm, not raw admin source.

I pushed a narrower version that keeps source aliasing for local repo development and uses the built admin package for external app installs. I also dropped the unrelated SSR externalization and `admin.astro` import change from this PR.

ascorbic:
We should probably just not be shipping the src folder to production

masonjames:
Thanks @ascorbic Agreed — that sounds like the cleanest fix.

One nuance from checking the manifests: `@emdash-cms/admin` already packs only `dist`, so this PR’s specific alias leak is mainly about linked/workspace installs where `src` still exists on disk. 

The source-shipping issue is in `emdash` itself: exports like `emdash/ui`, `emdash/ui/search`, and `emdash/routes/*` currently point at `src`, and those entries aren’t built into `dist` today.

I can keep this PR scoped as a small guard for consumer apps, then do a follow-up to move those exports to built outputs and remove `src` from the published package if that makes the most sense.

Happy to tackle that next if that’s the direction you want.

ascorbic:
Most of those UI exports are Astro components though, which are not compiled

masonjames:
Narrowed this to the admin package aliasing behavior only; no `emdash/ui*` or route source-export changes.

ascorbic (approved):
Thanks!

PR #218, opened 2026-04-04
Fix PluginBridge not exported from Cloudflare Worker entry point

### LINKED ISSUE #216: PluginBridge not exported from Worker entry point on Cloudflare — sandbox unavailable
When deploying an EmDash site to Cloudflare Workers with `sandboxRunner: sandbox()` configured, the marketplace plugin install endpoint returns `SANDBOX_NOT_AVAILABLE` even though the Worker Loader binding (`LOADER`) is correctly configured and the account is on the paid plan.

**Root cause:** The `PluginBridge` WorkerEntrypoint class is never exported from the Worker's main entry module (`entry.mjs`). The sandbox runner accesses it at runtime via `cloudflare:workers`'s `exports` object, which only contains exports from the Worker's main module. Since the Astro Cloudflare adapter generates `entry.mjs` with only a default export, `exports.PluginBridge` is always `undefined`, and `isAvailable()` returns `false`.

The class exists in the build output (inside `virtual_astro_middleware.mjs`) but is never re-exported from `entry.mjs`, so the Workers runtime can't find it.

**Environment:**
- `@astrojs/cloudflare` adapter
- `@emdash-cms/cloudflare` sandbox runner
- Cloudflare Workers paid plan with `worker_loaders` binding configured
- Deployed via `wrangler deploy` (not Pages)

**Proposed fix:** Two changes in `packages/core`:
1. The virtual `sandbox-runner` module re-exports `PluginBridge` from the sandbox runner package so it's included in the Rollup build graph
2. A Vite output plugin appends the `PluginBridge` named export to `entry.mjs` during `generateBundle`

PR coming from a fork shortly.

BenjaminPrice:
## What does this PR do?

The sandbox runner accesses PluginBridge via cloudflare:workers exports, which only contains exports from the Worker's main entry module. The Astro Cloudflare adapter generates entry.mjs with only a default export, so PluginBridge was never accessible at runtime — causing isAvailable() to return false and marketplace plugin installs to fail with SANDBOX_NOT_AVAILABLE.

Two changes:
1. The virtual sandbox-runner module now re-exports PluginBridge from the configured sandbox runner package, ensuring it's included in the Rollup build graph instead of being tree-shaken.
2. A new Vite output plugin (emdash-plugin-bridge-export) appends the PluginBridge named export to entry.mjs during generateBundle, making it visible to cloudflare:workers exports at runtime.

Closes #216

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes

## Screenshots / test output

Verified that after this change, the built `entry.mjs` includes `export { PluginBridge }` and the sandbox runner's `isAvailable()` returns true at runtime on Cloudflare Workers.

### DIFF
--- a/packages/core/src/astro/integration/virtual-modules.ts
+++ b/packages/core/src/astro/integration/virtual-modules.ts
@@ -273,6 +273,10 @@ export const sandboxEnabled = false;
 	return `
 // Auto-generated sandbox runner module
 import { createSandboxRunner as _createSandboxRunner } from "${sandboxRunner}";
+// Re-export PluginBridge so it's included in the build graph.
+// The entry-point export plugin will hoist this to entry.mjs at build time,
+// making it accessible via cloudflare:workers exports at runtime.
+export { PluginBridge } from "${sandboxRunner}";
 
 export const createSandboxRunner = _createSandboxRunner;
 export const sandboxEnabled = true;
--- a/packages/core/src/astro/integration/vite-config.ts
+++ b/packages/core/src/astro/integration/vite-config.ts
@@ -204,6 +204,54 @@ export function createVirtualModulesPlugin(options: VitePluginOptions): Plugin {
 }
 
 /**
+ * Creates a Vite plugin that ensures PluginBridge is exported from the
+ * Worker entry point when a sandbox runner is configured.
+ *
+ * The Cloudflare sandbox runner needs PluginBridge to be a named export
+ * of the main Worker module so it's accessible via cloudflare:workers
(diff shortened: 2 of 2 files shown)


ascorbic:
Hey. This isn't the right approach. The correct way is the one used in `demos/cloudflare/src/worker.ts`. The bug is that this is missing from the cloudflare templates, probably because of a bug in the sync script (it used to be in there).

PR #223, opened 2026-04-04
fix: chunk byline IN clauses to stay within D1 SQL variable limit

### LINKED ISSUE #219: hydrateEntryBylines exceeds D1 SQL variable limit on large collections
### Description

`getEmDashCollection` fails with `D1_ERROR: too many SQL variables` when the result set is large enough that the byline hydration step builds an `IN (?, ?, …)` clause exceeding D1's bound-parameter limit.

The root cause is in `hydrateEntryBylines` (`packages/core/src/query.ts`). It collects **all** entry IDs from the result set and passes them in a single call to `getBylinesForEntries`, which in turn issues unbounded `IN` queries through three code paths:

1. **`BylineRepository.getContentBylinesMany`** (`database/repositories/byline.ts`) — `WHERE cb.content_id IN (…)` with one variable per entry ID.
2. **`getAuthorIds`** (`bylines/index.ts`) — fallback path for entries without explicit bylines, `WHERE id IN (…)` with one variable per entry ID.
3. **`BylineRepository.findByUserIds`** (`database/repositories/byline.ts`) — `WHERE user_id IN (…)` for author-linked byline resolution.

Cloudflare D1 enforces a SQL bound-parameter limit (lower than SQLite's default 999). A collection query with `limit: 500` (or any sufficiently large result) will exceed this limit and trigger the error. The hydration runs inside `getEmDashCollection` before returning results, so callers have no way to work around it.

### Steps to reproduce

1. Seed a collection with ≥100 published entries (the exact threshold depends on D1's current variable limit).
2. Query the collection:
   ```ts
   const { entries, error } = await getEmDashCollection('posts', {
     status: 'published',
     limit: 500,
   });
   ```
3. The call fails with `D1_ERROR: too many SQL variables at offset NNN: SQLITE_ERROR`.

The error is caught by the `try/catch` in `hydrateEntryBylines` and logged as a warning — entries are returned but with no byline data.

### Environment

- emdash version: 0.1.0
- Runtime: Cloudflare Workers (D1)
- Node.js version: 22.x (local dev)

### Logs / error output

```
[emdash] Failed to hydrate bylines: D1_ERROR: too many SQL variables at offset 877: SQLITE_ERROR
```

(Observed on a production Cloudflare Worker serving a category page that returns ~100+ posts.)

### Suggested fix

Chunk the ID array in `hydrateEntryBylines` before calling `getBylinesForEntries`, then merge the result maps. This is the single point that controls the batch size for all three downstream queries. A conservative chunk size of 50 keeps each query well within D1's limits while minimizing round-trips.

```ts
// In hydrateEntryBylines (query.ts)
const CHUNK = 50;
const bylinesMap = new Map();
for (let i = 0; i < ids.length; i += CHUNK) {
  const chunkMap = await getBylinesForEntries(type, ids.slice(i, i + CHUNK));
  for (const [k, v] of chunkMap) bylinesMap.set(k, v);
}
```

Alternatively, the chunking could live in the repository methods themselves (`getContentBylinesMany`, `findByUserIds`) for defense-in-depth, since other callers could hit the same limit.

baezor:
## What does this PR do?

Fixes unbounded `IN (?, ?, …)` clauses in byline hydration that exceed Cloudflare D1's SQL bound-parameter limit when querying large collections.

Adds a `chunks()` utility (`utils/chunks.ts`) and applies it defense-in-depth at the repository level so any caller is protected:

- `BylineRepository.getContentBylinesMany` — deduplicates IDs, then chunks `content_id IN (…)`
- `BylineRepository.findByUserIds` — chunks `user_id IN (…)`
- `getAuthorIds` (bylines/index.ts) — chunks `id IN (…)` raw SQL

Each batch is capped at 50 IDs, well within D1's limit. Content IDs are also deduplicated before chunking to prevent duplicate credits when the same ID spans multiple chunks.

Closes #219

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes

## Screenshots / test output

All 26 tests pass across 3 test files (10 new):

```
 ✓ chunks > returns empty array for empty input
 ✓ chunks > returns single chunk when array fits within size
 ✓ chunks > splits array into even chunks
 ✓ chunks > handles remainder in last chunk
 ✓ chunks > handles chunk size of 1
 ✓ chunks > handles array exactly equal to chunk size
 ✓ SQL_BATCH_SIZE > is 50
 ✓ BylineRepository > getContentBylinesMany handles more IDs than SQL_BATCH_SIZE
 ✓ BylineRepository > getContentBylinesMany does not duplicate credits for repeated content IDs
 ✓ BylineRepository > findByUserIds handles more IDs than SQL_BATCH_SIZE
 ✓ getBylinesForEntries > handles batches larger than SQL_BATCH_SIZE across explicit and inferred bylines
```


baezor:
Confirming this addresses a real production bug — we just hit it at prepain.mx with `emdash@0.1.1` on Cloudflare D1, with 124 posts in the collection. When the admin bundle loads `/admin` and issues `GET /_emdash/api/content/posts?limit=100`, the worker logs:

```
Content list error: Error: D1_ERROR: too many SQL variables at offset 369: SQLITE_ERROR
```

Minimal standalone repro of the D1 limit itself (just to isolate the constraint from emdash):

```sh
wrangler d1 execute <db> --remote --command \
  "SELECT 1 WHERE 'x' = ? AND 1 IN (?,?,…×100)"
# → too many SQL variables at offset 231: SQLITE_ERROR [code: 7500]
```

Per the [D1 limits docs](https://developers.cloudflare.com/d1/platform/limits/), D1 caps bound parameters at **100 per query** — far below SQLite upstream's default of 32,766. Any `WHERE col = ? AND id IN (?×100)` clause produces 101 params and trips the limit. That's why this bug only reproduces on D1 at scale, and not on local SQLite fixtures.

Downstream (React #300 "rendered fewer hooks than expected") is a symptom — the admin bundle throws on the 500 response and the hook count goes inconsistent during the re-render. Fixing the query fixes the crash.

### Heads-up: one uncovered site

While tracing the failure in prod we noticed that `SeoRepository.getMany` (`packages/core/src/database/repositories/seo.ts:67-78`) has the **exact same** unbounded `IN` pattern:

```ts
const rows = await this.db
  .selectFrom("_emdash_seo")
  .selectAll()
  .where("collection", "=", collection)
  .where("content_id", "in", contentIds)  // ← 100 ids + 1 collection = 101 params
  .execute();
```

And it's called from `handleContentList` **before** `hydrateBylinesMany`:

```ts
// packages/core/src/api/handlers/content.ts:257-259
const hasSeo = await collectionHasSeo(db, collection);
await hydrateSeoMany(db, collection, result.items, hasSeo);
await hydrateBylinesMany(db, collection, result.items);
```

So any collection with `has_seo = 1` and ≥99 items will still 500 on the content list endpoint even after this PR merges. Our own posts collection has `has_seo = 1`, and it's the first function to fail in our trace.

Happy to open a follow-up PR that reuses the new `chunks()` helper + `SQL_BATCH_SIZE` for `SeoRepository.getMany`, or you can fold it into this PR if you'd prefer to land both together — whichever you prefer.

(As a short-term unblock we shipped a local `patch-package` hot-patch against our `node_modules/emdash/dist/**` that chunks all three functions — SEO, bylines, and `findByUserIds` — in batches of 90. We'll drop it the moment a release with the upstream fix ships.)

ascorbic (approved):
This and your other PR are great. Thank you. If possible, could you do the same for taxonomies?

PR #283, opened 2026-04-05
fix: Prevent draft autosave state loss and editor overwriting

orrc:
## What does this PR do?
Fixes a cluster of related draft/autosave bugs in the admin editor for revision-enabled content.

This started from a verified repro on a published post:
- Editing "Content" and waiting for autosave appeared to work locally
- However, the autosave `PUT` response returned stale content
- A redundant follow-up `GET` request also returned stale base-row data
- Any edits to "Excerpt" were overwritten immediately after autosave
- After page refresh, the rich text editor appeared to lose the saved draft content even though the draft revision had actually been updated

Root cause turned out to be three separate compounding issues:

1. `packages/core`
   - revision-backed autosave wrote the draft correctly, but the `PUT` response was built from the stale base content row instead of the active draft revision

2. `packages/admin`
   - autosave handling in the edit page was not safe for revision-backed content
   - initially this caused a redundant post-autosave `GET`
   - more importantly, an autosave response could overwrite newer local edits if the user kept typing while the request was in flight

3. `packages/admin`
   - `PortableTextEditor` treated its initial value as mount-only state
   - when parent state changed after mount (for example after draft load, reload, or reset), the live TipTap editor did not reconcile to that new value

What this PR changes:
- Returns the correct, draft-aware data from autosave responses for revision-backed content.
- Removes the redundant post-autosave `GET` call from the client (as the necessary data is in the autosave response).
- Only applies an autosave response back into edit-page queries when the local editor state still matches the snapshot that was sent.
- Reconciles post-mount external value changes into the rich-text editor.

Relates to #272, but extends beyond that admin-only reset loop to also fix the stale revision-backed autosave response, the redundant post-autosave refetch, and the in-flight autosave overwrite edge case.

## Type of change
- [x] Bug fix
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist
- [x] `pnpm typecheck` passes

But with a lot of prior verification, including with Playwright, and a lot of discussion on the solution

### DIFF
--- a/packages/admin/src/components/ContentEditor.tsx
+++ b/packages/admin/src/components/ContentEditor.tsx
@@ -103,6 +103,7 @@ export interface ContentEditorProps {
 		data: Record<string, unknown>;
 		slug?: string;
 		bylines?: BylineCreditInput[];
+		shouldApplyResponse?: () => boolean;
 	}) => void;
 	/** Whether autosave is in progress */
 	isAutosaving?: boolean;
@@ -300,6 +301,8 @@ export function ContentEditor({
 			}),
 		[formData, slug, activeBylines],
 	);
+	const currentDataRef = React.useRef(currentData);
+	currentDataRef.current = currentData;
 	const isDirty = isNew || currentData !== lastSavedData;
 
 	// Autosave with debounce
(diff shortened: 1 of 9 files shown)


ascorbic:
Hey, could you resolve the conflicts and add a changeset?

ascorbic:
Thanks for the thorough work here. Since this was opened, #302 has been merged, which fixes the admin-side form reset (replacing invalidateQueries with setQueryData cache patching). However, this PR contains two fixes that #302 doesn't address and that are still worth landing:                                                 
                           
 1. Server-side stale revision response – mergeDraftRevisionIntoItem in emdash-runtime.ts. The autosave PUT for revision-backed content was returning stale base-row data instead of draft revision data. 
 2. PortableText reconciliation – TipTap not updating when its value prop changes after mount. 

Would you be up for extracting those into one or two smaller PRs? Both would be good to land, but the server-side fix in particular is self-contained and should be straightforward to land on its own.

ascorbic:
Thanks for the contribution. I'm going to close this as it's been superseded by #302, #777 and #887. I hope you'll contribute again in future.

PR #302, opened 2026-04-06
Fix autosave form reset bug

### LINKED ISSUE #295: Bug: Autosave invalidates query cache and resets form fields to server state
## Description

When editing a content entry with many custom fields, autosave fires after 2 seconds of inactivity and resets all form fields to the previous server state, discarding any unsaved changes the user made.

## Steps to Reproduce

1. Create a collection with many custom fields (e.g. 80+ flat fields for a homepage)
2. Open the content editor in admin panel
3. Edit any field value
4. Wait ~2 seconds for autosave to trigger
5. **Observe:** The field reverts to its previous value

## Expected Behavior

Autosave should save the current form state without resetting the form. User edits should persist after autosave completes.

## Actual Behavior

After autosave completes, the form fields revert to the server state (the values before the edit). The user's changes are lost.

Manual save (clicking the Save button) works correctly — changes persist.

## Root Cause (from source analysis)

In the admin bundle (`@emdash-cms/admin/dist/index.js`):

1. **Autosave mutation** (`autosaveMutation`) calls `updateContent()` with `skipRevision: true`
2. **On success**, it calls `queryClient.invalidateQueries(["content", collection, id])` — this triggers a **refetch** of the entry from the server
3. The refetch updates the `item` prop, which triggers a `useEffect` that does:
   ```javascript
   setFormData(item.data);  // Resets form to server state
   ```
4. If the autosave payload didn't include all fields (or if there's a race condition between the save completing and the refetch), the form gets reset to stale server data

The `useEffect` dependency on `item?.updatedAt` / `itemDataString` causes a form reset every time the query cache is refreshed.

## Suggested Fix

Option A: Don't invalidate the query cache after autosave — just update `lastAutosaveAt` without refetching:
```javascript
onSuccess: () => {
    setLastAutosaveAt(new Date());
    // Don't invalidate queries — the local form state IS the source of truth
}
```

Option B: After refetch, merge server data with local unsaved changes instead of replacing form state.

## Environment

- EmDash v0.1.0
- Cloudflare Workers
- Collection with 85+ custom fields
- Chrome / Edge (latest)

## Workaround

Clicking the manual Save button immediately after editing works correctly. Only the debounced autosave exhibits this issue.

ideepakchauhan7:
## What does this PR do?

Fixes a bug where autosave invalidates the query cache and resets form fields to server state, causing user edits to be lost.

When editing a content entry with many custom fields, autosave fires after 2 seconds of inactivity. Upon success, it was calling `queryClient.invalidateQueries(["content", collection, id])`, which triggered a refetch from the server. This refetch updated the `item` prop, which triggered a `useEffect` in `ContentEditor` that resets form state (`formData`, `slug`, etc.) to server values.

The fix removes the `queryClient.invalidateQueries()` call from the `autosaveMutation` `onSuccess` handler. The local form state remains the source of truth during editing. Manual saves still properly invalidate and refresh the cache.

Closes #295

## Type of change

- [x] Bug fix

## Checklist

- [x] `pnpm typecheck` passes (pre-existing errors in admin package unrelated to this change)
- [x] `pnpm --silent lint:json | jq '.diagnostics | length'` returns 0 (pre-existing errors unrelated to this change)


ascorbic:
Can you fix the conflicts?

ideepakchauhan7:
I’m resolving the merge conflicts now and will push an updated branch shortly.

ideepakchauhan7:
Conflicts are fixed now. I merged the latest main into this branch and pushed the updated branch.

packages/admin/src/router.tsx
```diff
@@ -660,14 +660,9 @@ function ContentEditPage() {
...
-			// doesn't overwrite the form via the sync effect
-			void queryClient.invalidateQueries({ queryKey: ["content", collection, id] });
-			if (rawItem?.draftRevisionId) {
-				void queryClient.invalidateQueries({
-					queryKey: ["revision", rawItem.draftRevisionId],
-				});
-			}
+			// Keep the editor's local state as the source of truth during autosave.
+			// Invalidating here can refetch slightly older server data and reset the form
+			// while the user is still typing.
```
ascorbic: Instead of removing the invalidation, could you optimistically patch the cache insgtead. Soemthing liek this:
```suggestion
    // Optimistically update the cache so it stays fresh                                                                                          
    // without triggering a refetch
    queryClient.setQueryData(                                                                                                                     
      ["content", collection, id],                                                                                                                
      (old) => old ? { ...old, data: variables.data ?? old.data, slug: variables.slug ?? old.slug, updatedAt: new Date().toISOString() } : old
    ); 
```
(not checked)

packages/admin/tests/components/ContentEditor.test.tsx
```diff
@@ -296,6 +296,47 @@ describe("ContentEditor", () => {
 			const savedBtn = screen.getByRole("button", { name: "Saved" });
 			await expect.element(savedBtn).toBeDisabled();
 		});
+
+		it("does not queue another autosave after a successful autosave", async () => {
```
ascorbic: This tests doesn't actually test the bug that this is fixing. Could you add a regressoin test?

ascorbic (approved):
Great stuff. Thanks for getting this across the line

PR #325, opened 2026-04-06
fix: preserve post_date and publish status in WXR import

### LINKED ISSUE #322: WXR import does not preserve post dates or publish status
## Description

When importing a WordPress WXR export via the admin file-upload flow, two timestamps are silently discarded:

- `created_at` is set to the import time rather than the original `wp:post_date`
- `published_at` is `NULL` on every imported post, even those that were `publish` status in WordPress

The result is that a post originally written in 2019 appears in EmDash as if it was created today, and published posts have no `published_at` at all.

Related discussion: #285.

## Steps to reproduce

1. Create posts in WordPress with varied historical dates
2. Export as WXR
3. Import via Admin → Import → WordPress
4. Query `ec_posts` in the EmDash SQLite database

## Observed

All imported rows have `created_at` equal to the import timestamp and `published_at` NULL, regardless of original dates or publish status. Example from a real import (post originally dated 2019-01-01):

```
slug        | status    | created_at               | published_at
first-post  | published | 2026-04-06T18:18:37.810Z | NULL
```

Posts spanning 2019–2025 all received the same `created_at`.

## Expected

- `created_at` set from `wp:post_date` (fallback: `pubDate`)
- `published_at` set from `wp:post_date` when status is `publish`
- Draft posts: `created_at` from `wp:post_date`, `published_at` NULL

## Environment

- emdash version: 0.1.0
- Node.js version: 22.20.0
- Runtime: Node
- OS: Linux

MattieTK:
## What does this PR do?

WXR-imported posts were receiving `created_at` set to the import time and `published_at` set to `NULL`, regardless of the original WordPress post dates or publish status. A post written in 2019 would appear in EmDash as if it was created today, and all imported published posts had no `published_at`.

Closes #322. Related discussion: #285.

**Root cause:** `ContentRepository.create()` always used `new Date()` for `created_at` and had no way to accept an override. The `handleContentCreate()` handler and the WXR import route both lacked the plumbing to pass source dates through.

**Fix:**
- Added `createdAt` to `CreateContentInput` — if provided, used for `created_at`; otherwise falls back to `new Date()` (no behaviour change for existing callers)
- Threaded `createdAt` and `publishedAt` through `handleContentCreate()`
- In the WXR import route, extracted `wp:post_date_gmt` (fallback: `wp:post_date`) and passed it as `createdAt`; set `publishedAt` to the same value for posts with `status = "published"`

## Type of change

- [x] Bug fix

## Checklist

- [x] `pnpm typecheck` passes

## Screenshots / test output

```
✓ tests/database/repositories/content.test.ts (54 tests)
```

New tests:
- `should use provided createdAt instead of current time`
- `should use provided publishedAt when creating published content`

### DIFF
--- a/.changeset/pretty-crews-march.md
+++ b/.changeset/pretty-crews-march.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes WXR import not preserving original post dates or publish status. Imported content now uses `wp:post_date_gmt` for `created_at` and sets `published_at` for published posts, rather than defaulting to the import time.
--- a/packages/core/src/api/handlers/content.ts
+++ b/packages/core/src/api/handlers/content.ts
@@ -385,6 +385,8 @@ export async function handleContentCreate(
 		locale?: string;
 		translationOf?: string;
 		seo?: ContentSeoInput;
+		createdAt?: string | null;
+		publishedAt?: string | null;
 	},
 ): Promise<ApiResult<ContentResponse>> {
 	try {
@@ -423,6 +425,8 @@ export async function handleContentCreate(
 				authorId: body.authorId,
 				locale: body.locale,
 				translationOf: body.translationOf,
+				createdAt: body.createdAt,
+				publishedAt: body.publishedAt,
 			});
 
 			if (body.bylines !== undefined) {
--- a/packages/core/src/astro/routes/api/import/wordpress/execute.ts
+++ b/packages/core/src/astro/routes/api/import/wordpress/execute.ts
@@ -232,6 +232,12 @@ async function importContent(
 				bylineCache,
 			);
 
+			// Preserve original WordPress dates.
+			// postDateGmt is always UTC; fall back to postDate if absent.
+			const sourceDateStr = post.postDateGmt || post.postDate;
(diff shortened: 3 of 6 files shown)


barckcode:
Hey @MattieTK — nice catch on the date preservation issue! I was working on the same problem and noticed a few additional edge cases that might be worth considering:

1. **`0000-00-00 00:00:00` sentinel**: WordPress sets `post_date_gmt` to this value for unpublished drafts. Since it's a non-empty string, `post.postDateGmt || post.postDate` would pick it up, and `new Date("0000-00-00 00:00:00").toISOString()` throws a `RangeError`. This would crash the import when drafts are included.

2. **`wxrPostToNormalizedItem()` in `wxr.ts`**: The normalized import source still uses `post.postDate` (site-local, no timezone) instead of `postDateGmt`. This affects the plugin-based import path (`wordpress-plugin/execute.ts`) which goes through `NormalizedItem` rather than the direct WXR handler.

3. **`postModifiedGmt`**: Same issue with the `modified` field — uses the site-local variant instead of the GMT one.

I've opened #326 as a superset that includes your changes plus fixes for these additional cases. Happy to coordinate if you'd prefer to incorporate them into this PR instead — just let me know!

PR #354, opened 2026-04-07
fix(core): chunk IN-clause batch queries under D1's 100-param limit

ilicfilip:
## What does this PR do?

Fixes a 500 error on the admin content list page for collections with ~100+ items, caused by D1's 100-bound-parameter-per-statement limit.

**Repro** — list any collection with ≥ 99 items. Server logs:

```
D1_ERROR: too many SQL variables at offset 369: SQLITE_ERROR
    at D1DatabaseSession._sendOrThrow (cloudflare-internal:d1-api:139:19)
    at D1Connection.executeQuery
    at SeoRepository.getMany (packages/core/src/database/repositories/seo.ts:73)
    at hydrateSeoMany (packages/core/src/api/handlers/content.ts:84)
    at handleContentList (packages/core/src/api/handlers/content.ts:258)
```

**Cause** — `handleContentList` calls `hydrateSeoMany` and `hydrateBylinesMany`, which build single SELECTs with `where("content_id", "in", contentIds)`. Kysely binds one parameter per id; with the admin SPA's `limit: 100` and an extra slot for the `collection` filter, the statement hits 101 bound params on a full page. D1 (and stock SQLite — `SQLITE_LIMIT_VARIABLE_NUMBER` defaults to 100 on D1) rejects it.

After the 500 on SEO is fixed, the same query shape in `BylineRepository.getContentBylinesMany` hits it next, then `findByUserIds`.

**Fix** — chunk each IN-clause into batches of 90 ids in:

- `SeoRepository.getMany`
- `BylineRepository.getContentBylinesMany`
- `BylineRepository.findByUserIds`

90 leaves headroom under the 100-param ceiling for the collection/slug filter that shares the statement. Three functions touched, behavior is identical for id lists ≤ 90 (single SELECT as before) and correct for larger lists (unions the chunk results in the same map).

### Scope / known similar instances

Same `where(..., "in", ids)` pattern exists in a few other places that aren't on the content-list hot path and that I left out of this PR per CONTRIBUTING.md's "no bulk/spray fixes" rule. Worth a follow-up PR:

- `packages/core/src/database/repositories/comment.ts`
- `packages/core/src/database/repositories/plugin-storage.ts`
- `packages/core/src/taxonomies/index.ts`

### No regression test — why

I drafted one but pulled it. The test DB is in-memory `better-sqlite3`, whose default `SQLITE_LIMIT_VARIABLE_NUMBER` is 32766 — so a test that calls `getMany` with 150 ids passes against **both** the chunked and unchunked implementations. It would only prove "the chunking loop returns correct results if chunking exists", not "the chunking happens". A true regression test would need either (a) spying on db execution to assert chunk boundaries, or (b) lowering better-sqlite3's variable limit if that's possible. Happy to add either if a reviewer prefers — just wanted to avoid committing a test that doesn't actually guard the bug.

The bug itself is reproducible in any `wrangler dev` / `miniflare` environment that emulates D1's limit, and the stack trace above was captured from exactly such a run.

Closes #

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes (ran in `packages/core`)
- [x] `pnpm --silent lint:json | jq '.diagnostics | length'` returns 0 **for the files I touched**. Note: `main` currently has a baseline of 23 pre-existing lint diagnostics in demos/, templates/, and a few unrelated package files. I did not touch any of them and they are not introduced by this PR.
- [x] `pnpm test` passes — full `packages/core` suite, 2114/2114 tests across 114 files
- [ ] I have added/updated tests for my changes (if applicable) — see "No regression test" above
- [x] I have added a [changeset](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#changesets) (`.changeset/old-badgers-see.md`)

## Screenshots / test output

```
Test Files  114 passed (114)
     Tests  2114 passed (2114)
  Duration  5.41s
```


ascorbic:
Thanks! The changes in here have already been fixed, so I'm going to close this.

PR #371, opened 2026-04-08
fix(core): return OAuth discovery 401 for unauthenticated MCP POSTs

pejmanjohn:
## What does this PR do?

Fixes OAuth discovery for unauthenticated MCP `POST` requests on `/_emdash/api/mcp`.

Some MCP clients, including Codex, probe the MCP endpoint with `POST` when starting auth discovery. Before this change, unauthenticated `POST /_emdash/api/mcp` requests without `X-EmDash-Request: 1` were rejected by the generic CSRF middleware with `403 CSRF_REJECTED` before the MCP auth path could return `401 Unauthorized` with the `WWW-Authenticate` header.

That prevented OAuth-capable MCP clients from starting discovery even though EmDash already exposed valid OAuth metadata endpoints.

This change:
- allows tokenless MCP discovery `POST` requests with no authenticated session to reach the existing `401 + WWW-Authenticate` response
- preserves CSRF rejection for session/cookie-backed MCP `POST` requests without `X-EmDash-Request: 1`
- keeps non-MCP API CSRF behavior unchanged
- adds regression tests for unauthenticated MCP `POST`, invalid bearer token handling, and the non-regression CSRF cases

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [ ] `pnpm --silent lint:json | jq '.diagnostics | length'` returns 0

## Screenshots / test output

Ran:
- `pnpm build`
- `pnpm typecheck`
- `pnpm typecheck:demos`
- `pnpm --filter emdash exec vitest run tests/unit/auth/mcp-discovery-post.test.ts tests/unit/auth/discovery-endpoints.test.ts tests/unit/api/csrf.test.ts tests/unit/mcp/authorization.test.ts`

Notes:
- `pnpm --silent lint:json | jq '.diagnostics | length'` currently reports 3 existing warnings in unrelated files.

### DIFF
--- a/packages/core/src/astro/middleware/auth.ts
+++ b/packages/core/src/astro/middleware/auth.ts
@@ -211,19 +211,25 @@ export const onRequest = defineMiddleware(async (context, next) => {
 
 	const isTokenAuth = bearerResult === "authenticated";
 
+	// MCP discovery/tooling is bearer-only. Session/external auth should never
+	// be consulted for this endpoint, and unauthenticated requests must return
+	// the OAuth discovery-style 401 response.
+	const method = context.request.method.toUpperCase();
+	const isMcpEndpoint = url.pathname === MCP_ENDPOINT_PATH;
+	if (isMcpEndpoint && !isTokenAuth) {
+		return mcpUnauthorizedResponse(url);
+	}
+
 	// CSRF protection: require X-EmDash-Request header on state-changing requests.
 	// Skip for token-authenticated requests (tokens aren't ambient credentials).
 	// Browsers block cross-origin custom headers, so this prevents CSRF without tokens.
 	// OAuth authorize consent is exempt: it's a standard HTML form POST that can't
 	// include custom headers. The consent flow is protected by session + single-use codes.
-	const method = context.request.method.toUpperCase();
 	const isOAuthConsent = url.pathname.startsWith("/_emdash/oauth/authorize");
-	const isMcpEndpoint = url.pathname === MCP_ENDPOINT_PATH;
 	if (
 		isApiRoute &&
 		!isTokenAuth &&
 		!isOAuthConsent &&
-		!isMcpEndpoint &&
 		isUnsafeMethod(method) &&
 		!isPublicApiRoute
 	) {
@@ -576,20 +582,14 @@ async function handlePasskeyAuth(
 	next: Parameters<Parameters<typeof defineMiddleware>[0]>[1],
 	isApiRoute: boolean,
 ): Promise<Response> {
-	const { url, locals, request, session } = context;
+	const { url, locals, session } = context;
 	const { emdash } = locals;
-	const isMcpEndpoint = url.pathname === MCP_ENDPOINT_PATH;
-	const method = request.method.toUpperCase();
 
 	try {
 		// Check session for user (session.get returns a Promise)
 		const sessionUser = await session?.get("user");
 
 		if (!sessionUser?.id) {
-			// Not authenticated
-			if (isMcpEndpoint) {
-				return mcpUnauthorizedResponse(url);
-			}
 			if (isApiRoute) {
 				return Response.json(
 					{ error: { code: "NOT_AUTHENTICATED", message: "Not authenticated" } },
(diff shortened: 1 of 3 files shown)


packages/core/src/astro/middleware/auth.ts
ascorbic, line 236:
```suggestion
const isOAuthConsent = url.pathname.startsWith("/_emdash/oauth/authorize");
const isMcpEndpoint = url.pathname === "/_emdash/api/mcp";
if (
isApiRoute &&
!isTokenAuth &&
!isOAuthConsent &&
!isMcpEndpoint &&
```

ascorbic, line 258:
```suggestion
  if (isMcpEndpoint) {                                                                                                    
      return Response.json(                 
          { error: { code: "NOT_AUTHENTICATED", message: "Not authenticated" } },                                                               
          { status: 401, headers: {                                              
              "WWW-Authenticate": `Bearer resource_metadata="${url.origin}/.well-known/oauth-protected-resource"`,                              
              ...MW_CACHE_HEADERS,                                                                                
          }},                                                                                                                                   
      );     
  }               
```

ascorbic (commented):
Thanks for the PR. The fix is in the right area, but I think there's a simpler approach. No MCP client will ever authenticate via session cookies, so we don't need to preserve CSRF for session-backed MCP requests: we can just make MCP bearer-only. I've made a couple of inline suggesitons, but it might need a few more as I've not tested. Basically for MCP requests we want to exempt from CSRF, but only allow token auth.

packages/core/src/astro/middleware/auth.ts
pejmanjohn: @ascorbic Updated this to match the broader bearer-only direction. MCP now short-circuits before session or external auth: requests without a valid bearer token return the discovery-style `401` with `WWW-Authenticate`, and session-backed MCP requests are no longer accepted even if the CSRF header is present.

I also updated the regression test to assert that MCP POST discovery no longer reads session state at all. Verified with `pnpm --filter emdash test tests/unit/auth/mcp-discovery-post.test.ts`, `pnpm test tests/unit/auth/*.test.ts`, `pnpm typecheck`, and `pnpm --silent lint:json | jq .diagnostics | length`.

packages/core/src/astro/middleware/auth.ts
```diff
@@ -80,6 +81,33 @@ function buildEmDashCsp(marketplaceUrl?: string): string {
...
+	);
+}
+
+function mcpUnauthorizedResponse(url: URL): Response {
+	return Response.json(
+		{ error: { code: "NOT_AUTHENTICATED", message: "Not authenticated" } },
+		{
+			status: 401,
+			headers: {
+				"WWW-Authenticate": `Bearer resource_metadata="${url.origin}/.well-known/oauth-protected-resource"`,
```
ascorbic: This will need to use `getPublicOrigin` instead of `url.origin`, meaning `mcpUnauthorizedResponse` needs to be passed the config.

pejmanjohn: @ascorbic Updated this to route the anonymous MCP discovery `401` through `getPublicOrigin(...)`, so the `WWW-Authenticate` resource metadata URL now respects `siteUrl` / reverse-proxy deployments instead of using the raw request origin.

I also added a regression covering the configured public-origin case on the anonymous MCP POST path.

Verified with:
- `pnpm --filter emdash test tests/unit/auth/mcp-discovery-post.test.ts`
- `pnpm test tests/unit/auth/*.test.ts`
- `pnpm typecheck`
- `pnpm --silent lint:json | jq .diagnostics | length`

pejmanjohn:
@ascorbic rebased this branch onto current `main` to clear the `needs-rebase` conflict, and the MCP public-origin follow-up is included in the rebased history. I reran:
- `pnpm --filter emdash test tests/unit/auth/mcp-discovery-post.test.ts`
- `pnpm test tests/unit/auth/*.test.ts`
- `pnpm typecheck`

The PR was already ready for review, so it remains ready.

ascorbic (approved):
Great! Thanks

PR #380, opened 2026-04-08
fix(admin): ensure content list pagination works correctly

kombuchapunk:
## Summary

The "Load More" button in the admin content list never appeared — even when collections had more items than a single page. Two bugs in `packages/admin/src/`:

1. **`getNextPageParam` returned an empty string instead of `undefined`** (`router.tsx`) — React Query treats any truthy return as "there's another page", but an empty-string cursor also prevented it from correctly setting `hasNextPage = false`. Now explicitly returns `undefined` when there is no next cursor.

2. **`limit` was conditionally omitted from API requests** (`lib/api/content.ts`) — when no `limit` was passed, the query param was skipped entirely, causing inconsistent server responses. Now always sends `limit` (defaults to 100, capped server-side at 100).

Additionally, the explicit `limit: 100` in the query call was removed because D1 throws SQL variable limit errors when hydrating SEO data for 100 items. The server default of 50 works correctly with cursor-based pagination.

## Test plan

1. Start a dev server with a seeded collection that has **>50 items** (e.g. posts)
2. Open the content list: `http://localhost:4321/_emdash/admin/content/posts`
3. Verify:
   - [ ] "Load More" button appears below the list
   - [ ] Clicking "Load More" appends the next batch of items
   - [ ] After all items are loaded, the "Load More" button disappears
   - [ ] Client-side page navigation (arrow buttons, "1 / N") still works within each loaded batch
   - [ ] Collections with <50 items do **not** show a "Load More" button

### DIFF
--- a/packages/admin/src/lib/api/content.ts
+++ b/packages/admin/src/lib/api/content.ts
@@ -140,7 +140,9 @@ export async function fetchContentList(
 ): Promise<FindManyResult<ContentItem>> {
 	const params = new URLSearchParams();
 	if (options?.cursor) params.set("cursor", options.cursor);
-	if (options?.limit) params.set("limit", String(options.limit));
+	// Always include limit to ensure pagination works correctly
+	// Use provided value, or default to 100 for better UX in admin lists
+	params.set("limit", String(options?.limit ?? 100));
 	if (options?.status) params.set("status", options.status);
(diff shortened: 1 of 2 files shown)


MA2153:
Hello. This PR has been open for more than a month, yet the CLA has not been signed. If you intend to follow through, kindly open a new PR and sign the CLA. Thank you.

Issue #210, opened 2026-04-04
Title: getPasskeyConfig() ignores configured rpId, always uses url.hostname — breaks passkeys behind reverse proxy


sebhosting:
## Bug

When self-hosting EmDash behind a reverse proxy (Traefik, nginx, etc), 
the internal request URL is `localhost`, causing `getPasskeyConfig()` to 
set `rpId: "localhost"` regardless of the configured rpId.

## Root Cause

In `packages/auth/src/passkey/config.ts`, `getPasskeyConfig()` derives 
rpId from `url.hostname` instead of reading from config:

\```ts
function getPasskeyConfig(url, siteName) {
  return {
    rpName: siteName || url.hostname,
    rpId: url.hostname,   // ← ignores config
    origin: url.origin
  };
}
\```

## Fix

Read rpId from config when available:

\```ts
function getPasskeyConfig(url, siteName, config?) {
  return {
    rpName: config?.rpName || siteName || url.hostname,
    rpId: config?.rpId || url.hostname,
    origin: config?.origin || url.origin
  };
}
\```

## Workaround

Patch the built chunk directly and set explicit values.

## Environment
- Self-hosted Node.js + SQLite
- Behind Traefik reverse proxy
- Cloudflare DNS proxy

## Root Cause

In `packages/auth/src/passkey/config.ts`, `getPasskeyConfig()` derives 
rpId from `url.hostname` instead of reading from config:

\```ts
function getPasskeyConfig(url, siteName) {
  return {
    rpName: siteName || url.hostname,
    rpId: url.hostname,   // ← ignores config
    origin: url.origin
  };
}
\```

## Environment
- Self-hosted Node.js + SQLite
- Behind Traefik reverse proxy
- Cloudflare DNS proxy

### Logs / error output

```shell

```

PR #382, opened 2026-04-08
fix: add siteUrl to resolve origin mismatch behind reverse proxies

UpperM:
## What does this PR do?

Adds a `siteUrl` config option that fixes a class of reverse-proxy origin mismatches affecting Node.js self-hosted deployments. When EmDash runs behind a TLS-terminating proxy (nginx, Traefik, Caddy -- typical for Coolify, Dokploy, or any Docker setup), the server sees `http://localhost:4321` while browsers reach it via `https://mysite.example.com`. PR #225 fixed this for passkeys; this PR extends the same fix to all affected call sites.

**Affected subsystems (before this fix):**
- CSRF checks (form-like POSTs return 403 behind a proxy)
- Auth and setup wizard redirects (point to the internal address)
- OAuth redirect URIs (don't match the registered public URL)
- MCP/CLI well-known discovery endpoints (advertise internal URLs)
- `Secure` cookie flag (omitted when TLS is terminated by the proxy)
- Snapshot export, theme preview, sitemap, robots.txt, JSON-LD

**What this adds:**

```ts
// astro.config.mjs
emdash({
  siteUrl: "https://mysite.example.com",
})
```

A single `getPublicOrigin(url, config)` helper wraps every affected `url.origin` call. When `siteUrl` is set, it returns that; otherwise it falls back to `url.origin` unchanged, so non-proxied deployments are not affected. The public origin comes from operator config or environment variables, not from request headers.

Also supports `EMDASH_SITE_URL` / `SITE_URL` env vars for container deployments where the domain is only known at runtime (using `process.env` rather than `import.meta.env` so it works after the build).

**Breaking:** `passkeyPublicOrigin` is removed. Since this is pre-release, renaming felt cleaner than maintaining both. Migration is just renaming the key in `astro.config.mjs`.

### Decision point: `checkOrigin: false`

This PR disables Astro's built-in `security.checkOrigin` and relies on EmDash's own CSRF layer (`checkPublicCsrf` + `X-EmDash-Request` header). I want to be transparent about why and open this to discussion.

**The problem:** Astro's `checkOrigin` and `allowedDomains` are baked into the build manifest. Behind a reverse proxy, `url.origin` is the internal address and Astro's origin check blocks form-like POSTs and requests without content-type with "Cross-site POST form submissions are forbidden" -- before EmDash middleware runs. (Note: `application/json` requests are not affected by Astro's check, but some EmDash flows use form-like content types.) The only Astro-native solutions are:

- `allowedDomains` scoped to a hostname -- works, but requires the domain at build time, which breaks Docker "build once, deploy anywhere" images where `SITE_URL` is set at runtime
- `allowedDomains: [{}]` wildcard -- enables host header poisoning on GET responses (login redirects, OAuth, WWW-Authenticate) when `SITE_URL` isn't configured

**The approach:** Disable `checkOrigin` and let EmDash handle CSRF entirely. EmDash's CSRF layer handles the proxy case that Astro's check cannot: dual-origin support and runtime siteUrl resolution via env vars. When `siteUrl` is known at build time, `allowedDomains` is still set so `Astro.url` reflects the public origin for user template code.

**The gap:** User-authored form-handling routes on the same Astro site would lose Astro's CSRF protection for form submissions. (Astro's `checkOrigin` only covers form-like content types and requests without content-type -- it does not protect `application/json` API routes regardless.) In practice, EmDash sites typically don't add custom form POST endpoints (all state changes go through EmDash's API), but this is worth flagging.

**Alternatives I considered but rejected:**
- Middleware interception (Astro's origin check is prepended via `unshift` before user middleware in the pipeline)
- Runtime manifest patching (integration can't control the adapter)
- Custom adapter (too invasive for this fix)

If you'd prefer a different approach -- keeping `checkOrigin` enabled and documenting that Docker users must set `siteUrl` at build time, or another path I haven't considered -- I'm happy to rework this.

> Discussion #315 was opened before this implementation. I'm happy to adjust scope, naming, or approach based on maintainer feedback -- this can be split, renamed, or otherwise reshaped.

Related: #210, #225, #253, #393

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [ ] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/315

## Screenshots / test output

<details>
<summary>Test output</summary>

```
115 test files passed (2132 tests)
✓ packages/core/tests/unit/api/public-url.test.ts
✓ packages/core/tests/unit/api/csrf.test.ts
✓ packages/core/tests/unit/auth/passkey-config.test.ts
✓ packages/core/tests/unit/auth/discovery-endpoints.test.ts
```

</details>

### DIFF
--- a/.changeset/site-url-reverse-proxy.md
+++ b/.changeset/site-url-reverse-proxy.md
@@ -1,0 +1,11 @@
+---
+"emdash": minor
+---
+
+Adds `siteUrl` config option to fix reverse-proxy origin mismatch. Replaces `passkeyPublicOrigin` with a single setting that covers all origin-dependent features: passkeys, CSRF, OAuth, auth redirects, MCP discovery, snapshots, sitemap, robots.txt, and JSON-LD.
+
+Supports `EMDASH_SITE_URL` / `SITE_URL` environment variables for container deployments where the domain is only known at runtime.
+
+Disables Astro's `security.checkOrigin` (EmDash's own CSRF layer handles origin validation with dual-origin support and runtime siteUrl resolution). When `siteUrl` is set in config, also sets `security.allowedDomains` so `Astro.url` reflects the public origin in templates.
+
+**Breaking:** `passkeyPublicOrigin` is removed. Rename to `siteUrl` in your `astro.config.mjs`.
--- a/demos/simple/astro.config.mjs
+++ b/demos/simple/astro.config.mjs
@@ -30,8 +30,8 @@ export default defineConfig({
 				baseUrl: "/_emdash/api/media/file",
 			}),
 			plugins: [auditLogPlugin()],
-			// HTTPS reverse proxy: uncomment so passkey verify matches browser origin
-			// passkeyPublicOrigin: "https://emdash.local:8443",
+			// HTTPS reverse proxy: uncomment so all origin-dependent features match browser
+			// siteUrl: "https://emdash.local:8443",
 		}),
 	],
 	devToolbar: { enabled: false },
--- a/docs/src/content/docs/reference/configuration.mdx
+++ b/docs/src/content/docs/reference/configuration.mdx
@@ -200,13 +200,13 @@ Use Cloudflare Access as the authentication provider instead of passkeys.
 	magic links, and self-signup are disabled.
 </Aside>
 
-### `passkeyPublicOrigin`
+### `siteUrl`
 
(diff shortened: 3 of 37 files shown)


ascorbic (approved):
Great work. Thanks for this.

PR #424, opened 2026-04-10
fix(middleware): allow public access to search API (#104)

### LINKED ISSUE #104: Search API requires authentication, preventing frontend search
## Description

The `/_emdash/api/search` endpoint requires `search:read` permission, which means unauthenticated visitors cannot use the `LiveSearch` component on the frontend.

## Root cause

In `astro/routes/api/search/index.ts`:

```ts
const denied = requirePerm(user, "search:read");
if (denied) return denied;
```

This returns 401 for any unauthenticated request. The same applies to `/_emdash/api/search/suggest`.

## Impact

The `LiveSearch` component (which hardcodes `/_emdash/api/search` as its endpoint) doesn't work for public visitors. Search is a common frontend feature that should work without authentication.

## Workaround

Create a custom public API route that calls the `search()` function directly:

```ts
// src/pages/api/search.ts
import type { APIRoute } from "astro";
import { search } from "emdash";

export const GET: APIRoute = async ({ url }) => {
  const q = url.searchParams.get("q")?.trim();
  if (!q || q.length < 2) {
    return new Response(JSON.stringify({ data: { items: [] } }), {
      headers: { "Content-Type": "application/json" },
    });
  }

  const result = await search(q, {
    collections: url.searchParams.get("collections")?.split(","),
    status: "published",
    limit: Math.min(Number(url.searchParams.get("limit")) || 10, 20),
  });

  return new Response(JSON.stringify({ data: result }), {
    headers: { "Content-Type": "application/json" },
  });
};
```

This requires building a custom search UI since `LiveSearch` hardcodes the emdash API endpoint.

## Suggested fix

Either:
1. Make `/_emdash/api/search` publicly accessible (with `status: "published"` enforced for unauthenticated users), or
2. Add a configurable `endpoint` prop to `LiveSearch` so themes can point it at a custom route, or
3. Add search to the public API prefixes list with appropriate rate limiting

csfalcao:
## What does this PR do?

Fixes the public search bug from #104 at the middleware layer.

#107 removed the handler-level `requirePerm("search:read")` check from `packages/core/src/astro/routes/api/search/index.ts` and `.../suggest.ts`, but the auth middleware at `packages/core/src/astro/middleware/auth.ts` (`handlePasskeyAuth`, ~line 587) still returned `401 NOT_AUTHENTICATED` before the handler ran, because `/_emdash/api/search` was not in `PUBLIC_API_EXACT` or `PUBLIC_API_PREFIXES`. The handler-level change from #107 therefore never executed for anonymous callers, and the shipped `<LiveSearch>` component (which fetches without credentials per its JSDoc example) silently showed "No results found" on every query.

Reproducible on `main` at commit 59bdca1:

```console
$ curl "http://localhost:4321/_emdash/api/search?q=design"
{"error":{"code":"NOT_AUTHENTICATED","message":"Not authenticated"}}
HTTP 401
```

This PR adds `/_emdash/api/search` to `PUBLIC_API_EXACT`. The `searchWithDb` query layer already hardcodes `status = 'published'` and the SQL in `query.ts` repeats the filter, so anonymous callers only see published content — matching the safety justification given in #104's closing comment. Admin endpoints (`/search/enable`, `/search/rebuild`, `/search/stats`) remain gated because they are not added to the set — the exact-match check in `isPublicEmDashRoute` means the sub-paths do not inherit access.

`/search/suggest` is deliberately **not** included: while investigating I hit a separate FTS5 syntax bug in `getSuggestions` (double `*` prefix, `query.ts:279`) that makes it return 500 regardless of auth. I'll file that as a separate issue with a failing test so it stays scoped.

The existing E2E test `"search endpoint requires authentication"` asserted the buggy behavior. It is replaced with:
- `"search endpoint is publicly accessible"` — asserts `200` for unauthenticated GET
- `"search admin endpoints still require authentication"` — asserts `/stats` and `/enable` stay gated

Verified locally against a fresh `demos/simple` with the seeded blog content:

```console
$ curl "http://localhost:4321/_emdash/api/search?q=design"
HTTP 200
# 4 results with <mark>Designing</mark> FTS5 highlighting

$ curl "http://localhost:4321/_emdash/api/search/stats"
HTTP 401   # admin-only, as intended

$ curl -X POST "http://localhost:4321/_emdash/api/search/enable" \
    -H "Content-Type: application/json" -d '{"collection":"posts"}'
HTTP 403   # CSRF-rejected (correct — no X-EmDash-Request header)
```

Closes #104. Follows up on #107.

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [ ] `pnpm --silent lint:json | jq '.diagnostics | length'` returns 0
- [ ] New features link to an approved Discussion — _n/a, this is a bug fix_

**On the unchecked boxes — full transparency:**

- `lint:json` returns `2` diagnostics on both `main` (commit 59bdca1) and this branch. Both are pre-existing `typescript-eslint` warnings in files this PR does not touch (`packages/admin/vitest.config.ts`, `packages/core/src/astro/integration/vite-config.ts`). This PR introduces zero new lint findings. I followed `AGENTS.md`'s "do not touch code outside the scope of your change" guidance and left them alone.
- `pnpm --filter emdash test` shows 2 pre-existing failures in `tests/integration/plugins/capabilities.test.ts` (HTTP capability timeouts at 5000ms). Reproduced on clean `main` at 59bdca1. Unrelated to auth or search. Targeted test run for the openapi test (`tests/unit/api/openapi.test.ts`, which references `/search`) passes 22/22.
- E2E tests for the modified `search.spec.ts` were not run locally because Playwright browsers aren't installed on this machine. The behavior was verified directly via `curl` against a running `demos/simple` instance — the same assertions the tests make. CI will execute the full Playwright suite.

Investigation, diff, test rewrites, and PR body were produced with Claude Opus 4.6 acting as an agent in this repo. Every change was reviewed line by line and all claims in this PR (repro commands, file paths, line numbers, pre-existing warnings, test names) were independently verified against the running demo and `git log` on `main`.

## Screenshots / test output

End-to-end verification via `curl` against `pnpm --filter emdash-demo dev` from a clean clone at commit 59bdca1 (confirming bug), and after applying this patch (confirming fix):

**Before (on main):**
```
$ curl -o /dev/null -w "%{http_code}\n" "http://localhost:4321/_emdash/api/search?q=design"
401
```

**After (this branch):**
```
$ curl -o /dev/null -w "%{http_code}\n" "http://localhost:4321/_emdash/api/search?q=design"
200

$ curl "http://localhost:4321/_emdash/api/search?q=design" | jq '.data.items | length'
4
```

The `<LiveSearch>` header widget in `demos/simple` / `templates/blog` now returns live results in the dropdown instead of "No results found" for any query.


csfalcao:

ascorbic (approved):
Thanks!

PR #426, opened 2026-04-10
feat: Node.js plugin isolation via workerd sandbox

BenjaminPrice:
## What does this PR do?

Adds workerd-based plugin sandboxing for Node.js deployments, closing EmDash's most significant architectural gap: plugins that run sandboxed on Cloudflare Workers now run sandboxed on Node.js too, with matching capability enforcement and isolation guarantees.

Discussion: https://github.com/emdash-cms/emdash/discussions/425

## How it works

A new `@emdash-cms/workerd` package implements the `SandboxRunner` interface (same contract as `@emdash-cms/cloudflare`). In production it spawns workerd as a child process with a generated capnp config. In development it uses miniflare for faster startup. Plugin workers run in V8 isolates and communicate with Node via an authenticated HTTP backing service.

**Architecture:**
- `WorkerdSandboxRunner` (production): spawns workerd, generates capnp config per plugin, manages lifecycle with exponential backoff restart, SIGTERM forwarding, intentional-stop guard, and epoch-based stale handle detection
- `MiniflareDevRunner` (dev): uses miniflare's `outboundService` to route bridge calls to Node handler functions; non-bridge fetches are blocked to prevent capability bypass
- `bridge-handler.ts`: shared bridge dispatch logic used by both runners, with capability enforcement matching the Cloudflare PluginBridge
- Backing service: HTTP server with per-plugin HMAC tokens, server-side capability validation, and constant-time auth comparison

**Security model matches Cloudflare:**
- Per-plugin auth tokens encoding capabilities, validated server-side
- Per-startup invoke token authenticates inbound hook/route calls (prevents same-host attackers from invoking the per-plugin TCP listener directly)
- Constant-time comparison everywhere (`timingSafeEqual` in Node, hand-rolled XOR in workerd where Web Crypto doesn't expose it)
- Network fetch uses core's shared `createHttpAccess` (redirect revalidation on each hop, credential stripping on cross-origin redirects, SSRF blocking, max 5 redirects)
- Strict capability enforcement: `write:content` does NOT imply `read:content` (matches Cloudflare bridge)
- Cross-plugin data isolation (KV and storage scoped per plugin ID)
- Direct `fetch()` from plugin code is blocked in dev (miniflare `outboundService` returns 403) and routed through the backing service in prod (capnp `globalOutbound`), forcing all network access through `ctx.http.fetch` and capability/host checks

**Honest about resource limits:** `cpuMs`, `memoryMb`, and `subrequests` are Cloudflare platform features, not standalone workerd. Only `wallTimeMs` is enforced (via `Promise.race`). Operators get a startup warning if they configure unenforced limits, and the docs include a caution box.

**Sandbox bypass mode (`sandbox: false`):** Debugging escape hatch that loads sandboxed plugin entries in-process via `adaptSandboxEntry` + data URL `import()`. Build-time and marketplace plugins both work under bypass; marketplace install/update routes skip the `SANDBOX_NOT_AVAILABLE` gate when bypassed and rebuild the hook pipeline so changes take effect immediately.

## Type of change

- [x] Feature (requires [approved Discussion](https://github.com/emdash-cms/emdash/discussions/categories/ideas))

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm --silent lint:json | jq '.diagnostics | length'` returns 0 errors
- [x] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/425

## Tests

**36 new tests** in `@emdash-cms/workerd` across 3 files, plus 2163 existing core tests passing:

- `miniflare-isolation.test.ts` (6 tests): V8 isolate isolation, service bindings, KV namespaces, dynamic code loading, worker reconfiguration
- `bridge-handler.test.ts` (19 tests): KV CRUD + isolation, strict capability enforcement (read/write content do not imply each other; `network:fetch:any` satisfies `network:fetch`), storage collection validation, error message parity with Cloudflare bridge
- `plugin-integration.test.ts` (11 tests): Real plugin operations modeled after the `sandboxed-test` plugin (KV round-trip, storage round-trip, content lifecycle with ULID/versioning/soft-delete, cross-plugin isolation, write-only plugins cannot read)

All tests use real in-memory SQLite with production schema, no mocking.

```
 ✓ test/bridge-handler.test.ts (19 tests)
 ✓ test/plugin-integration.test.ts (11 tests)
 ✓ test/miniflare-isolation.test.ts (6 tests)

 Test Files  3 passed (3)
      Tests  36 passed (36)
```

## Review iterations

This PR went through ~20 rounds of automated and human review. Notable hardening from review feedback:

- **Security:** invoke-token authentication on hook/route calls, constant-time comparisons (`timingSafeEqual` + hand-rolled for workerd), miniflare `outboundService` blocks non-bridge fetches, strict capability enforcement (`write:content` ≠ `read:content`)
- **Lifecycle:** intentional-stop flag suppresses crash recovery on planned reloads, deferred startup serialization via `startupPromise`, `isHealthy()` returns false when restart pending, `stopWorkerd()` fast-paths on already-exited and uses local `exited` flag for SIGKILL fallback
- **Binary handling:** `ctx.http.fetch` returns base64-encoded bytes preserving binary content, RequestInit marshaling preserves multi-value headers (`[[name, value]]` pairs), Blob/File/FormData/URLSearchParams bodies, ArrayBuffer with `byteOffset`/`byteLength` window
- **Storage parity (also fixed in Cloudflare bridge):** delegated `storageQuery`/`storageCount` to `PluginStorageRepository` so `where`/`orderBy`/`cursor`/`count` work correctly. Fixes infinite-loop pagination on shipped plugins like `forms-submissions`.
- **Media:** upload writes to configured `Storage` adapter, sets `status='ready'`, rolls back storage object on DB failure (best-effort cleanup); delete removes the storage object too
- **Workerd binary:** resolved from package `bin/workerd` (no `npx` runtime download), `execFileSync` for paths with spaces, stdout/stderr drained to prevent pipe deadlock
- **Marketplace bypass:** `loadMarketplacePluginsBypassed` runs before pipeline creation on cold start; `syncMarketplacePluginsBypassed` handles runtime install/update/uninstall and rebuilds the hook pipeline

### New: `packages/workerd/`

| File | Purpose |
|---|---|
| `src/sandbox/runner.ts` | `WorkerdSandboxRunner` implementing `SandboxRunner`, child process lifecycle, auth token generation, intentional-stop guard, epoch-based stale handle detection |
| `src/sandbox/dev-runner.ts` | `MiniflareDevRunner` for development, auto-selected when `NODE_ENV === "development"` |
| `src/sandbox/bridge-handler.ts` | Shared bridge dispatch with capability enforcement, matching Cloudflare PluginBridge parity, delegating storage to `PluginStorageRepository` |
| `src/sandbox/backing-service.ts` | HTTP server wrapper adding auth token validation around bridge handler |
| `src/sandbox/capnp.ts` | workerd capnp config generator from plugin manifests, `globalOutbound` routes through backing service |
| `src/sandbox/wrapper.ts` | Plugin wrapper code generator (runs inside workerd), constant-time invoke-token check, RequestInit marshaling, base64 binary preservation |

### Modified: `packages/core/`

| File | Change |
|---|---|
| `plugins/sandbox/types.ts` | Added `isHealthy()` to `SandboxRunner` interface, `SandboxUnavailableError` class, `mediaStorage` in `SandboxOptions` |
| `plugins/sandbox/noop.ts` | Implements `isHealthy()`, error message mentions both CF and workerd |
| `astro/integration/runtime.ts` | Added `sandbox?: boolean` config option (escape hatch) |
| `astro/integration/virtual-modules.ts` | Generates `sandboxBypassed` flag when `sandbox: false` |
| `astro/middleware.ts` | Reads `sandboxBypassed` via namespace import (handles missing export) |
| `emdash-runtime.ts` | `loadBypassedPlugins` + `loadMarketplacePluginsBypassed`, runtime sync, `isSandboxBypassed()`, threads `mediaStorage` into sandbox runners |
| `api/handlers/marketplace.ts` | `handleMarketplaceInstall`/`handleMarketplaceUpdate` accept `sandboxBypassed`, skip `SANDBOX_NOT_AVAILABLE` gate when set |
| `index.ts` | Exports `SandboxUnavailableError`, `createHttpAccess`, `createUnrestrictedHttpAccess`, repositories used by platform adapters |

### Modified: `packages/cloudflare/`

| File | Change |
|---|---|
| `sandbox/runner.ts` | Implements `isHealthy()` (delegates to `isAvailable()`), passes `storageConfig` to bridge bindings |
| `sandbox/bridge.ts` | Storage operations delegate to `PluginStorageRepository` (parity fix) |

### Docs

| File | Change |
|---|---|
| `plugins/sandbox.mdx` | Added Node.js + workerd setup guide, 3-column security comparison, escape hatch docs, honest resource-limit caution box |
| `plugins/creating-plugins.mdx` | Added "Testing in the Sandbox" section with capability behavior table (methods throw, not "missing from ctx") |

### Changeset

`.changeset/bumpy-crabs-nail.md` — `emdash` minor, `@emdash-cms/cloudflare` patch, `@emdash-cms/workerd` minor.


packages/workerd/src/sandbox/dev-runner.ts
```diff
@@ -0,0 +1,217 @@
...
+				modules: true,
+				script: wrapperCode,
+				outboundService: async (request: Request) => {
+					const url = new URL(request.url);
+					if (url.hostname === "bridge") {
+						return bridgeHandler(request);
+					}
+					return globalThis.fetch(request);
+				},
+			});
```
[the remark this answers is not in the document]
BenjaminPrice: Fixed in commit 7b9f914 (and subsequent review-fix commits). `MiniflareDevRunner.rebuild()` now passes plugin code as a named module: the worker config includes `modules: [{ path: "worker.js", contents: wrapperCode }, { path: "sandbox-plugin.js", contents: code }]`, so the wrapper's `import pluginModule from "sandbox-plugin.js"` resolves correctly.

packages/workerd/src/sandbox/runner.ts
```diff
@@ -0,0 +1,630 @@
...
+	 */
+	isAvailable(): boolean {
+		try {
+			// Check if workerd binary exists
+			const { execSync } = require("node:child_process") as typeof import("node:child_process");
+			execSync("npx workerd --version", { stdio: "ignore", timeout: 5000 });
+			return true;
+		} catch {
+			return false;
+		}
```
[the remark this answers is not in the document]
BenjaminPrice: Fixed in commit a846039. All `require()` calls in `runner.ts` were replaced with proper ESM imports — `createHmac` from `node:crypto`, `execFileSync` and `spawn` from `node:child_process`, and `createRequire(import.meta.url)` (from `node:module`) for resolving the workerd binary.

packages/workerd/src/sandbox/runner.ts
```diff
@@ -0,0 +1,630 @@
...
+		// Handle workerd exit with auto-restart on crash
+		this.workerdProcess.on("exit", (code) => {
+			this.healthy = false;
+			if (this.shuttingDown) return;
+			if (code !== 0 && code !== null) {
+				console.error(`[emdash:workerd] workerd exited with code ${code}`);
+				this.scheduleRestart();
+			}
+		});
+
```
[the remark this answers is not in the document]
BenjaminPrice: Fixed in commit a846039. Stdout/stderr are now drained — `proc.stdout?.pipe(process.stdout)` and `proc.stderr?.pipe(process.stderr)` so log output never fills the pipe buffer and deadlocks the child.

packages/workerd/src/sandbox/runner.ts
```diff
@@ -0,0 +1,630 @@
...
+	/**
+	 * Check if workerd is available on this system.
+	 */
+	isAvailable(): boolean {
+		try {
+			// Check if workerd binary exists
+			const { execSync } = require("node:child_process") as typeof import("node:child_process");
+			execSync("npx workerd --version", { stdio: "ignore", timeout: 5000 });
+			return true;
```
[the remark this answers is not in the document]
BenjaminPrice: Fixed in commit a846039. `isAvailable()` and `restart()` no longer use `npx`. The workerd binary is resolved from the `workerd` package's installed `bin/workerd` via `createRequire(import.meta.url).resolve("workerd")` and `execFileSync(bin, ["--version"], ...)` (no shell, no runtime download).

packages/workerd/src/sandbox/runner.ts
```diff
@@ -0,0 +1,630 @@
...
+/**
+ * Default resource limits for sandboxed plugins.
+ * Matches Cloudflare production limits.
+ */
+const DEFAULT_LIMITS = {
+	cpuMs: 50,
+	memoryMb: 128,
+	subrequests: 10,
+	wallTimeMs: 30_000,
+} as const;
```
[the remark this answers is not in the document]
BenjaminPrice: Addressed honestly in commit a846039. Resource limits other than `wallTimeMs` are not actually enforceable in standalone workerd (`cpuMs`/`memoryMb`/`subrequests` are Cloudflare platform features). The `limits` field was removed from the capnp generator, the runner emits a startup warning if operators configure unenforced limits, and `docs/src/content/docs/plugins/sandbox.mdx` now has a caution box explaining that only `wallTimeMs` is enforced via `Promise.race`. The PR description's "How it works" section calls this out under "Honest about resource limits".

packages/workerd/src/sandbox/bridge-handler.ts
```diff
@@ -0,0 +1,988 @@
...
+	if (!opts.capabilities.includes(capability)) {
+		throw new Error(`Plugin ${opts.pluginId} does not have capability: ${capability}`);
+	}
+}
+
+function validateStorageCollection(opts: BridgeHandlerOptions, collection: string): void {
+	if (!opts.storageCollections.includes(collection)) {
+		throw new Error(`Plugin ${opts.pluginId} does not declare storage collection: ${collection}`);
+	}
+}
```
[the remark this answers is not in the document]
BenjaminPrice: Fixed in commit a846039. Error messages now match the Cloudflare PluginBridge format exactly: `Missing capability: {cap}` and `Storage collection not declared: {collection}`. There is a regression test in `bridge-handler.test.ts` asserting the exact error strings.

packages/workerd/src/sandbox/bridge-handler.ts
```diff
@@ -0,0 +1,988 @@
...
+			return mediaList(db, body);
+		case "media/delete":
+			requireCapability(opts, "write:media");
+			return mediaDelete(db, requireString(body, "id"));
+
+		// ── HTTP ────────────────────────────────────────────────────────
+		case "http/fetch":
+			requireCapability(opts, "network:fetch");
+			return httpFetch(requireString(body, "url"), body.init as RequestInit | undefined, opts);
+
```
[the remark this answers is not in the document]
BenjaminPrice: Fixed in commit a846039. `requireCapability` now treats `network:fetch:any` as satisfying `network:fetch` (a plugin with the broader capability is allowed through the narrower check). The reverse is not true. Strict capability semantics also enforced: `write:content` does NOT imply `read:content`.

packages/workerd/src/sandbox/wrapper.ts
```diff
@@ -0,0 +1,246 @@
...
+		delete: (collection, id) => bridgeCall("content/delete", { collection, id }),
+	};
+
+	const media = {
+		get: (id) => bridgeCall("media/get", { id }),
+		list: (opts) => bridgeCall("media/list", opts || {}),
+		upload: (filename, contentType, bytes) => bridgeCall("media/upload", { filename, contentType, bytes: Array.from(bytes) }),
+		getUploadUrl: () => { throw new Error("getUploadUrl is not available in sandbox mode. Use media.upload() instead."); },
+		delete: (id) => bridgeCall("media/delete", { id }),
+	};
```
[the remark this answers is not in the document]
BenjaminPrice: Fixed in commit a846039. `dispatch()` now has a `media/upload` case (`mediaUpload`) that writes the bytes to the configured `Storage` adapter, inserts the media row with `status='ready'`, and rolls back the storage object on DB-insert failure. The wrapper's `media.upload()` correctly preserves `byteOffset`/`byteLength` for ArrayBuffer/TypedArray/DataView inputs so non-Uint8Array binary data is not corrupted.

packages/workerd/src/sandbox/capnp.ts
```diff
@@ -0,0 +1,97 @@
...
+		lines.push(`  modules = [`);
+		lines.push(`    (name = "worker.js", esModule = embed "${wrapperFile}"),`);
+		lines.push(`    (name = "sandbox-plugin.js", esModule = embed "${pluginFile}"),`);
+		lines.push(`  ],`);
+		lines.push(`  compatibilityDate = "2025-01-01",`);
+		lines.push(`  compatibilityFlags = ["nodejs_compat"],`);
+		// globalOutbound allows the plugin wrapper to fetch() the backing service
+		// The wrapper code uses absolute URLs to the backing service
+		lines.push(`);`);
+		lines.push(``);
```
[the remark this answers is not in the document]
BenjaminPrice: Fixed in commit a846039. The capnp generator now sets `globalOutbound = "emdash-backing"` so any outbound `fetch()` from a plugin worker is routed through the backing service, where capability + host validation runs. A code comment documents that direct `fetch()` (not via `ctx.http.fetch`) returns 500 "Unknown bridge method" by design — this forces all network access through the capability-checked path.

packages/workerd/package.json
```diff
@@ -0,0 +1,59 @@
...
+	},
+	"devDependencies": {
+		"@types/better-sqlite3": "catalog:",
+		"better-sqlite3": "catalog:",
+		"kysely": "^0.27.0",
+		"miniflare": "^4.20250408.0",
+		"tsdown": "catalog:",
+		"typescript": "catalog:",
+		"vitest": "catalog:"
+	},
```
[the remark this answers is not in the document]
BenjaminPrice: Fixed in commit a846039. `MiniflareDevRunner` is statically imported by the package entry, but its `rebuild()` method only `await import("miniflare")` lazily — so production installs that don't use the dev runner never need the dependency. For users who do want hot reload, miniflare is documented as a peer/optional dependency in the docs. Production users only need the `workerd` binary, which is declared as a peerDependency.

docs/src/content/docs/plugins/creating-plugins.mdx
```diff
@@ -380,6 +380,63 @@ Test plugins by creating a minimal Astro site with the plugin registered:
...
+
+| What | Trusted mode | Sandboxed mode |
+|---|---|---|
+| **Undeclared capabilities** | `ctx.content`, `ctx.media`, etc. are always present | Missing from `ctx` if not declared in `capabilities` |
+| **Network access** | `fetch()` works globally | Only via `ctx.http.fetch()`, restricted to `allowedHosts` |
+| **Node.js builtins** | `fs`, `path`, `child_process` available | Not available (V8 isolate, no Node APIs) |
+| **Environment variables** | `process.env` accessible | Not accessible |
+| **CPU time** | Unbounded | Limited (default 50ms per invocation) |
+| **Wall-clock time** | Unbounded | Limited (default 30s per invocation) |
+| **Direct DB access** | Possible (but discouraged) | Not possible, all access via `ctx.*` |
```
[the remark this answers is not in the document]
BenjaminPrice: Fixed in commit a846039 along with the docs update. The capability behavior table in `creating-plugins.mdx` now correctly states that `ctx.content`, `ctx.media`, etc. are present on `ctx` regardless of capabilities — calling them without the capability throws an error (`Missing capability: read:content`). The `users` and `email` namespaces remain `undefined` when not declared (matching the wrapper code).

docs/src/content/docs/plugins/sandbox.mdx
```diff
@@ -145,34 +145,80 @@ Sandboxing requires Dynamic Worker Loader. Add to your `wrangler.jsonc`:
...
+       }),
+     ],
+   });
+   ```
+
+3. Restart your dev server. Sandboxed plugins will now run in workerd isolates.
+
+</Steps>
+
+In development, the runner uses miniflare (bundled with workerd) for faster startup. In production (`NODE_ENV=production`), it spawns workerd as a child process with a generated configuration.
```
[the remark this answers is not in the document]
BenjaminPrice: Fixed in the docs cleanup. `sandbox.mdx` no longer says miniflare is "bundled with workerd". It correctly describes miniflare as an optional dev-only dependency for hot reload, and `workerd` as the production peer dependency.

packages/workerd/src/sandbox/runner.ts
```diff
@@ -0,0 +1,630 @@
...
+
+		const payload = Buffer.from(payloadB64, "base64url").toString();
+		const { createHmac } = require("node:crypto") as typeof import("node:crypto");
+		const expectedHmac = createHmac("sha256", this.masterSecret)
+			.update(payload)
+			.digest("base64url");
+
+		if (hmacB64 !== expectedHmac) return null;
+
+		try {
```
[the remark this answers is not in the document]
BenjaminPrice: Fixed in commit a846039. Auth token validation in `backing-service.ts` now uses `timingSafeEqual` on Buffers (with a length-equality precheck to avoid throwing on mismatched lengths). The workerd-side invoke token check uses a hand-rolled constant-time XOR comparison since workerd doesn't expose `timingSafeEqual`.

packages/workerd/src/sandbox/runner.ts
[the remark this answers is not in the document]
BenjaminPrice: Fixed in commit a846039. The exit handler now clears `this.workerdProcess = null` and increments the epoch so stale handles fail fast with `SandboxUnavailableError`. `stopWorkerd()` short-circuits if `proc.exitCode !== null`, and `intentionalStop` is set before calling `stopWorkerd()` so the exit handler skips the crash-recovery branch on planned reloads (preventing restart-loop cascades on plugin install/uninstall).

packages/workerd/src/sandbox/runner.ts
```diff
@@ -0,0 +1,630 @@
...
+
+		return new Promise((resolve) => {
+			proc.on("exit", () => resolve());
+			proc.kill("SIGTERM");
+			// Force kill after 5 seconds
+			setTimeout(() => {
+				if (!proc.killed) proc.kill("SIGKILL");
+			}, 5000);
+		});
+	}
```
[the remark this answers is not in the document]
BenjaminPrice: Fixed in commit a846039. `stopWorkerd()` now fast-paths on `proc.exitCode !== null` (resolves immediately if already exited). The SIGKILL fallback uses a local `exited` flag set inside the exit listener, not `proc.killed` (which flips on signal queue, not actual exit), so the timeout reliably escalates from SIGTERM to SIGKILL.

packages/workerd/src/sandbox/runner.ts
```diff
@@ -0,0 +1,630 @@
...
+
+		// Handle workerd exit with auto-restart on crash
+		this.workerdProcess.on("exit", (code) => {
+			this.healthy = false;
+			if (this.shuttingDown) return;
+			if (code !== 0 && code !== null) {
+				console.error(`[emdash:workerd] workerd exited with code ${code}`);
+				this.scheduleRestart();
+			}
+		});
```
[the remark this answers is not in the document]
BenjaminPrice: Fixed in commit a846039. The crash-restart guard now triggers on signal-based termination (`code === null && signal !== null`) too, so OOM kills and other signal-based exits get the exponential backoff restart treatment. The `intentionalStop` flag suppresses recovery for planned reloads (SIGTERM from `terminateAll`/`unloadPlugin`).

packages/workerd/src/sandbox/bridge-handler.ts
```diff
@@ -0,0 +1,1115 @@
...
+}> {
+	validateCollectionName(collection);
+
+	const now = new Date().toISOString();
+
+	// Build update: always bump updated_at and version
+	let query = db
+		.updateTable(`ec_${collection}` as keyof Database)
+		.set({ updated_at: now } as never)
+		.set(sql`version = version + 1` as never)
```
ascorbic:
```suggestion
		.set({version: sql`version + 1` as never})
```

packages/workerd/src/sandbox/wrapper.ts
```diff
@@ -0,0 +1,414 @@
...
+				out.body = viewToBase64(toBytes(init.body));
+			} else if (typeof Blob !== "undefined" && init.body instanceof Blob) {
+				// Blob/File (without going through FormData): read bytes and
+				// preserve content type if not already set
+				const bytes = new Uint8Array(await init.body.arrayBuffer());
+				out.bodyType = "base64";
+				out.body = viewToBase64(bytes);
+				if (init.body.type) {
+					out.headers = out.headers || {};
+					if (!out.headers["content-type"] && !out.headers["Content-Type"]) {
```
ascorbic: `out.headers` is set to an array inside the init.headers block above, so this won't work.

packages/workerd/src/sandbox/wrapper.ts
```diff
@@ -0,0 +1,414 @@
...
+				view = new Uint8Array(bytes);
+			} else if (ArrayBuffer.isView(bytes)) {
+				view = new Uint8Array(bytes.buffer, bytes.byteOffset, bytes.byteLength);
+			} else {
+				throw new TypeError("media.upload: bytes must be ArrayBuffer or ArrayBufferView");
+			}
+			return bridgeCall("media/upload", {
+				filename,
+				contentType,
+				bytes: Array.from(view),
```
ascorbic: This will make a lot of JSON. Either use base64 instead, or (probably better) move uploads out of the JSON bridge entirely.

packages/workerd/src/sandbox/runner.ts
```diff
@@ -0,0 +1,831 @@
...
+	 * Removes the plugin from the in-memory map and marks needsRestart so
+	 * the next invocation rebuilds workerd without it. We don't restart
+	 * eagerly here because update/uninstall flows often unload immediately
+	 * before loading the new version, and back-to-back restarts are wasteful.
+	 */
+	unloadPlugin(pluginId: string): void {
+		if (this.plugins.delete(pluginId)) {
+			// Mark for restart so the next load() or invocation regenerates
+			// the capnp config without this plugin's port/listener.
+			this.needsRestart = true;
```
ascorbic, line 498: Probably better toi eagerly restart here so there isn't the delay duirng the next request

packages/core/src/emdash-runtime.ts
```diff
@@ -665,6 +844,41 @@ export class EmDashRuntime {
 			console.warn("[comments] Failed to register default moderator:", error);
 		}
 
+		// sandbox: false escape hatch - load sandboxed plugin entries in-process
+		// as trusted plugins (no isolation) so they participate in the hook pipeline.
+		if (deps.sandboxBypassed && deps.sandboxedPluginEntries.length > 0) {
+			console.info(
```
ascorbic, line 997: This would mean there's no warning on first run if plugins haven't been installed yet. There should be a loud warning whether or not there are plugins already installed

packages/core/src/emdash-runtime.ts
```diff
@@ -554,6 +585,149 @@ export class EmDashRuntime {
...
+				const existing = this.allPipelinePlugins.find((p) => p.id === pluginId);
+				if (existing && existing.version === bundle.manifest.version) continue;
+
+				// Remove any older version
+				if (existing) {
+					this.removePluginFromLists(pluginId);
+				}
+
+				try {
+					const dataUrl = `data:text/javascript;base64,${Buffer.from(bundle.backendCode).toString("base64")}`;
```
ascorbic, line 799: I'm pretty sure this won't work in workerd, so sandbox: false in Cloudflare will fail in a confusing way. It should probably not be supported anyway tbh, so we should detect than and fail in a more useful way

packages/core/src/emdash-runtime.ts
```diff
@@ -554,6 +585,149 @@ export class EmDashRuntime {
...
+				if (state.status !== "active") continue;
+				desired.set(state.pluginId, state.marketplaceVersion ?? state.version);
+			}
+
+			// Drop metadata for plugins no longer active.
+			const toRemove: string[] = [];
+			for (const pluginId of marketplaceManifestCache.keys()) {
+				if (!desired.has(pluginId)) toRemove.push(pluginId);
+			}
+			for (const pluginId of toRemove) {
```
ascorbic, line 739: This doesn't fire `plugin:deactivate`

packages/workerd/src/sandbox/backing-service.ts
```diff
@@ -0,0 +1,98 @@
...
+import type { WorkerdSandboxRunner } from "./runner.js";
+
+/**
+ * Create an HTTP request handler for the backing service.
+ */
+export function createBackingServiceHandler(
+	runner: WorkerdSandboxRunner,
+): (req: IncomingMessage, res: ServerResponse) => void {
+	// Cache bridge handlers per plugin token to avoid re-creation
+	const handlerCache = new Map<string, (request: Request) => Promise<Response>>();
```
ascorbic, line 32: this sticks around forever, even if a plugin is removed

packages/workerd/src/sandbox/backing-service.ts
```diff
@@ -0,0 +1,98 @@
...
+			res.end(responseBody);
+		} catch (error) {
+			const message = error instanceof Error ? error.message : "Internal error";
+			res.writeHead(500, { "Content-Type": "application/json" });
+			res.end(JSON.stringify({ error: message }));
+		}
+	};
+}
+
+async function readBody(req: IncomingMessage): Promise<Record<string, unknown>> {
```
ascorbic, line 108: This needs some size limits

packages/workerd/src/sandbox/bridge-handler.ts
```diff
@@ -0,0 +1,1115 @@
...
+	opts: BridgeHandlerOptions,
+	collection: string,
+	queryOpts: Record<string, unknown>,
+): Promise<{ items: Array<{ id: string; data: unknown }>; hasMore: boolean; cursor?: string }> {
+	const repo = getStorageRepo(opts, collection);
+	// eslint-disable-next-line typescript-eslint/no-unsafe-type-assertion -- WhereClause is structurally Record<string, unknown>
+	const result = await repo.query({
+		where: queryOpts.where as never,
+		orderBy: queryOpts.orderBy as Record<string, "asc" | "desc"> | undefined,
+		limit: typeof queryOpts.limit === "number" ? queryOpts.limit : undefined,
```
ascorbic: We should probably limit this and other similar queries like list, getMany. I think this nees to be changed in the cloudflare sandbox too

packages/workerd/src/sandbox/runner.ts
```diff
@@ -0,0 +1,831 @@
...
+		) {
+			console.warn(
+				"[emdash:workerd] cpuMs, memoryMb, and subrequests limits are not enforced " +
+					"by standalone workerd. Only wallTimeMs is enforced on the Node path. " +
+					"For full resource isolation, deploy on Cloudflare Workers.",
+			);
+		}
+
+		// Forward SIGTERM to workerd child for clean shutdown
+		this.sigHandler = () => {
```
ascorbic: this should be deduped - just one per process

packages/workerd/src/sandbox/runner.ts
```diff
@@ -0,0 +1,831 @@
...
+				// to get past the auth check first or we'd see 401.
+				const res = await fetch(`http://127.0.0.1:${firstPlugin.port}/__health`, {
+					signal: AbortSignal.timeout(1000),
+					headers: { Authorization: `Bearer ${this.invokeToken}` },
+				});
+				// Any response from the worker (404 from unknown route, or
+				// any 2xx) means workerd is up and serving requests. 401
+				// would mean the worker is up but rejecting our auth, which
+				// shouldn't happen since we're sending the right token.
+				if (res.status === 404 || res.ok) {
```
ascorbic: This feels a bit fragile. Can we use an expliclit ready endpoint?

packages/workerd/src/sandbox/bridge-handler.ts
```diff
@@ -0,0 +1,1115 @@
...
+
+/** Per-collection storage config (matches manifest.storage entries) */
+export interface BridgeStorageCollectionConfig {
+	indexes?: Array<string | string[]>;
+	uniqueIndexes?: Array<string | string[]>;
+}
+
+export interface BridgeHandlerOptions {
+	pluginId: string;
+	version: string;
```
ascorbic, line 56: Is this used anywhere?

packages/workerd/src/sandbox/dev-runner.ts
```diff
@@ -0,0 +1,275 @@
...
+	private manifest: PluginManifest;
+	private runner: MiniflareDevRunner;
+
+	constructor(id: string, manifest: PluginManifest, runner: MiniflareDevRunner) {
+		this.id = id;
+		this.manifest = manifest;
+		this.runner = runner;
+	}
+
+	async invokeHook(hookName: string, event: unknown): Promise<unknown> {
```
ascorbic, line 244: These should have `withWallTimeLimit` too

packages/workerd/src/sandbox/runner.ts
```diff
@@ -0,0 +1,831 @@
...
+ * deployments) — uses the production WorkerdSandboxRunner.
+ *
+ * The dev runner skips production hardening (wall-time wrapper, child
+ * process supervision, crash/restart with backoff), so falling back to
+ * it silently in production would be a security regression.
+ *
+ * Operators who want the dev runner explicitly should set NODE_ENV=development.
+ */
+export const createSandboxRunner: SandboxRunnerFactory = (options) => {
+	const isDev = process.env.NODE_ENV === "development";
```
ascorbic: I think there should be an explicit env var for this, rather than using `NODE_ENV`

packages/workerd/src/sandbox/runner.ts
```diff
@@ -0,0 +1,831 @@
...
+	 */
+	private resolveWorkerdBinary(): string {
+		try {
+			// workerd package: main is lib/main.js, bin is bin/workerd
+			const esmRequire = createRequire(import.meta.url);
+			const workerdMain = esmRequire.resolve("workerd");
+			// workerdMain = .../node_modules/workerd/lib/main.js
+			// binary = .../node_modules/workerd/bin/workerd
+			const pkgDir = join(workerdMain, "..", "..");
+			return join(pkgDir, "bin", "workerd");
```
ascorbic: Could we support Windows here?

packages/workerd/src/sandbox/runner.ts
```diff
@@ -0,0 +1,831 @@
...
+		this.runner = runner;
+	}
+
+	/**
+	 * Ensure workerd is running before invoking a hook or route.
+	 * On first call, this triggers deferred workerd startup (batching
+	 * all plugins registered via load() into a single workerd start).
+	 */
+	private async ensureReady(): Promise<void> {
+		await this.runner.ensureRunning();
```
ascorbic, line 906: I think it would be better to start workerd when Node is started. Otherwise there's a delay on first invocation

ascorbic (commented):
Amazing work on this! I've spent the morning going through this with Claude and it's neary there but I just have a few comments.

I think the fact there's now a bit more latency on the bridge means we should add content.createMany/updateMany/deleteMany to the API to avoid so many round-trips.

For media uploads I think it's best if we move the files out of the JSON bridge entirely, as it's not very efficient. Better to use somehting like the way it's done for public storage, where it gets a signed URL and then is uploaded directly. This could be handled  inside the wrapper with a binding so the API appears the same to the plugin itself.

One thing to consider too is defaulting to using Unix sockets instead of TCP (with fallback in Windows dev).

ascorbic:
Also, we definitely need some real workerd tests for this. Finally: it should be called sandbox-workerd not just workerd

ascorbic:
@BenjaminPrice it would be great to get this in! Will you be able to look at the comments, or do you want me to take it on instead?

BenjaminPrice:
I somehow missed your previous comment. I’ll get this done.

packages/workerd/src/sandbox/bridge-handler.ts
BenjaminPrice: Fixed. Changed to `.set({ version: sql\`version + 1\` } as never)` — the idiomatic Kysely object form.

packages/workerd/src/sandbox/wrapper.ts
BenjaminPrice: Fixed. The Blob branch now uses array-based header operations: `out.headers.some(([k]) => k.toLowerCase() === "content-type")` and `.push()` instead of object property access, matching the `[name, value]` pair format set at line 174.

packages/workerd/src/sandbox/wrapper.ts
BenjaminPrice: Fixed. Media upload bytes are now base64-encoded instead of `Array.from(view)`. The wrapper sends `{ bytes: btoa(binary), encoding: "base64" }` and the bridge handler decodes with `atob()`. This is ~3x smaller over the wire. Signed URL uploads (bypassing the JSON bridge entirely) are plumbed but deferred — the sandbox blocks direct `fetch()` via `globalOutbound`, so it needs networking changes to allow direct storage access.

packages/workerd/src/sandbox/runner.ts
BenjaminPrice: Fixed. `unloadPlugin()` now eagerly restarts workerd via `scheduleEagerStart()` (50ms debounce to batch rapid unload-then-load sequences during plugin upgrades). If no plugins remain after unload, it calls `stopWorkerd()` instead.

packages/core/src/emdash-runtime.ts
BenjaminPrice: Fixed. The sandbox availability check is now separated from the plugin count check. `sandboxRunner.isAvailable()` is evaluated and warns even when `sandboxedPluginEntries.length === 0`, so operators see the warning on first run before any plugins are installed.

packages/core/src/emdash-runtime.ts
BenjaminPrice: Fixed. Before attempting `await import(dataUrl)` in bypass mode, we now detect the Cloudflare Workers runtime via `navigator.userAgent === "Cloudflare-Workers"` and log a clear error: "sandbox: false is not supported in Cloudflare Workers. Dynamic import() of data URLs is not available in this runtime."

packages/core/src/emdash-runtime.ts
BenjaminPrice: Fixed. The `toRemove` loop in `syncMarketplacePluginsBypassed()` now calls `this._hooks.runPluginDeactivate(pluginId)` (with try/catch) before removing the plugin from caches and pipeline lists, giving plugins a chance to clean up.

packages/workerd/src/sandbox/backing-service.ts
BenjaminPrice: Fixed. `createBackingServiceHandler()` now returns a `BackingServiceHandler` with a `removePlugin(pluginId)` method. The cache is keyed by pluginId (not token), and `unloadPlugin()` calls `backingService.removePlugin()` to evict stale entries on plugin removal.

packages/workerd/src/sandbox/backing-service.ts
BenjaminPrice: Fixed. `readBody()` now enforces a 10MB limit (`MAX_BRIDGE_BODY_BYTES`). Each chunk increments a counter; exceeding the limit throws with `statusCode: 413`, which the error handler surfaces as a 413 response.

packages/workerd/src/sandbox/bridge-handler.ts
BenjaminPrice: Fixed. `storageQuery()` now caps the limit with `Math.min(queryOpts.limit, 100)`, matching the enforcement in content/media/users list queries. Also added `storageGetMany` — it accepts an array of IDs so there is an implicit limit, but batch operations (`contentCreateMany`/`updateMany`/`deleteMany`) are also capped at 100 items per call via `MAX_BATCH_SIZE`. The Cloudflare sandbox needs the same storageQuery fix — noted for a follow-up.

packages/workerd/src/sandbox/runner.ts
BenjaminPrice: Fixed. SIGTERM handling is now a module-level singleton: a `Set<WorkerdSandboxRunner>` tracks active instances, and a single `process.on("SIGTERM")` listener iterates them. Constructors call `registerSigHandler(this)`, `terminateAll()` calls `unregisterSigHandler(this)`.

packages/workerd/src/sandbox/runner.ts
BenjaminPrice: Fixed. Added an explicit `/__ready` endpoint to the wrapper template that returns `200 "ok"`. `waitForReady()` now polls `/__ready` and checks `res.ok` instead of accepting 404 from a nonexistent `/__health` path.

packages/workerd/src/sandbox/bridge-handler.ts
BenjaminPrice: Yes, it is used. `BridgeHandlerOptions` is the parameter type for `createBridgeHandler()` (bridge-handler.ts:74), and is consumed in `backing-service.ts` (line 52) and `dev-runner.ts` (line 144). It is the public API surface for creating a bridge handler.

packages/workerd/src/sandbox/dev-runner.ts
BenjaminPrice: Fixed. `MiniflareDevPlugin.invokeHook()` and `invokeRoute()` are now wrapped with `withWallTimeLimit()`. The limit comes from `this.runner.wallTimeMs` (new getter on `MiniflareDevRunner`, defaults to 30s). The implementation mirrors `WorkerdSandboxedPlugin.withWallTimeLimit` — `Promise.race` with a timeout that throws on expiry.

packages/workerd/src/sandbox/runner.ts
BenjaminPrice: Fixed. Added `EMDASH_SANDBOX_DEV` env var. The dev runner selection is now: `EMDASH_SANDBOX_DEV=1` (explicit) OR `NODE_ENV=development` (fallback for existing workflows).

packages/workerd/src/sandbox/runner.ts
BenjaminPrice: Fixed. `resolveWorkerdBinary()` now appends `.exe` on Windows: `const binaryName = process.platform === "win32" ? "workerd.exe" : "workerd"`. The rest of the transport uses TCP on `127.0.0.1` which is platform-agnostic. Unix socket support is plumbed but disabled pending capnp address format validation — when enabled, it falls back to TCP on Windows automatically.

packages/workerd/src/sandbox/runner.ts
BenjaminPrice: Fixed. `load()` now calls `scheduleEagerStart()` after registering the plugin, which fires a 50ms debounced `ensureRunning()`. The debounce batches sequential `load()` calls during init (the runtime loads plugins one by one) into a single workerd start. Same for `unloadPlugin()` — it eagerly restarts after removal so remaining plugins serve with the updated config immediately.

BenjaminPrice:
Both addressed:

**Real workerd tests:** Added `test/workerd-integration.test.ts` with 6 tests that spawn a real workerd process: route invocation, hook invocation, KV operations, plugin unload/reload, wall-time enforcement, and multiple simultaneous plugins. Tests skip gracefully via `describe.skipIf(!workerdAvailable)` when workerd is not installed. All 42 tests pass (36 existing + 6 new).

**Package rename:** Renamed from `@emdash-cms/workerd` to `@emdash-cms/sandbox-workerd` across package.json, all source imports, docs, and changesets.

Also addressed the review-level items:
- **Batch content APIs** (`content.createMany/updateMany/deleteMany`) added to reduce bridge round-trips, capped at 100 items per call.
- **Media uploads**: now base64-encoded instead of JSON number arrays (~3x smaller). Signed URL uploads are plumbed but deferred — the sandbox blocks direct `fetch()` via `globalOutbound`, so it needs transport changes to allow direct storage access.
- **Unix sockets**: plumbing is in place (backing server, capnp config, cleanup) but disabled pending validation of workerd's capnp external address format for `unix:` paths. TCP fallback works and is the current default.

BenjaminPrice:
Follow-up: Unix sockets are now fully enabled. The earlier disable was a red herring — the actual blocker was a capnp identifier naming bug (hyphens in const names like `plugin_debug-test_1_0_0` are invalid capnp). Once that was fixed with `toCapnpId()` (camelCase conversion), Unix sockets worked immediately. The backing service now uses Unix domain sockets by default on non-Windows, with TCP fallback on Windows. All 42 tests pass.

ascorbic:
@BenjaminPrice I've pushed a couple of fixes to issues that I ran into while testing, but it's working well for me now! Can you look at the review comments too.

ascorbic:
@BenjaminPrice this is really close! It would be great if we could address the last few points and get it out

BenjaminPrice:
> @BenjaminPrice this is really close! It would be great if we could address the last few points and get it out

Sorry, work has been quite busy. I'll try to get to this during this week.

packages/workerd/src/sandbox/wrapper.ts
```diff
@@ -0,0 +1,425 @@
...
+				}
+				out.bodyType = "formdata";
+				out.body = parts;
+			} else if (init.body instanceof URLSearchParams) {
+				out.bodyType = "string";
+				out.body = init.body.toString();
+				out.headers = out.headers || {};
+				if (!out.headers["content-type"] && !out.headers["Content-Type"]) {
+					out.headers["content-type"] = "application/x-www-form-urlencoded";
+				}
```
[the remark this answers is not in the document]
BenjaminPrice: Fixed in 153b421. The URLSearchParams branch of `marshalRequestInit` was assigning `out.headers["content-type"]` via bracket-on-array, so the property was invisible to `unmarshalRequestInit`'s pair-iteration and dropped by `JSON.stringify` on the wire. Switched to the same array-shape pattern as the Blob branch above. Regression test in `packages/workerd/test/wrapper-marshal.test.ts` (3 cases including a JSON round-trip check).

packages/workerd/src/sandbox/bridge-handler.ts
```diff
@@ -0,0 +1,1207 @@
...
+		type: string;
+		data: Record<string, unknown>;
+		createdAt: string;
+		updatedAt: string;
+	}>;
+	cursor?: string;
+	hasMore: boolean;
+}> {
+	validateCollectionName(collection);
+	const limit = Math.min(Number(opts.limit) || 50, 100);
```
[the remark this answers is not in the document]
BenjaminPrice: Fixed in 11d7b3f. `Math.min(Number(opts.limit) || 50, 100)` accepted negatives — `{ limit: -5 }` produced `query.limit(-4)`, which SQLite silently treats as zero rows and PostgreSQL rejects. Added the same `Math.max(1, ...)` floor used in `userList` to `contentList` (line 485) and `mediaList` (line 769); `storageQuery` (line 1157) keeps its undefined-passthrough but now clamps when a number is supplied. Regression coverage in `bridge-handler.test.ts` covers all three endpoints.

packages/workerd/src/sandbox/runner.ts
```diff
@@ -0,0 +1,899 @@
...
+	 * Wait for workerd to be ready by polling plugin ports.
+	 */
+	private async waitForReady(): Promise<void> {
+		const startTime = Date.now();
+		const timeout = 10_000;
+
+		while (Date.now() - startTime < timeout) {
+			try {
+				// Try to reach the first plugin
+				const firstPlugin = this.plugins.values().next().value;
```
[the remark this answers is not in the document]
BenjaminPrice: Fixed in 11ca6db. Extracted a `probeAllReady(plugins, invokeToken)` helper that fetches `/__ready` on every plugin in parallel and returns true only when every probe is ok. `waitForReady()` calls it each tick instead of only probing `plugins.values().next().value`. Regression coverage in `runner-ready-probe.test.ts` spins up two HTTP servers — one healthy, one returning 503 / unreachable — and verifies the helper returns `false` until both are healthy.

packages/workerd/src/sandbox/bridge-handler.ts
```diff
@@ -0,0 +1,1207 @@
...
+	}>
+> {
+	if (items.length > MAX_BATCH_SIZE) {
+		throw new Error(`Batch size ${items.length} exceeds maximum of ${MAX_BATCH_SIZE}`);
+	}
+	const results = [];
+	for (const data of items) {
+		results.push(await contentCreate(db, collection, data));
+	}
+	return results;
```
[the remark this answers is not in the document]
BenjaminPrice: Fixed in 3d5a8bd. `contentCreateMany`, `contentUpdateMany`, and `contentDeleteMany` now wrap their per-item loops in `db.transaction().execute(async (trx) => ...)`, so a mid-batch failure rolls the whole batch back. `MAX_BATCH_SIZE` guard stays outside the transaction. Regression test in `bridge-handler.test.ts` triggers a UNIQUE-constraint violation on item 3 of a 4-item create batch and asserts none of the rows landed.

packages/workerd/src/sandbox/runner.ts
```diff
@@ -0,0 +1,899 @@
...
+			process.stdout.write(`[emdash:workerd] ${chunk.toString()}`);
+		});
+		this.workerdProcess.stderr?.on("data", (chunk: Buffer) => {
+			process.stderr.write(`[emdash:workerd] ${chunk.toString()}`);
+		});
+
+		// Handle workerd exit with auto-restart on crash
+		this.workerdProcess.on("exit", (code, signal) => {
+			this.healthy = false;
+			this.workerdProcess = null;
```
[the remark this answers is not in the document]
BenjaminPrice: Fixed in 47779a3. Capture `proc` into a local immediately after `spawn()` and route the exit listener through a new `makeWorkerdExitHandler(host, proc)` helper, which short-circuits with `if (host.workerdProcess !== proc) return;` before touching shared state. A late exit from a prior workerd can no longer null out a newer one or clobber `healthy`. Helper is module-exported for unit testing — `runner-exit-handler.test.ts` has 6 tests including the stale-handler-is-a-no-op case.

packages/workerd/src/sandbox/runner.ts
```diff
@@ -0,0 +1,899 @@
...
+
+	/**
+	 * Schedule eager workerd start with a short debounce.
+	 * Batches rapid load/unload sequences into a single restart.
+	 */
+	private scheduleEagerStart(): void {
+		if (this.eagerStartTimer) clearTimeout(this.eagerStartTimer);
+		this.eagerStartTimer = setTimeout(() => {
+			this.eagerStartTimer = null;
+			void this.ensureRunning();
```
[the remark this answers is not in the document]
BenjaminPrice: Fixed in bec7033. `scheduleEagerStart` now attaches a real `.catch`: it logs `[emdash:workerd] eager start failed: <err>` and calls `this.scheduleRestart()`, which already handles exponential backoff, the 5-failure-window cap, and shutdown gating. So ENOENT-on-binary / capnp parse errors / waitForReady timeouts no longer surface as `unhandledRejection` and no longer leave the runner silently down. Regression test in `runner-eager-start.test.ts` stubs `ensureRunning` to reject, advances fake timers past the 50ms debounce, and asserts both the log and the restart call.

packages/workerd/src/sandbox/runner.ts
```diff
@@ -0,0 +1,899 @@
...
+			proc.kill("SIGTERM");
+			// Force kill after 5 seconds if SIGTERM was ignored.
+			// Use the local `exited` flag (not proc.killed, which flips
+			// to true as soon as a signal is queued, not when the process
+			// actually exits).
+			setTimeout(() => {
+				if (!exited) {
+					proc.kill("SIGKILL");
+				}
+			}, 5000);
```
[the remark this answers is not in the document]
BenjaminPrice: Fixed in 3d82462. Extracted the SIGTERM-then-SIGKILL dance into a `waitForProcessExit(proc, timeoutMs = 5000)` helper that captures the kill timer and `clearTimeout`s it inside the exit listener, so a clean exit no longer leaves a 5s pending timer holding the Node event loop alive past termination. The escalation check now also uses `proc.exitCode === null` rather than a local `exited` flag — that matches actual process state more precisely. Regression tests in `runner-stop-workerd.test.ts` assert `vi.getTimerCount() === 0` after a clean exit (load-bearing) and cover the SIGKILL-escalation path.

packages/workerd/src/sandbox/runner.ts
```diff
@@ -0,0 +1,899 @@
...
+		const pluginId = `${manifest.id}:${manifest.version}`;
+
+		// Return cached plugin if already loaded
+		const existing = this.plugins.get(pluginId);
+		if (existing) {
+			return new WorkerdSandboxedPlugin(pluginId, manifest, existing.port, this.limits, this);
+		}
+
+		// Assign port and generate auth token
+		const port = this.nextPluginPort++;
```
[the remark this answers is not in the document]
BenjaminPrice: Fixed in 064e90d. Added a `freePorts: number[]` stack alongside `nextPluginPort`. `load()` does `this.freePorts.pop() ?? this.nextPluginPort++`, and `unloadPlugin()` now reads the entry first (since `Map.delete` only returns a boolean), pushes the port back onto the stack, then deletes. No `EADDRINUSE` probe yet — that's a bigger change as you noted — but the leak you flagged is gone. `runner-port-reclaim.test.ts` verifies p3 reuses the port freed by p1, and that `nextPluginPort` doesn't climb across 10 load/unload cycles.

packages/workerd/src/sandbox/backing-service.ts
```diff
@@ -0,0 +1,129 @@
...
+					storageCollections: claims.storageCollections,
+					storageConfig: runner.getPluginStorageConfig(claims.pluginId, claims.version) as
+						| Record<string, { indexes?: Array<string | string[]> }>
+						| undefined,
+					db: runner.db,
+					emailSend: () => runner.emailSend,
+					storage: runner.mediaStorage,
+				});
+				handlerCache.set(cacheKey, bridgeHandler);
+				tokenToPluginId.set(token, cacheKey);
```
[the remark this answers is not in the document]
BenjaminPrice: Fixed in 9579288. You're right — the cache was keyed by `claims.pluginId` throughout, the map was only written and iterated-on-delete, never read. Deleted it entirely (and the iteration loop in `removePlugin`). `removePlugin` is now a one-liner that just calls `handlerCache.delete(pluginId)`.

packages/workerd/src/sandbox/runner.ts
```diff
@@ -0,0 +1,899 @@
...
+		});
+
+		const configPath = join(this.configDir, "workerd.capnp");
+		await writeFile(configPath, capnpConfig);
+
+		// Spawn workerd using resolved binary (not npx)
+		const workerdBin = this.resolveWorkerdBinary();
+		this.workerdProcess = spawn(workerdBin, ["serve", configPath], {
+			stdio: ["ignore", "pipe", "pipe"],
+			env: { ...process.env },
```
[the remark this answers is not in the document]
BenjaminPrice: Verified and fixed in 845d088. I ran an experimental probe plugin that tried to read `globalThis.process.env` from inside a real workerd worker (compatibility flag `nodejs_compat`), with the parent setting an obviously-unique env var. Result: with workerd 1.x the polyfill returns an empty `process.env` to plugins regardless of the child's actual environment — `Object.keys(process.env)` is `[]`, PATH/HOME/the probe var all came back `null`. So no active leak today.

That said, it's an undocumented implementation detail. Added a `minimalWorkerdEnv()` helper that allowlists only what workerd / its dependencies might need to start (`PATH`, `HOME`, `TMPDIR`/`TMP`/`TEMP`, `LANG`, `LC_ALL`) with an `EMDASH_WORKERD_PASSTHROUGH_ENV` comma-separated escape hatch for operators who need to extend. `workerd-integration.test.ts` continues to spawn a real workerd under the new minimal env and all six integration tests pass.

ascorbic (approved):
Amazing work! Great to get this in

neuhaus:
Great work guys! Someone still has to press the merge button I guess?

PR #453, opened 2026-04-11
fix(plugins): route ctx.content seo writes to the SEO panel

all3f0r1:
## What does this PR do?

Fixes `ctx.content.create()` / `ctx.content.update()` so plugins can write to the core SEO panel. Previously, passing a `seo` key in the input failed with `SQLite error: no such column: seo` because SEO fields live in the separate `_emdash_seo` table, not on the content row. Plugins could not programmatically populate the SEO panel, which blocked WordPress/Yoast importers, AI-assisted SEO tools, and content automation plugins.

The plugin content adapter now mirrors the REST API behavior:

- `createContentAccessWithWrite.create` / `update` extract the reserved `seo` key from the input and route it to `SeoRepository.upsert`, running content + SEO writes inside a single `withTransaction` so either both succeed or neither does.
- Writing `seo` to a collection without `has_seo = 1` throws a validation error matching the REST API (`Collection "<x>" does not have SEO enabled`).
- `createContentAccess.get` / `list` hydrate `ContentItem.seo` for SEO-enabled collections, so plugins can read current SEO state before doing non-destructive updates (also requested in the issue).
- Partial SEO updates preserve existing fields (via the existing `SeoRepository.upsert` `COALESCE` logic).
- A small `SeoRepository.isEnabled(collection)` helper encapsulates the `has_seo` lookup used by both the plugin context and the new tests.
- `ContentItem`, `ContentItemSeo`, `ContentItemSeoInput`, and `ContentWriteInput` are exported from `plugins/types.ts` so plugin authors get full typings.

Closes #374

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm --silent lint:json | jq '.diagnostics | length'` returns 0 (for files touched by this PR)
- [x] `pnpm test` passes (2257 tests; 8 new regression tests for #374)
- [ ] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/... _(n/a — bug fix)_

Implemented with Claude Code (Opus 4.6). A human reviewed every change, ran the test suite, and wrote the regression tests against the exact repro from the issue. Reviewers may want to pay extra attention to: the transaction boundary in `createContentAccessWithWrite.update` (the seo-only path intentionally skips `ContentRepository.update` so it does not bump `updated_at`/`version` when only SEO changed — open to flipping this if the project prefers always bumping on any write), and the decision to reserve `seo` as a magic key inside the flat `data` map rather than introducing a new envelope shape.

## Screenshots / test output

```
 ✓ tests/integration/plugins/capabilities.test.ts (60 tests) 2081ms

 Test Files  1 passed (1)
      Tests  60 passed (60)
```

New tests added in `packages/core/tests/integration/plugins/capabilities.test.ts` under `Content Access > SEO panel integration`:

- `get()` returns SEO defaults for SEO-enabled collections
- `get()` omits seo for non-SEO collections
- `update()` routes `seo` to the SEO panel (direct regression test for #374)
- `update()` accepts field updates alongside seo in one call
- `update()` partial seo updates do not clobber existing fields
- `create()` routes `seo` to the SEO panel
- `update()` throws when seo is provided on a non-SEO collection
- `list()` hydrates seo for each item in SEO-enabled collections

### DIFF
--- a/.changeset/fix-plugin-content-seo.md
+++ b/.changeset/fix-plugin-content-seo.md
@@ -1,0 +1,9 @@
+---
+"emdash": patch
+---
+
+Fixes `ctx.content.create()` and `ctx.content.update()` so plugins can write
+to the core SEO panel. When the input `data` contains a reserved `seo` key,
+it is now extracted and routed to `_emdash_seo` via the SEO repository,
+matching the REST API shape. `ctx.content.get()` and `ctx.content.list()`
+also hydrate the `seo` field on returned items for SEO-enabled collections.
--- a/packages/core/src/database/repositories/seo.ts
+++ b/packages/core/src/database/repositories/seo.ts
@@ -37,6 +37,19 @@ export class SeoRepository {
 	constructor(private db: Kysely<Database>) {}
 
 	/**
+	 * Check whether a collection has SEO enabled (`has_seo = 1`).
+	 * Returns `false` if the collection does not exist.
+	 */
+	async isEnabled(collection: string): Promise<boolean> {
+		const row = await this.db
+			.selectFrom("_emdash_collections")
+			.select("has_seo")
+			.where("slug", "=", collection)
+			.executeTakeFirst();
+		return row?.has_seo === 1;
+	}
+
+	/**
 	 * Get SEO data for a content item. Returns null defaults if no row exists.
 	 */
 	async get(collection: string, contentId: string): Promise<ContentSeo> {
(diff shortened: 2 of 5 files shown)


ascorbic (approved):
This is great! Thanks

PR #457, opened 2026-04-11
feat(storage): resolve S3 config from S3_* env vars at runtime

UpperM:
## What does this PR do?

`astro.config.mjs` runs at Astro build time, so values passed to `s3({...})` get baked into the virtual `emdash/config` module. That blocks the "build the image once, inject credentials at boot" pattern.

This PR resolves any field omitted from `s3({...})` from the matching `S3_*` env var at runtime (Node only). Existing deployments passing explicit values are unaffected,  explicit always wins. `s3()` with no args is now valid for fully env-driven configs.

Precedence: explicit `s3({...})` > `S3_*` env var > absent. Empty strings are treated as absent in both sources. Endpoint values are validated at both sources with the source named in the error (`S3_ENDPOINT is not a valid http/https URL` vs `s3({ endpoint }) is not a valid http/https URL`). 

Scope is Node; Workers binding resolution is out of scope and documented as such.

```js
// astro.config.mjs
storage: s3(),                                    // all fields from S3_* env
storage: s3({ publicUrl: "https://cdn.ex.com" }), // mix
storage: s3({ endpoint, bucket, accessKeyId, secretAccessKey }), // unchanged
```

Docs updated in `storage.mdx` / `configuration.mdx`, including a **prerequisite note** that users must install `@aws-sdk/client-s3` and `@aws-sdk/s3-request-presigner` themselves since core deliberately does not bundle the SDK (the ambient `packages/core/src/aws-sdk.d.ts` establishes the "bring-your-own-SDK" contract but was previously undocumented). 

26 new unit tests cover precedence, empty-string coercion, URL validation, credential pairing, and the Workers `typeof process` guard.

## Type of change

- [ ] Bug fix
- [x] Feature (requires [approved Discussion](https://github.com/emdash-cms/emdash/discussions/categories/ideas))
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm --silent lint:json`  0 diagnostics in files touched by this PR (3 pre-existing warnings in unrelated files left alone per scope rules)
- [x] `pnpm test` passes (2280 / 2280)
- [x] Tests added (26 in `tests/unit/storage/s3.test.ts`)
- [x] Changeset added (`.changeset/feat-s3-runtime-env-fallback.md`, `minor`)
- [x] New features link to an approved Discussion https://github.com/emdash-cms/emdash/discussions/459

### DIFF
--- a/packages/core/src/storage/s3.ts
+++ b/packages/core/src/storage/s3.ts
@@ -51,48 +51,40 @@ const s3ConfigSchema = z.object({
 	publicUrl: z.string().optional(),
 });
 
-function isRecord(value: unknown): value is Record<string, unknown> {
-	return typeof value === "object" && value !== null;
-}
-
 function isConfigKey(key: unknown): key is keyof S3StorageConfig {
 	return typeof key === "string" && key in ENV_KEYS;
 }
 
-/** Treat empty strings and undefined as absent so `S3_X=""` / `s3({ x: "" })` never mask config, and absent keys don't clobber env during merge. */
-function stripAbsent(raw: unknown): Record<string, unknown> {
-	if (!isRecord(raw)) return {};
-	const out: Record<string, unknown> = {};
-	for (const [key, value] of Object.entries(raw)) {
-		if (value === "" || value === undefined) continue;
-		out[key] = value;
-	}
-	return out;
-}
-
-function parseSource(raw: unknown, source: "env" | "explicit"): Partial<S3StorageConfig> {
-	const result = s3ConfigSchema.safeParse(stripAbsent(raw));
-	if (result.success) return result.data;
-	const issue = result.error.issues[0];
-	const pathKey = issue?.path[0];
-	if (!issue || !isConfigKey(pathKey)) fail("S3 config validation failed");
-	const label = source === "env" ? ENV_KEYS[pathKey] : `s3({ ${pathKey} })`;
-	fail(`${label} ${issue.message}`);
-}
-
-function readEnv(): Partial<S3StorageConfig> {
-	if (typeof process === "undefined" || !process.env) return {};
+/**
+ * Build the merged config: for each field, use the explicit value if present,
+ * otherwise fall back to the corresponding S3_* env var.  Validate once on the
+ * final merged result so a malformed env var never breaks the build when the
+ * caller provides that field explicitly.
+ */
+export function resolveS3Config(partial: Record<string, unknown>): S3StorageConfig {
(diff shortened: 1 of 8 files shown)


UpperM:
I stubbed` @aws-sdk/*` with `vi.mock` in the s3 unit tests to keep core dependency-free. Open to adding them as devDependencies instead if you'd prefer real-SDK tests. Any preference ?

packages/core/src/storage/s3.ts
```diff
@@ -28,6 +29,95 @@ import type {
...
+}
+
+function readEnv(): Partial<S3StorageConfig> {
+	if (typeof process === "undefined" || !process.env) return {};
+	const raw: Record<string, unknown> = {};
+	for (const [field, envKey] of Object.entries(ENV_KEYS)) {
+		raw[field] = process.env[envKey];
+	}
+	return parseSource(raw, "env");
+}
```
ascorbic: I don't think we should be parsing and validating all the env vars here. We should parse them individually, only if they're used. If somebody has an explicit value set in config, a malformed env var shouldn;t break the build. Instead of this eager parsing, I'd construct the config object manually, defaulting to any manually set values and falling back to the env var if not set, and only validating it after that.

packages/core/src/storage/s3.ts
```diff
@@ -49,16 +139,21 @@ export class S3Storage implements Storage {
 		this.publicUrl = config.publicUrl;
 		this.endpoint = config.endpoint;
 
+		// eslint-disable-next-line typescript-eslint(no-unsafe-type-assertion) -- SDK type declares credentials as required but accepts omission at runtime when the caller provides credentials through an alternative mechanism
```
ascorbic: Is this an error in the S3 client types?

packages/core/src/storage/s3.ts
UpperM: You're right.
Reworked: resolveS3Config now loops through each field, takes the explicit value if present, falls back to the env var otherwise, then validates once on the merged result. A malformed env var is never seen if the caller provides that field. 
Removed readEnv, parseSource, stripAbsent, it's one loop + one safeParse now.

packages/core/src/storage/s3.ts
UpperM: That eslint-disable comment was not needed, my bad.

ascorbic (approved):
Great stuff, thanks

PR #512, opened 2026-04-13
fix(visual-editor): guard publish against in-flight blur-save (#446)

### LINKED ISSUE #446: Save/publish race: blur-triggered saveField is fire-and-forget, publish promotes unchanged revision (EmDash 0.1.1)
## Summary

In EmDash 0.1.1 visual editing mode, inline-editing a `string` field and clicking the "Publish" toolbar button fast enough to race the blur-triggered `saveField` PUT causes the user's edit to be silently dropped. The user sees the old value on reload, and if there's a stale draft_revision_id it gets wiped as part of publish. No error surfaces — it looks like "publish reverted my change."

## Reproduction

1. Set up an EmDash 0.1.1 site with a `string`-type field on a collection that supports drafts + revisions (e.g., `pages.hero_heading`).
2. Render the field on a public page with `{...entry.edit.hero_heading}` spread so the element becomes inline-editable when the admin logs in.
3. Log in as an admin, click "Edit" in the EmDash toolbar (set `emdash-edit-mode` cookie).
4. Click the hero heading. It goes `contenteditable="plaintext-only"`.
5. Type a new value.
6. **Click the "Publish" button in the EmDash toolbar immediately** — don't click anywhere else first.

**Expected:** the edit is saved and published. The live page renders the new value on reload.

**Actual:** the live page reverts to the old value. If there was a prior draft, the `draft_revision_id` column gets cleared (because publish promotes and wipes it). The edit is lost.

## Root cause

The race is between `handleBlur` → `saveField()` (async, fire-and-forget) and the publish button's click handler firing `POST /_emdash/api/content/{collection}/{id}/publish`.

### 1. `saveField` is fire-and-forget in the overlay

`packages/core/src/visual-editing/toolbar.ts`, around the `handleBlur` function inside `startTextEdit`:

```js
function handleBlur() {
  // ...
  var newValue = (element.textContent || "").trim();
  if (newValue !== originalText.trim()) {
    saveField(annotation.collection, annotation.id, annotation.field, newValue);
    // ^ return value ignored — the save is fire-and-forget
  } else {
    setSaveState("idle");
  }
}
```

`saveField` returns a promise but `handleBlur` doesn't await it, and the publish button's click handler has no awareness of the save being in flight.

### 2. Publish handler fires immediately regardless

```js
function publish(collection, id) {
  publishBtn.disabled = true;
  publishBtn.textContent = "Publishing\u2026";
  ecFetch("/_emdash/api/content/" + encodeURIComponent(collection) + "/"
    + encodeURIComponent(id) + "/publish", { method: "POST", credentials: "same-origin" })
    .then(/* ... */);
}
```

No check for in-flight PUTs against the same entry.

### 3. Server-side publish promotes live revision when draft is null

`packages/core/src/content-repository.ts` (or wherever `ContentRepository.publish` lives):

```js
async publish(type, id) {
  const existing = await this.findById(type, id);
  // ...
  let revisionToPublish = existing.draftRevisionId || existing.liveRevisionId;
  // ...
  await sql`UPDATE ${tableName} SET live_revision_id = ${revisionToPublish},
             draft_revision_id = NULL, status = 'published', ... WHERE id = ${id}`;
}
```

If the PUT hasn't committed the draft yet, `existing.draftRevisionId` is null, publish promotes the unchanged `liveRevisionId` to itself, clears draft_revision_id, and returns success. The user's late-arriving PUT then creates a draft that immediately gets orphaned on the next publish — or the next page load if the user reloads first.

## Timing window

The race window is small but hit consistently during QA when testing "click-to-edit then publish" as a single motion. Our diagnostic logging captured a case where:

- `PUT /_emdash/api/content/pages/01KNS28Z7BYP8TK41WE7FCEQYC` — status 200, body `{"data":{"hero_subheading":"UPDATED SUB HEADING"}}`
- `POST /_emdash/api/content/pages/01KNS28Z7BYP8TK41WE7FCEQYC/publish` — status 200, fired ~100ms later

Both succeeded, but the resulting live revision had the OLD value because publish reached `ContentRepository.publish` before the D1 write from the PUT was visible. (Cloudflare D1 has its own propagation delay, which may make the race worse on Workers compared to local SQLite.)

### Option 1 — Client-side (simpler, works with any backend)

Make the publish handler await any in-flight `saveField` for the same entry before firing the POST:

```js
// In the overlay, track in-flight saves per entry
var inflightSaves = Object.create(null);

// In saveField:
function saveField(collection, id, field, value) {
  var key = collection + "/" + id;
  var p = ecFetch(/* ... PUT ... */).then(/* ... */);
  inflightSaves[key] = p;
  p.finally(function() {
    if (inflightSaves[key] === p) delete inflightSaves[key];
  });
  return p;
}

// In publish:
async function publish(collection, id) {
  var key = collection + "/" + id;
  if (inflightSaves[key]) {
    try { await inflightSaves[key]; } catch (_) {}
  }
  // ... existing publish logic ...
}
```

This fixes the race without touching the server. It's what we're using as a workaround (see below).

### Option 2 — Server-side (more robust, affects all clients)

In `ContentRepository.publish`, add a check: if the caller provided a specific `revisionId` OR if the most-recent revision in the revisions table is newer than the entry's `draft_revision_id` pointer, use that instead:

```js
async publish(type, id) {
  const existing = await this.findById(type, id);
  // ...
  let revisionToPublish = existing.draftRevisionId;
  if (!revisionToPublish) {
    // Check for a very recent revision that hasn't been linked yet
    const latest = await revisionRepo.findMostRecent(type, id);
    if (latest && latest.createdAt > existing.updatedAt - 2000 /* ms grace */) {
      revisionToPublish = latest.id;
    }
  }
  revisionToPublish = revisionToPublish || existing.liveRevisionId;
  // ...
}
```

This handles the case where the PUT wrote the revision but the entry row's `draft_revision_id` pointer hasn't propagated yet. The grace window is small enough to not affect normal usage.

### Option 3 — Disable publish button while saves in flight (UX belt-and-braces)

Regardless of Option 1 or 2, the toolbar should visually disable the Publish button while `setSaveState("unsaved")` or `setSaveState("saving")` is active, and only re-enable after `setSaveState("saved")`. This prevents users from triggering the race at all.

## Workaround (current)

A client-side `fetch` interceptor in our Astro `<body>` patches `window.fetch` to track in-flight PUTs per `<collection>/<id>` key and makes publish POSTs await the pending promise. Full source: [src/layouts/Base.astro lines 40–133 in our repo](https://github.com/johnp3004/medema-studios/blob/main/src/layouts/Base.astro) (condensed):

```js
var inflightSaves = Object.create(null);
var origFetch = window.fetch;

window.fetch = function (input, init) {
  var url = typeof input === "string" ? input : (input && input.url) || "";
  var method = (init && init.method) || "GET";
  var isContentPut = method === "PUT" && /\/_emdash\/api\/content\//.test(url);
  var isPublish = method === "POST" && /\/_emdash\/api\/content\/.*\/publish$/.test(url);

  var m = url.match(/\/_emdash\/api\/content\/([^\/?]+)\/([^\/?]+)/);
  var key = m ? m[1] + "/" + m[2] : null;

  if (isContentPut && key) {
    var p = origFetch.apply(this, arguments);
    inflightSaves[key] = p;
    p.then(function () { if (inflightSaves[key] === p) delete inflightSaves[key]; },
           function () { if (inflightSaves[key] === p) delete inflightSaves[key]; });
    return p;
  }

  if (isPublish && key) {
    var pending = inflightSaves[key];
    var fire = function () { return origFetch.apply(this, arguments); }.bind(this);
    return pending ? pending.then(fire, fire) : fire();
  }

  return origFetch.apply(this, arguments);
};
```

Verified working end-to-end against EmDash 0.1.1 on Cloudflare Workers + D1. Before: "UPDATED SUB HEADING" edit silently lost on publish. After: new live revision created with the edited value.

## Environment

- EmDash 0.1.1 (latest npm)
- Astro 6 (`output: "server"`, `@astrojs/cloudflare` adapter)
- Cloudflare Workers + D1 + R2
- Authenticated admin via passkey
- Chrome + Safari both reproduce

## Impact

**Severity: High** — this is silent data loss during the most common visual editing workflow (type a change, click Publish). Users have no indication the save didn't land. Only affects users who click Publish fast enough to race the blur-triggered PUT, but that's a natural workflow — "edit this text, publish it" is exactly what the inline editor is designed for.

## Related

- Bug #1 filed separately: "EmDash 0.1.0 — Visual editing overlay fails to detect field kinds (manifest response has extra data wrapper)". Both bugs affect visual editing and should be fixed together in the next patch release.

mahesh-projects:
handleBlur() was discarding the Promise returned by saveField(), leaving publish() with no way to know a save was in progress. publish() now chains onto pendingSavePromise when set, deferring the POST until the PUT commits. Closes #446.

## What does this PR do?

Fixes a save/publish race condition in the visual editor toolbar. Closes #446.

  When a user blurred a field and immediately clicked **Publish**, `handleBlur()` was discarding the `Promise` returned by `saveField()`. This left
  `publish()` with no way to know a save was in progress — so the publish POST could arrive at the server before the save PUT committed, silently
  promoting a stale revision.

  The fix introduces `pendingSavePromise`. When set, `publish()` chains onto it and defers the POST until the PUT resolves rather than firing
  immediately.

  > **Background:** This race was investigated in depth using [rp1 on EmDash — hunting a race condition you can't reproduce in a
  test](https://medium.com/@inbox4mahesh/rp1-on-emdash-part-3-hunting-a-race-condition-you-cant-reproduce-in-a-test-3493714d35f1), which traces how
  the root cause was identified through static analysis rather than a reproducible test.

  **Files changed:**
  - `packages/core/src/visual-editing/toolbar.ts` — add `pendingSavePromise` guard
  - `.changeset/fix-446-save-publish-race.md` — patch changeset

  ## Type of change

  - [x] Bug fix

  ## Checklist

  - [x] `pnpm typecheck` passes
  - [x] I have added a [changeset](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#changesets) (if this PR changes a published
  package)

  Generated with AI — [rp1](https://rp1.run/)

  Co-Authored-By: rp1 bot@rp1.run


mahesh-projects:

ascorbic (approved):
Thanks!

PR #540, opened 2026-04-13
feat(plugins): add where filter to ContentListOptions

jdevalk:
## What does this PR do?

Adds `where: { status?, locale? }` to the plugin-facing `ContentListOptions`, letting plugins narrow `ContentAccess.list()` results at the database layer instead of pulling every row and filtering in userland.

The underlying `ContentRepository.findMany` already accepts `where.status` / `where.locale` / `where.authorId` ([`packages/core/src/database/repositories/types.ts:66`](https://github.com/emdash-cms/emdash/blob/main/packages/core/src/database/repositories/types.ts#L66)) — this PR only plumbs the commonly-needed subset through the public API. `authorId` is left out for now since no current ask requires it; easy follow-up if someone needs it.

**Why:** today, a plugin that wants "published items only" has to pull every row (drafts, trashed, everything), filter in memory, and discard most of them. On a site with thousands of drafts this is wasteful and paginates wrong (the cursor counts pre-filter rows). A `where.status` filter pushes the narrowing into the SQL query, which is what the repository was designed to do.

Discussion: [#530](https://github.com/emdash-cms/emdash/discussions/530). Companion PR #539 handles the `locale` field addition to `ContentItem`.

## Type of change

- [x] Feature (additive — new optional field on an existing options type, no behavior change for consumers that don't pass it)

## Checklist

- [x] `pnpm format` run
- [x] Changeset added (`emdash` minor)
- [ ] Tests — following the precedent set by #536 (no new tests for the same plugin API surface). Happy to add one if reviewers prefer.

### DIFF
--- a/packages/core/src/plugins/context.ts
+++ b/packages/core/src/plugins/context.ts
@@ -235,6 +235,7 @@ export function createContentAccess(db: Kysely<Database>): ContentAccess {
 				limit: options?.limit ?? 50,
 				cursor: options?.cursor,
 				orderBy,
+				where: options?.where,
 			});
 
 			const items: ContentItem[] = result.items.map((item) => ({
--- a/packages/core/src/plugins/types.ts
+++ b/packages/core/src/plugins/types.ts
@@ -208,13 +208,6 @@ export interface ContentItem {
 	publishedAt: string | null;
 }
 
-/**
- * Filters applied at the database layer to `ContentAccess.list()`.
- *
- * Applied via SQL `WHERE` clauses, so they narrow the result set before
- * pagination — unlike filtering the returned array in userland, which
- * would still pay for pulling every row across the boundary.
- */
 export interface ContentListWhere {
 	/** Exact match on `status` (e.g. `"published"`, `"draft"`). */
 	status?: string;
--- a/packages/core/src/plugins/types.ts
+++ b/packages/core/src/plugins/types.ts
@@ -209,12 +209,27 @@ export interface ContentItem {
 }
 
 /**
+ * Filters applied at the database layer to `ContentAccess.list()`.
+ *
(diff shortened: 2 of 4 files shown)


packages/core/src/plugins/types.ts
```diff
@@ -208,13 +208,28 @@ export interface ContentItem {
 	publishedAt: string | null;
 }
 
+/**
+ * Filters applied at the database layer to `ContentAccess.list()`.
+ *
+ * Applied via SQL `WHERE` clauses, so they narrow the result set before
+ * pagination — unlike filtering the returned array in userland, which
+ * would still pay for pulling every row across the boundary.
```
ascorbic:
```suggestion
```

This isn't needed

ascorbic (commented):
Can you add a test too please

packages/core/src/plugins/types.ts
jdevalk: Done — removed in 4ff98d0.

packages/core/src/plugins/context.ts
```diff
@@ -235,6 +235,7 @@ export function createContentAccess(db: Kysely<Database>): ContentAccess {
 				limit: options?.limit ?? 50,
 				cursor: options?.cursor,
 				orderBy,
+				where: options?.where,
 			});
```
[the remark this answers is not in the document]
jdevalk: Added 5 integration tests in tests/integration/plugins/capabilities.test.ts covering where.status, where.locale, combined filters, and cursor pagination with filters applied (4ff98d0).

packages/core/src/plugins/types.ts
jdevalk: Done.

jdevalk:
> Can you add a test too please

Done.

ascorbic (approved):
Thanks

PR #558, opened 2026-04-14
fix(core): /search/suggest crash from double-star FTS5 prefix (#547)

### LINKED ISSUE #547: /search/suggest returns 500 — double "*" prefix produces invalid FTS5 syntax
## What happened

`GET /_emdash/api/search/suggest?q=<any-term>` returns HTTP 500 with `SqliteError: fts5: syntax error near "*"`. The endpoint is unusable for the autocomplete use case it was built for.

## Root cause

`getSuggestions` in `packages/core/src/search/query.ts` double-appends the prefix operator `*`:

- `escapeQuery()` (line 395) already returns each term with a trailing `*`:
  ```ts
  return terms.map((t) => `"${t}"*`).join(" ");
  ```
  → `escapeQuery("des")` returns `"des"*`.

- `getSuggestions()` (line 279) then adds another `*`:
  ```ts
  const prefixQuery = `${escapeQuery(query)}*`;
  ```
  → `prefixQuery` becomes `"des"**`, which is invalid FTS5 and raises `fts5: syntax error near "*"`.

Affected commit on `main`: `59bdca1`.

## Expected behavior

`/search/suggest?q=des` returns a 200 with up to `limit` suggestions whose title starts with "des".

## Reproduction

Against a fresh `demos/simple` (or any demo) with search enabled and FTS indexed:

```console
$ curl "http://localhost:4321/_emdash/api/search/suggest?q=des"
HTTP 500
{"error":{"code":"INTERNAL_ERROR","message":"..."}}
# Server log: SqliteError: fts5: syntax error near "*"
```

For comparison, `/search?q=des` works correctly (it uses `escapeQuery` directly without the extra `*`) after PR #424.

## Suggested fix

Remove the trailing `*` from line 279, since `escapeQuery` already produces prefix-matched terms:

```diff
-		const prefixQuery = `${escapeQuery(query)}*`;
+		const prefixQuery = escapeQuery(query);
```

Also worth adding: a regression test in `packages/core/tests/` that calls `getSuggestions` with a plain term and asserts no throw + returns shape.

## Context

This was noticed while working on PR #424 (public-search middleware). I kept it out of #424's scope per `AGENTS.md` ("do not touch code outside the scope of your change"). Happy to open a follow-up PR with the fix + failing-then-passing test if that works for the maintainers.

csfalcao:
## What does this PR do?

Fixes the `/_emdash/api/search/suggest` 500 error filed in #547. `getSuggestions` in `packages/core/src/search/query.ts` was appending `*` on top of the `*` that `escapeQuery` already appends to each term, producing invalid FTS5 like `"des"**` and raising `SqliteError: fts5: syntax error near "*"` on every request.

```diff
-		// Use prefix search for autocomplete
-		const prefixQuery = `${escapeQuery(query)}*`;
-		if (!prefixQuery || prefixQuery === "*") {
+		// Use prefix search for autocomplete. `escapeQuery` already appends `*`
+		// to each term for prefix matching, so we must not append another one.
+		const prefixQuery = escapeQuery(query);
+		if (!prefixQuery) {
 			continue;
 		}
```

`escapeQuery` returns `"des"*` for `"des"` (line 395 in the same file), so simply passing its output to `MATCH` is the correct thing to do. The `prefixQuery === "*"` guard is no longer reachable now that the literal `*` is gone, so the check is simplified to `!prefixQuery`.

A regression test is added at `packages/core/tests/integration/search/suggest.test.ts`:
- `returns matching suggestions for a plain prefix query` — reproduces the crash on `main`, passes after the fix.
- `returns empty array for a non-matching query` — guards against the returned shape regressing.

Closes #547. Follow-up to PR #424 (where I intentionally kept this out of scope per AGENTS.md).

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes (baseline on `main` is 11 pre-existing diagnostics in unrelated files; this PR does not add any)
- [x] `pnpm test` passes (new suggest tests green; the one pre-existing failure in `tests/integration/plugins/capabilities.test.ts "does not throw for any host"` reproduces on pristine `main` too and is unrelated to this change — verified by checking out `origin/main` and running the same test)
- [ ] User-visible strings in the admin UI are wrapped for translation (N/A — no UI changes)

## Screenshots / test output

Before (`main` at `82c6345`):

```console
$ pnpm vitest run tests/integration/search/suggest.test.ts
FAIL  tests/integration/search/suggest.test.ts > getSuggestions (Integration) > returns matching suggestions for a plain prefix query
SqliteError: fts5: syntax error near "*"
 ❯ getSuggestions src/search/query.ts:284:19
Tests  2 failed (2)
```

After this PR:

```console
$ pnpm vitest run tests/integration/search/suggest.test.ts
✓ tests/integration/search/suggest.test.ts (2 tests) 52ms
Tests  2 passed (2)
```

### DIFF
--- a/.changeset/shy-keys-check.md
+++ b/.changeset/shy-keys-check.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes `/_emdash/api/search/suggest` 500 error. `getSuggestions` no longer double-appends the FTS5 prefix operator `*` on top of the one `escapeQuery` already adds, so autocomplete queries like `?q=des` now return results instead of raising `SqliteError: fts5: syntax error near "*"`.
--- a/packages/core/src/search/query.ts
+++ b/packages/core/src/search/query.ts
@@ -275,9 +275,10 @@ export async function getSuggestions(
 		const ftsTable = ftsManager.getFtsTableName(collection);
 		const contentTable = ftsManager.getContentTableName(collection);
 
-		// Use prefix search for autocomplete
-		const prefixQuery = `${escapeQuery(query)}*`;
-		if (!prefixQuery || prefixQuery === "*") {
+		// Use prefix search for autocomplete. `escapeQuery` already appends `*`
+		// to each term for prefix matching, so we must not append another one.
+		const prefixQuery = escapeQuery(query);
+		if (!prefixQuery) {
 			continue;
 		}
 
--- a/packages/core/tests/integration/search/suggest.test.ts
+++ b/packages/core/tests/integration/search/suggest.test.ts
@@ -1,0 +1,57 @@
+import type { Kysely } from "kysely";
+import { describe, it, expect, beforeEach, afterEach } from "vitest";
+
+import { ContentRepository } from "../../../src/database/repositories/content.js";
+import type { Database } from "../../../src/database/types.js";
+import { SchemaRegistry } from "../../../src/schema/registry.js";
+import { FTSManager } from "../../../src/search/fts-manager.js";
+import { getSuggestions } from "../../../src/search/query.js";
+import { createPostFixture } from "../../utils/fixtures.js";
+import { setupTestDatabaseWithCollections, teardownTestDatabase } from "../../utils/test-db.js";
+
+describe("getSuggestions (Integration)", () => {
+	let db: Kysely<Database>;
+	let repo: ContentRepository;
+
+	beforeEach(async () => {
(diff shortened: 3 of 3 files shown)


ascorbic (approved):
Thanks!

PR #607, opened 2026-04-16
fix(import): bring WordPress menus, settings, and custom-block images through

mohamedmostafa58:
## What does this PR do?

Fixes several silent-drop paths in the WordPress (WXR) import so exports land with their menus, site identity, button links, and custom-theme imagery intact.

- **Navigation menus are never imported.** `nav_menu_item` and `wp_navigation` are filtered out as internal post types in the analyzer, so the admin UI's "Import navigation menus" toggle has no server-side effect. Any site being imported loses its Primary/Footer/etc. menus.
- **Site identity is never imported.** `<channel><title>`, `<description>`, and the base URL are parsed but never written to site settings. The admin's `importSiteTitle` toggle is tracked in component state but dropped in `executeImport()`.
- **Button hrefs are dropped.** `core/buttons` only reads `attrs.url`. Any content authored in the classic editor or older Gutenberg keeps the URL inside the inner `<a href>` and the container transformer doesn't fall back to it, so buttons come through with labels but no links.
- **Custom theme block imagery is dropped.** Themes ship custom blocks that wrap content around their own image attributes — hero sections holding an `imageUrl`, logo marquees / sliders / carousels holding an `items[]` array of image URLs, and similar patterns. The unknown-block fallback only recurses into `innerBlocks`, so any imagery declared on the wrapper's attrs silently disappears.
- **Internal links still point at the source WordPress host after import.** Navigation buttons and inline link marks that referenced other pages on the source site keep their absolute URLs, so clicking them bounces back to the WordPress origin instead of navigating within the imported site.

## What changed

- `packages/core/src/astro/routes/api/import/wordpress/analyze.ts`
  - Surface `navMenus`, `tagline`, and a detected `homePageSlug` on the analysis response (the admin UI's existing check against `data.navMenus` now actually hits).
- `packages/core/src/astro/routes/api/import/wordpress/execute.ts`
  - New `importNavMenus` helper — writes into `_emdash_menus` / `_emdash_menu_items`, resolves `post_type` menu items that reference imported pages to their emdash IDs, falls back to a custom URL when the target wasn't imported, and replaces same-name menus on re-import.
  - New `importSiteSettings` helper — writes `title` / `tagline` / `url` via `setSiteSettings`.
  - `ImportConfig` gains `importMenus` and `importSiteSettings` flags (default true); `ImportResult` reports `menus` and `siteSettings` summaries.
  - Menu items whose WordPress title is empty (the case where WordPress renders the linked page's current title at runtime) fall back to the imported page's title instead of the raw slug.
  - New `rewriteInternalLinks` post-pass — button URLs and inline link-mark hrefs that point at the source WordPress host are rewritten to site-relative paths. Scoped narrowly: image asset URLs, linked-image hrefs, and gallery items stay absolute and continue to flow through the existing media-URL rewrite step.
- `packages/gutenberg-to-portable-text/src/transformers/core.ts`
  - `core/buttons` falls back to the inner `<a href>` when `attrs.url` is missing, mirroring the fallback already in `core/file`.
- `packages/gutenberg-to-portable-text/src/transformers/index.ts`
  - Unknown-block fallback walks `block.attrs` and emits standard `image` blocks for any image-shaped URLs it finds (single-image attrs and `items[].url` array shapes), in addition to recursing into inner blocks. Emits standard image blocks so the existing media-import + URL-rewrite steps pick them up normally.
- Admin UI (`packages/admin`) sends the `importMenus` / `importSiteTitle` toggle state in the import config, and types are extended to match the new response fields.


ascorbic:
Hey, @mohamedmostafa58 ! It'd be good to get this in. Could oyu look at the review comments?

packages/core/src/astro/routes/api/import/wordpress/execute.ts
```diff
@@ -13,36 +13,45 @@ import {
...
 import { sanitizeSlug } from "./analyze.js";
 
+const MENU_NAME_INVALID_CHARS = /[^a-z0-9_-]/g;
+const MENU_NAME_REPEAT_DASHES = /-+/g;
+const MENU_NAME_TRIM_DASHES = /^-|-$/g;
+
 export const prerender = false;
 
 export interface ImportConfig {
-	/** Map WordPress post types to EmDash collections */
```
ascorbic, line 31: Why have the jsdoc comments been removed?

ascorbic:
Thanks for the contribution. I'm going to close this as it's been fixed in #1830. I hope you'll contribute again in future.

PR #612, opened 2026-04-16
fix(core): hydrate SEO on entries from getEmDashEntry/getEmDashCollection

drudge:
## What does this PR do?

`entry.data.seo` is always `undefined` when using `getEmDashEntry` / `getEmDashCollection`, because the live loader in `loader.ts` only reads columns from the content table — SEO lives in the separate `_emdash_seo` table. Today the only paths that attach SEO are the REST handlers (`api/handlers/content.ts`) and the plugin context, so storefront templates that call `getSeoMeta(entry, …)` (as the `seo/index.ts` docstring already suggests) silently get empty values.

This PR adds a `hydrateEntrySeo` helper in `query.ts` that mirrors the existing `hydrateEntryBylines` / `hydrateEntryTerms`:

- Early-returns when the collection does not have `has_seo = 1`.
- Batch-loads SEO rows via the existing `SeoRepository.getMany` (no N+1).
- Attaches each entry's resolved record to `entry.data.seo`.
- Swallows "no such table" errors for pre-migration databases (same pattern as bylines/terms).

Wired into both `getEmDashEntry` (single) and `getEmDashCollection` (batch). Runs in parallel with the existing byline/term hydration via `Promise.all`, so the added cost is one extra batched query — only on SEO-enabled collections.

Closes #

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

## Test output

`packages/core/tests/integration/seo/seo-query-hydration.test.ts` adds a regression test that mocks `astro:content` to feed a controlled raw entry shape, then drives the real `getEmDashEntry` / `getEmDashCollection` against an in-memory SQLite + `SeoRepository`. Reverting `hydrateEntrySeo` flips the positive cases red, so the test guards the fix from regression. Existing `tests/integration/seo/seo.test.ts` (55 cases) still passes alongside it.

## Open question for reviewers

Should SEO hydration be always-on (this PR's choice) or opt-in (e.g. `{ includeSeo: false }` for callers that want to skip the extra query)? The cost is one extra batched query per call — scoped to SEO-enabled collections only — so the blast radius feels small enough to justify always-on, matching the existing byline/term hydration pattern. Happy to gate it either way.


ascorbic:
Thanks for the contribution. I'm going to close this as we've taken a different approach to this.

PR #633, opened 2026-04-17
feat: AI Search plugin with indexing, search modal, and demo site

ttmx:
## What does this PR do?

Adds a Cloudflare-native AI Search plugin for EmDash, powered by Cloudflare AI Search. This includes:

- A Cloudflare integration in `packages/cloudflare/` that indexes content into an AI Search namespace via the native worker binding, with content-sync hooks, a cron-flushed reindex queue, query synonyms, an index-status endpoint, and a settings admin UI
- Updates to the existing Cloudflare demo (`demos/cloudflare/`) that register the AI Search binding, add a cron trigger to flush the reindex queue, and add a `/search` page backed by the plugin's query endpoint

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] New features link to an approved Discussion (https://github.com/emdash-cms/emdash/discussions/2260)

- [x] This PR includes AI-generated code — model/tool: OpenCode with GPT 5.5 and later GPT-5.6 Sol

## Screenshots / test output
image


packages/cloudflare/src/plugins/ai-search-admin.tsx
```diff
@@ -0,0 +1,542 @@
...
+		} catch (err) {
+			setSynonymsError(err instanceof Error ? err.message : "Failed to save synonyms");
+		} finally {
+			setIsSavingSynonyms(false);
+		}
+	};
+
+	return (
+		<div className="space-y-6">
+			<h1 className="text-3xl font-bold">AI Search</h1>
```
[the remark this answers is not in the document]
ttmx: This is a plugin, there is no plugin infra for intl, and lingui does not work here without declaring the translations in core, which is not workable for an optional plugin.

packages/cloudflare/src/plugins/ai-search-admin.tsx
```diff
@@ -0,0 +1,542 @@
...
+	Trash,
+} from "@phosphor-icons/react";
+import type { PluginAdminExports } from "emdash";
+import { apiFetch, parseApiResponse } from "emdash/plugin-utils";
+import * as React from "react";
+
+const API_BASE = "/_emdash/api/plugins/ai-search";
+const SCHEMA_API_BASE = "/_emdash/api/schema";
+
+const indexedLabel = (n: number) => `${n} item${n !== 1 ? "s" : ""}`;
```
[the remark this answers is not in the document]
ttmx: Also an intl issue, addressed before.

scottbuscemi:
I deployed this branch to a real Workers site to exercise AI Search end-to-end, seeded ~10 posts, and hit a blocking bug in the public search path. I pushed a fix + regression tests to the branch (`9392a73`).

## Bug: public site search is broken for every anonymous visitor

`createAISearchSnippetEndpoint` read `locals.emdash.db` to load synonyms:

```ts
const emdash = locals.emdash;
const options = new OptionsRepository(emdash.db); // emdash.db is undefined here
```

On the **anonymous request path** EmDash only attaches the partial fast-path facade, where the `db` getter is not populated (it exists only on the full, authenticated facade). The type system missed it because `EmDashHandlers.db` is declared non-optional and the endpoint casts `locals.emdash` to `EmDashRuntime`.

So on a normal unauthenticated visit, `OptionsRepository.get()` → `this.db.selectFrom("options")` throws:

```
TypeError: Cannot read properties of undefined (reading 'selectFrom')
```

…and the endpoint returns `503 { success: false, error: "Search is temporarily unavailable" }`. Since the snippet endpoint is only ever hit anonymously (it's the public site search), **search never returns results in production**. Repro on my deploy: the search modal/`/api/ai-search/search` 503'd for every query even though items were indexed (`status: completed`) and a raw binding query returned them.

## Fix

Resolve a request-scoped runtime via `withEmDashRuntime()` (whose `db` is a real ALS-scoped handle regardless of auth) instead of reading `locals.emdash.db`:

```ts
const { withEmDashRuntime } = await import("emdash/middleware");
return withEmDashRuntime(async (runtime) => {
  const options = new OptionsRepository(runtime.db);
  ...
});
```

The import is **lazy on purpose**. A top-level `import { withEmDashRuntime } from "emdash/middleware"` pulls the `astro:middleware` virtual module into the ai-search module graph. Because `astro.config.mjs` imports `aiSearch()` from `@emdash-cms/cloudflare/plugins`, that specifier then gets evaluated by the Astro **config loader** (plain Node ESM), which can't resolve the `astro:` scheme — `astro build` fails with:

```
Only URLs with a scheme in: file, data, and node are supported... Received protocol 'astro:'
```

Loading it via dynamic `import()` inside the handler (and keeping the top-level reference as `import type`) keeps config evaluation clean.

After the fix, search works on the deployed site — e.g. "space station thriller" → the *Orbit* post, "lighthouse survival" → *The Long Quiet*, etc.

## Tests

The existing suite exercised `handleAISearchSnippetRequest` (the pure handler) but never the `createAISearchSnippetEndpoint` wrapper, which is where the db was acquired — that's why this slipped through. Added two files:

- **`ai-search-endpoint.test.ts`** — drives the endpoint wrapper with an anonymous-style `locals` (`emdash` present but no `db`, and `emdash` absent entirely). Asserts it returns 200 and constructs `OptionsRepository` from the `withEmDashRuntime` db, never from `undefined`. This test fails on the pre-fix code with the exact `selectFrom` TypeError.
- **`ai-search-build-safety.test.ts`** — guards that `emdash/middleware` is only referenced as `import type` or a dynamic `import()` in `ai-search.ts`, so a future top-level import can't silently reintroduce the `astro:` build break.

All 31 ai-search tests pass; typecheck and `oxlint --type-aware` are clean. No changeset added — the existing `.changeset/ai-search-plugin.md` already covers `@emdash-cms/cloudflare`.

PS - the cron pr was merged, and the admin UI can't be i18n yet due to how core handles plugins. This should be good to go pending additional testing.

packages/cloudflare/src/plugins/ai-search.ts
```diff
@@ -0,0 +1,1427 @@
...
+	metadata: Record<string, string>,
+	ctx: PluginContext,
+	options: UploadItemOptions = {},
+): Promise<void> {
+	const mirrorKey = `item:${key}`;
+	const previousId = await ctx.kv.get<string>(mirrorKey);
+	let previousRemoved = false;
+	if (previousId) {
+		try {
+			await instance.items.delete(previousId);
```
khoinguyenpham04: Could we make this a safe swap, perhaps with a versioned upload key? Deleting the current item first means that if every upload retry fails, the live result and its mirror are gone until another save or reindex.

packages/cloudflare/src/plugins/ai-search.ts
```diff
@@ -0,0 +1,1427 @@
...
+					filters: {
+						visible_after: { $lte: nowSeconds },
+						locale: { $eq: locale },
+					},
+					metadata_only: true,
+				},
+			},
+		});
+
+		const collections = input.collection?.split(",").map((value) => value.trim());
```
khoinguyenpham04: could we apply the collection filter in the AI Search request? filtering after `max_num_results` can discard valid matches when another collection ranks higher, so a posts-only search can return no results even when matching posts exist.

packages/cloudflare/src/plugins/ai-search.ts
```diff
@@ -0,0 +1,1427 @@
...
+
+			"content:afterSave": {
+				handler: async (event: ContentHookEvent, ctx: PluginContext): Promise<void> => {
+					const { content, collection } = event;
+					if (!(await shouldSync(collection, ctx))) return;
+
+					// Sync based on the current status: published content is visible
+					// immediately (visible_after=0), scheduled content is indexed but
+					// gated until its scheduledAt timestamp, and drafts are removed.
+					return waitForSync(syncSearchIndex(content, collection, ctx));
```
khoinguyenpham04, line 1175: could we use the live data when a published item has a pending draft? A draft-only save keeps the status as published, so this currently replaces the public search document with unpublished draft text.

packages/cloudflare/src/plugins/ai-search.ts
```diff
@@ -0,0 +1,1427 @@
...
+		// `locals.emdash.db` throws when we load synonyms. Resolve a real,
+		// request-scoped runtime via withEmDashRuntime() so `runtime.db` works
+		// regardless of auth.
+		//
+		// Imported lazily: `emdash/middleware` references `astro:` virtual
+		// modules, which the Astro config loader (plain Node ESM) cannot
+		// resolve. A top-level import would drag that into config evaluation
+		// via `aiSearch()` and break `astro build`.
+		const { withEmDashRuntime } = await import("emdash/middleware");
+		return withEmDashRuntime(async (runtime: EmDashRuntime) => {
```
khoinguyenpham04, line 1551: could we check that ai-search is active before serving this request? this endpoint still searches and applies saved synonyms after the plugin is disabled, so the toggle does not actually turn public search off.

khoinguyenpham04 (commented):
could you please address these below? cheers

packages/cloudflare/src/plugins/ai-search.ts
ttmx: Done

packages/cloudflare/src/plugins/ai-search.ts
ttmx: Fixed

packages/cloudflare/src/plugins/ai-search.ts
ttmx: I just removed the limit since this should be handled on the ai search instance tbh

packages/cloudflare/src/plugins/ai-search.ts
ttmx: Super simplified this function, no need to delete the previous item, we can just overwrite it.

ttmx:
The public snippet endpoint returns { success: false, error: <string> } instead of the required { error: { code, message } } envelope in two places. -> Yes, because the ai search snippet uses this shape.

packages/cloudflare/src/plugins/ai-search.ts
```diff
@@ -0,0 +1,1373 @@
...
+				},
+			},
+
+			"content:afterUnpublish": {
+				handler: async (
+					event: ContentPublishStateChangeEvent,
+					ctx: PluginContext,
+				): Promise<void> => {
+					const { content, collection } = event;
+					if (!(await shouldSync(collection, ctx))) return;
```
khoinguyenpham04: I indexed a page, changed the config from `["pages", "posts"]` to `["posts"]`, then unpublished and deleted the page. It stayed in `AI Search` and its mirror key remained. Pls let removal hooks run regardless of the selected collections; keep `shouldSync` for indexing, but do not let it skip removal.

packages/cloudflare/src/plugins/ai-search.ts
```diff
@@ -0,0 +1,1373 @@
...
+}
+
+async function ensureAISearchInstance(
+	ns: AiSearchNamespace,
+	config: AISearchConfig,
+): Promise<AiSearchInstance> {
+	const instanceName = config.instanceName ?? "emdash-content";
+	const handle = ns.get(instanceName);
+	try {
+		await handle.info();
```
khoinguyenpham04, line 683: I tested this end to end with a real bare `AI Search` instance. The upload completed and status reported the item as indexed, but public filtered search returned nothing while an unfiltered search found the item. Please reconcile the required `custom_metadata` here and report when a reindex is needed.

packages/cloudflare/src/plugins/ai-search.ts
```diff
@@ -0,0 +1,1373 @@
...
+	const searchLocale = async (locale: string): Promise<AISearchSnippetResponse> => {
+		const response = await instance.search({
+			messages: [{ role: "user", content: effectiveQuery }],
+			ai_search_options: {
+				retrieval: {
+					filters: {
+						visible_after: { $lte: nowSeconds },
+						locale: { $eq: locale },
+					},
+					metadata_only: true,
```
khoinguyenpham04, line 740: `max_num_results` is accepted by the public request but never reaches `instance.search`. The shipped snippet client and the demo request 30 results, but a real request with `max_num_results: 1` returned 2 results. Pls validate and forward the requested limit.

packages/cloudflare/src/plugins/ai-search.ts
```diff
@@ -0,0 +1,1373 @@
...
+		version: "1.0.0",
+		capabilities: ["read:content"],
+		admin: {
+			entry: "@emdash-cms/cloudflare/plugins/ai-search-admin",
+			pages: [{ path: "/settings", label: "Cloudflare AI Search", icon: "search" }],
+		},
+
+		hooks: {
+			"plugin:install": {
+				handler: async (_event: unknown, ctx: PluginContext): Promise<void> => {
```
khoinguyenpham04: On a fresh setup, the `ai-search` plugin was active but had no configured collections and no `reindex` job. This `plugin:install` handler is not running for configured plugins, so initial indexing never starts automatically; the operator has to run a manual sync. Please move this setup to a lifecycle path that runs.

packages/cloudflare/src/plugins/ai-search.ts
```diff
@@ -0,0 +1,1373 @@
...
+const SYNONYM_CACHE_TTL_MS = 60_000;
+
+let synonymCache = {
+	rewriter: compileSynonyms([]),
+	refreshAfter: 0,
+};
+
+async function getSynonymRewriter(kv: KVReader): Promise<SynonymRewriter> {
+	const now = Date.now();
+	if (now < synonymCache.refreshAfter) return synonymCache.rewriter;
```
khoinguyenpham04: I saved a `synonym`, removed it, and immediately searched again. The old rewrite was still applied for up to 60 seconds. Pls make this propagation behavior explicit, or use a version check so updates can be picked up without adding a full database read to every search.

packages/cloudflare/src/plugins/ai-search.ts
```diff
@@ -0,0 +1,1373 @@
...
+				const slug = typeof metadata.slug === "string" && metadata.slug ? metadata.slug : parsed.id;
+				const image =
+					typeof metadata.image === "string" && metadata.image ? metadata.image : undefined;
+				return {
+					id: chunk.id,
+					type: chunk.type,
+					score: chunk.score,
+					item: {
+						key: renderResultUrl(config, { ...parsed, slug }, locale),
+						...(chunk.item.timestamp === undefined ? {} : { timestamp: chunk.item.timestamp }),
```
khoinguyenpham04: The returned date is the `AI Search` index time, not the content publish time. After reindexing, older content shows a recent, year-less date in the search results. Please omit this `timestamp` or return the content publish date.

khoinguyenpham04 (commented):
Thanks for the fixes, could you address these comments below? Cheers

packages/cloudflare/src/plugins/ai-search.ts
khoinguyenpham04: I re-tested this after the `remove the limit` change. That change does not address the ordering issue: with 19 matching documents, the post ranked outside the first 10, so `collection=posts` still returned zero even though the post was indexed. The missing client limit and this post-retrieval filter are the same root cause. Pls apply the collection filter during retrieval or search each collection separately.

packages/cloudflare/src/plugins/ai-search-admin.tsx
```diff
@@ -0,0 +1,542 @@
...
+				setAvailable(options);
+
+				const valid = new Set(options.map((c) => c.slug));
+				const saved = (configData.collections ?? []).filter((slug) => valid.has(slug));
+				// Restore the last configured selection; fall back to the common
+				// content collections when nothing has been configured yet.
+				const initial =
+					saved.length > 0
+						? saved
+						: options.filter((c) => c.slug === "posts" || c.slug === "pages").map((c) => c.slug);
```
khoinguyenpham04: The admin UI falls back to selecting `posts` and `pages` here, but it does not persist that selection. When no config exists, `shouldSync` treats `null` as every collection, so the UI can appear scoped to two collections while content hooks index all collections. I reproduced this on a fresh setup. Pls persist the default selection or make the UI match the hook behavior.

packages/cloudflare/src/plugins/ai-search.ts
ttmx: Added this since it might be used in the future by the ai search snippet.

packages/cloudflare/src/plugins/ai-search.ts
ttmx: I thought about this, and my idea was that this would give the wrong idea, since it would only clear it in the worker isolate in which you do the search I believe, and people from other locations would still hit the cache. I can change it regardless, I agree this would make it easier to test synonyms.

packages/cloudflare/src/plugins/ai-search.ts
ttmx: Passed the collection filtering to the ai search query itself now.

packages/cloudflare/src/plugins/ai-search.ts
ttmx: Fixed, I hadn't thought about this.

packages/cloudflare/src/plugins/ai-search.ts
ttmx: Now reports in settings page that metadata is desynced and lets you press a button to fix it.

packages/cloudflare/src/plugins/ai-search.ts
ttmx: Removed plugin:install
I don't think there is an appropriate install path for config plugins?
We do create the instance whenever a post is added or a search is made or a reindex is triggered.

packages/cloudflare/src/plugins/ai-search.ts
ttmx: Added clearing in this worker isolate.

packages/cloudflare/src/plugins/ai-search.ts
ttmx: Removed timestamp

packages/cloudflare/src/plugins/ai-search-admin.tsx
ttmx: Fixed by actually equaling no collections saved to posts+pages and refactoring so that this is set centrally.

khoinguyenpham04 (commented):
thanks for the changes, can you fixes these please?

packages/cloudflare/src/plugins/ai-search.ts
```diff
@@ -0,0 +1,1475 @@
...
+
+// =============================================================================
+// Helpers
+// =============================================================================
+
+function isRecord(value: unknown): value is Record<string, unknown> {
+	return typeof value === "object" && value !== null && !Array.isArray(value);
+}
+
+function isMissingAISearchInstanceError(error: unknown): boolean {
```
khoinguyenpham04, line 274: Cloudflare documents missing instances as `AiSearchNotFoundError` with the message `ai_search_not_found`, but this helper only checks for a 404/`NOT_FOUND` property or an “instance not found” message. The metadata route therefore returns 503 for a fresh or deleted instance instead of allowing the operator to repair it.

Could u recognize the documented error shape and add a regression test using the production-shaped error?

packages/cloudflare/src/plugins/ai-search.ts
```diff
@@ -0,0 +1,1475 @@
...
+async function ensureAISearchInstance(
+	ns: AiSearchNamespace,
+	config: AISearchConfig,
+): Promise<AiSearchInstance> {
+	const instanceName = config.instanceName ?? "emdash-content";
+	const handle = ns.get(instanceName);
+	try {
+		await handle.info();
+		return handle;
+	} catch {
```
khoinguyenpham04: This catch treats every `info()` failure as a missing instance. A transient failure or concurrent first request can therefore trigger `create()` and surface another 503.

Can you restrict creation to recognized not-found errors and handle concurrent initialization safely?

packages/cloudflare/src/plugins/ai-search.ts
```diff
@@ -0,0 +1,1475 @@
...
+	async function getEffectiveCollections(ctx: PluginContext): Promise<string[]> {
+		return (await getConfiguredCollections(ctx)) ?? DEFAULT_COLLECTIONS;
+	}
+
+	/** Persist the operator's collection selection from the dashboard. */
+	async function saveConfiguredCollections(
+		ctx: PluginContext,
+		collections: string[],
+	): Promise<void> {
+		await ctx.kv.set(CONFIG_COLLECTIONS_KEY, collections);
```
khoinguyenpham04, line 836: After changing the configured collections from `["posts", "pages"]` to `["pages"]`, an already-uploaded `posts/foo.md` remains in AI Search. This function only persists the selection; `shouldSync()` affects future hooks, and unscoped searches do not apply the saved collection set.

Please reconcile removed collections by deleting their mirrored AI Search items, or enforce the effective collection set in every search path. Please add a regression test for an item indexed before deselection.

packages/cloudflare/src/plugins/ai-search.ts
```diff
@@ -0,0 +1,1475 @@
...
+					}
+
+					for (let page = 0; page < REINDEX_PAGES_PER_TICK; page++) {
+						if ((await processReindexBatch(job, ctx)).done) break;
+					}
+					if (reindexResult(job).done) await ctx.cron?.cancel(REINDEX_CRON_TASK);
+				},
+			},
+
+			"content:afterSave": {
```
khoinguyenpham04, line 1158: Thanks for clarifying that configured native plugins do not have an install hook. The hooks below only respond to future content events; none of them enqueues a backfill for existing published content when the plugin becomes active.

Could u either enqueue an initial reindex from a supported lifecycle path, or make manual sync an explicit required setup step in the status UI and documentation? A regression test should cover existing published content.

packages/cloudflare/src/plugins/ai-search.ts
```diff
@@ -0,0 +1,1475 @@
...
+}
+
+async function getSynonymRewriter(kv: KVReader): Promise<SynonymRewriter> {
+	const cache = synonymCacheHolder();
+	const now = Date.now();
+	if (now < cache.refreshAfter) return cache.rewriter;
+
+	const generation = cache.generation;
+	cache.refreshAfter = now + SYNONYM_CACHE_TTL_MS;
+	const synonyms = (await kv.get<Synonym[]>(CONFIG_SYNONYMS_KEY)) ?? [];
```
khoinguyenpham04, line 626: The 60-second cache reduces the frequency, but the first anonymous request after expiry still reads synonyms here. In the public endpoint, this `kv.get()` is backed by `OptionsRepository`, so it adds a database query to the logged-out search path.

please move refresh or invalidation out of the request path, or explicitly document and approve this tradeoff. A query-count test would also help prevent this from becoming more frequent.

ttmx: I think this is fine. We need to get something from KV if its not in memory, the max we could do is increase the timeout.

packages/cloudflare/src/plugins/ai-search.ts
ttmx: I added this in the docs more explicitly. There is no reliable way to make it index on install, besides perhaps hooking onto cron, but this seems really hacky, so I'd rather not.

packages/cloudflare/src/plugins/ai-search.ts
ttmx: Fixed.

packages/cloudflare/src/plugins/ai-search.ts
ttmx: Fixed

packages/cloudflare/src/plugins/ai-search.ts
ttmx: Fixed

PR #647, opened 2026-04-18
Feature/i18n: Add support for Farsi/Persian in admin package

arashackdev:
## What does this PR do?

- Adds support for Farsi in EmDash admin.
- Adds full coverage for translations of the admin package in Farsi. 
- Adds https://github.com/rastikerdar/vazirmatn font to the core astro package.
- Adds `vazirmatn` as the default font family for Farsi in the default Noto Sans font families.

Note:
- `vazirmatn` isn't a Noto Sans font (to my understanding), but it is in the Sans font family. ([Google Fonts Link](https://fonts.google.com/specimen/Vazirmatn))
- Although the Farsi font support is added to the astro core package, the font isn't applied in the admin when switching language. This is due to the font script not being loaded on the admin, though it applies to all languages.

## Type of change

- [ ] Bug fix
- [x] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

The changes were iterated over with Claude Code, both for code adjustments and translations. However, this PR isn't done or owned by an agent.

The 3rd-4th iteration on translation onwards were done more granularly, to replace specific terminologies with their correct Farsi corresponding terms and verbs that are already established in the Iranian web community. The current translations are verified by traversing through the dashboard, though mistakes are possible and could be updated later for the existing translations.

---

## Screenshots / Test Outputs

580428744-9c9a61a2-534f-42a5-a218-082b5251f6d0


arashackdev:

ascorbic:
Hi. Before you go too far in this: I am deliberately *not* including all fonts by default. That would cause very large builds, and most sites don't need them all. Also can we keep translation PRs (.po and locale config files) separate from other code changes, including fonts, CSS etc.

arashackdev:
> Hi. Before you go too far in this: I am deliberately _not_ including all fonts by default. That would cause very large builds, and most sites don't need them all. Also can we keep translation PRs (.po and locale config files) separate from other code changes, including fonts, CSS etc.

Hi, Matt! That sounds like a great idea, I would create a separate PR for translations only and try to see how css/font changes could be refined. Initially, I started with translations, but verifying them in the dashboard with default fonts got difficult, since Farsi font support is limited in browsers. I might be a bit picky here (😅), but below is the difference:
emdash-dashboard-farsi-default
emdash-dashboard-farsi-vazirmatn

Unfortunately, Iran (the community behind this language) has been shut away from the internet for over 50 days now. After they get connected, I think they'll have better ideas for improvements.

While doing this PR, I have to touch the core astro package to enable loading admin fonts (in case the admin locale is different from the astro website/instance admin is interacting with) If this is a real issue (after more verifying (🤞), I can add the fix both separately or with the css/font improvements.

Anyhow, the first change will be translations after verifying them through the dashboard, one more iteration.

ascorbic:
It should use Noto Sans Arabic if you enable Arabic script in the config, which should at least be better than browser default. That said, I am open to adding a proper Farsi font – just not enabled by default. Right now we only enable the main Noto Sans font by default, and requite opt-in for other scripts (such as Arabic and CJK scripts). I suggest that the approach you use in font-provider.ts is fine, but just not the auto-enabling part. This should enable you to pass "farsi" as an enabled font script in the config.

packages/core/src/astro/integration/font-provider.ts
```diff
@@ -57,7 +58,7 @@ const ALL_GOOGLE_SUBSETS = [
 ];
 
 /**
- * Known Noto Sans script families on Google Fonts.
+ * Known Noto Sans and Sans script families on Google Fonts.
```
arashackdev, line 61: I changed this comment since https://github.com/rastikerdar/vazirmatn isn't a **Noto Sans** font, but still in the **Sans** font family. I thought this comment would me more accurate, but okay to revert it too.

packages/core/src/astro/integration/font-provider.ts
```diff
@@ -69,6 +70,7 @@ const NOTO_SCRIPT_FAMILIES: Record<string, string> = {
 	"chinese-hongkong": "Noto Sans HK",
 	devanagari: "Noto Sans Devanagari",
 	ethiopic: "Noto Sans Ethiopic",
+	farsi: "Vazirmatn",
```
arashackdev, line 73: Although the font is defined here, it is not applied automatically in the admin. This needs some rework on how font scripts are loaded and is out of scope here.

PS: https://github.com/emdash-cms/emdash/pull/647#issuecomment-4275849897 illustrates the visual change when the font is loaded.

arashackdev:
The changes for force loading the fonts are reverted. I also got familiar with configuring emdash locally and got the font working without a problem. 😅

The PR covers translations for farsi and adding a default font for it.

ascorbic (approved):
Great, thanks for getting this across the line. I've tested it locally and it works great. The only issue is that if somebody enables both Farsi and Arabic script, it uses the Arabic font for both. I am guessing this is because they have overlapping unicode ranges, which make it pretty hard to fix. I'd like to think about if there are solutions there, but it's not a blocker and I'm happy to get this in.

Issue #643, opened 2026-04-18
feat(auth): server-side session validation helper for custom API routes


Vallhalen:
## Problem

When building custom API endpoints in an EmDash site (e.g., `/api/leads`, `/api/brief-upload`), there is no way to validate whether a request comes from an authenticated admin session.

Currently developers resort to checking cookie names:

```typescript
// INSECURE - only checks if cookie NAME exists, not its value
const cookie = request.headers.get("cookie") || "";
const isAdmin = cookie.includes("astro-session");
```

An attacker can bypass this with `Cookie: astro-session=fake`.

## Security impact

This was identified during a security audit (RAPTOR framework). Any custom API route that needs admin-only access has no secure way to validate sessions without direct database queries against EmDash's internal session tables.

## Proposed solution

Export a session validation helper from the emdash package:

```typescript
import { validateSession } from "emdash/auth";

export const GET: APIRoute = async ({ request, locals }) => {
  const user = await validateSession(request, locals.emdash);
  if (!user) return new Response("Unauthorized", { status: 401 });
  // user.id, user.email, user.role available
};
```

Or provide a middleware/guard:

```typescript
import { requireAdmin } from "emdash/auth";

export const GET: APIRoute = requireAdmin(async ({ request }) => {
  // Only reaches here if session is valid admin
});
```

## Current workaround

```typescript
// Validate session value length (not just name) - still fragile
const match = cookie.match(/astro-session=([^;]+)/);
const isAdmin = match && match[1] && match[1].length >= 20;
```

This is better than name-only check but still doesn't verify the session against the database.

## Environment

- EmDash 0.5.0
- Cloudflare Workers deployment
- Custom API routes in src/pages/api/

Vallhalen:
**Status check on 0.13.0 — looks addressed via `locals.user`.** 0.13 ships a `./locals` subpath that augments `App.Locals` with:

```ts
interface Locals {
  /** Authenticated user - set by auth middleware when a valid session exists */
  user?: User;
}
```

paired with the `emdash/middleware/auth` export (`onRequest`). So a custom API route can now securely check `context.locals.user` instead of the insecure `cookie.includes('astro-session')` pattern from the report — the middleware validates the session and populates `user` only when it's genuinely valid.

[...]

PR #667, opened 2026-04-19
docs: add guide for securing custom API routes

Vallhalen:
## Summary

Adds a new documentation guide explaining how to properly authenticate admin requests in custom API endpoints built on EmDash sites.

## Context

During a security audit of our EmDash-based site (using RAPTOR and gstack /cso), we discovered that a common pattern of checking `cookie.includes("astro-session")` is insecure — an attacker can bypass it with a forged cookie header.

This guide documents the secure approach: validating the session token against the SESSION KV namespace.

## What the guide covers

- Why cookie-name-only auth is insecure (with concrete exploit scenario)
- KV-backed session validation pattern (full code example)
- Fail-closed principle (deny when KV unavailable)
- Additional security recommendations (CSRF, rate limiting, path traversal)
- Reference to #643 for the built-in `validateSession()` feature request

## Related issues

- #643 — feat(auth): server-side session validation helper

## Checklist

- [x] New documentation follows Starlight MDX format
- [x] Code examples are complete and copy-pasteable
- [x] Security recommendations based on real audit findings

### DIFF
--- a/docs/src/content/docs/guides/custom-api-auth.mdx
+++ b/docs/src/content/docs/guides/custom-api-auth.mdx
@@ -1,0 +1,111 @@
+---
+title: Securing Custom API Routes
+description: How to authenticate admin requests in custom API endpoints using EmDash session tokens.
+---
+
+import { Aside, Code, Tabs, TabItem } from "@astrojs/starlight/components";
+
+When building custom API endpoints in your EmDash site (e.g., admin dashboards, file managers, lead trackers), you need to verify that incoming requests come from authenticated admin users. This guide shows how to do it securely.
+
+## The Problem
+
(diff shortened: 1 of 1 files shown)


ascorbic:
Hey! The correct way to access the session is to use the Astro session API. See https://docs.astro.build/en/guides/sessions/#api-endpoints

You want to do `session.get('user')`

Vallhalen:
You're absolutely right — the Astro Session API is the proper way to handle this. Our guide was essentially reinventing the wheel by manually checking KV when `context.session` already provides a clean, built-in abstraction.

We'll update our own endpoints to use the session API instead of the manual cookie/KV pattern.

Closing this PR — thanks for pointing us in the right direction! 🙏

PR #675, opened 2026-04-20
Fix R2 publicUrl being ignored for rendered media

### LINKED ISSUE #508: R2 storage publicUrl config is not used by Image component or media API endpoints
## Summary

The `r2()` storage config accepts a `publicUrl` option, but it has no effect on how images are served to visitors. The Image component and media API endpoints hardcode `/_emdash/api/media/file/` instead of consulting the storage adapter's `getPublicUrl()` method.

This means all media requests go through the Worker (which reads from R2 and proxies the bytes), even when a public R2 custom domain is configured. The `Cache-Control: public, max-age=31536000, immutable` header mitigates this for repeat visits, but every cache miss still hits Worker CPU/memory to proxy R2 bytes.

## Config

```ts
storage: r2({
  binding: "MEDIA",
  publicUrl: "https://media.example.com",  // accepted but unused
}),
```

## Where it breaks

**1. `EmDashImage.astro` (lines 58–64)** — hardcodes the API route:

```ts
function buildLocalImageUrl(img: MediaValue): string {
  const storageKey = (img.meta?.storageKey as string) || img.id;
  if (storageKey) {
    return `/_emdash/api/media/file/${storageKey}`;
  }
  return "";
}
```

**2. Portable Text `Image.astro` (line 128)** — same hardcoded fallback:

```ts
src = asset.url || `/_emdash/api/media/file/${asset._ref}`;
```

**3. Media API endpoints** (`api/media/[id]/confirm.ts:27`, `api/media.ts:35`) — return hardcoded URLs instead of calling `storage.getPublicUrl()`:

```ts
url: `/_emdash/api/media/file/${item.storageKey}`,
```

## Where it works

The `R2Storage` class correctly implements it:

- `upload()` returns `url: this.getPublicUrl(options.key)` ✅
- `getPublicUrl()` returns `${publicUrl}/${key}` when configured ✅

But nothing downstream uses the upload result's URL or calls `getPublicUrl()`.

## Expected behavior

When `publicUrl` is configured, rendered `` tags should use the public URL (e.g., `https://media.example.com/storageKey.jpg`) instead of `/_emdash/api/media/file/storageKey.jpg`. This would bypass the Worker entirely for media serving.

## Suggested fix

Expose the storage's public URL base to the rendering layer — either:

1. Make `buildLocalImageUrl()` in the Image components aware of the configured `publicUrl`, or
2. Have the media API endpoints call `storage.getPublicUrl(key)` so the URL stored in content fields reflects the public URL

Option 2 would also fix existing content on next re-publish/re-save.

## Environment

- EmDash 0.1.x on Cloudflare Workers + R2
- `@emdash-cms/cloudflare` storage adapter
- R2 bucket with custom domain configured

eyupcanakman:
## What does this PR do?

Closes #508

The storage `publicUrl` option is documented but has no effect on how images render. `EmDashImage`, the Portable Text image block, and `Gallery` all hardcoded the media proxy path, so media requests went through the Worker even when a custom R2 domain is configured.

### Change

- New `getPublicMediaUrl(storageKey)` on `locals.emdash` that calls `storage.getPublicUrl()`. `EmDashImage`, `Image`, and `Gallery` all route through a shared `buildRenderMediaUrl` helper and fall back to `/_emdash/api/media/file/{key}` when no adapter is attached.
- The url-rewrite branch of `buildRenderMediaUrl` only accepts `[A-Za-z0-9._-]+` for the extracted key, so attacker-controlled portable-text `asset.url` can't traverse to sibling CDN prefixes or inject query strings. Unrecognised shapes pass through to the internal proxy, which validates by id.
- `S3Storage.getPublicUrl` aligns its fallback with R2: returns the proxy path when no `publicUrl` is configured. The old endpoint+bucket string had no in-tree consumer.

### Scope note

`EmDashMedia`, admin thumbnails, `InlinePortableTextEditor`, and the visual-editing toolbar keep their existing behavior. A follow-up can route those surfaces through the same capability.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] User-visible strings in the admin UI are wrapped for translation and `pnpm locale:extract` has been run (if applicable). N/A, no admin UI changes
- [ ] New features link to an approved Discussion. N/A, bug fix

## Screenshots / test output

```
 ✓ tests/unit/media/url.test.ts (16 tests) 4ms
 ✓ tests/unit/storage/s3.test.ts (28 tests) 6ms
 ✓ tests/unit/astro/middleware-prerender.test.ts (4 tests) 21ms
 Test Files  3 passed (3)
      Tests  48 passed (48)
```

### DIFF
--- a/.changeset/quiet-rivers-bloom.md
+++ b/.changeset/quiet-rivers-bloom.md
@@ -1,0 +1,5 @@
+---
+"emdash": minor
+---
+
+Renders local media through storage `publicUrl` when configured. `EmDashImage` and the Portable Text image block now call a new `locals.emdash.getPublicMediaUrl()` helper, so R2 and S3 deployments with a custom domain serve images from that domain. `S3Storage.getPublicUrl` now returns the `/_emdash/api/media/file/{key}` path when no `publicUrl` is set (previously `{endpoint}/{bucket}/{key}`).
--- a/packages/core/src/astro/middleware.ts
+++ b/packages/core/src/astro/middleware.ts
@@ -43,7 +43,7 @@ import {
 } from "../emdash-runtime.js";
 import { setI18nConfig } from "../i18n/config.js";
 import type { Database, Storage } from "../index.js";
-import { resolvePublicMediaUrl } from "../media/url.js";
+import { createPublicMediaUrlResolver } from "../media/url.js";
 import type { SandboxRunner } from "../plugins/sandbox/types.js";
 import type { ResolvedPlugin } from "../plugins/types.js";
 import { getRequestContext, runWithContext } from "../request-context.js";
@@ -306,7 +306,7 @@ export const onRequest = defineMiddleware(async (context, next) => {
 						locals.emdash = {
 							collectPageMetadata: runtime.collectPageMetadata.bind(runtime),
 							collectPageFragments: runtime.collectPageFragments.bind(runtime),
-							getPublicMediaUrl: (key: string) => resolvePublicMediaUrl(runtime.storage, key),
+							getPublicMediaUrl: createPublicMediaUrlResolver(runtime.storage),
 						} as EmDashHandlers;
 					} catch {
 						// Non-fatal — EmDashHead will fall back to base SEO contributions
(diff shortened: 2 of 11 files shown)


ascorbic:
@eyupcanakman thanks for this. Could you look at the copilot review comments?

eyupcanakman:
Went through them. `buildRenderMediaUrl` now keeps bare `_ref` / `img.id` on the internal proxy and the rewrite branch only accepts plain filenames, so portable-text content can't reroute the CDN origin. Gallery routes through the same helper, and the middleware closure moved into `createPublicMediaUrlResolver`.

ascorbic (approved):
Thanks!

PR #676, opened 2026-04-20
fix(seo): emit locale-prefixed URLs in per-collection sitemap

drudge:
## Summary

When Astro i18n is enabled with `prefixDefaultLocale: false`, the per-collection sitemap emits every translated row under the default-locale path. With a bilingual site (e.g. EN at `/`, FR at `/fr/`), the FR entries appear in the sitemap as `/a-propos` instead of `/fr/a-propos` and 404.

Root cause: `handleSitemapData` doesn't select the row's `locale`, and the route builds URLs from `url_pattern` alone.

## Changes

- `packages/core/src/api/handlers/seo.ts` — select `c.locale` and expose it on `SitemapContentEntry`. Safe to select unconditionally since migration `019_i18n` backfills `locale` on every `ec_*` table.
- `packages/core/src/astro/routes/sitemap-[collection].xml.ts` — mirror the prefix rule from `src/loader.ts`: prepend `/<locale>` when the row's locale isn't the default, or always when `prefixDefaultLocale: true`.
  - Reads i18n straight from `virtual:emdash/config` (same source the loader uses) rather than `getI18nConfig()`, so the route doesn't depend on middleware having initialized the i18n singleton before the injected route runs.
- `packages/core/tests/integration/seo/seo.test.ts` — two cases covering default-locale rows and translated rows.

## Notes

- No schema changes, no new `url_pattern` placeholder — locale prepending matches the loader's existing convention.
- `prefers-reduced-motion`: urlset output is unchanged when i18n is disabled (single-locale or no config).
- A follow-up PR addresses a separate pre-existing encoding bug where `encodeURIComponent(slug)` turned multi-segment slugs like `faq/products` into `faq%2Fproducts`. Kept out of this PR to keep the diff focused.

## Test plan

- [x] `pnpm --filter emdash test -- tests/integration/seo/seo.test.ts` — 57 tests pass
- [x] Verified live against a bilingual Astro site (EN at `/`, FR at `/fr/`): `/sitemap-pages.xml` emits both `/about` and `/fr/a-propos`; FR URLs load 200.

### DIFF
--- a/packages/core/src/api/handlers/seo.ts
+++ b/packages/core/src/api/handlers/seo.ts
@@ -18,6 +18,8 @@ export interface SitemapContentEntry {
 	slug: string | null;
 	/** ISO date of last modification */
 	updatedAt: string;
+	/** Row locale (e.g. "en", "fr"). Always set — i18n migration backfills to "en". */
+	locale: string;
 }
 
 /** Per-collection sitemap data with entries and URL pattern */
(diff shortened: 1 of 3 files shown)


ascorbic:
Thanks. This was superseded by #1195.

PR #694, opened 2026-04-21
RFC: Decentralized Plugin Registry

ascorbic:
Draft RFC for decentralised plugin registry.

Rendered markdown: https://github.com/emdash-cms/emdash/blob/wip/plugin-rfc/rfcs/0001-plugin-registry.md

---

### Try this PR

**[Open a fresh playground →](https://wip-plugin-rfc-emdash-playground.emdash-cms.workers.dev)**

<sub>Tracks `wip/plugin-rfc`. Updated automatically when the playground redeploys.</sub>


BenjaminPrice:
One immediate thought is that supporting native plugins that bypass the sandbox may be worth digging into. 

It somewhat goes against the idea of EmDash bringing secure plugins as a replacement to Wordpress. Mainly because the majority of end users likely won’t know/understand/care about the difference. 

All end users are likely to look at is functionality and more than likely just blindly install the plugin if it looks to do what they expect, regardless of if it’s sandboxed. At that point it negates the whole idea of the sandbox. 

Also, it disincentivizes plugin authors from building for the sandbox and to rather just take the easier, native, path.

ascorbic:
@BenjaminPrice yeah, that's a valid concern. The reason I thought that it was ok on balance is because the install process for a sandboxed plugin is so much easier: just click and it works, vs needing to npm install and add it to the config. There's also the fact that this RFC doesn't specify how these will actually be displayed – and some clients may choose to not display native plugins at all. I'd say this would probably be the case for managed hosts.

jimray:
Oh this is exciting, fantastic work @ascorbic. Reading this mostly through an atproto lens...

Overall, this feels very protocol-native: publishing to author's PDS, using the Relay and App View to aggregate, with clients to query. Love it! There's also the old familiar problem of recentralization creeping in via the biggest App View being the one Emdash maintains. It's a familiar one!

The [atproto primer](https://github.com/emdash-cms/emdash/blob/wip/plugin-rfc/rfcs/0001-plugin-registry.md#at-protocol-primer) is so good.

**Lexicons**
The Lexicon designs overall look solid. 

Are you certain you want to tie everything to `com.emdash`. A plugins.pub domain (and `pub.plugins` Lexicon root) could be more general purpose if you want to expand beyond the Emdash project. At the very least, you could have a higher level set of Lexicons that would be general purpose for repositories and then use `com.emdash` for anything specific to the CMS project; this is similar to the distinction between `com.atproto` and `app.bsky`.

The longer text fields for `readme` and `changelog` strike me as a potential problem. 50k on a README with a 100k record size limit is rough, plus updating the README (a not-infrequent event) has a cascading effect where you have to rewrite the entire signed record. Maybe a blob of type `text/markdown` makes more sense here?

Bundling the icon and screenshots into the tarball feels like it could be problematic. Does this mean the App View, or any directory client, would need to download the full bundle and extract the images in order to display them? I think a more idiomatic approach would be to include them in the Lexicon, maybe similar to how a [bsky record includes embeds](https://docs.bsky.app/docs/advanced-guides/posts#images-embeds:~:text=A%20complete%20post%20record%2C%20containing%20two%20images%2C%20would%20look%20something%20like).

Setting `security` as `Contact[]` is probably fine but possibly not very useful. Consider aligning with https://github.com/.well-known/security.txt conventions or just linking to a file. A free-form URL to a security policy might be a better fit.

There's a lot of variability in the Lexicon based on the type of package it is (native vs. sandboxed). Feels like a good candidate for a `union`? What this might look like in practice:

A sandboxed release:
```json
 {
    "$type": "com.emdashcms.registry.release",
    "package": "at://did:plc:abc/com.emdashcms.registry.package/gallery",
    "version": "1.0.0",
    "integrity": "sha256-...",
    "source": {
      "$type": "com.emdashcms.registry.defs#sandboxedSource",
      "url": "https://...",
      "capabilities": ["read:content"],
      "allowedHosts": ["images.example.com"]
    },
    "createdAt": "..."
  }
```

A native release:
```json
  {
    "$type": "com.emdashcms.registry.release",
    "package": "at://did:plc:abc/com.emdashcms.registry.package/seo",
    "version": "1.0.0",
    "integrity": "sha256-...",
    "source": {
      "$type": "com.emdashcms.registry.defs#nativeSource",
      "npmVersion": "@example/emdash-seo@1.0.0"
    },
    "createdAt": "..."
  }
```

One way to test this would be to add a hypothetical third type — if it doesn't break existing records, it's probably correct. If it requires new upstream fields, not there yet.

Consider making `url` an array that would allow the author to specify multiple (such as a CDN) to avoid linkrot.

The Lexicon evolution section is smart and a good set of best practices.

**CLI**

Really love the commitment to using OAuth over app passwords. Not always an easy thing on the CLI, excited to see your implementation!

**Misc**

The moderation/squatters issue is potentially quite real; the directory landing before review/reports/labelers could be problematic. There's a period where `did:plc:bad-guy/emdash-gallery-plugin` sits next to the legitimate one with zero UI signal to distinguish them, AND the directory is small enough that users are clicking into everything.

Is there a migration story for existing plugins? Or a plan for communicating to them directly?

Related, what's the discovery mechanism for updates once a plugin is installed? How does an admin know there's a new version, if there are breaking changes, security issues, etc.

Consider adding a deprecation mechanism (maybe that's just a field in the lexicon?)

Multiple authors feels like something worth resolving sooner rather than later.

BenjaminPrice:
A few additional thoughts:

---

**Native plugins in the registry**

Expanding on my earlier comment, I think this needs more design attention before landing. The RFC currently presents sandboxed and native plugins side by side in a unified registry, with the install flow adapting to the type. The concern is what this looks like to the end user.

A non-technical site owner browsing the directory sees two plugins that do the same thing. One is sandboxed with declared capabilities. One is native with full system access. Both show up in the same search results, with the same visual treatment. The type difference is a metadata field, not a prominent trust signal.

The RFC says "some clients may choose to not display native plugins at all" and that the install friction difference (one-click vs npm install + config change) is a natural deterrent. I think that underestimates the problem. If the directory website (not the admin UI) lists both types, someone Googling "emdash gallery plugin" might land on the native one and follow the install instructions without understanding what they're opting into. Managed hosts can hide native plugins in their admin UI, but the public directory is where discovery happens for new users.

@jimray's point about squatting compounds this. Pre-labelers, a native plugin squatting on a name like `gallery-plugin` sits next to the legitimate sandboxed one with zero trust signal and *also* runs without sandbox enforcement. That's a worse outcome than a sandboxed squatter.

Not arguing to remove native plugins from the registry, the unified discovery is genuinely useful. But I think the RFC should say more about how the directory UI distinguishes them and what the default presentation should be. At minimum:
- Prominent visual distinction between sandboxed and native in all listing contexts
- A clear explanation of what "native" means in non-technical language ("this plugin has full access to your site" or similar)
- A recommendation that public directory implementations default to showing sandboxed-only results, with native behind a filter toggle

---

**Reviews, labelers, and the pre-labeler gap**

The RFC defers reviews, moderation, and labelers to follow-on RFCs, and cites #296 as the design basis for that work. I'd like to suggest that at least the initial labeler work happens in parallel with the registry implementation rather than as a follow-on. Launching the directory without any trust signals, even basic ones, creates a window where the squatting and abuse problems @jimray flagged are real and unmitigated. Ideally the registry and at least one labeler ship together.

Happy to help draft or contribute to that work. Some things I've been thinking about that might be useful:

- **Install-gated reviews using transparent site identity.** The RFC already reserves `did:web` derived from the site's domain for signed install/review records (called out in the #296 attribution). The idea is that EmDash auto-generates a `did:web` identity for every site on first boot. Keypair generated, DID document served at `/.well-known/did.json`, site owner never sees any of it. Reviews are signed with this key. No accounts, no signup, no friction. You installed the plugin, you can rate it.

- **Install context on reviews (Steam-inspired).** Each review carries how long the plugin has been installed, whether it's currently active, and the version at review time. "Active for 3 months, v1.2.0" is a different trust signal than "installed 10 minutes ago."

- **Review quality labeler as the first labeler.** Built on the ATProto labeler spec, using patterns from the existing `ai-moderation` plugin (Llama Guard categories, enrichment + decision separation). Categories like review bombing detection (flag anomalous volume + sentiment windows, exclude from aggregate score), spam/low-effort flagging, and vote manipulation detection. Downrank rather than delete. Labels are informational, the AppView decides how to act on them.

- **Publisher verification as a lightweight launch labeler.** Even basic checks (DID resolves, PDS is accessible, signing key matches) could ship as a simple labeler alongside the registry itself. This gives the directory *something* to differentiate legitimate publishers from squatters on day one, without waiting for the full trust RFC.

Happy to write any of this up in more detail or take a first pass at the follow-on RFC.

---

**Key management and the PDS/MST approach**

@chuckadams' comment in #296 about FAIR's key management challenges is worth flagging here. They found that private keys were too sensitive to store in a WP database, which led to passphrase encryption requirements and UX complexity around key prompting. Still an open design issue in FAIR Beacon.

The RFC's approach of using repo-level MST signing, where the PDS handles the keys and publishers never directly manage them, sidesteps this entirely. The publisher authenticates via OAuth, the PDS signs the records, and there's no plaintext key file sitting on disk. This is a genuine advantage of building on ATProto's signing infrastructure rather than rolling explicit per-publisher key management.

Worth calling this out as a design rationale in the RFC itself, since it's a concrete example of where the ATProto approach pays off vs. the alternatives.

schiste:
I hope I'm not too late to the party. 

1. in the RFC wordpress plugin system depending on wordpress.org is flagged as an issue/risk and the RFC is leaning a lot on a similar situation with atproto. Not a blocker, just flagging it as something that felt odd when I read it. 

**3. Sandboxed/native plugins:** supporting @BenjaminPrice comment, I understand the rational but one of the "selling point" of emdash was a better/more secure plugin ecosystem. I would expect the two type of plugins to have clear boundaries for that very reason. And this should not be only UX/UI 

On my second point, why not prepare the place for a community driven review system (not implementing it with the first batch but setting us up for that). I'm currently working on that issue (not for emdash plugins) and I'm leaning toward using a public git repository to store assessments and have automated + community driven assessment. For our plugins it could mean a emdash-plugins repo. Plugin authors can publish .md with all relevant informations and source code. TBD process to elevate a plugin from unreviewed to stage 1, 2 3 reviewed. 

Not saying to set it up from the get go, but at least having a first step and extend the current setup with small, stable trust-tier enum carried alongside the runtime variant (something like tier: "sandboxed" | "native-unreviewed" | "native-reviewed" | "native-firstparty"). And that way avoid having to retroffit the model for the first party plugins you mention. 

**4. Every plugins are considered the same, same as above, we may need at least to scaffold authoring / trust / security early on.** 

I would add a bit of parity between ssandbox and native, even if not enforced for native, #nativeRuntime should require a structured declaredAccess block: things like filesystem: "read" | "write" | "none", network: HostPattern[] | "any", processes: bool, env: string[], astroHooks: string[], addsRoutes: bool, addsMiddleware: bool. That way at install, and afterwards, admins can have a clear vision of what plugins "do". 

I am not super expert on Astro, and that should be for later, but shouldn't Astro include a "failsafe" in case a plugin doesn't match the signed release? Probably not a day one thing though. 

Lastly, though it's related to the topic it's probably a bigger discussion, but the separation between sandbox and native should be stronger, that may mean going down the road of having their own endpoints. To be able to have different security settings/rules later down the road.

But it's great to see this topic moving forward thank you so much :)

ascorbic:
I'd done a rewrite that should bring it a lot closer to FAIR, and brings in some of the changes in the protocol recently. I'd really appreciate feedback, particularly from @BenjaminPrice and @toderash! I've got a bit more detail on the atproto side too, so @jimray's thoughts would be great too.

BenjaminPrice:
The FAIR unification is a strong rewrite — adopting `pm.fair.package.profile` / `pm.fair.package.release` directly and layering EmDash-specific concerns into `com.emdashcms.*` cleanly resolves the dual-namespace ambiguity the previous draft had. Deferring native plugins out of v1 (and out of the registry-facing admin UI) also addresses the immediate trust-conflation concern that @schiste and I raised. Thanks for the careful response.

I'd still like to see a small amount of forward-compatible scaffolding land in v1, on the EmDash extension specifically. The cost is two fields in a namespace you fully control; the upside is avoiding a retrofit against records already sitting in publishers' PDSs by the time native plugins or graduated trust tiers come back into scope.

---

**1. Reserve `tier` on `com.emdashcms.package.releaseExtension`**

Add the field now, even though v1 only ever writes `"sandboxed"`:

```
tier: "sandboxed" | "native-unreviewed" | "native-reviewed" | "native-firstparty"
```

Clients treat unrecognised values as untrusted. When the native follow-on RFC lands, every existing record is already shape-correct — no migration, no inferred-default ambiguity, no question of how a v1 client should treat a record without the field.

---

**2. Promote `capabilities` / `allowedHosts` into a structured `declaredAccess` block**

Today the extension carries flat `capabilities` and `allowedHosts`. Lifting them into a structured block extends cleanly to native and matches @schiste's earlier ask without committing to native semantics now:

```ts
declaredAccess: {
  network?: "none" | { hosts: HostPattern[] } | "any";
  filesystem?: "none" | "read" | "write";
  processes?: boolean;
  env?: string[];
  astroHooks?: string[];
  addsRoutes?: boolean;
  addsMiddleware?: boolean;
  // sandbox-specific (current fields, retained here)
  capabilities?: Capability[];
  allowedHosts?: HostPattern[];
}
```

For sandboxed plugins, the sandbox enforces these — no behaviour change in v1. For native plugins later, the same block is purely declarative; the admin install UI surfaces what the plugin claims to do, with no enforcement guarantee. Same renderer, same shape, different trust model.

---

**3. Aggregator XRPC: take a `tier` filter parameter, default to `sandboxed`**

The XRPC adoption is great. While the surface is being defined, please reserve a `tier` query parameter on the discovery endpoints and have v1 implementations only return `sandboxed` results. Opting into native plugins later then becomes `?tier=sandboxed,native-firstparty` rather than a new endpoint or breaking parameter shape.

---

**4. Capability-consent UI renders structured-ready**

The capability-consent dialog should render `declaredAccess` field-by-field, not as a flat capability list. This is a one-time piece of work that future-proofs the UI for native attestations without committing the registry to displaying them.

---

The combined ask is small — `tier` + `declaredAccess` on the extension, `tier` query param on the aggregator, structured rendering in the consent UI — and lands the protocol in a state where the harder follow-on questions don't require touching v1 records.

ascorbic:
@BenjaminPrice thanks for the feedback! Some good ideas. 

For 1 and 3, I'd say yes I think these would almost certainly be the shape we'd end up with. However I would say that because atproto lexicon fields can be added but never removed, we should hold off on adding them until we're actually using them. While your suggestion is probably the route we'd take, there's no harm in waiting til we need them if we say they default to "sandboxed". Regarding the tier specifically, I'd say it would probably be just "native" and "sandboxed", with reviewed etc handled via labels. What I'll do is add these to the future plans section, but not normative for now.

With 2, yes that sounds like a good idea. Are you suggesting that we add extra fields sandboxed plugins now too though? As with 1 and 3 I would leave the other fields undefined for now, and just define the sandboxed fields. We can then add the rest later when we spec native plugins. One thing I would be wary of though is anything that would make it appear like the native permissions are actually enforced. This is why I'm unsure of whether we should declare them in the same way. However in any case, this can be decided when we spec native plugins.

4: yes, I think generally the consent UI will need some work. We don't need to spec that though – that's a design decision in the admin UI, or for marketplace implementations.

toderash:
Hey @ascorbic, thanks for this update - some very good work and discussion here that nicely narrows the gap with the FAIR Protocol. I'll try to offer some thoughts in specific areas, hopefully without needing to go too deep (or into the weeds) on any of them!

I'll mostly refrain from items that are very EmDash-specific concerns, but deferring native plugins for post-v1 removes the trust conflation issues until they can be dealt with more fully is a better approach than rushing it now. I think that plus adopting `pm.fair.*` as the primary Lexicon roots is a good direction.

## AT URI ≠ DID

The FAIR spec requires `id` to be a valid DID, so an AT URI would fail strict validation. That said, the primary concern is to have a universally unique identifier for the package, so _in theory_ it could be treated that way, defined in `ext-emdash.md` as accepted for the package type with resolution rules defined. I'll need the FAIR TSC to weigh in on whether that'd be a reasonable approach, or if accepting a non-DID identifier breaks the spirit of it. Registering a DID per package isn't really a "legacy requirement", it's pretty foundational to how the protocol resolves the package in the supply chain.

Having said all that, AT URI is arguably stronger than a web DID, but I'd suggest its format require the use of a package slug rather than the auto-generated `rkey`. The slug is publisher-controlled and (can be) stable, whereas the `rkey` is typically formed with a TID that would change the `rkey` in thee AT URI if the record is  deleted & recreated. It'd be good to have @jimray weigh in here, but one thing to note is that the client must support ATproto; https alone isn't enough, so you can't validate a signature without ATproto access. Not a showstopper, but could have implications for corporate firewall rules, depending on the BOFH you're dealing with.

This brings me to

## Trust Models

It may help to understand the _why_ behind a couple of differences here - not that it needs to change anything, but it may help in fitting the pieces together.

Trust in the **Package** is pretty integral to the FAIR Protocol, which explains the per-package DID. There are a few reasons for this:

* In a mature ecosystem, packages can change publishers, either through a leadership handover in an OSS project (new author) or through the sale of a commercial product. A universal identifier for the package itself ensures that it is transferable while maintaining certainty that "this is that".
* Separation of signing authority is inherent, so a publisher with many packages has signing on a per-project (package) basis, narrowing the list of who can access signing keys. This is where @chuckadams comment comes in: you have to protect the keys to protect the supply chain, and I'd suggest it's hard to have too much paranoia in that arena.
* A DID per package facilitates signing individual assets associated with the package.

The FAIR Protocol does anticipate Publisher-level trust, which is described to some extent in our docs on moderation, but not yet scoped within the Protocol itself. I've filed an issue to document that, and a PR is forthcoming - the upshot is that establishing Publisher identity with a DID aligns with FAIR. It's currently "buried" in our moderation docs because the intent is to assign a trust label that helps inform users about a package's publisher - you might not trust a package from an untrusted publisher, or you might trust a new package with no track record if you trust the publisher.

To this point, you've likely seen the recent disclosures that a batch of WP plugins were acquired by a new publisher who inserted backdoors that went undetected for months.

Trust in the **Publisher** is primary for EmDash in the RFC, with trust in the package inherited from that root. This is the more direct path to doing the same thing with a trust label in FAIR's design. I'd argue publisher trust is a necessity.

Practically speaking, the MST signatures have a couple of disadvantages that may or may not be concerns for the EmDash ecosystem.

* The signature is at the commit level rather than the asset level, so verifying a specific artifact requires traversing the Merkle proof path for that record over a live connection.
* A cached artifact can't be verified offline, where per-artifact Ed25519 signatures can. (Not a huge lift to add per-artifact signatures using the Publisher's signing key.)

## Capability declaration/enforcement & granular capabilities

I like @BenjaminPrice's `declaredAccess` structure: a block like this should distinguish between declared/enforced and declared/advisory before native plugins are adopted, but would already apply for the sandbox. @schiste touched on this as well with some examples.

What the RFC may be missing is what the runtime actually enforces at execution. Perhaps this is left out intentionally, but should document what EmDash actually does wrt enforcement. A minimal section in v1 may help here at least for the sandbox. To further the point a bit, I'd suggest that `declaredAccess` support some quantitative constraint fields to set enforceable limits.

It's been suggested that EmDash doesn't need the same level or type of plugin scrutiny as WordPress because that's provided by the sandbox: this is true, but only partially. Thinking through the sandbox, istm that binary permissions could pose a risk in certain situations - my example here is a plugin that legitimately needs smtp access, which is granted as a binary `true`. What this opens up is the ability to send any number of emails to any recipient. If it's also got permission to do an http request, it's now a potential spam relay without breaching the sandbox. All it needs to do is include a legitimate function for the one it intends to abuse; the exposure also applies to a vulnerable plugin, so publisher intent need not be malicious. otoh, you could limit that with something like

```
email?: {
  send?: { rateLimit?: { perHour: number }; recipientDomains?: string[] }
}
```

This isn't pure spec territory, since it needs to match the implementation, and is a core design question for EmDash; here it would just need to reflect whatever it actually does. Anyway, I'd recommend including all this for sandboxed capabilities now so it doesn't need to be reined in later by adding new constraints, and can form the pattern for speccing native plugin handling.

## Pre-Label Gap

In FAIR, we made the decision not to open up the WordPress plugin listing too widely until our trust labeller is in place. TYPO3 already has this sorted, but in the wp ecosystem, we're applying some correctives. For one, the size of the ecosystem has seen the official plugin review team go quickly from caught-up to swamped, largely with a lot of AI-generated submissions. An ecosystem that ramps up slowly won't have the same issue, but EmDash may not have the luxury of a slow-and-steady ramp-up. This is to say simply that the pre-label gap can be a real issue, so working out a phased approach for opening up the ecosystem may be warranted based on growth rate rather than legacy factors.

Here, @BenjaminPrice's "Review quality labeller" may be applicable, or at least offer a similar parallel to the way we're approaching it in FAIR, at least for the WordPress ecosystem. There, we're tooling up some processes to automate the biggest steps in addressing that 

## Overall

I really like how this is shaping up. To help identify remaining gaps and where/how they're best addressed, I've got some homework to do on the FAIR Protocol side drafting some of the language to describe what's already been done, decided, or proposed. 

One example: @chuckadams is the lead on our aggregator implementation, and while its API endpoints are not yet well-documented, they go further than what the protocol outlines, and we extended it as recently as the CloudFest Hackathon in March.

Another is an outline of how trust shifts between publisher and repo, how publisher "ownership" of a package is verifiably transferred, and disclosures where publisher trust can't be established.

I need more hours in the week! More to follow.

ascorbic:
Thanks, @toderash. I've looked in more dertail at your points, as well as @BenjaminPrice's, and made some updates to the RFC:

- 56050449bf0c1832ff29de3679384f1ddc279b91 I'm convinced by the `declaredAccess` idea, and have adopted it. The shape I went with was `category.operation.constraints..`, so e.g. `network: { request: { allowedHosts: [...] } }`, or `"content": { "read": true }` and left the constraints as an open definition. 
- 7d98f6c6d03cbd752ed05b712d21c803efc41532 adds a bit more about the trust model tradeoffs. I spent quite a bit of time thinking about the offline verification/MST traversal issue and after drafting various solutions I ended up back where I started! I actually think the threat model there is narrow enough that it is best not included directly in the spec, at least for EmDash. If an artifact is mirrored offline, you verify it at the time of mirroring, and optionally cache the proof chain at the same time. 
- 1decbe1dd76e14d410b0f2f11edb2e56c5f4ce6e I've added a bit about labellers, because even though I've deferred the details to another RFC I wanted to make it clear that we want to add these early, and certainly before any stable release. I also touch on a web of trust approach, similar to Bluesky trusted verifiers, that could make it more scalable early on to verify publishers.

ascorbic:
I have now added draft lexicons! They are under `com.emdashcms.experimental.*` for now, including ones that will hopefully end up under `pm.fair.*`

schiste:
Thanks for the hard work!

On declaredAccess, I realize in this first version we may be missing how to manage/what happens when a plugin updates and expand its access. Without that, "privilege escalation" through plugin updates feels quite easy, but I may be missing something.

toderash:
@ascorbic you read me correctly on the trust model an MST signing - mainly just fyi material for where the philosophies differ slightly and the implications that context can swing one way or the other. ymmv! Nothing outright incompatible yet, with a narrowing gap. Explaining the 'why' helps clarify what's movable and what's not, so will continue that approach.

Back to my PR draft...

ascorbic:
@schiste it's a good point. We must ensure that updated packages need to ask permission if they change their declared access

PR #716, opened 2026-04-22
fix(core): hash full plugin descriptor in manifest cache key

lsngmin:
## What does this PR do?

Fixes stale manifest cache when plugin descriptor fields change without a version bump. Previously only id@version was hashed; now the full PluginDescriptor is serialized into the cache key.

No tests added. The cache key logic is embedded in EmDashRuntime.create() so existing integration tests cover it.

Supersedes #707

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes


ascorbic:
Thanks for picking this up. Sorry for the slow followup. 

The direction is right but the implementation has a few issues. `JSON.stringify(p)` doesn't work as a hash on `ResolvedPlugin`: hooks and routes are records of functions. The cleaner fix is to hash the slice that actually feeds the manifest, since that's what we're caching. Roughly:
```js
function manifestFingerprint(p: ResolvedPlugin) {
	return {
		id: p.id,
		version: p.version,
		capabilities: [...p.capabilities].sort(),
		allowedHosts: [...p.allowedHosts].sort(),
		storageKeys: Object.keys(p.storage).sort(),
		admin: {
			pages: p.admin.pages,
			widgets: p.admin.widgets,
			portableTextBlocks: p.admin.portableTextBlocks,
			fieldWidgets: p.admin.fieldWidgets,
		},
	};
}
```
Then a stable stringify of that. Hooks and routes don't appear in the manifest, so they correctly don't appear in the key.

Could you also bring back the integration test from #707? "Existing tests cover it" isn't true: nothing in the suite exercises a config change between two runs sharing a DB, which is the only path this code matters on. Worth covering at least: no change preserves the cache, an admin.portableTextBlocks change busts it, a capabilities change busts it.
Separately, I think we still want #707's dev-mode behaviour as a follow-up. Even with a correct hash, "edit plugin file, no schema change" should hot-reload in dev rather than rely on bumping the version.

lsngmin:
Updated, thanks.

ascorbic:
Good news! We're not caching the manifest anymore, so this isn't needed.

PR #749, opened 2026-04-24
fix(admin): make content list search, sort, and pagination work across the whole collection

edrpls:
## What does this PR do?

Fixes three compounding UX failures in the admin collection list view (`/_emdash/admin/content/{collection}`). Each one looks like a separate paper-cut in isolation, but together they make the admin functionally broken for any collection larger than the first API page.

### 1. Search didn't find existing content

Typing `Bleeding` in the search box on a collection with 100+ entries returned zero results even though a published entry titled exactly `Bleeding` existed. The admin filtered client-side against `items` already in memory, and the API accepted neither a `search` nor a `q` param. For entries past the first fetch, the user saw "No results" and had no way to know the item was there.

**Fix:** `GET /_emdash/api/content/{collection}` now accepts `q` — case-insensitive substring match against whichever of `title`, `name`, and `slug` actually exist on the collection's table (introspected via `pragma_table_info` / `information_schema.columns`). LIKE wildcards are escaped (`%`, `_`, `\`), inputs are trimmed + capped at 200 chars, and the MCP `content_list` tool accepts `q` too.

### 2. Column headers weren't sort controls

`Title`, `Status`, `Locale`, `Date` were plain `<th>` labels with no click handler, no `aria-sort`, no indication they could do anything. The repository's `FindManyOptions` already supported `orderBy`, but nothing in the admin sent it.

**Fix:** headers are now real `<button>`s wrapped in `<th aria-sort>`. Click cycles direction; screen readers get the sort state via `aria-sort` on the header (which they announce automatically), so the button's accessible name stays clean. Extended the server's `mapOrderField` whitelist to include `status`, `locale`, and `name` alongside the existing date fields.

### 3. Pagination denominator grew as you paged forward

Pages 1–4 showed `1/5`, `2/5`, …; page 5 showed `5/10`; page 10 showed `10/15`; and so on. `PAGE_SIZE = 20` paginates over what's been fetched; `limit = 100` per API call means 5 client pages per fetch. The auto-fetch effect fired when you reached the last client page, another 100 arrived, `totalPages` jumped, and the denominator lurched forward.

**Fix:** list response now returns `total` alongside `items` + `nextCursor`. Admin uses it as the stable denominator (a 143-entry collection reads `1/8` on page 1). When a server doesn't return `total`, the admin falls back to the pre-change behavior, so old servers keep working.

### While I was there

- `ContentPickerModal` had the same client-side search bug — it now passes `q` to the API too. Added `placeholderData: keepPreviousData` so the picker list doesn't flash to empty between keystrokes while the debounced query refetches.
- `clampedPage` logic in `ContentList` handles the edge case where filters collapse the result count below the current page (previously rendered an empty table until the next refetch).

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

The symptoms are user-facing (search doesn't find existing content; pagination denominator is non-monotonic) and the fixes are scoped to the list endpoint + admin list component + picker. Happy to split into three separate PRs if preferred, but I think the change is easier to review together since the three symptoms share the same root cause (admin treating the list as a thin client over a cursor-only API with no server-side query support).

## Checklist

- [x] `pnpm typecheck` passes (admin + core)
- [x] `pnpm lint` passes (no new findings; pre-existing workspace-wide warnings unchanged)
- [x] `pnpm test` passes for touched areas: `packages/core` repo + handler suites (92 tests), `packages/admin` component suite (39 tests including 5 new ones)
- [x] Added regression tests that reproduce each of the three bugs before the fix
- [x] New admin strings were wrapped for translation and `pnpm locale:extract` was run; the diff simplifies the a11y strings (net -3 new messages across all locales)
- [x] Added a changeset (`patch` for `emdash` and `@emdash-cms/admin`)

Produced in a single Claude Code session. Ran Code Reviewer, Security Engineer, and Accessibility Auditor sub-agents against the committed-but-unpushed branch and applied their findings in commit 2 before pushing — specifically:
- a11y: dropped a verbose `aria-label` in favor of `aria-sort` (which screen readers announce automatically) and added an explicit `focus-visible` outline since kumo resets browser defaults.
- security: confirmed `pragma_table_info(?)` is parameterized by better-sqlite3 and `validateIdentifier` is belt-and-braces, not load-bearing; LIKE escape order is correct; added `ESCAPE '\\'` pairing verified.
- code review: added tests for slug-only collections (collection with no `title`/`name`) and for multi-page search preserving `total` across cursor pages; documented the auto-fetch effect's reliance on the caller deduping `onLoadMore`.

## Testing

From a fresh `pnpm install` + `pnpm build`:

```
packages/core $ pnpm test tests/database/repositories/content.test.ts tests/unit/api/content-handlers.test.ts
  ✓ tests/unit/api/content-handlers.test.ts (24 tests)
  ✓ tests/database/repositories/content.test.ts (68 tests)

packages/admin $ pnpm test tests/components/ContentList.test.tsx
  ✓ tests/components/ContentList.test.tsx (39 tests)

packages/admin $ pnpm test   # full suite
  Test Files  54 passed (54)
  Tests       791 passed (791)
```

New regression tests (all fail on `main`, pass on this branch):
- `ContentRepository > findMany > search > finds items by a title substring match`
- `ContentRepository > findMany > search > is case-insensitive`
- `ContentRepository > findMany > search > searches slug-only collections that have no title/name columns`
- `ContentRepository > findMany > search > keeps list items and total consistent across pagination`
- `ContentRepository > findMany > search > escapes LIKE wildcards so they don't act as patterns`
- `ContentRepository > findMany > total > reports total rows regardless of limit`
- `ContentRepository > findMany > orderBy > accepts status as an order field`
- `Content Handlers — list search and total > filters items by q (case-insensitive substring)`
- `Content Handlers — list search and total > returns total independent of limit`
- `ContentList > pagination > uses \`total\` as a stable denominator instead of items.length`
- `ContentList > sortable headers > calls onSortChange when a header is clicked`
- `ContentList > sortable headers > exposes sort state via aria-sort on the active header`
- `ContentList > controlled search > forwards the typed query to onSearchChange without local filtering`


edrpls:
Closing at my own request — this PR bundles three distinct fixes and is too large to review as one unit. Splitting into three independent PRs, each with its own changeset, tests, and scope so they can be reviewed/merged on their own cadence:

1. **Sortable column headers** — `SortableTh`, `aria-sort`, server `orderBy` whitelist extension. Smallest, fully independent.
2. **Stable pagination denominator** — `total` field on the list response; admin uses it. Independent.
3. **Server-side search** — `q` query param with column introspection + LIKE escaping, picker migration, MCP tool. Largest surface.

Each will link to the others for context. All three share the same branch root and tests; no one depends on another being merged first, though the UX benefits compound. Thanks — and sorry for the noise.

PR #751, opened 2026-04-24
fix(admin): stable pagination denominator via total on the list response

edrpls:
## What does this PR do?

Fixes the admin collection list's non-monotonic pagination denominator. Observed against the films collection (~143 entries) at `/_emdash/admin/content/films`:

| Page | Current | Expected |
|------|---------|----------|
| 1–4  | `N / 5` | `N / 8`  |
| 5    | `5 / 10`| `5 / 8`  |
| 10   | `10 / 15`| out of range  |

**Root cause:** the admin calculates `totalPages = ceil(items.length / PAGE_SIZE)` where `items` accumulates 100 per API fetch and `PAGE_SIZE` is 20 (5 client pages per fetch). Each auto-load bumps the denominator.

**Fix:** `GET /_emdash/api/content/{collection}` response now includes an optional `total` field with the full filtered row count. The admin uses it as the stable denominator.

Split from the (now closed) #749 — this is the second of three independent PRs. The other two are [#750 sortable headers](https://github.com/emdash-cms/emdash/pull/750) (merged?) and server-side search (separate PR to follow).

### Changes

- `FindManyResult<T>.total?` added to `packages/core/src/database/repositories/types.ts`.
- `ContentRepository.findMany` runs the existing `count()` in parallel with the list query (`Promise.all`) and sets `total` on the result. On SQLite the extra query is essentially free.
- `ListResponse<T>.total?` + `contentListResponseSchema.total` (optional) so clients that ignore it keep working.
- `handleContentList` forwards `total` through to the API envelope.
- Admin's `FindManyResult.total?` mirrors the server shape.
- `<ContentList>` accepts `total`; when set it drives `effectiveTotal` used by `totalPages`. Falls back to the loaded count when older servers don't return it.
- `clampedPage` in `<ContentList>` — if filters collapse the count below the active page, clamp instead of rendering an empty table while waiting for a refetch.

### Backward compat

- `total` is `optional` everywhere. An older server omitting it leaves the admin's denominator behavior unchanged (grows with loaded items).
- Consumers that destructure `{ items, nextCursor }` keep working — Zod preserves declared optional fields and no existing code paths depend on the shape being closed.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests
- [ ] Chore

## Checklist

[...]

PR #750, opened 2026-04-24
fix(admin): make content list column headers sortable

edrpls:
## What does this PR do?

Turns the admin collection list's column headers (`Title`, `Status`, `Locale`, `Date`) into real sort controls. Previously they were plain `<th>` labels — no click handler, no sort indicator, no `aria-sort`. The backend's `FindManyOptions.orderBy` already supported this; nothing in the admin sent it.

**Before:** clicking a header does nothing. The only ordering is the server's default (publishedAt desc).
**After:** clicking toggles direction, an icon reflects current state, and `aria-sort` on the `<th>` surfaces it to assistive tech.

Split from the (now closed) #749 so this piece can be reviewed on its own. The other two pieces (stable pagination denominator, server-side search) will follow as separate PRs that can be reviewed/merged independently.

### Changes

- **`SortableTh` component** in `packages/admin/src/components/ContentList.tsx`. `<th aria-sort><button>Title ↓</button></th>` — button's accessible name is just the column label; sort state comes from `aria-sort` (which screen readers announce automatically, matching the WAI-ARIA APG sortable-table pattern).
- **Focus-visible outline** on the sort button — kumo resets browser defaults and the button had none.
- **Extended `mapOrderField`** in `packages/core/src/database/repositories/content.ts` to accept `status`, `locale`, and `name` alongside the existing date fields. Unknown fields still throw `EmDashValidationError` — the whitelist remains the defense against column enumeration.
- **Router state** — `ContentListPage` holds sort state and passes it through, plus `orderBy`/`order` in the query key so direction flips invalidate the cursor chain.
- **Backward compat** — callers of `<ContentList>` that don't pass `onSortChange` get the pre-change static-label headers. `ContentPickerModal` is unaffected.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests
- [ ] Chore

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md)
- [x] `pnpm typecheck` passes (admin + core)
- [x] `pnpm lint` passes (no new findings)
- [x] Targeted tests pass: 57 repo tests in `packages/core/tests/database/repositories/content.test.ts`, 37 component tests in `packages/admin/tests/components/ContentList.test.tsx`
- [x] `pnpm format` has been run
[...]

PR #752, opened 2026-04-24
fix(admin): extend server-side content search to the content picker and MCP

edrpls:
## What does this PR do?

Extends the server-side content list search (`?q=`) — which shipped via #1226 — to two surfaces that still post-filtered in memory:

- **`ContentPickerModal`** (used when linking content from the editor) now pushes its search box to the server (`search` option → `?q=`) instead of filtering only the rows already loaded, so it finds entries anywhere in a large collection. It uses `keepPreviousData` so the list doesn't flash to empty between keystrokes, and keeps load-more available while searching (results can span multiple pages).
- **MCP `content_list` tool** gains a `q` parameter, so agents can search a collection server-side rather than post-filtering a single page of results.

This branch was rebased onto `main` after #1226 merged an independent implementation of the core feature. The overlapping core-search work (handler, repository, schemas, `ContentList`, router, API client, index migration) is now provided by #1226; this PR keeps only the picker + MCP extensions on top of it.

Related: #1219, #1226

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

## Screenshots / test output

MCP integration tests pass, including a new case asserting `content_list` filters by `q`:

```
Test Files  1 passed (1)
     Tests  27 passed (27)
```


edrpls:
Rebased on upstream main (`f28eb7b`). Resolved orthogonal conflicts with #750 (sort headers, since merged) — `<ContentList>` now takes both `sort`/`onSortChange` and `searchQuery`/`onSearchChange` props; router threads sort + debounced search through the same TanStack Query key; `fetchContentList` accepts both `orderBy`/`order` and `q`; `describe("orderBy")`, `describe("search")`, and `describe("sortable headers")` test blocks coexist. Locale `.po` files reset to upstream for Lunaria re-extraction.

Local test runs after rebase, on top of `f28eb7b`:
- `packages/core/tests/database/repositories/content.test.ts`: 51/51 pass (includes 6 new `search` tests)
- `packages/core/tests/unit/api/content-handlers.test.ts`: 33/33 pass
- `packages/admin/tests/components/ContentList.test.tsx`: 38/38 pass

Bumping out of stale-warning territory.

edrpls:
Rebased on upstream `main` (aaf021c) to clear conflicts after #751 merged.

**Conflicts resolved** (all orthogonal to #751's now-landed total/pagination machinery — same files, adjacent additions):

- `packages/admin/src/components/ContentList.tsx` — kept `total` + `clampedPage`/`paginatedItems` from main alongside new `searchQuery`/`onSearchChange` props. The auto-fetch effect now bails on client-side search only (`!serverSideSearch && searchQuery`), so server-driven search can still page forward into the filtered set.
- `packages/admin/src/router.tsx` — both `total={total}` and `searchQuery`/`onSearchChange` props pass through; the existing `useInfiniteQuery` key already threads `sort` + `q: debouncedSearch` so cache invalidation is correct.
- `packages/core/src/database/repositories/content.ts` — auto-merged. `findMany` now runs `count(type, where)` in parallel via `Promise.all`, and `count` itself resolves `searchSpec` so the denominator respects an active query.
- `packages/core/tests/database/repositories/content.test.ts` + `tests/unit/api/content-handlers.test.ts` — kept both `describe("total" / "list total")` and `describe("search" / "list search")` blocks side-by-side instead of choosing one.

**Local verification** against `aaf021c`:

| Suite | Result |
|---|---|
| `packages/core` repo + handler tests (`content.test.ts` + `content-handlers.test.ts`) | 88 / 88 |
| `packages/core` dialect tests (`dialect-compat` + `dialect-runtime-imports`) | 17 / 17 |
| `packages/admin` `ContentList.test.tsx` | 39 / 39 |
| `pnpm --filter emdash --filter @emdash-cms/admin typecheck` | clean |

edrpls:
Thanks for the careful review — all four UI edge cases are fixed in 37749f37:

1. **Pagination denominator** — `effectiveTotal` now trusts the server `total` whenever server-side search is active (`serverSideSearch || !searchQuery`). The server's `total` already reflects the filtered set, since `findMany` passes `search` through to `count`, so the denominator no longer collapses to the loaded page size mid-search.
2. **Search input visibility** — the input now stays mounted when there's content **or** an active server-side query (`items.length > 0 || (serverSideSearch && searchQuery)`), so a zero-result search no longer hides the box and traps the user.
3. **Empty-collection vs. no-results** — the "create your first one" CTA is now gated on `items.length === 0 && !searchQuery`; a zero-result server-side search falls through to the "No results for …" message instead.
4. **ContentPickerModal load-more** — dropped the `!searchQuery` guard on the load-more button so paginated search results can be loaded further.

Added three `ContentList` tests covering the server-side denominator, input persistence, and the no-results-vs-empty-collection distinction. Full suite green (42/42).

packages/core/src/database/repositories/content.ts
```diff
@@ -779,10 +820,51 @@ export class ContentRepository {
...
+		rawSearch: string | undefined,
+	): Promise<SearchSpec | null> {
+		if (typeof rawSearch !== "string") return null;
+		const trimmed = rawSearch.trim();
+		if (!trimmed) return null;
+
+		// Clamp the input; a 200-char search is already well past anything
+		// a human would type and stops pathological inputs from reaching SQL.
+		const bounded = trimmed.slice(0, SEARCH_MAX_LENGTH);
+		const pattern = `%${escapeLike(bounded.toLowerCase())}%`;
```
edrpls: **Non-ASCII case-insensitive search is broken on SQLite.** The pattern is lowercased in JS (`bounded.toLowerCase()`, full Unicode) and the column is wrapped in SQL `LOWER()`, but stock `better-sqlite3` `LOWER()` folds **ASCII only** (no ICU/COLLATE configured anywhere in the repo). So a row titled `ÉCOLE` keeps `LOWER('ÉCOLE') = 'ÉCOLE'` on the column side while the pattern becomes `%école%` → the `LIKE` never matches. Users can't find non-English content they can plainly see in the list — notable for a CMS that ships i18n/RTL as a headline feature. The code comment here even assumes "LIKE is already case-insensitive" on SQLite, which is exactly the ASCII-only limitation `toLowerCase()` walks past. Postgres `lower()` also diverges from JS for some codepoints (ß, İ). Consider normalizing both sides consistently (e.g. an FTS-backed path, or `unaccent`/ICU on Postgres and a documented ASCII-only caveat on SQLite).

packages/core/src/database/repositories/content.ts
```diff
@@ -92,6 +93,37 @@ function escapeRegExp(s: string): string {
...
+ * certainly not what a human typed and would just waste a SELECT.
+ */
+const SEARCH_MAX_LENGTH = 200;
+
+/**
+ * Columns that content search targets, in priority order. A column only
+ * participates if it actually exists on the collection's table — every
+ * collection has `slug`, but `title` and `name` are user-defined fields.
+ */
+const SEARCHABLE_COLUMNS = ["title", "name", "slug"] as const;
```
edrpls: **Hardcoding `["title","name","slug"]` silently breaks search for collections whose primary text field is named differently** (e.g. `headline`, `body`, `excerpt`). Those become real TEXT columns via `FIELD_TYPE_TO_COLUMN`, but they're not in this list, so they're never searched — and if a collection has none of title/name/slug as a *matching* column, `resolveSearchSpec` returns `null` and the WHERE is dropped entirely (see separate comment). The codebase already has a first-class notion of searchability: `_emdash_fields.searchable` (migration `012_search.ts`) read via `FTSManager.getSearchableFields()`. This new path reimplements "which columns are searchable" with a divergent hardcoded list that will drift from the schema flag. Consider deriving the searchable columns from the field schema instead of a constant.

packages/admin/src/components/ContentList.tsx
```diff
@@ -111,29 +118,45 @@ export function ContentList({
...
 	// When the server reports a total, it's the source of truth for the
-	// denominator. Otherwise fall back to the size of the (possibly partial)
-	// client list, matching pre-refactor behavior. Client-side search always
-	// defers to `filteredItems` because `total` reflects the unfiltered set.
-	const effectiveTotal = typeof total === "number" && !searchQuery ? total : filteredItems.length;
+	// denominator. In server-side search mode `total` already reflects the
+	// filtered result set, so trust it even while a query is active.
+	// Client-side search always defers to `filteredItems` because there
+	// `total` reflects the unfiltered set.
+	const effectiveTotal =
```
edrpls, line 238: **The row-count line contradicts the pagination denominator in server-side search mode.** `effectiveTotal` (here) correctly uses the server `total`, so pagination shows e.g. `1 / 8` for 143 matches. But `renderItemCount` (~line 533) takes the `if (searchQuery)` branch and returns `plural(filteredCount, …)` where `filteredCount = filteredItems.length` = the *loaded* item count (≤100), so the line reads "100 items matching …" while the pager says 8 pages of 20. The two disagree until every page is fetched. In server-side mode the count line should use `total` (the server's filtered count), same as the denominator — only fall back to `filteredCount` in the client-side path.

packages/admin/src/components/ContentPickerModal.tsx
```diff
@@ -55,13 +55,21 @@ export function ContentPickerModal({ open, onOpenChange, onSelect }: ContentPick
...
+		// refetches, so the list doesn't flash to empty between keystrokes.
+		placeholderData: keepPreviousData,
 	});
 
-	// Sync initial page into accumulated items
+	// Sync initial page into accumulated items. The query re-runs when the
+	// debounced search changes, so we reset the accumulator each time.
 	React.useEffect(() => {
 		if (contentResult) {
 			setAllItems(contentResult.items);
```
edrpls: **A background refetch wipes out load-more results.** This effect resets the accumulator (`setAllItems(contentResult.items)`) on every new `contentResult`. The base query has no `staleTime`, so TanStack defaults apply (`staleTime: 0`, `refetchOnWindowFocus: true`). After the user clicks "Load more" (which appends pages into `allItems`), any refetch of `["content-picker", collection, {q}]` — window refocus, or a content mutation invalidating `["content", …]` while the modal is open — re-settles to a first-page-only `contentResult`, and this effect discards every appended row and resets the cursor. The scrolled-in results just vanish. Consider deriving the displayed list from the query pages (e.g. `useInfiniteQuery`) rather than mirroring into effect-driven state.

edrpls (commented):
Automated high-effort review (`/code-review high`) of the server-side search change. 9 findings below as inline comments, ranked by severity: 2 correctness gaps that matter for an i18n CMS, a count/denominator mismatch, two picker state races, a perf regression on D1, a Postgres dialect-parity issue, a debounce UI flash, and one reuse cleanup. None block the core fix; the non-ASCII and custom-field ones are the most worth addressing before merge.

packages/core/src/database/repositories/content.ts
```diff
@@ -758,9 +798,10 @@ export class ContentRepository {
 	 */
 	async count(
 		type: string,
-		where?: { status?: string; authorId?: string; locale?: string },
+		where?: { status?: string; authorId?: string; locale?: string; search?: string },
 	): Promise<number> {
 		const tableName = getTableName(type);
+		const searchSpec = await this.resolveSearchSpec(tableName, where?.search);
```
edrpls: **Double column-introspection per search request.** `findMany` resolves a `searchSpec` (line 506) and then calls `this.count(type, options.where)` inside its `Promise.all`, and `count` resolves the spec *again* here — each call runs `listTableColumns` (a `pragma_table_info` / `information_schema` query). So a searched list issues data + count + **2×** introspection = 4 round-trips (vs 2 for non-search) on the D1 primary, on every debounced keystroke. The `SearchSpec` doc comment even says the spec exists so list+count "can share the same normalized pattern + column list" — but it isn't shared. Resolve once in `findMany` and pass the spec into `count` (optional param). (The no-search path is fine — `resolveSearchSpec` early-returns before the introspection.)

packages/core/src/database/dialect-helpers.ts
```diff
@@ -160,6 +160,30 @@ export async function listTablesLike(db: Kysely<any>, pattern: string): Promise<
...
+ * List column names for a table. Returns an empty array if the table
+ * does not exist. Used by callers that need to build SQL fragments
+ * referencing columns whose presence varies between collections.
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- accepts any Kysely instance
+export async function listTableColumns(db: Kysely<any>, tableName: string): Promise<string[]> {
+	if (isPostgres(db)) {
+		const result = await sql<{ column_name: string }>`
+			SELECT column_name FROM information_schema.columns
+			WHERE table_schema = 'public' AND table_name = ${tableName}
```
edrpls: **Postgres dialect parity:** this hardcodes `table_schema = 'public'`, but the sibling `columnExists` (line 128) and `indexExists` (line 102) use `current_schema()` — and migration `038` documents this exact footgun. On a Postgres deployment whose content tables live in a non-`public` schema, `listTableColumns` returns `[]` → `resolveSearchSpec` finds no columns → the search WHERE is skipped → a query returns the **entire unfiltered collection** instead of filtered/empty results, silently. (`tableExists`/`listTablesLike` share the same hardcode, so there's mixed precedent, but the safe fix is to match `columnExists`: `table_schema = current_schema()`.)

packages/admin/src/components/ContentPickerModal.tsx
```diff
@@ -76,6 +84,7 @@ export function ContentPickerModal({ open, onOpenChange, onSelect }: ContentPick
 			const result = await fetchContentList(selectedCollection, {
 				limit: 50,
 				cursor: nextCursor,
+				q: searchParam,
```
edrpls: **Load-more can fetch with a mismatched cursor/query.** `searchParam` is derived synchronously from `debouncedSearch`, but `nextCursor` is only updated in the `[contentResult]` effect; with `placeholderData: keepPreviousData`, during a debounced refetch `contentResult` still holds the OLD page so `nextCursor` is stale while `searchParam` is already new. A "Load more" click in that window sends an old-query cursor with the new `q`. Cursors carry no query identity and pagination is order-based, so it won't throw — it just appends an incoherent page (skipped/duplicated rows). Gating load-more on a settled (non-placeholder) result, or keying the cursor to the active query, would close the window.

packages/admin/src/components/ContentList.tsx
```diff
@@ -188,7 +213,10 @@ export function ContentList({
 			</div>
 
 			{/* Search */}
-			{items.length > 0 && (
+			{/* Keep the box mounted whenever there's content OR an active
+			    server-side query — otherwise a zero-result search would hide
+			    the input and trap the user with no way to edit or clear it. */}
+			{(items.length > 0 || (serverSideSearch && searchQuery)) && (
```
edrpls: **Transient flash when clearing a zero-result search.** The router passes the *immediate* `searchQuery` as a prop while `items` is driven by the *300ms-debounced* query, so the two are skewed. After clearing a zero-result query: `searchQuery` becomes `""` instantly but `items` stays `[]` for the debounce window. During that window this mount condition (`items.length > 0 || (serverSideSearch && searchQuery)`) is `false` → the search `<Input>` unmounts/remounts (focus loss), and simultaneously the empty-state at line 304 (`items.length === 0 && !searchQuery`) becomes `true` → the "No … yet. Create your first one" CTA flashes for a non-empty collection. Passing the debounced value (or keying the empty-state on `isFetching`) would avoid the skew. Cosmetic, not data loss.

packages/core/src/database/repositories/content.ts
```diff
@@ -92,6 +93,37 @@ function escapeRegExp(s: string): string {
 	return s.replace(REGEX_ESCAPE_PATTERN, "\\$&");
 }
 
+/**
+ * Escape LIKE wildcard characters so user-supplied search input can't
+ * accidentally act as a wildcard. Paired with `ESCAPE '\\'` at the SQL
+ * call site so the backslash itself works as the escape character.
+ */
+function escapeLike(value: string): string {
```
edrpls: **Cleanup (reuse):** this `escapeLike` is byte-for-byte identical to the ones already in `repositories/media.ts` (line 9) and `repositories/options.ts` (line 5) — `value.replaceAll("\\","\\\\").replaceAll("%","\\%").replaceAll("_","\\_")` — and `redirect.ts`/`comment.ts` implement the same escape via regex. That's now 3 verbatim copies + 2 equivalents of a security-sensitive routine. Worth extracting one shared helper (e.g. exported from `repositories/types.ts` or `dialect-helpers.ts`) so a future fix lands in one place.

edrpls:
@emdashbot — ran a high-effort multi-agent review on this branch and pushed fixes for the two low-risk findings. Summary below, with a couple of design calls that need a maintainer decision before I go further.

### ✅ Fixed in `d087bc86`
- **Search count line contradicted the pagination denominator.** In server-side mode `renderItemCount` reported the *loaded* item count ("20 items matching…") while the pager used the server `total` ("1 / 8"). Now uses `total` for the match count in server-side mode; client-side mode still uses the local filtered count. Test added.
- **Double table-column introspection per search.** `findMany` resolved the search predicate and then `count` resolved it again, so each searched request ran `listTableColumns` twice (4 D1 round-trips instead of 2). `count` now takes an optional pre-resolved `searchSpec` and `findMany` passes its own through — introspection runs once.

### ❓ Need a maintainer decision (design altitude)
1. **Non-ASCII case-insensitive search is broken on SQLite.** The pattern is lowercased in JS (full Unicode) but the column uses SQLite's `LOWER()`, which folds ASCII only (no ICU/COLLATE configured). A row titled `ÉCOLE` won't match `école`. For a CMS that ships i18n/RTL as a headline feature this seems significant — but the fix (FTS-backed search, or ICU/`unaccent` on PG + a documented SQLite caveat) is a bigger call than I want to make unilaterally. **Worth blocking the merge, or acceptable as a known limitation + follow-up?**
2. **Search only targets hardcoded `["title","name","slug"]`,** ignoring the existing per-field `_emdash_fields.searchable` flag (read elsewhere via `FTSManager.getSearchableFields`). A collection whose primary text field is `headline`/`body` is silently unsearchable. **Should the list `q` filter derive searchable columns from the field schema instead of a constant?** (This also overlaps conceptually with the FTS layer — want them unified or kept separate?)

### 🟡 Lower-priority findings (left for your call — happy to fix any in this PR)
- **Postgres dialect parity:** `listTableColumns` hardcodes `table_schema = 'public'` while sibling `columnExists`/`indexExists` use `current_schema()` (migration 038 documents this footgun). On a non-public-schema PG deployment, search silently returns the *entire unfiltered collection*. One-line fix; flagging since `tableExists`/`listTablesLike` share the hardcode (mixed precedent).
- **ContentPickerModal: load-more results wiped by a background refetch** (no `staleTime`; refocus/invalidation re-settles to page 1 and the accumulator-reset effect discards appended rows). Cleaner as `useInfiniteQuery`.
- **ContentPickerModal: load-more cursor/query mismatch** during the 300ms debounce window (old cursor + new `q`).
- **ContentList: search-box unmount + empty-collection CTA flash** for ~300ms after clearing a zero-result query (router passes the immediate `searchQuery` while `items` lag on the debounce). Cosmetic.
- **Cleanup:** `escapeLike` is duplicated verbatim in `media.ts` and `options.ts` (+ regex equivalents in `redirect.ts`/`comment.ts`) — worth one shared helper.

Full details are in the inline comments on review [#4405665999](https://github.com/emdash-cms/emdash/pull/752#pullrequestreview-4405665999). Let me know how you'd like to handle #1 and #2 and I'll proceed.

packages/admin/src/components/ContentPickerModal.tsx
edrpls: Fixed in c4cb0e99 — the picker now derives its list from `useInfiniteQuery` pages instead of mirroring into an `allItems`/`nextCursor` accumulator. A background refetch can no longer drop loaded pages, and `search` lives in the query key so a search change starts a fresh page chain rather than firing load-more with a stale cursor. Resolving.

packages/admin/src/components/ContentPickerModal.tsx
edrpls: Fixed in c4cb0e99 — the picker now derives its list from `useInfiniteQuery` pages instead of mirroring into an `allItems`/`nextCursor` accumulator. A background refetch can no longer drop loaded pages, and `search` lives in the query key so a search change starts a fresh page chain rather than firing load-more with a stale cursor. Resolving.

packages/core/src/database/repositories/content.ts
edrpls: Outdated — superseded by #1226, which resolves exactly this: its `resolveSearchColumns` introspects the collection's `_emdash_fields` rather than hardcoding a column list, so non-standard primary text fields are searchable. This branch no longer touches that code. Resolving.

packages/core/src/database/repositories/content.ts
edrpls: Outdated — this branch was rebased onto `main` after #1226 merged an independent implementation of the core server-side search. The flagged code is no longer part of this PR; the handler/repository/ContentList search now ship via #1226. Resolving.

packages/core/src/database/repositories/content.ts
edrpls: Outdated — this branch was rebased onto `main` after #1226 merged an independent implementation of the core server-side search. The flagged code is no longer part of this PR; the handler/repository/ContentList search now ship via #1226. Resolving.

packages/core/src/database/repositories/content.ts
edrpls: Outdated — this branch was rebased onto `main` after #1226 merged an independent implementation of the core server-side search. The flagged code is no longer part of this PR; the handler/repository/ContentList search now ship via #1226. Resolving.

packages/core/src/database/dialect-helpers.ts
edrpls: Outdated — this branch was rebased onto `main` after #1226 merged an independent implementation of the core server-side search. The flagged code is no longer part of this PR; the handler/repository/ContentList search now ship via #1226. Resolving.

packages/admin/src/components/ContentList.tsx
edrpls: Outdated — this branch was rebased onto `main` after #1226 merged an independent implementation of the core server-side search. The flagged code is no longer part of this PR; the handler/repository/ContentList search now ship via #1226. Resolving.

packages/admin/src/components/ContentList.tsx
edrpls: Outdated — this branch was rebased onto `main` after #1226 merged an independent implementation of the core server-side search. The flagged code is no longer part of this PR; the handler/repository/ContentList search now ship via #1226. Resolving.

edrpls:
**Update — scope narrowed after rebase.** Since my review note above, #1226 merged an independent implementation of the core server-side content search (`?q=`), including the locale-aware list index migration. I've rebased this branch onto `main` and dropped the overlapping core work; this PR is now just the two surfaces #1226 didn't cover:

- **`ContentPickerModal`** — pushes its search box to the server instead of in-memory filtering.
- **MCP `content_list`** — gains a `q` parameter so agents search server-side.

That moots most of the items above. Re-mapping each:

### Design calls — now #1226's territory
1. **Non-ASCII case-insensitive search on SQLite** — no longer in this PR; the core search ships via #1226. If the ASCII-only `LOWER()` folding is still worth addressing, it's a follow-up against #1226's implementation rather than a blocker here. Happy to open a separate issue.
2. **Hardcoded `["title","name","slug"]` vs the field schema** — #1226 partly addresses this: its `resolveSearchColumns` derives columns from `_emdash_fields` (adds `title`/`name` only when the collection defines them) rather than a flat constant. Whether to further key off the `searchable` flag / unify with the FTS layer is best tracked against that code.

### Lower-priority findings
- **Postgres `table_schema = 'public'` hardcode** — moot; the `listTableColumns` helper that had it was part of the discarded implementation and is gone.
- **`ContentPickerModal` load-more wiped by background refetch** ✅ **fixed** in `c4cb0e99` — the picker now derives its list from `useInfiniteQuery` pages instead of a manual accumulator, so a refocus/invalidation can't drop loaded pages.
- **`ContentPickerModal` load-more cursor/query mismatch** ✅ **fixed** in the same commit — `search` lives in the query key, so a search change starts a fresh page chain rather than firing load-more with a stale cursor.
- **`ContentList` zero-result CTA flash** — moot; that component's search now ships via #1226.
- **`escapeLike` duplication** — moot here (the copy was in the discarded code); still a reasonable standalone cleanup if anyone wants to pick it up.

All inline threads have been resolved accordingly. No maintainer decision is blocking this PR anymore — it's a focused follow-up on top of #1226. 🤖 update drafted with Claude Code.

packages/core/src/mcp/server.ts
```diff
@@ -555,6 +555,12 @@ export function createMcpServer(): McpServer {
 					.string()
 					.optional()
 					.describe("Filter by locale (e.g. 'en', 'fr'). Only relevant when i18n is enabled."),
+				q: z
+					.string()
+					.trim()
```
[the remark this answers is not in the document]
edrpls: Applied in 5c80fee9 — the MCP `q` schema now carries `.min(1)` after `.trim()`, matching `contentListQuery`.

Following up in 37040d3f with the regression test the constraint was missing: `content_list` with `q: "   "` now asserts an error response. Verified it fails without the `.min(1)` (the blank `q` is silently accepted and treated as "no search"), so the two interfaces can't drift apart again unnoticed.

packages/core/tests/integration/mcp/content-misc.test.ts
```diff
@@ -508,6 +508,25 @@ describe("soft-delete visibility", () => {
 		expect(items).toHaveLength(1);
 		expect(items[0]?.id).toBe(extractJson<{ item: { id: string } }>(b).item.id);
 	});
+
+	it("content_list filters by the q search parameter", async () => {
```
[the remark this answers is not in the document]
edrpls: Applied in 5c80fee9 — the search test moved out of `soft-delete visibility` into its own `describe("content_list search")` block with its own harness setup, so a failure names the behavior it covers. The file's header comment lists it alongside the other groups.

37040d3f adds a second case to that block for the blank-`q` rejection.

edrpls:
Merged current `main` (36e9e664) — the branch was 2 commits behind after the last sync.

On the two red E2E checks from the previous run: they're the `invite-flow` spec (`user appears in the users list` — `text=invited-user@example.com` not visible within 15s), which this PR cannot plausibly affect. The diff against `main` is 4 files — `ContentPickerModal.tsx`, `mcp/server.ts`, the MCP test, and the changeset — and touches nothing in the users, invite, or auth path. Main's own CI is green on 11 of its last 12 runs, and the same E2E suite passes on my other open PRs, so this reads as a flaky/timing failure rather than a regression. I don't have rerun rights on the repo, so this merge doubles as a fresh CI trigger — if the shard goes green on this run that settles it.

Local verification on the merged head: full `packages/core` suite green (5094 passing), MCP integration suite 321 passing, lint and typecheck clean for both `emdash` and `@emdash-cms/admin`.

edrpls:
@ascorbic gentle ping on this one — it's been open since April and is in a ready state: rebased onto current main, CI fully green, both emdashbot review passes addressed (the second concluded "clean and ready to land after two nits", and those nits landed with regression tests). Zero unresolved threads.

If you'd rather route it through the bot first, I can't apply `bot:review` myself (no label permissions) — reapplying it would kick a fresh pass. Happy to rebase again if it drifts before you get to it.

edrpls:
All three findings from the fresh review pass are addressed in 12f892e0:

- **Picker empty-state** now gates on the effective `searchParam` — a whitespace-only query no longer claims a search is filtering.
- **Narrative JSX comment** above the load-more condition removed.
- **MCP test header** now lists the `content_list` search-by-`q` coverage.

Verified locally: lint clean, MCP suite 29/29, full admin browser suite 1363/1363.

PR #755, opened 2026-04-24
fix(import): invalidate manifest cache after wp prepare creates schema

### LINKED ISSUE #747: WordPress importer creates new collections but does not invalidate manifest cache — subsequent items fail with `Collection "<slug>" does not exist`
### Description

When importing a WXR file that contains a custom post type the site doesn't already have as a collection (e.g. `tablepress_table`), the importer successfully creates the collection + content table + field registry, but **does not call `invalidateManifest()`**. Every subsequent item destined for that collection then fails with `Collection "<slug>" does not exist` — because the check at [`execute.ts:176`](https://github.com/emdash-cms/emdash/blob/main/packages/core/src/astro/routes/api/import/wordpress/execute.ts#L176) reads from a stale manifest.

The stale manifest is persisted to the `options` table under `emdash:manifest_cache`, so the bug **survives `npx emdash dev` restarts and `emdash types` regeneration**. The only recovery is to delete the row manually or to re-run the same import a second time (which is what bisected it for me).

### Steps to reproduce

1. `npm create emdash@latest` → `starter` template → `npx emdash init`
2. Complete admin setup → open the WordPress importer
3. Upload a WXR containing at least one custom post type *not* already in the default schema (e.g. `tablepress_table`)
4. Check the custom post type on the Review step → import

**Expected:** items land in the newly-created collection.
**Actual:** Schema Changes section reports `Collections created: tablepress_table` **and** Content Errors shows N × `Collection "tablepress_table" does not exist` — one per item.

Re-running the same import (even with only the failing collection selected, even after server restart) reproduces the same N errors. Only clearing `emdash:manifest_cache` manually + restarting resolves it:

```sql
DELETE FROM options WHERE name = 'emdash:manifest_cache';
```

### Environment

- EmDash `0.6.0` (via `create-emdash@0.6.0`, `starter` template, Node / SQLite / local filesystem)
- Node 24.14.1, npm 11.11.0
- WXR: several thousand items, including a custom post type (`tablepress_table`, 
  from the TablePress plugin) not present in the default schema

### Logs / error output

```shell
[`execute.ts:176`](https://github.com/emdash-cms/emdash/blob/main/packages/core/src/astro/routes/api/import/wordpress/execute.ts#L176) gates item writes on `manifest.collections[collection]`. The manifest is produced by `EmDashRuntime.getManifest()` and is backed by a DB-persisted cache keyed on a config fingerprint (see #606, #616). The fingerprint does not change when the schema registry is mutated mid-request, so post-creation reads return the pre-creation cached object.

The importer needs to call `runtime.invalidateManifest()` (or the equivalent) after creating new collections/fields, before iterating items. The same fix applies to the `wordpress-plugin` variant at [`wordpress-plugin/execute.ts:190`](https://github.com/emdash-cms/emdash/blob/main/packages/core/src/astro/routes/api/import/wordpress-plugin/execute.ts#L190).

## Suggested fix

After the schema-creation phase of the import completes (and before the item-writing loop begins), call `runtime.invalidateManifest()` so the subsequent `getManifest()` rebuild picks up the new collections.

## Workaround for current users


DELETE FROM options WHERE name = 'emdash:manifest_cache';

Then restart the dev server (`npx emdash dev`) and re-run the import with only the failing collection selected. The retry will succeed because the rebuilt manifest now includes the collection.
```

mvanhorn:
## What does this PR do?

Fix the WordPress importer so freshly created collections are visible to the subsequent `execute` phase.

`POST /_emdash/api/import/wordpress/prepare` creates collections and fields via `SchemaRegistry`, but never invalidated the manifest. The manifest is backed by a DB-persisted cache in the `options` table (`emdash:manifest_cache`), so when the browser then called `POST .../execute` — a separate request — that request read the **pre-creation** manifest from the DB cache and reported `Collection "<slug>" does not exist` for every item destined for the newly created collection. The bug survived `npx emdash dev` restarts because the cache was on disk, not in memory.

Fix is minimal: call `emdash.invalidateManifest()` from the `prepare` POST handler when `prepareImport` reports any `collectionsCreated` or `fieldsCreated`. Skipped when prepare makes no schema changes, so runs that only tweak existing mappings don't thrash the cache.

Closes #747

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (or targeted tests for my change) — `pnpm test tests/unit/import/` → 142 tests pass including the two new cases
- [x] I have added/updated tests for my changes (if applicable) — new `tests/unit/import/wp-prepare-invalidate.test.ts`, TDD-verified by reverting the fix and watching the test fail
- [ ] User-visible strings in the admin UI are wrapped for translation — N/A, no UI strings
- [x] I have added a changeset (`.changeset/fix-wp-import-invalidate-manifest.md`, `emdash` patch)
- [ ] New features link to an approved Discussion — N/A, bug fix

## Test output

```
$ pnpm test tests/unit/import/wp-prepare-invalidate.test.ts
 ✓ tests/unit/import/wp-prepare-invalidate.test.ts (2 tests) 45ms
   ✓ POST /api/import/wordpress/prepare
     ✓ invalidates the manifest cache after creating a new collection (regression for #747)
     ✓ does not invalidate the manifest when prepareImport makes no schema changes

$ pnpm test tests/unit/import/
 Test Files  7 passed (7)
 Tests       142 passed (142)
```

Regression check: removing the `emdash.invalidateManifest()` call makes the first new test fail with `expected "spy" to be called 1 times, but got 0`.

## Scope notes

The issue also mentions `packages/core/src/astro/routes/api/import/wordpress-plugin/execute.ts:190` as an "apply the same fix" site. That path is different: the plugin variant creates **fields** inline during `execute`, not **collections**, and the check at that line (`manifest?.collections[collection]`) isn't affected by field creation. If field-level manifest staleness is also a bug there, it's a separate issue — I kept this PR scoped to the reproducer in the issue body.

### DIFF
--- a/.changeset/fix-wp-import-invalidate-manifest.md
+++ b/.changeset/fix-wp-import-invalidate-manifest.md
@@ -1,0 +1,7 @@
+---
+"emdash": patch
+---
+
+Fixes the WordPress importer so collections created mid-import are visible to the subsequent execute phase.
+
+`POST /_emdash/api/import/wordpress/prepare` now calls `emdash.invalidateManifest()` when it creates new collections or fields. Without this, the DB-persisted manifest cache (`emdash:manifest_cache` in the `options` table) stays stale and the `execute` request reports `Collection "<slug>" does not exist` for every item destined for a freshly created collection — a bug that survived dev-server restarts and required manually deleting the cache row.
--- a/packages/core/src/astro/routes/api/import/wordpress/prepare.ts
+++ b/packages/core/src/astro/routes/api/import/wordpress/prepare.ts
@@ -58,6 +58,16 @@ export const POST: APIRoute = async ({ request, locals }) => {
 		// eslint-disable-next-line typescript-eslint(no-unsafe-type-assertion) -- Zod schema output narrowed to PrepareRequest
 		const result = await prepareImport(emdash.db, body as PrepareRequest);
 
+		// If prepare created any new collections or fields, invalidate the
+		// persisted manifest cache (`emdash:manifest_cache` in the options
+		// table) so that the execute endpoint -- a separate request -- sees
+		// the new schema. Without this the execute step reads a stale
+		// manifest and reports `Collection "<slug>" does not exist` for
+		// every item destined for a freshly-created collection. See #747.
+		if (result.collectionsCreated.length > 0 || result.fieldsCreated.length > 0) {
+			emdash.invalidateManifest();
+		}
+
 		return apiSuccess(result, result.success ? 200 : 400);
 	} catch (error) {
 		return handleError(error, "Failed to prepare import", "WXR_PREPARE_ERROR");
--- a/packages/core/tests/unit/import/wp-prepare-invalidate.test.ts
+++ b/packages/core/tests/unit/import/wp-prepare-invalidate.test.ts
@@ -1,0 +1,89 @@
+/**
+ * Regression test for #747: WordPress importer must invalidate the manifest
+ * cache after creating new collections/fields. Without this the execute
+ * endpoint reads a stale DB-persisted manifest and reports
+ * `Collection "<slug>" does not exist` for every item destined for a freshly
+ * created collection.
+ */
+
+import { describe, expect, it, vi } from "vitest";
+
+import { POST } from "../../../src/astro/routes/api/import/wordpress/prepare.js";
+import { setupTestDatabase } from "../../utils/test-db.js";
+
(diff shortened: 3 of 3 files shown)


ascorbic (approved):
Thanks!

PR #800, opened 2026-04-27
fix(auth,core): accept multiple origins for passkey verification

### LINKED ISSUE #799: Passkey verification rejects valid origins under shared rpId (multi-subdomain deployments)
### Description

EmDash's passkey verification is strictly bound to a single origin via `siteUrl`, which makes it impossible to verify a passkey when the deployment is reachable under multiple hostnames sharing a registrable parent domain — a topology WebAuthn explicitly supports.

The most visible failure: the setup wizard rejects valid passkey registrations when `EMDASH_SITE_URL` is set to the apex (e.g. `https://example.com`) but the wizard is opened on a subdomain (e.g. `https://preview.example.com`). The browser correctly registers the passkey with `rpId=example.com` (a valid registrable parent), but EmDash's strict origin-equality check rejects the response.

Same issue blocks subsequent passkey logins after switching subdomains, even though the credential itself is still valid for the rpId.

### Reproduction

1. Deploy `templates/blog-cloudflare` to a Cloudflare Worker
2. Add `preview.example.com` as a Custom Domain (in addition to `example.com`)
3. Set `EMDASH_SITE_URL=https://example.com` in `wrangler.jsonc`
4. Open `https://preview.example.com/_emdash/admin/setup`
5. Run the setup wizard
6. Step 3 (passkey registration) returns 500 "Failed to verify admin setup"
7. Server log: `Invalid origin: expected https://example.com, got https://preview.example.com`

### Expected behaviour

`siteUrl` defines the canonical origin and `rpId`; passkey verification should accept any explicitly-allowed origin that shares that `rpId`. WebAuthn allows this — operator just needs a way to express "these other origins are also valid."

### Root cause

`packages/auth/src/passkey/{register,authenticate}.ts` does:

```ts
if (clientData.origin !== config.origin) {
  throw new Error(`Invalid origin: expected ${config.origin}, got ${clientData.origin}`);
}
```

`PasskeyConfig.origin` is a single string derived from `siteUrl` in `packages/core/src/auth/passkey-config.ts`. There is no escape hatch.

### Suggested fix

1. Replace `PasskeyConfig.origin: string` with `PasskeyConfig.origins: string[]`. First entry is canonical; verification accepts any entry via `.includes()`.
2. Add a 4th optional `allowedOrigins?: string[]` parameter to `getPasskeyConfig()`.
3. Add `EMDASH_ALLOWED_ORIGINS` (comma-separated list) as the runtime knob, mirroring the existing `EMDASH_SITE_URL` pattern in `packages/core/src/api/public-url.ts`.
4. Plumb it through verify routes only (the options-generating routes don't read origins).

This is a fix-side change only: existing single-origin deployments are unaffected (the array contains one origin) and `siteUrl` continues to control `rpId`.

### Workaround for now

Don't set `siteUrl`. Passkey gets bound to the request hostname (`preview.example.com` or whatever subdomain the user is on). Re-register the passkey on each origin you need to use. Same workaround as the one mentioned in #409.

### Files involved

- `packages/auth/src/passkey/types.ts` — `PasskeyConfig`
- `packages/auth/src/passkey/register.ts:124` — origin equality check
- `packages/auth/src/passkey/authenticate.ts:159` — origin equality check
- `packages/core/src/auth/passkey-config.ts` — `getPasskeyConfig()`
- `packages/core/src/api/public-url.ts` — env-var resolution helpers
- 5 verify route handlers under `packages/core/src/astro/routes/api/`

### Related

- Discussion #511 — same pattern, unanswered
- Discussion #409 — workaround documented (re-register per origin)
- PR #382 — introduced `siteUrl` for single-origin reverse-proxy use; multi-origin wasn't in scope at the time

### Environment

- emdash: 0.7.0
- @emdash-cms/auth: 0.7.0
- @emdash-cms/cloudflare: 0.7.0
- Adapter: `@astrojs/cloudflare`
- Astro: 6.x
- Deployment: Cloudflare Workers + Custom Domains via Cloudflare DNS

csfalcao:
## What does this PR do?

Lifts the single-origin restriction on passkey verification so deployments serving the same EmDash instance from multiple hostnames (apex + preview/staging subdomains, all sharing one `rpId`) can verify passkeys without re-registering on each origin.

The strict equality check `clientData.origin !== config.origin` in `register.ts` and `authenticate.ts` is replaced with a list-membership check. `PasskeyConfig.origin: string` becomes `origins: string[]`. A new `EMDASH_ALLOWED_ORIGINS` env var (comma-separated) lets operators declare the additional accepted origins; `siteUrl` continues to control `rpId`.

Closes #799

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes (12 pre-existing diagnostics on main, none added by this PR)
- [x] `pnpm test` passes for affected packages (`@emdash-cms/auth` 49/49; `emdash` core matches main baseline — no regressions added; `@emdash-cms/admin` failure is pre-existing on main and unrelated to passkey auth)

## Implementation notes

**Type change.** `packages/auth/src/passkey/types.ts` now declares `PasskeyConfig.origins: string[]` instead of `origin: string`. Per the project's pre-release rules (no backward-compat, no migration paths), the rename is direct.

**Verification.** `register.ts` and `authenticate.ts` both replace `clientData.origin !== config.origin` with `!config.origins.includes(clientData.origin)`. Same security posture — origin still has to be in an explicit list — just multi-entry instead of single.

**Helper.** `getPasskeyConfig()` in `packages/core/src/auth/passkey-config.ts` gains a fourth optional argument `allowedOrigins?: string[]`. The canonical origin (from `siteUrl` or `url.origin`) goes first; allowed origins follow, deduplicated; empty/falsy entries are dropped.

**Env var.** `getEnvAllowedOrigins()` in `packages/core/src/api/public-url.ts` mirrors the existing `getEnvSiteUrl()` pattern: reads `process.env.EMDASH_ALLOWED_ORIGINS`, parses each comma-separated entry through `new URL()`, normalizes to `origin`, drops non-http(s)/unparseable entries, caches.

**Plumbed only where it's read.** Only the 5 verify routes (`setup/admin-verify`, `auth/passkey/verify`, `auth/passkey/register/verify`, `auth/signup/complete`, `auth/invite/complete`) pass `getEnvAllowedOrigins()` into `getPasskeyConfig`. The 4 options-generating routes were left alone — they don't read `origins`.

**Tests.** Following "TDD for bugs":

- `packages/auth/src/passkey/authenticate.test.ts` — cross-subdomain reject and accept tests (the reject case fails on the old code; the accept case is the fix's reason for existing). `register.ts` is not separately tested because its origin-check is the identical 3-line block — the auth tests cover the shared semantics.
- `packages/core/tests/unit/auth/passkey-config.test.ts` — 5 new tests covering append, canonical-first when no `siteUrl`, dedupe of repeated canonical, empty-entry skip, empty `allowedOrigins`. Existing assertions migrated from `config.origin` to `config.origins[0]`.
- `packages/core/tests/unit/api/public-url.test.ts` — 7 new tests covering `getEnvAllowedOrigins()`: unset, comma-list, whitespace trim, origin normalization, protocol filter, unparseable skip, cache invalidation.

**Docs.** `docs/src/content/docs/reference/configuration.mdx` gets a new "Multi-origin passkey verification" subsection under Site URL, plus an `EMDASH_ALLOWED_ORIGINS` row in the Environment Variables table.

## Test output

```
@emdash-cms/auth — 49 tests pass
  src/passkey/authenticate.test.ts (4 tests)  ✓ rejects an origin that is not in the accepted list
                                              ✓ accepts an assertion from a subdomain when its origin is listed under a shared rpId

emdash (core)
  tests/unit/auth/passkey-config.test.ts      ✓ 22 tests (was 17)
  tests/unit/api/public-url.test.ts           ✓ 21 tests (was 14)
```


packages/core/src/api/public-url.ts
```diff
@@ -70,10 +68,48 @@ function getEnvSiteUrl(): string | undefined {
...
+ * Reads `EMDASH_ALLOWED_ORIGINS` (comma-separated list of origins) for
+ * multi-origin deployments where the same RP is reachable under several
+ * hostnames sharing the registrable parent domain (e.g. apex + preview).
+ *
+ * Each entry is parsed via `new URL()` and reduced to its `origin`. Entries
+ * with non-http(s) protocols or unparseable values are silently skipped.
+ *
+ * Like `getEnvSiteUrl`, uses `process.env` (Vite leaves it untouched at
+ * runtime) and is unavailable on Cloudflare Workers, where `env` bindings
+ * provide values instead.
```
ascorbic: process.env *is* available on Workers, because we have nodejs_compat enabled

packages/core/src/api/public-url.ts
```diff
@@ -70,10 +68,48 @@ function getEnvSiteUrl(): string | undefined {
...
+ *
+ * Caches after first call.
+ */
+let _envAllowedOrigins: string[] | null = null;
+
+export function getEnvAllowedOrigins(): string[] {
```
ascorbic, line 97: Can we do the same as siteUrl and allow passing the value in config too.

packages/core/src/auth/passkey-config.ts
```diff
@@ -18,27 +23,44 @@ export interface PasskeyConfig {
  * @param url The request URL (typically `new URL(Astro.request.url)` or `new URL(request.url)`)
  * @param siteName Optional site name for rpName (defaults to hostname from `url` or public origin)
  * @param siteUrl Optional browser-facing origin (see `EmDashConfig.siteUrl`).
- *        When set, **origin** and **rpId** are taken from this URL so they match WebAuthn `clientData.origin`.
+ *        When set, the canonical **origin** and **rpId** are taken from this URL.
+ * @param allowedOrigins Optional list of additional accepted origins for verification.
+ *        Each must share `rpId` with the canonical origin (WebAuthn requirement).
+ *        Typical use: apex + preview subdomain on the same registrable domain.
  * @throws If `siteUrl` is non-empty but not parseable by `new URL()`.
```
ascorbic, line 30: We need to validate this: require siteUrl to be set, and for all allowedOrigins to be subdomains.

ascorbic (commented):
Thanks! Good addition, I just have a few comments

packages/core/src/api/public-url.ts
csfalcao: Fixed in V2 (f580326) — dropped the wrong claim from `getEnvAllowedOrigins`'s docstring.

Heads-up: the same incorrect claim still sits in the pre-existing `getEnvSiteUrl` docstring (`public-url.ts` lines 22-23). Out of scope for this PR per the no-drive-by rule, but happy to fold that one-line fix in if you'd prefer — let me know.

packages/core/src/api/public-url.ts
csfalcao: Done in V2 (f580326). Added `EmDashConfig.allowedOrigins?: string[]` next to `siteUrl` in `integration/runtime.ts`. Config and `EMDASH_ALLOWED_ORIGINS` are merged at runtime (rather than precedence-style like `siteUrl`) — the rationale: each entry is an independent permission, so list-union is the natural operation. This lets versioned `astro.config.mjs` declare stable origins while env adds environment-specific extras (e.g. ephemeral PR-preview URLs).

Plumbing follows your earlier architectural note about env vars being runtime: `getConfiguredAllowedOrigins(config)` (in the new `auth/allowed-origins.ts`) merges both sources tagged-by-source, then `validateAllowedOrigins(siteUrl, tagged)` validates the effective set at the runtime chokepoint. Config-side validation also runs at integration init — but only the shape rules (parseable, http/https, no trailing dots, no empty labels), since `siteUrl` may come from `EMDASH_SITE_URL` post-build. When `config.siteUrl` IS present at init, the full validator runs for fail-fast.

packages/core/src/auth/passkey-config.ts
csfalcao: Done in V2 (f580326). Validation lives in `auth/allowed-origins.ts` as a shared module with two entry points:

- `validateOriginShape(tagged)` — per-entry rules only (parseable URL, http/https, no trailing-dot hostname, no empty labels). Used by integration init when `config.siteUrl` is absent (since env can supply it post-build).
- `validateAllowedOrigins(siteUrl, tagged)` — full validation, including Rule A (`allowedOrigins` non-empty ⇒ `siteUrl` required), Rule B (each origin must be the same hostname as `siteUrl` or a subdomain — leading-dot suffix check, so `fakeexample.com` and `example.com.evil.com` both reject), and rejecting IP-literal `siteUrl` or trailing-dot `siteUrl` when `allowedOrigins` is non-empty.

Errors are source-attributed (`config.allowedOrigins` vs `EMDASH_ALLOWED_ORIGINS`) so operators see where the offending value came from. Coverage: 27 unit tests in `tests/unit/auth/allowed-origins.test.ts` — including suffix-attacker (`example.com.evil.com`), prefix-attacker (`fakeexample.com`), apex-listed-under-subdomain-siteUrl, IP-literal v4/v6, dedupe across sources, and source-attribution-of-the-offending-entry.

Two related judgment calls I held off on, both worth flagging:
- **`http://` rejection for non-localhost origins**: the browser refuses `http://` WebAuthn outside localhost anyway, so listing one is dead config. Adding ~5 lines of rejection would catch it at startup. Held off because it goes beyond your ask — happy to add if you want it.
- **Trailing-dot rejection in `siteUrl` outside the multi-origin path**: same class of mismatch, but pre-existing concern beyond this PR. Could be a separate one-liner in `integration/index.ts`'s siteUrl validator.

csfalcao:
## V2 pushed (f580326)

Addresses all three of your comments. Summary:

**Architecture changes**
- New `packages/core/src/auth/allowed-origins.ts` module — pure resolution + validation, source-tagged
- `getPasskeyConfig()` stays a pure builder (validation moved out)
- Validation runs at the runtime chokepoint (5 verify routes); config-side shape rules also run at integration init when `config.siteUrl` is present

**Surface additions**
- `EmDashConfig.allowedOrigins?: string[]` (declared in `astro.config.mjs`)
- Existing `EMDASH_ALLOWED_ORIGINS` env var still works; merges with config
- `getEnvAllowedOrigins()` now throws on invalid entries (was: silent skip — wrong for an allowlist)
- Internal `_resetEnvSiteUrlCache` → `_resetEnvCache` (it now resets both caches)

**Validation rules** (all with source-attributed errors)
- Per-entry: parseable URL, http/https only, no trailing-dot hostname, no empty labels in hostname
- Rule A: `allowedOrigins` non-empty ⇒ `siteUrl` required
- Rule B: each origin must be the same hostname as `siteUrl` or a true subdomain (leading-dot suffix check)
- Reject IP-literal `siteUrl` and trailing-dot `siteUrl` when `allowedOrigins` is non-empty

**Test coverage**
- 27 new tests in `tests/unit/auth/allowed-origins.test.ts` (security cases: suffix-attacker, prefix-attacker, apex-under-subdomain, IP literals, source attribution, dedupe across sources)
- New `packages/auth/src/passkey/register.test.ts` locks in origin-check parity with `authenticate.ts` (Bot review #4)
- Updated `public-url.test.ts` for the rename + new throw cases (21/21 ✓)

**Local verification**
- `pnpm --filter emdash typecheck` ✓
- `pnpm --filter @emdash-cms/auth typecheck` ✓
- `pnpm --filter @emdash-cms/auth test` 5/5 files ✓
- `pnpm --filter emdash test` matches main baseline (2 pre-existing failures in `invite.test.ts` and `passkey-verify-route.test.ts`, none added by V2)
- `pnpm --silent lint:json` 12 diagnostics (one fewer than V1; nothing new in files I touched)
- `pnpm format` applied

Replied inline to each of your three comments above with where the change landed and the rationale where I deviated (merge vs precedence for `allowedOrigins`, validation placement).

ascorbic (approved):
Thanks!

PR #803, opened 2026-04-28
fix: make migrations 034 and 035 idempotent for partial reapply

### LINKED ISSUE #788: Migration runner crashes on partially-applied schema and the middleware silently degrades the admin to a permissionless "ghost" user
### Description

Upgrading from 0.1.1 to 0.6.0 (and again from 0.1.1 to 0.7.0 after a rollback) left the production D1 database in a state where the schema changes from migrations 034_published_at_index and 035_bounded_404_log were applied, but the corresponding rows in _emdash_migrations were not present. From that point onward, every admin request to the deployed Worker fails with:

`EmDash middleware error: Error: Migration failed: D1_ERROR: index idx_ec_about_page_deleted_published_id already exists: SQLITE_ERROR (migration: 034_published_at_index)`

…and the auth middleware swallows the error in a way that renders the admin in a confusing degraded state instead of surfacing a clear failure.

### Steps to reproduce

I cannot reproduce reliably from a clean install — it appears to depend on something interrupting the original 0.1.1 → 0.6.0 / 0.7.0 migration run mid-way. Likely candidates: a transient D1 error during one of the migrations, or a Worker subrequest/CPU limit while the runner is processing a long batch. Once the system is in this state, however, it stays stuck across redeploys until the missing rows are written manually.

Workaround
Once the schema changes are confirmed already present, manually re-record the migrations:

`INSERT INTO _emdash_migrations (name, timestamp) VALUES
  ('034_published_at_index', '2026-04-27T16:00:00.000Z'),
  ('035_bounded_404_log',    '2026-04-27T16:00:00.000Z');`
After this, the middleware error stops, auth resolves, and admin loads normally.

Suggested fixes

- Make the database updates safe to run more than once. For example, only create an index or add a column if it doesn’t already exist. That way, if something only half-finished before, rerunning it will fix things instead of breaking.

### Environment

- Started on emdash 0.1.1, attempted upgrade to 0.6.0, rolled back to 0.1.1, attempted upgrade to 0.7.0
- @emdash-cms/cloudflare matched the corresponding emdash version in each step
- Astro 6.1.2 + @astrojs/cloudflare 13.1.6
- Cloudflare Workers (paid plan), D1, R2, KV

### Logs / error output

```shell
EmDash middleware error: Error: Migration failed: D1_ERROR: index idx_ec_about_page_deleted_published_id already exists: SQLITE_ERROR (migration: 034_published_at_index)
```

mvanhorn:
## What does this PR do?

Closes #788

Migrations 034 and 035 weren't safe to rerun against a partially-applied schema. If a previous run's schema commit succeeded but the row write to `_emdash_migrations` failed (D1 subrequest limit, transient error mid-batch, Worker CPU limit), the next migration run reapplies the schema and crashes with `index already exists: SQLITE_ERROR (migration: 034_published_at_index)` — which is the exact failure the reporter hit upgrading from 0.1.1 to 0.6.0/0.7.0.

This makes both migrations safe to re-run on a partially-applied schema while preserving accumulated `_emdash_404_log` data.

- `034_published_at_index.ts`: `CREATE INDEX` → `CREATE INDEX IF NOT EXISTS` so re-creating an existing per-table index is a no-op.
- `035_bounded_404_log.ts`: capture `hitsExists` once before the column adds. Guard `addColumn("hits")` and `addColumn("last_seen_at")` with `columnExists` checks. On rerun (when `hits` already exists), skip the dedup rollup `UPDATE`/`DELETE` block so live counters and `last_seen_at` aren't reset back to `COUNT(*) = 1` / `created_at`. Add `.ifNotExists()` to the new index creates and `.ifExists()` to the path-index drop (and the symmetric calls in `down()`).
- `dialect-helpers.ts`: add `columnExists` and `indexExists` helpers. Both branch on `isPostgres` and follow the same pattern as the existing `tableExists`. Use `pragma_table_info(table)` on SQLite, `information_schema.columns` / `pg_indexes` on Postgres.
- Added integration test `migrations.test.ts` that runs all migrations, deletes the rows for 034 and 035 from `_emdash_migrations`, and re-runs `runMigrations` against the partially-applied schema. Both names appear in `applied`, no error is thrown, and the migration table count returns to `MIGRATION_COUNT`.
- Patch-level changeset.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes (no new diagnostics)
- [x] `pnpm test` passes (`packages/core/tests/integration/database/migrations.test.ts` 15/15)
- [ ] User-visible strings in the admin UI are wrapped for translation (n/a)
- [ ] New features link to an approved Discussion: n/a (bug fix)

## Screenshots / test output

```
Test Files  1 passed (1)
     Tests  15 passed (15)
  Duration  341ms
```

The new test exercises the partial-reapply path:

```ts
it("should re-run migrations 034 and 035 when schema changes were partially applied", async () => {
    await db.destroy();
    db = await setupTestDatabaseWithCollections();

    await db
        .deleteFrom("_emdash_migrations")
        .where("name", "in", ["034_published_at_index", "035_bounded_404_log"])
        .execute();

    const { applied } = await runMigrations(db);

    expect(applied).toContain("034_published_at_index");
    expect(applied).toContain("035_bounded_404_log");

    const migrations = await db.selectFrom("_emdash_migrations").selectAll().execute();
    expect(migrations).toHaveLength(MIGRATION_COUNT);
});
```

## Out of scope

The middleware behavior the reporter also flagged (auth middleware swallowing the migration error and falling through to a permissionless ghost user) is a separate failure mode in `auth.ts` and worth its own issue. Per CLAUDE.md's "no drive-by changes" rule, that's not addressed here.

### DIFF
--- a/.changeset/safe-idempotent-migrations.md
+++ b/.changeset/safe-idempotent-migrations.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes migrations 034 and 035 so they can safely re-run when a previous attempt left the schema partially applied without recording it in `_emdash_migrations`. Resolves the "index already exists" error reported on upgrade from 0.1.1 to 0.6.0+.
--- a/packages/core/src/database/dialect-helpers.ts
+++ b/packages/core/src/database/dialect-helpers.ts
@@ -91,6 +91,56 @@ export async function tableExists(db: Kysely<any>, tableName: string): Promise<b
 }
 
 /**
+ * Check if an index exists in the database.
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- accepts any Kysely instance
+export async function indexExists(db: Kysely<any>, indexName: string): Promise<boolean> {
+	if (isPostgres(db)) {
+		const result = await sql<{ exists: boolean }>`
+			SELECT EXISTS(
+				SELECT 1 FROM pg_indexes
+				WHERE schemaname = current_schema() AND indexname = ${indexName}
+			) as exists
+		`.execute(db);
+		return result.rows[0]?.exists === true;
+	}
+
+	const result = await sql<{ name: string }>`
+		SELECT name FROM sqlite_master
+		WHERE type = 'index' AND name = ${indexName}
+	`.execute(db);
+	return result.rows.length > 0;
+}
+
+/**
+ * Check if a column exists in the database.
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- accepts any Kysely instance
+export async function columnExists(
+	db: Kysely<any>,
+	tableName: string,
+	columnName: string,
+): Promise<boolean> {
+	if (isPostgres(db)) {
+		const result = await sql<{ exists: boolean }>`
+			SELECT EXISTS(
+				SELECT 1 FROM information_schema.columns
+				WHERE table_schema = current_schema()
+					AND table_name = ${tableName}
+					AND column_name = ${columnName}
+			) as exists
+		`.execute(db);
+		return result.rows[0]?.exists === true;
+	}
+
+	const result = await sql<{ name: string }>`
+		SELECT name FROM pragma_table_info(${tableName})
+		WHERE name = ${columnName}
+	`.execute(db);
+	return result.rows.length > 0;
+}
+
+/**
  * List tables matching a LIKE pattern.
  */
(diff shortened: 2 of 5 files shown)


ascorbic (approved):
Thanks!

PR #817, opened 2026-04-29
fix(redirects): fire for unauthenticated visitors

### LINKED ISSUE #808: Redirect middleware silently skipped for public   visitors (locals.emdash.db missing on the anonymous path)
### Description

The redirect middleware (`emdash/middleware/redirect`) silently no-ops for unauthenticated visitors, even though its source comment says it "should handle ALL routes, including public ones". As a result, **none of the rules in `_emdash_redirects` actually fire for real site traffic** — only logged-in admins, sessions on `/_emdash`, and requests carrying the `emdash-edit-mode=true` cookie ever see redirects.

This affects every EmDash site relying on redirects (WP migration 301s, manual prefix rewrites, the `Auto: slug change` rows EmDash itself inserts on slug edits). The hits counter and `_emdash_404_log` don't increment either, so it stays invisible from the admin UI.

#### Root cause

`onRequest$5` (runtime init in `src/astro/middleware.ts`) takes a different branch for **public** requests and only attaches `collectPageMetadata` / `collectPageFragments` to `locals.emdash` — not `db`:

```ts
    // ...
    // eslint-disable-next-line typescript-eslint(no-unsafe-type-assertion) -- partial object; getPageRuntime() only checks for these two methods
    locals.emdash = {
      collectPageMetadata: runtime.collectPageMetadata.bind(runtime),
      collectPageFragments: runtime.collectPageFragments.bind(runtime),
    } as EmDashHandlers;            // ← no `db`
    // ...
  }
}
```

Then the redirect middleware (`onRequest$4`) bails on `if (!emdash?.db) return next()`, so the lookup never runs for public visitors.

The eslint-disable comment ("partial object; only checks for these two methods") suggests the omission of `db` on the public branch is deliberate — the authenticated `locals.emdash` exposes ~30 admin-capable handlers (`handleContentDelete`, `handleMediaCreate`, full Kysely `db`, `setPluginStatus`, etc.) and there's no reason for those to be reachable from a public page render. So this is a defense-in-depth boundary, not an oversight.

The bug is on the redirect middleware side: it reads `db` from `locals.emdash` only, even though the runtime middleware also installs an anon-scoped DB into the request context (`runWithContext({ db: anonScoped.db })`) which is reachable via `getRequestContext()` from inside the redirect middleware too.

#### Two possible fixes

**(A) Expose `runtime.db` on the public-path `locals.emdash`** — minimal, 1-line diff:

```diff
   locals.emdash = {
     collectPageMetadata: runtime.collectPageMetadata.bind(runtime),
     collectPageFragments: runtime.collectPageFragments.bind(runtime),
+    db: runtime.db,
   } as EmDashHandlers;
```

Pro: trivially simple, fixes redirects without touching the redirect middleware. Verified working in production (see Steps to reproduce).
Con: weakens the defense-in-depth boundary above — page render code on public requests can now reach `Astro.locals.emdash.db` and run arbitrary queries (including writes). Existing pages don't do this, but it removes a guardrail.

**(B) Make the redirect middleware fall back to the request-scoped DB** — architecturally cleaner:

```diff
 const onRequest = defineMiddleware(async (context, next) => {
   ...
-  const { emdash } = context.locals;
-  if (!emdash?.db) return next();
+  const db = context.locals.emdash?.db ?? getRequestContext()?.db;
+  if (!db) return next();
   try {
-    const repo = new RedirectRepository(emdash.db);
+    const repo = new RedirectRepository(db);
     ...
```

Pro: keeps the public-path `locals.emdash` minimal as originally intended. Uses the `anonScoped.db` that the runtime middleware already prepared (which is the right DB handle for public requests — read-replica routing, anon scoping, etc.).
Con: small structural change in the redirect middleware (extra import, fallback logic).

(B) feels more in line with the existing design intent and doesn't broaden the public-render attack surface. (A) is what we're running locally because it's a smaller patch to maintain via `pnpm patch` — happy to switch once an upstream direction is decided.

Either way, also worth mirroring the same fallback in `onRequest$3` (setup middleware) which has the same `if (!emdash?.db) return next()` pattern, though that one is admin-route-only so the impact is narrower.

Happy to open a PR for whichever direction feels right.

### Steps to reproduce

1. Fresh EmDash 0.7.0 (or 0.8.0) site, deployed to Cloudflare Workers or running locally with `npx emdash dev`
2. Insert a redirect via the admin API:
   ```bash
   curl -X POST "http://localhost:4321/_emdash/api/redirects/" \
     -H "X-EmDash-Request: 1" \
     -H "Authorization: Bearer <dev-bypass token>" \
     -H "Content-Type: application/json" \
     -d '{"source":"/old","destination":"/new","type":301,"isPattern":false,"enabled":true}'
   ```
3. Verify the row exists: `GET /_emdash/api/redirects/?limit=10` returns it with `enabled: true`, `hits: 0`
4. **Public visitor (no cookies, no session):**
   ```bash
   curl -I http://localhost:4321/old
   ```
   → response is `404` (or 302 to `/404` from a catch-all page). **No 301.** No Location header pointing to `/new`.
5. **With `emdash-edit-mode=true` cookie (forces the authenticated branch):**
   ```bash
   curl -I -b "emdash-edit-mode=true" http://localhost:4321/old
   ```
   → response is `301` with `Location: /new`. Redirect fires.
6. Re-query the row: `hits` stayed `0` after step 4, incremented after step 5. Confirms the middleware did not run in case 4.

### Environment

- emdash version: `0.7.0` (also reproduced by reading `0.8.0` source — same code path)
- Node.js version: 22 (production CI runners), 25 (local sandbox)
- Runtime: Cloudflare Workers (`@astrojs/cloudflare` adapter, D1 + R2 + KV bindings) and `astro dev` locally
- OS: Linux (production), Linux container locally
- Astro: 6.1.6
- pnpm: 10.33.0

### Logs / error output

```shell
There is no error output — that's part of why this is hard to notice. The redirect middleware's `try { ... } catch { return next() }` swallows the early bailout silently. Server-Timing on a public request shows the middleware ran (`mw;dur=10`) but didn't intercept; `_emdash_redirects.hits` and `_emdash_404_log` both stay at zero for the affected paths.
```

all3f0r1:
## What does this PR do?

Fixes the redirect middleware so 301/302 rules from `_emdash_redirects` actually fire for unauthenticated visitors.

The runtime middleware sets up two different shapes of `locals.emdash` depending on the caller — the authenticated/edit-mode/preview branch exposes the full handlers + `db`, while the public-visitor fast path intentionally exposes only `collectPageMetadata` / `collectPageFragments` (defense-in-depth, so page renders can't reach into admin-capable handlers). The redirect middleware was reading `db` from `locals.emdash` only, so on the public branch it bailed out at `if (!emdash?.db) return next()` — meaning WordPress migration 301s, manual prefix rewrites and the `Auto: slug change` rows EmDash inserts on slug edits did nothing for real site traffic. `hits` and `_emdash_404_log` also stayed at zero, so the bug was invisible from the admin UI.

The fix keeps the public-render boundary intact and routes the lookup through `getDb()` from the loader, which already does the right thing: it returns the per-request scoped Kysely from ALS (the `anonScoped.db` the runtime middleware installs for D1 read-replica sessions) when present, and falls back to the singleton instance otherwise. Same accessor the template helpers use.

Closes #808

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes (core package — pre-existing cloudflare typecheck failures are unrelated to this change and reproduce on `main`)
- [x] `pnpm lint` passes (no new diagnostics introduced — diagnostic count unchanged at 25 vs. baseline)
- [x] `pnpm test` passes for redirect/middleware tests (5 new + 45 redirect-repo + 7 redirect-handler + middleware-prerender all green; 3 pre-existing auth test failures on `main` are unrelated)
- [ ] User-visible strings in the admin UI are wrapped for translation and `pnpm locale:extract` has been run (n/a — server middleware, no UI strings)

## Screenshots / test output

New regression test covers the public-visitor branch and verifies the fallback chain:

```
✓ tests/unit/astro/middleware-redirect.test.ts (5 tests)
  ✓ fires for an unauthenticated visitor on a public path (no locals.emdash.db)
  ✓ fires pattern matches for unauthenticated visitors
  ✓ still uses locals.emdash.db when present (authenticated/edit-mode/preview path)
  ✓ skips silently when no database is available at all
  ✓ does not intercept /_emdash routes
```

Reproducer from the issue (curl `/old` with no cookies) now returns `301` with `Location: /new` and `_emdash_redirects.hits` increments — same result as the edit-mode case.


ascorbic:
Thanks @all3f0r1, the correctness fix is right.

Snapshot shows +1 query on every public request. That's because `findExactMatch` runs uncached, while the pattern path already uses `getCachedPatternRules()` / `setCachedPatternRules()` in `redirects/cache.ts`, with `invalidateRedirectCache()` wired into every redirect write site (and the slug-change auto-redirect path in `content.ts`).

Before merging, can we extend the cache to cover exact matches too? Roughly:

- Add a `Map<string, Redirect>` to `redirects/cache.ts` alongside the pattern array.
- One repo method that loads `WHERE enabled = 1`, splits into exact Map + pattern array, populates both caches in one query.
- `invalidateRedirectCache()` clears both.
- Middleware does `cachedExactMap.get(pathname)` instead of `repo.findExactMatch`.

Cold isolate: 1 query loads both. Warm: 0 queries. Empty-redirect sites: empty Map, no probe needed.

Happy to take a follow-up look once it's pushed.

all3f0r1:
Alright, I applied your request the cleanest way I could come up with. LMK if it's up to your standard.

ascorbic (approved):
Thanks! I have a feeling we'll end up needing to optimise this at some point, but it's good for now

PR #822, opened 2026-04-29
fix(core): exclude drafts and trashed entries from taxonomy term counts

### LINKED ISSUE #581: Unpublished Posts are included in Category counts
### Description

When posts are in unpublished and still draft mode they are being included in Category counts (as of version 0.1.0)

### Steps to reproduce

1. Create and deploy a new site from the  "blog" template
2. Create and publish at least one post to view the "Categories" side bar widget. 
3. Add new posts, and save in "Draft" mode without publishing
4. See Category post counts in the sidebar of a published post increment with unpublished post counts

### Environment

- emdash version 0.1.0
- cloudflare workers deployment

### Logs / error output

```shell
No error logs
```

r2sake:
## What does this PR do?

`getTaxonomyTerms` and `getTerm` counted assignments straight from `content_taxonomies` with no filter on entry status or `deleted_at`. As a result, drafts and soft-deleted entries showed up in the Categories / Tags sidebar widget counts on public sites — the post count next to a category name didn't match the number of posts you'd actually find when clicking through.

This joins the underlying `ec_{collection}` table for each collection declared on the taxonomy definition and restricts counts to `status = 'published' AND deleted_at IS NULL`. Dynamic table names are validated via `validateIdentifier` and bound through `sql.ref()` per the project's SQL conventions. A new helper `countPublishedTermAssignments` is shared between `getTaxonomyTerms` and `getTerm`, with a `termIds` option so the single-term path doesn't scan the whole taxonomy. Pre-migration databases (or drift where a referenced `ec_*` table doesn't exist) contribute zero rows via the existing `isMissingTableError` branch.

Closes #581

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/...

## Screenshots / test output

```
$ pnpm --filter emdash test --run tests/unit/taxonomies
 ✓ tests/unit/taxonomies/term-counts.test.ts (4 tests) 245ms
 ✓ tests/unit/taxonomies/get-all-terms-for-entries.test.ts (7 tests) 389ms
 ✓ tests/unit/taxonomies/taxonomies.test.ts (35 tests) 1702ms

 Test Files  3 passed (3)
      Tests  46 passed (46)

$ pnpm --filter emdash test
 Test Files  183 passed (183)
      Tests  3037 passed (3037)
```

The 4 new test cases in `packages/core/tests/unit/taxonomies/term-counts.test.ts` cover:

- drafts excluded from `getTaxonomyTerms` counts
- drafts excluded from `getTerm` count
- soft-deleted entries excluded from term counts
- term with only draft attachments returns count `0`


ascorbic:
Thanks @r2sake. Correctness-wise this is exactly the right fix, and I want to merge it. But I'd like us to address the per-page query cost before it lands. Two concerns:

1. Round-trip count scales with `def.collections.length`

`countPublishedTermAssignments` loops over `collections` and runs one query per collection. The snapshot diff in this PR shows +1 cold/+1 warm because the sample taxonomies span one collection each, but on real sites where a `tags` taxonomy is shared across `posts` and `products` (the canonical multi-collection case), this becomes +N round-trips on every category/tag page render.

If we keep the per-collection loop, we should at least merge them into a single query via `UNION ALL`. Roughly:

```ts
const unionParts = collections.map((c) => sql`
  SELECT ct.taxonomy_id, COUNT(*) AS count
  FROM content_taxonomies ct
  INNER JOIN ${sql.ref(`ec_${c}`)} e ON e.id = ct.entry_id
  WHERE ct.collection = ${c}
    AND e.status = 'published'
    AND e.deleted_at IS NULL
    ${termsFilter}
  GROUP BY ct.taxonomy_id
`);
const result = await sql<{ taxonomy_id: string; count: number }>`
  SELECT taxonomy_id, SUM(count) AS count FROM (${sql.join(unionParts, sql` UNION ALL `)}) GROUP BY taxonomy_id
`.execute(db);
```

That keeps the cost at one round-trip regardless of collection count. We'd still need the `validateIdentifier` / `isMissingTableError` handling per collection.

2. The result is a strong cache candidate

`getTaxonomyTerms` is already wrapped in `requestCached`, so within a single render this only fires once. But across renders, and especially across colos where the warm count is the steady-state, counts change only when entries are published, unpublished, trashed, or restored. `invalidateTermCache()` is already exported and called from the right places (seed apply, all `taxonomies` handlers, the `content/[collection]/[id]/terms/[taxonomy]` route), but the function body is currently a no-op (vestigial from the old "has any assignments?" probe).

If we put the term-counts map behind a worker-lifetime holder and wire `invalidateTermCache()` to clear it, the warm cost goes from "+1 query every render that calls a taxonomy widget" to "0 queries until a content write invalidates". The pattern to follow is `bylinesHolder` (`packages/core/src/bylines/index.ts`) or `SECRETS_CACHE_KEY` (`packages/core/src/config/secrets.ts`):

```ts
const TERM_COUNTS_KEY = Symbol.for("@emdash-cms/core/term-counts@1");
interface TermCountsHolder { byTaxonomy: Map<string, Map<string, number>> }
function holder() {
  const g = globalThis as Record<symbol, TermCountsHolder | undefined>;
  if (!g[TERM_COUNTS_KEY]) g[TERM_COUNTS_KEY] = { byTaxonomy: new Map() };
  return g[TERM_COUNTS_KEY]!;
}
// inside getTaxonomyTerms, after fetching `def`:
const cached = holder().byTaxonomy.get(taxonomyName);
if (cached) {
  // build from cache
} else {
  const counts = await countPublishedTermAssignments(db, def.collections);
  holder().byTaxonomy.set(taxonomyName, counts);
}

// invalidateTermCache becomes:
export function invalidateTermCache(): void {
  holder().byTaxonomy.clear();
}
```

This makes content writes responsible for clearing the cache, which is exactly what the existing invalidation call sites do already. Stale-read window across colos exists but is bounded by the eventual-consistency window of the write itself.

What I'd love to see in this PR:

1. Single UNION ALL query for the cold path (item 1)
2. Worker-lifetime cache wired to the existing `invalidateTermCache()` (item 2)

ascorbic:
Thanks for your work on this. There are currently several overlapping PRs that each manage part of this, but none of them manage the lot. I've found the simplest solution is to create a new PR inspired by all of them. For that reason I'm going to close this in favour of #1892

PR #448, opened 2026-04-11
add site image and favicon from settings to templates

grexe:
## What does this PR do?

Closes #430 

## Type of change

- [X] Bug fix
- [ ] Feature (requires [approved Discussion](https://github.com/emdash-cms/emdash/discussions/categories/ideas))
- [ ] Refactor (no behavior change)
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests
- [ ] Chore (dependencies, CI, tooling)

## Checklist

- [X] I have read [CONTRIBUTING.md](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md)
- [X] `pnpm typecheck` passes
- [X] `pnpm --silent lint:json | jq '.diagnostics | length'` returns 0
- [X] `pnpm test` passes (or targeted tests for my change)
- [X] `pnpm format` has been run
- [X] I have added/updated tests for my changes (if applicable)
- [X] I have added a [changeset](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#changesets) (if this PR changes a published package)
- [ ] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/...

## AI-generated code disclosure

- [X] This PR includes AI-generated code

PR #847, opened 2026-04-30
fix(core): inject favicon link with correct MIME type from site settings

### LINKED ISSUE #831: Favicon functionality fails against vectorized images
### Description

On https://greenstars.cc my favicon despite being set within /_emdash/admin/settings/general is not actually used or displayed, this is improper behavior.
The media I provided was an SVG.
RE: https://web.dev/articles/building/an-adaptive-favicon
and my comment regarding such in https://github.com/emdash-cms/emdash/issues/430#issuecomment-4226476591

My repo if useful in any aspect: https://github.com/thegreatGreenstar/greenstars-great-emdash-site

### Environment

EmDash CMS v0.8.0 (5c0195a)
Emdash latest all-round, per dependabot, run manually before this

ascorbic:
## What does this PR do?

User-configured site favicons were never injected into the public site's `<head>` by core. Per-template `Base.astro` layouts each rendered their own `<link rel="icon">`, but pre-#448 templates lacked the line entirely, and even current templates dropped the `type` attribute. SVG favicons therefore failed to render in Chromium browsers, which require `type="image/svg+xml"` when the URL has no `.svg` extension.

This PR centralizes favicon injection in `EmDashHead` and emits the correct MIME type from the stored media row.

Closes #831

### Approach

A new `renderSiteIdentity()` helper in `packages/core/src/page/site-identity.ts` emits a `<link rel="icon">` tag with the optional `type` attribute. It deliberately lives outside the plugin contribution pipeline (`page/metadata.ts`) because that pipeline's `isSafeHref` allowlist rejects same-origin paths like `/_emdash/api/media/file/...`. That restriction is correct for sandboxed plugin contributions but blocks our own first-party favicon URLs.

`MediaReference` now carries `url`, `contentType`, `width`, and `height` when resolved via `resolveMediaReference`, so callers can emit correct head tags without a second round-trip to the media table. The stored shape is unchanged: only the in-memory resolved value picks up the extra fields. Zod schemas at the REST/MCP boundary still define just `{ mediaId, alt? }` and rely on default strip-mode parsing to discard the resolved fields if a client posts them back. A docstring on `MediaReference` calls this out so a future contributor doesn't accidentally relax the schema and start writing the resolved fields back to storage.

`EmDashHead.astro` now calls `getSiteSettings()` instead of `getSiteSetting("seo")` so it can pick up `favicon` alongside the SEO settings. Both helpers are request-cached and worker-cached; the standard `Base.astro` pattern in every shipped template already calls `getSiteSettings()` first, so this is a free read in practice.

### Backwards compatibility

Templates that already render their own favicon link continue to work. Browsers tolerate the duplicate. A follow-up cleanup can drop the per-template line once this has shipped, but it does not need to be coordinated.

### Query-count impact

Zero net delta from this change. `getSiteSettings()` is request-cached; `Base.astro` already calls it before `EmDashHead` renders. Verified by running `pnpm query-counts` with and without the change applied: identical numbers either way. The repository's snapshot is currently stale relative to actual behaviour on `main` (every public route is +1 to +6 queries above the recorded values), but that pre-existing drift is unrelated to this PR.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm test` passes (3053 / 3053 in core)
- [x] I have added/updated tests for my changes (7 new unit tests in `tests/unit/page/site-identity.test.ts`)
- [x] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/... (n/a, bug fix)

## Screenshots / test output

```
$ pnpm --filter emdash test --run tests/unit/page/site-identity
 ✓ tests/unit/page/site-identity.test.ts (7 tests) 2ms

 Test Files  1 passed (1)
      Tests  7 passed (7)

$ pnpm --filter emdash test --run
 Test Files  184 passed (184)
      Tests  3053 passed (3053)
```

The 7 new tests cover:

- Empty / undefined input is a no-op
- `MediaReference` without resolved `url` is a no-op (handles deleted media row gracefully)
- Favicon with `url` and `contentType` renders the correct tag
- SVG favicons get `type="image/svg+xml"` (the #831 bug)
- Missing `contentType` falls back to a typeless `<link>` tag (legacy stored media)
- Hostile content in `url` and `contentType` is HTML-attribute-escaped via `escapeHtmlAttr`

### DIFF
--- a/.changeset/fix-favicon-injection.md
+++ b/.changeset/fix-favicon-injection.md
@@ -1,0 +1,7 @@
+---
+"emdash": patch
+---
+
+Fixes site favicon injection so user-configured favicons render on the public site, including SVG favicons in Chromium browsers (#831). `EmDashHead` now emits a `<link rel="icon">` tag with the correct `type` attribute (e.g. `image/svg+xml`) sourced from the stored media's MIME type. Templates that already render their own favicon link continue to work; browsers tolerate the duplicate, and a follow-up cleanup can drop the per-template line.
+
+`MediaReference` now carries `url`, `contentType`, `width`, and `height` when resolved via `resolveMediaReference`, so callers can emit correct head tags without a second round-trip to the media table.
--- a/.changeset/fix-favicon-injection.md
+++ b/.changeset/fix-favicon-injection.md
@@ -2,6 +2,6 @@
 "emdash": patch
 ---
 
-Fixes site favicon injection so user-configured favicons render on the public site, including SVG favicons in Chromium browsers (#831). `EmDashHead` now emits a `<link rel="icon">` tag with the correct `type` attribute (e.g. `image/svg+xml`) sourced from the stored media's MIME type. Templates that already render their own favicon link continue to work; browsers tolerate the duplicate, and a follow-up cleanup can drop the per-template line.
+Fixes site favicon injection so user-configured favicons render on the public site, including SVG favicons in Chromium browsers (#831). `EmDashHead` now emits a `<link rel="icon">` tag with the correct `type` attribute (e.g. `image/svg+xml`) sourced from the stored media's MIME type. The bundled templates and demos have been updated to drop their per-template favicon link in favour of the centralized injection; existing user sites that still emit their own `<link rel="icon">` continue to work because browsers tolerate the duplicate.
 
 `MediaReference` now carries `url`, `contentType`, `width`, and `height` when resolved via `resolveMediaReference`, so callers can emit correct head tags without a second round-trip to the media table.
--- a/demos/cloudflare/src/layouts/Base.astro
+++ b/demos/cloudflare/src/layouts/Base.astro
@@ -40,7 +40,7 @@ const {
 	author,
 	content,
 } = Astro.props;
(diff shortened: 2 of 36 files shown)


ascorbic:
Addressing review feedback in 2eb7071a:

- Dropped `<link rel="icon">` from all 16 bundled `Base.astro` files (templates, demos, fixtures, infra) and removed `siteFavicon` from the matching `site-identity.ts` helpers. Core now emits the link via `renderSiteIdentity()`, so the per-template line was producing a duplicate tag.
- Pulled back from emitting `apple-touch-icon` and `application-name` automatically. They need their own configurable fields in `SiteSettings`: an SVG favicon would produce a broken iOS home-screen icon, and a small PNG favicon would scale up blurry. Tracked separately.
- Re-verified query counts after rebuilding core: `pnpm query-counts` matches the snapshot. The earlier "drift" I'd noted was a stale build in my local fixture, not real.

PR #849, opened 2026-04-30
fix(admin): round-trip datetime widget values through datetime-local input

drudge:
## What does this PR do?

Fixes the `datetime` field widget so existing values display in the editor and new values pass server validation. Two universal symptoms:

1. **Empty display.** The widget passed full ISO 8601 (`YYYY-MM-DDTHH:mm:ss.sssZ`) directly into `<input type="datetime-local">`, which only accepts `YYYY-MM-DDTHH:mm` and silently renders empty on anything else. Every existing datetime entry showed an empty input on edit.
2. **Save rejected.** `datetime-local` emits `YYYY-MM-DDTHH:mm`. The per-field zod schema in `packages/core/src/schema/zod-generator.ts` is `z.string().datetime().or(z.string().date())`, neither branch of which accepts that shape. Every save of a non-empty datetime was rejected.

The fix strips the seconds/ms/Z suffix for display and appends `:00.000Z` on save so the stored shape matches the validator's `datetime()` branch. Date-only stored values (`YYYY-MM-DD`, the schema's other accepted shape) are padded to UTC midnight for the input. The widget is treated as UTC for a stable round-trip — using `new Date().toISOString()` (the convention in the publish-schedule UI in the same file) would shift values by the local-UTC offset on every save, so each round-trip would mutate the persisted time.

The same bug existed on the `datetime` sub-field inside `RepeaterField`; fixed there in the same shape.

Closes #

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes (`pnpm --silent lint:json | jq '.diagnostics | length'` returns 0)
- [x] `pnpm test` passes (or targeted tests for my change) — `ContentEditor.test.tsx` (35/35) and new `RepeaterField.test.tsx` (2/2) green
- [x] User-visible strings in the admin UI are wrapped for translation (if applicable). N/A — no new strings.
- [ ] New features link to an approved Discussion: N/A, bug fix

## Screenshots / test output

```
 ✓ tests/components/ContentEditor.test.tsx (35 tests)
   ✓ field rendering > renders datetime fields as datetime-local inputs
   ✓ field rendering > displays a stored ISO datetime in the datetime-local input
   ✓ field rendering > saves datetime fields back as full ISO 8601 with Z and milliseconds
   ... (32 existing)

 ✓ tests/components/RepeaterField.test.tsx (2 tests)
   ✓ datetime sub-field > displays a stored ISO datetime in the datetime-local input
   ✓ datetime sub-field > emits a full ISO 8601 value with Z and milliseconds on change

 Test Files  2 passed (2)
      Tests  37 passed (37)
```

### Adversarial review notes (out of scope, flagged for future work)

- **Schedule picker (`ContentEditor.tsx:814`) uses `new Date().toISOString()` directly** for its datetime-local round-trip, which means a user entering a wall-clock time has it interpreted as local-TZ then converted to UTC for storage. This widget treats the field as UTC; the scheduler treats it as local. Two different conventions for the same input type in the same file. Not addressed here to keep the PR scoped to the reported bug; worth a follow-up Discussion to unify.
- **Stored values bearing a non-Z offset** (e.g. `2026-02-26T09:30:00+02:00`, valid per `z.string().datetime()`) display the bare 16-char prefix in the input, which the browser treats as local. Saving without changes would store `…+02:00` as `…Z` — a different moment. EmDash always stores `Z`-suffixed ISO today, so this only surfaces on data migrated from other systems. Not addressed.
- **Date-only stored values normalize to datetime on save.** A stored `2026-02-26` becomes `2026-02-26T00:00:00.000Z` after one round-trip through this widget. The validator accepts both shapes so persistence doesn't break, but downstream consumers that distinguish dates from datetimes by string shape would see a change. Documented in the changeset.

### DIFF
--- a/.changeset/datetime-widget-roundtrip.md
+++ b/.changeset/datetime-widget-roundtrip.md
@@ -1,0 +1,5 @@
+---
+"@emdash-cms/admin": patch
+---
+
+Fixes the `datetime` field widget so existing values display in the editor and new values pass server validation. The widget passed raw ISO 8601 (`YYYY-MM-DDTHH:mm:ss.sssZ`) into `<input type="datetime-local">`, which silently rendered empty, and emitted `YYYY-MM-DDTHH:mm` on save, which the field's zod schema rejected. Strips the suffix for display, appends `:00.000Z` on save, and normalizes date-only stored values to UTC midnight for the input. Applies to the top-level `datetime` widget in the content editor and the `datetime` sub-field type inside `RepeaterField`.
--- a/packages/admin/src/components/ContentEditor.tsx
+++ b/packages/admin/src/components/ContentEditor.tsx
@@ -36,6 +36,7 @@ import type {
 	TranslationSummary,
 } from "../lib/api";
 import { getPreviewUrl, getDraftStatus } from "../lib/api";
+import { fromDatetimeLocalInputValue, toDatetimeLocalInputValue } from "../lib/datetime-local.js";
 import { usePluginAdmins } from "../lib/plugin-context.js";
 import { contentUrl } from "../lib/url.js";
 import { cn, slugify } from "../lib/utils";
@@ -1264,8 +1265,8 @@ function FieldRenderer({
 					label={label}
 					id={id}
 					type="datetime-local"
-					value={typeof value === "string" ? value : ""}
-					onChange={(e) => handleChange(e.target.value)}
+					value={toDatetimeLocalInputValue(value)}
+					onChange={(e) => handleChange(fromDatetimeLocalInputValue(e.target.value))}
 					required={field.required}
 				/>
 			);
--- a/packages/admin/src/components/RepeaterField.tsx
+++ b/packages/admin/src/components/RepeaterField.tsx
@@ -20,6 +20,7 @@ import { useLingui } from "@lingui/react/macro";
 import { Plus, Trash, DotsSixVertical, CaretDown } from "@phosphor-icons/react";
 import * as React from "react";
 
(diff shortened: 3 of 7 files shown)


drudge:
Addressed both Copilot review notes — extracted the round-trip logic into a shared helper module so `ContentEditor` and `RepeaterField` no longer have parallel implementations.

- New: `packages/admin/src/lib/datetime-local.ts` exports `toDatetimeLocalInputValue` (stored value → input attribute) and `fromDatetimeLocalInputValue` (input value → stored ISO).
- `ContentEditor.tsx` and `RepeaterField.tsx` now call those two helpers — datetime case is back to a one-liner in each.
- New: `packages/admin/tests/lib/datetime-local.test.ts` with 7 focused unit tests pinning the string-shape assumptions (10/16-char inputs, the round-trip-stable invariant, empty/null handling) — much faster than the browser-mode tests, and covers them in one place.

The browser-mode tests in `ContentEditor.test.tsx` (3) and `RepeaterField.test.tsx` (2) remain to verify the integrated widget behavior.

`pnpm typecheck`, `pnpm format` clean. `pnpm --silent lint:json` shows the same 5 pre-existing warnings (in `packages/blocks/src/validation.ts` and `packages/core/src/astro/routes/sitemap-[collection].xml.ts` — none introduced by this PR). 44/44 admin tests green.

ascorbic (approved):
Thanks!

PR #853, opened 2026-04-30
fix(schema): coerce SQLite-shaped numeric booleans in zod field schema

drudge:
## What does this PR do?

Saving any content entry on a collection with a boolean field fails with a validation error after a GET → edit → POST round-trip. The repository converts booleans to `0/1` on write but never converts them back on read, so reads surface boolean columns as numbers. The per-field zod schema is `z.boolean()`, which rejects numbers. The admin re-includes untouched booleans in its form payload, so a user toggling one switch sees every other boolean field's `0/1` value re-submitted and rejected.

Trace:

- `packages/core/src/database/repositories/content.ts:54` (`serializeValue`) — boolean → `0/1`.
- `packages/core/src/database/repositories/content.ts:71` (`deserializeValue`) — only parses JSON, never coerces numerics.
- `packages/core/src/database/repositories/content.ts:1157` (`mapRow`) — runs every column through `deserializeValue`.
- `packages/core/src/schema/zod-generator.ts:70` — emits `z.boolean()` and rejects the read shape.
- `packages/core/src/api/handlers/validation.ts:71` — calls `generateZodSchema(...).partial()` for updates.

Wrap the boolean field schema in a `z.preprocess` that coerces the SQLite shape (`0/1`) to `false/true`, leaving every other input (other numbers, strings, `BigInt`) to fall through to `z.boolean()`'s standard rejection. Surgical fix at the validation boundary that restores symmetry with the write path without touching the repository contract or the existing `ContentRepository(db)` call sites that read `data.*` directly. Boolean fields map to `INTEGER` columns on every dialect via `FIELD_TYPE_TO_COLUMN` (`schema/types.ts:64`), so the fix covers SQLite and Postgres.

Closes #

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes (`pnpm --silent lint:json | jq '.diagnostics | length'` returns 0)
- [x] `pnpm test` passes (or targeted tests for my change) — full zod-generator suite (20/20 incl. 3 new cases) green; content-handlers suite (19/19) unaffected
- [x] User-visible strings in the admin UI are wrapped for translation (if applicable). N/A — no admin UI changes.
- [ ] New features link to an approved Discussion: N/A, bug fix

## Screenshots / test output

```
 ✓ tests/unit/schema/zod-generator.test.ts (20 tests)
   ✓ generateFieldSchema > should generate boolean schema
   ✓ generateFieldSchema > should coerce SQLite-shaped numeric booleans (0/1) to real booleans
   ✓ generateFieldSchema > should preserve `.default(false)` chaining through the boolean preprocess
   ✓ generateFieldSchema > should accept SQLite-shaped booleans in partial-mode validation
   ... (16 existing)

 Test Files  1 passed (1)
      Tests  20 passed (20)
```

### Adversarial review notes (out of scope, flagged for future work)

- **Asymmetric serialization at the repo layer.** The root cause is that `deserializeValue` does not invert `serializeValue` for booleans. A more comprehensive fix would track per-collection field types in `mapRow` and coerce there, returning real booleans on `data.*` access. That changes the repository's read contract and would require an audit of every `ContentRepository(db)` consumer that currently reads `data.field as number` (no enforcement today, but easy to slip in). This PR keeps that contract intact and patches only the validation boundary.
- **Type-generation drift.** `generateTypeScript` emits `boolean` for boolean fields, but consumers reading `data.*` from `getEmDashEntry` / `getEmDashCollection` still see `0/1` until the repo-layer fix lands. This PR doesn't widen the emitted TS type, on the assumption that fixing the read path is preferable to documenting a `boolean | number` contract.
- **BigInt** rejection is intentional. No known SQLite driver returns BigInt for `INTEGER` columns containing 0/1, and silently coercing would mask a potential driver bug.

### DIFF
--- a/.changeset/zod-boolean-sqlite-coercion.md
+++ b/.changeset/zod-boolean-sqlite-coercion.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes content saves on collections with boolean fields. Boolean fields map to `INTEGER` columns and the repository writes booleans as `0/1`, but never converts them back on read, so a GET → edit → POST round-trip surfaced numbers where the per-collection zod schema expected booleans, and every save was rejected. The boolean field schema now coerces the `0/1` shape to real booleans at the validation boundary; other numbers and strings still fail validation as before.
--- a/packages/core/src/schema/zod-generator.ts
+++ b/packages/core/src/schema/zod-generator.ts
@@ -68,7 +68,15 @@ function getBaseSchema(type: FieldType, field: Field): ZodTypeAny {
 			return z.number().int();
 
 		case "boolean":
-			return z.boolean();
+			// Boolean fields map to `INTEGER` columns (`FIELD_TYPE_TO_COLUMN`
+			// in `schema/types.ts`) and `serializeValue` in
+			// `database/repositories/content.ts` writes booleans as 0/1.
+			// `deserializeValue` never converts them back, so reads return
+			// numbers. Coerce the stored 0/1 shape here so a GET → POST
+			// round-trip on a boolean field passes validation. Other inputs
+			// (strings, other numbers) fall through to `z.boolean()` and
+			// produce its standard rejection.
+			return z.preprocess((v) => (v === 0 || v === 1 ? Boolean(v) : v), z.boolean());
 
 		case "datetime":
 			return z.string().datetime().or(z.string().date());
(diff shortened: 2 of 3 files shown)


drudge:
Addressed all three Copilot notes — the bug is dialect-agnostic (any `INTEGER`-backed boolean), not SQLite-specific. Reworded the changeset, the inline code comment, and the test names + comments to say "INTEGER columns" / "stored 0/1 booleans" instead of "SQLite shape". Pushed as a force-update with no behavior change. `pnpm format` clean, `lint:json` 0, 20/20 zod-generator tests still passing.

ascorbic (approved):
Thanks!

PR #896, opened 2026-05-02
fix: dashboard 500 on D1 with many title-bearing collections (#895)

### LINKED ISSUE #895: Dashboard 500 on D1: fetchRecentItems UNION ALL exceeds SQLITE_LIMIT_COMPOUND_SELECT
## Summary

`GET /_emdash/api/dashboard` returns HTTP 500 with `{"error":{"code":"DASHBOARD_STATS_ERROR","message":"Failed to load dashboard statistics"}}` on Cloudflare D1 (and miniflare's D1 emulator) when the project has enough title-bearing collections. The admin dashboard fails to render for every authenticated user.

Root cause is in `fetchRecentItems` (`packages/core/src/api/handlers/dashboard.ts`): it composes a single chained `UNION ALL` across every title-bearing collection and wraps it in an outer `SELECT * FROM (combined) ORDER BY updated_at DESC LIMIT 10`. D1 enforces a `SQLITE_LIMIT_COMPOUND_SELECT` cap that is tighter than the default 500 used by `better-sqlite3`, so the query fails on D1 even though the same SQL runs fine against a local sqlite file.

This is a regression from 0.7.0 (where the dashboard worked on D1) to 0.8.0, and is unchanged in 0.9.0.

## Environment

- emdash: `0.8.0` (also reproduced on `0.9.0`, source unchanged)
- @emdash-cms/cloudflare: `0.8.0` (and `0.9.0`)
- Astro: 5.x with `@astrojs/cloudflare` adapter, `output: "server"`
- Runtime: Cloudflare Workers (production) and miniflare via `pnpm dev` (local)
- Project: 12 collections total, 9 of them title-bearing (pages, hero, cta, faq, logo_grid, feature_list, stat_block, posts, legal_pages)

## Reproduction

1. Create or use an emdash 0.8 project on Cloudflare D1 with at least 9 title-bearing collections (each defining a `title` field).
2. Run `pnpm seed` and `pnpm dev`.
3. Log into `/_emdash/admin` (passkey).
4. Watch the network tab and dev-server stderr.

Result: every dashboard load fires `GET /_emdash/api/dashboard` twice and both return 500.

## Server stack trace (local `pnpm dev`)

```
[ERROR] [vite] Dashboard stats error: Error: D1_ERROR: too many terms in compound SELECT: SQLITE_ERROR
    at D1DatabaseSessionAlwaysPrimary._sendOrThrow (cloudflare-internal:d1-api:139:19)
    at cloudflare-internal:d1-api:366:41
    ... 7 lines matching cause stack trace ...
    at renderEndpoint (.../chunk-AI5DR33A.js:46:18)
  [cause]: Error: too many terms in compound SELECT: SQLITE_ERROR
      at D1DatabaseSessionAlwaysPrimary._sendOrThrow (cloudflare-internal:d1-api:140:24)
      at cloudflare-internal:d1-api:366:41
      at D1Connection.executeQuery (.../chunk-V24LZAOZ.js:73:21)
      at .../chunk-O7MX2KDY.js:5895:22
      at DefaultConnectionProvider.provideConnection (.../chunk-O7MX2KDY.js:11936:14)
      at _DefaultQueryExecutor.executeQuery (.../chunk-O7MX2KDY.js:5894:12)
      at fetchRecentItems (.../emdash@0.8.0/.../api/handlers/dashboard.ts:184:17)
      at handleDashboardStats (.../emdash@0.8.0/.../api/handlers/dashboard.ts:82:23)
      at Module.GET (.../emdash@0.8.0/.../astro/routes/api/dashboard.ts:26:18)
```

The same stack reproduces on staging Cloudflare D1 via `wrangler tail`.

## Root cause analysis

`fetchRecentItems` builds a Kysely raw-SQL composition roughly equivalent to:

```sql
SELECT * FROM (
  SELECT id, 'col1' AS slug_lit, 'Label 1' AS label_lit, COALESCE(title, slug, id) AS title,
         slug, status, updated_at, author_id
    FROM ec_col1 WHERE deleted_at IS NULL
  UNION ALL
  SELECT id, 'col2' AS slug_lit, 'Label 2' AS label_lit, COALESCE(title, slug, id) AS title, ...
    FROM ec_col2 WHERE deleted_at IS NULL
  UNION ALL
  ... one arm per title-bearing collection ...
) ORDER BY updated_at DESC LIMIT 10;
```

When the chained arm count crosses D1's compound SELECT cap, D1 rejects the query. The cap is undocumented for D1 specifically but is clearly lower than SQLite's upstream default of 500. We hit it at 9 arms.

The schema, migrations, seed and auth path are all correct: I verified by running the entire handler logic against the local sqlite via `better-sqlite3` plus Kysely 0.27 (the same dialect family). All five steps inside `handleDashboardStats` complete cleanly under `better-sqlite3`. The failure is specific to the D1 driver path.

This bug was introduced in 0.8.0 because the equivalent 0.7.0 code path did not use this single-statement UNION ALL composition. 0.9.0 ships an identical `dashboard.ts` (verified by md5: `fa72b1b1b2dc0f601e27174470857b12` for both `v0.8.0` and `v0.9.0`).

## Proposed fix

Replace the chained UNION ALL composition with N per-collection queries executed concurrently, then merge, sort and slice in JS. Equivalent semantics, no schema changes, no D1-specific branching, fetches at most `N * limit` rows for a small final `limit`.

Sketch:

```ts
const perCollection = await Promise.all(
  titleBearingCollections.map(({ slug, label }) =>
    db
      .selectFrom(getTableName(slug))
      .select(['id', 'slug', 'status', 'updated_at', 'author_id', 'title'])
      .where('deleted_at', 'is', null)
      .orderBy('updated_at', 'desc')
      .limit(10)
      .execute()
      .then(rows =>
        rows.map(r => ({
          ...r,
          collection_slug: slug,
          collection_label: label,
          title: r.title ?? r.slug ?? r.id,
        })),
      ),
  ),
);

const merged = perCollection
  .flat()
  .sort((a, b) => (b.updated_at > a.updated_at ? 1 : -1))
  .slice(0, 10);
```

This also avoids the per-arm `sql.lit(label)` interpolation which had its own UTF-8 risk for non-ASCII labels.

If the maintainer prefers to keep a single-statement query, an alternative is to chunk the UNION ALL into fixed-size groups (for example 4 arms per group), `Promise.all` the groups, and merge in JS. The per-collection variant is simpler and likely cheaper for D1.

## Workaround

We are running a local `pnpm patch emdash` against the affected file to apply the per-collection fix. Happy to send it as a PR if helpful.

## Severity

This blocks the emdash admin dashboard for any project on Cloudflare D1 with enough title-bearing collections (we hit it at 9). Auth works, the rest of the admin loads, but the first screen after login returns the failed-stats panel.

cristianuibar:
## What does this PR do?

Fixes the `GET /_emdash/api/dashboard` 500 error on Cloudflare D1 (and the miniflare D1 emulator) when a project has enough title-bearing collections. The admin dashboard fails to render for every authenticated user once the threshold is crossed (reproduced at 9 arms in the issue).

**Root cause:** `fetchRecentItems` in `packages/core/src/api/handlers/dashboard.ts` composes a single chained `UNION ALL` across every title-bearing collection and wraps it in an outer `SELECT * FROM (combined) ORDER BY updated_at DESC LIMIT 10`. D1 enforces a `SQLITE_LIMIT_COMPOUND_SELECT` cap that is tighter than upstream SQLite's default of 500, so the query fails on D1 even though the same SQL runs fine against `better-sqlite3`. This is a 0.7.0 → 0.8.0 regression and is unchanged in 0.9.0.

**Fix:** Replace the chained `UNION ALL` with N per-collection queries fired in parallel via `Promise.all`, then merge / sort / slice in JS. Equivalent semantics, no schema changes, no D1-specific branching, and at most `N * 10` rows materialized before the final `slice(0, 10)`. Per-arm `sql.lit(label)` interpolation is preserved (kept on the outer side of each per-collection query, identical to the previous arm shape).

This patch was developed and validated in production at https://github.com/cristianuibar/wishday-web (12 collections, 9 title-bearing) by running `pnpm patch emdash@0.8.0` against the affected file. The dashboard has rendered cleanly under that patch through staging deploys; this PR upstreams the same change.

Closes #895

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes (no new diagnostics introduced; pre-existing main-branch warnings unchanged)
- [x] `pnpm test` passes (or targeted tests for my change) — `tests/unit/api/dashboard-handlers.test.ts` 11/11 pass; the 8 pre-existing failures in `vite-config.test.ts`, `wordpress-slug-sanitization.test.ts`, and `wp-prepare-invalidate.test.ts` reproduce on `origin/main` and are unrelated to this change
- [x] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable). Do not include `messages.po` changes except in translation PRs — a workflow extracts catalogs on merge to `main`. _N/A — no admin UI strings touched._
- [x] I have added a [changeset](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#changesets) (if this PR changes a published package) — `.changeset/fix-dashboard-d1-compound-select.md`, `emdash: patch`
- [ ] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/... _N/A — bug fix._

### Test notes

The added regression test (`merges recent items across many title-bearing collections (#895)`) creates 12 collections each with a `title` field and one entry, then asserts the merged top-10 contract: count = 10, ordered by `updated_at` desc, drawn across all collections.

Caveat — the original D1 `SQLITE_LIMIT_COMPOUND_SELECT` cap cannot be reproduced under `better-sqlite3` (its compound-SELECT limit is upstream SQLite's default of 500), so this test will not _fail_ against the previous `UNION ALL` code path. It exists as a behavioral regression guard for the per-collection merge contract introduced by this fix and would catch a future revert. The actual 500 was reproduced and fixed against live Cloudflare D1 in https://github.com/cristianuibar/wishday-web via `pnpm patch`; the patch shipped here is equivalent.

- [x] This PR includes AI-generated code — model/tool: Claude Opus 4.7 (1M context) via Claude Code

## Screenshots / test output

```
$ cd packages/core && pnpm exec vitest run tests/unit/api/dashboard-handlers.test.ts
 ✓ tests/unit/api/dashboard-handlers.test.ts (11 tests) 463ms

 Test Files  1 passed (1)
      Tests  11 passed (11)
```

### DIFF
--- a/.changeset/fix-dashboard-d1-compound-select.md
+++ b/.changeset/fix-dashboard-d1-compound-select.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes 500 error on `GET /_emdash/api/dashboard` when running on Cloudflare D1 with many title-bearing collections. `fetchRecentItems` now issues one query per collection in parallel and merges results in JS instead of building a single chained `UNION ALL`, which trips D1's `SQLITE_LIMIT_COMPOUND_SELECT` cap once enough collections are present (#895).
--- a/packages/core/src/api/handlers/dashboard.ts
+++ b/packages/core/src/api/handlers/dashboard.ts
@@ -140,21 +140,22 @@ async function fetchRecentItems(
 
 	const collectionsWithTitle = new Set(titleFields.map((r) => r.collection_slug));
 
-	// Build a UNION ALL query across all content tables.
-	// Each branch is wrapped in SELECT * FROM (...) so the inner
-	// ORDER BY + LIMIT is valid SQLite (bare ORDER BY inside UNION
-	// branches is a syntax error in SQLite).
-	const subQueries = collections.map((col) => {
-		validateIdentifier(col.slug);
-		const table = `ec_${col.slug}`;
-		const hasTitle = collectionsWithTitle.has(col.slug);
+	// Issue one query per collection in parallel, then merge in JS.
+	// A single UNION ALL across N collections trips D1's
+	// SQLITE_LIMIT_COMPOUND_SELECT cap when N is large enough (#895);
+	// per-collection queries side-step that. Each query fetches at most
+	// 10 rows, so the merge handles at most N * 10 rows before slicing.
+	const perCollection = await Promise.all(
+		collections.map(async (col) => {
+			validateIdentifier(col.slug);
+			const table = `ec_${col.slug}`;
+			const hasTitle = collectionsWithTitle.has(col.slug);
 
-		// Use title column if it exists, otherwise fall back to slug → id.
-		// All output uses snake_case to avoid SQLite quoting issues on D1.
-		const titleExpr = hasTitle ? sql`COALESCE(title, slug, id)` : sql`COALESCE(slug, id)`;
+			// Use title column if it exists, otherwise fall back to slug, id.
+			// All output uses snake_case to avoid SQLite quoting issues on D1.
+			const titleExpr = hasTitle ? sql`COALESCE(title, slug, id)` : sql`COALESCE(slug, id)`;
 
-		return sql<RecentItemRow>`
-			SELECT * FROM (
+			const result = await sql<RecentItemRow>`
 				SELECT
 					id,
 					${sql.lit(col.slug)} AS collection,
@@ -168,27 +169,19 @@ async function fetchRecentItems(
 				WHERE deleted_at IS NULL
 				ORDER BY updated_at DESC
 				LIMIT 10
-			)
-		`;
-	});
+			`.execute(db);
+			return result.rows;
+		}),
+	);
 
-	// Combine with UNION ALL
-	// eslint-disable-next-line typescript-eslint(no-unnecessary-type-assertion) -- noUncheckedIndexedAccess
-	let combined = subQueries[0]!;
-	for (let i = 1; i < subQueries.length; i++) {
-		// eslint-disable-next-line typescript-eslint(no-unnecessary-type-assertion) -- noUncheckedIndexedAccess
-		combined = sql<RecentItemRow>`${combined} UNION ALL ${subQueries[i]!}`;
-	}
+	// Merge across collections, sort by updated_at desc, take top 10.
+	const merged = perCollection
+		.flat()
+		.toSorted((a, b) => (a.updated_at < b.updated_at ? 1 : a.updated_at > b.updated_at ? -1 : 0))
+		.slice(0, 10);
 
-	// Final sort + limit across all branches
-	const result = await sql<RecentItemRow>`
-		SELECT * FROM (${combined})
-		ORDER BY updated_at DESC
-		LIMIT 10
-	`.execute(db);
-
-	// Map snake_case DB rows → camelCase API shape
-	return result.rows.map((row) => ({
+	// Map snake_case DB rows to camelCase API shape
+	return merged.map((row) => ({
 		id: row.id,
 		collection: row.collection,
 		collectionLabel: row.collection_label,
--- a/packages/core/tests/unit/api/dashboard-handlers.test.ts
+++ b/packages/core/tests/unit/api/dashboard-handlers.test.ts
@@ -214,6 +214,54 @@ describe("Dashboard Handlers", () => {
 			expect(postStats!.total).toBe(1);
 		});
 
+		it("merges recent items across many title-bearing collections (#895)", async () => {
+			// Regression for #895: with enough title-bearing collections, the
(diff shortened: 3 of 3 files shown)


cristianuibar:

ascorbic (approved):
Thanks for catching this!

PR #988, opened 2026-05-11
Harden plugin MCP tool registration

masonjames:
## What does this PR do?

Scoped plugin IDs previously produced MCP tool names containing `@` and `/`, and a malformed serialized JSON schema could stop plugin tool enumeration. This change generates deterministic MCP-safe names and falls back to an object schema when schema conversion fails, while preserving route permissions, consent, scopes, and audit behavior.

This refresh builds on the plugin MCP contract already merged in #2002. Names for existing MCP-safe plugin IDs remain unchanged. The route remains responsible for input validation when a serialized schema cannot be converted.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] Changed files formatted with the repository formatter

- [x] This PR includes AI-generated code — model/tool: OpenAI Codex (GPT-6); earlier contributions as recorded in PR history.

## Screenshots / test output

Screenshots and admin translation checks are not applicable: no UI changes. Prior Discussion approval is not required for this bug fix.

Validation on upstream `fc4a7be4c`:
- MCP authorization, naming, and schema regression suites: 52 tests passed.
- Real sandboxed-plugin and registry-fixture MCP integration: 5 tests passed.
- Full package typecheck passed; type-aware lint reported zero diagnostics.
- Changed code formatted with the repository formatter; current MCP documentation retained and updated.

These are local checks; hosted CI is reported separately by GitHub.


packages/core/src/mcp/server.ts
```diff
@@ -232,6 +235,34 @@ function jsonResult(data: unknown): SuccessEnvelope {
 	return respondData(data);
 }
 
+function pluginToolName(pluginId: string, toolName: string): string {
+	const encodedPluginId = Array.from(new TextEncoder().encode(pluginId), (byte) =>
+		byte.toString(16).padStart(2, "0"),
+	).join("");
+	return `plugin_${encodedPluginId}_${toolName}`;
+}
```
ascorbic: Is this unavoidable? Presumably this is to avoid collisions

ascorbic (commented):
Excellent work. One thing that would be good would be if there was a way we could provide a schema for tool inputs so we can validate and describe them. While we can't serialise a Zod schema directly, maybe a JSON schema would work?

Can you include docs in the PR too, as it will also make it easier to review.

packages/core/src/mcp/server.ts
masonjames: Not strictly unavoidable, but some encoding/escaping step is. Plugin IDs can include characters that are not valid in MCP tool names, like @, /, and -, and we also need to avoid collisions between plugins exposing the same local tool name. Hex was the conservative choice because it is deterministic, reversible, and collision-free.

I agree it is ugly as a client-facing name, though. For this PR I leaned toward keeping it because it is safe and stable. 

I plan to follow this up with a more readable generated-name scheme, probably slugified plugin ID plus a short hash fallback for collisions. Happy to switch in this PR if you’d rather settle the naming surface before merge.

The followup design would be something like:

1. Slugify plugin IDs into MCP-safe names:
`@emdash-cms/plugin-forms` -> `emdash_cms_plugin_forms` or maybe `plugin_forms`.
2. Generate:
`plugin_${safePluginSlug}_${toolName}`.
3. Detect collisions at registration time.
4. If two plugins produce the same safe name, append a short stable hash derived from the original plugin ID: `plugin_forms_submit_form_a13f9c`.
5. Document that the visible MCP tool name is stable but generated, and plugin authors control only the local toolName.

masonjames:
> Excellent work. One thing that would be good would be if there was a way we could provide a schema for tool inputs so we can validate and describe them. While we can't serialise a Zod schema directly, maybe a JSON schema would work?
> 
> Can you include docs in the PR too, as it will also make it easier to review.

Agreed on both.

I’ll add docs in this PR, and I’ll look at adding a constrained `inputSchema` field to `ManifestMcpToolEntry` for client-facing tool introspection. I’m inclined to keep route handlers as the runtime validation source of truth for now, so plugin tools still execute through the existing route validation/context path, while JSON Schema gives MCP clients useful parameter descriptions.

I’ll keep the JSON Schema subset conservative: object root, `properties`, `required`, `additionalProperties`, scalar/array types, `enum`, `description`, `default`, and basic string/number/array constraints. 

No `$ref`/recursive schemas in this pass unless you think we should make that part of the initial contract.

packages/core/src/mcp/server.ts
ascorbic: We could make it collision-proof by using double-underscore and forbidding them in tool names. So it would be `emdash_cms__plugin_forms__submit_form`

masonjames:
Done. I added docs for plugin MCP tools and added manifest-side `inputSchema` support using a constrained JSON Schema subset. Trusted plugins can still pass Zod via `input`; manifest/sandboxed plugins can now provide serializable JSON Schema, which is converted to Zod at MCP registration for client metadata and call validation.

packages/core/src/mcp/server.ts
masonjames: Done. Switched this to the double-underscore form, e.g. `emdash_cms__plugin_forms__submit_form`, and local tool names now forbid `__`. There’s also a stable hash fallback for ambiguous plugin ids.

packages/core/src/plugins/define-plugin.ts
```diff
@@ -123,6 +123,7 @@ function defineNativePlugin<TStorage extends PluginStorageConfig>(
 		allowedHosts = [],
 		hooks = {},
 		routes = {},
+		mcpTools = {},
```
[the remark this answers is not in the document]
masonjames: Addressed. `definePlugin` now validates MCP tool name shape, route existence, and the required `mcp:tools` capability.

packages/core/src/mcp/server.ts
```diff
@@ -419,6 +450,61 @@ export function createMcpServer(): McpServer {
...
+		const name = pluginToolName(tool.pluginId, tool.name);
+		if (registeredToolNames.has(name)) continue;
+		registeredToolNames.add(name);
+
+		server.registerTool(
+			name,
+			{
+				title: tool.title,
+				description: tool.description,
+				inputSchema: tool.input ?? z.record(z.string(), z.unknown()),
```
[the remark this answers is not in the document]
masonjames: Addressed with manifest `inputSchema` support and JSON Schema-to-Zod conversion during MCP registration.

packages/core/src/plugins/manifest-schema.ts
```diff
@@ -249,6 +259,7 @@ export const pluginManifestSchema = z.object({
 			manifestRouteEntrySchema,
 		]),
 	),
+	mcpTools: z.array(manifestMcpToolEntrySchema).optional().default([]),
```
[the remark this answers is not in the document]
masonjames: Addressed. Manifest validation now requires `mcp:tools` when tools are declared, and each MCP tool route must also be declared in `routes`.

packages/core/tests/integration/mcp/plugin-tools.test.ts
```diff
@@ -0,0 +1,194 @@
...
+		const connected = await connectWithPlugins([createPluginWithMcpCapability()], ["content:read"]);
+
+		const result = await connected.client.callTool({
+			name: pluginToolName("test-plugin", "summarize"),
+			arguments: { text: "hello" },
+		});
+
+		expect(isErrorResult(result)).toBe(true);
+		expect(extractText(result)).toContain("[INSUFFICIENT_SCOPE]");
+	});
```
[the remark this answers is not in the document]
masonjames: Added coverage for an admin-scoped token held by a non-ADMIN user.

packages/core/src/mcp/server.ts
```diff
@@ -419,6 +450,61 @@ export function createMcpServer(): McpServer {
...
+						err?.details && typeof err.details === "object"
+							? (err.details as Record<string, unknown>)
+							: undefined;
+					return respondError(code, message, details);
+				}
+
+				return respondData(result.data ?? null);
+			},
+		);
+	}
```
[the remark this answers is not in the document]
masonjames: Added a comment in `server.ts`; plugin MCP tools stay admin-only regardless of route `public` metadata.

packages/core/src/emdash-runtime.ts
```diff
@@ -2149,6 +2159,60 @@ export class EmDashRuntime {
...
+			for (const tool of meta.mcpTools ?? []) {
+				tools.push({
+					pluginId,
+					name: tool.name,
+					title: tool.title,
+					description: tool.description,
+					route: tool.route,
+				});
+			}
+		}
```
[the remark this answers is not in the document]
masonjames: Addressed with runtime dedup by `(pluginId, toolName)` before MCP registration.

packages/core/src/mcp/json-schema.ts
```diff
@@ -0,0 +1,95 @@
...
+import type {
+	ManifestJsonArraySchema,
+	ManifestJsonBooleanSchema,
+	ManifestJsonNumberSchema,
+	ManifestJsonObjectSchema,
+	ManifestJsonSchema,
+	ManifestJsonStringSchema,
+} from "../plugins/types.js";
+
+export function jsonSchemaObjectToZod(
```
ascorbic: Worth seeing if we can use native `fromJSONSchema`: https://zod.dev/json-schema?id=zfromjsonschema#zfromjsonschema

packages/core/src/plugins/types.ts
```diff
@@ -47,13 +47,65 @@ export {
...
+export type ManifestJsonSchema =
+	| ManifestJsonStringSchema
+	| ManifestJsonNumberSchema
+	| ManifestJsonBooleanSchema
+	| ManifestJsonArraySchema
+	| ManifestJsonObjectSchema;
+
+export interface ManifestMcpToolEntry extends SharedManifestMcpToolEntry {
+	/** JSON Schema object used for MCP input validation and client introspection. */
+	inputSchema?: ManifestJsonObjectSchema;
```
ascorbic: You could use `z.core.JSONSchema.ObjectSchema` from Zod if you use `z.fromJSONSchema`

packages/core/src/emdash-runtime.ts
```diff
@@ -2161,19 +2161,27 @@ export class EmDashRuntime {
 
 	getPluginMcpTools(): import("./plugins/types.js").PluginMcpToolRegistration[] {
 		const tools: import("./plugins/types.js").PluginMcpToolRegistration[] = [];
+		const seen = new Set<string>();
+		const addTool = (tool: import("./plugins/types.js").PluginMcpToolRegistration) => {
+			const key = `${tool.pluginId}\0${tool.name}`;
```
ascorbic: clever!

packages/core/tests/integration/mcp/plugin-tools.test.ts
```diff
@@ -15,10 +15,24 @@ import {
 import { setupTestDatabaseWithCollections, teardownTestDatabase } from "../../utils/test-db.js";
 
 function pluginToolName(pluginId: string, toolName: string): string {
```
ascorbic: Defining these here means you're not testing the real thing

packages/core/src/mcp/json-schema.ts
masonjames: Done. Switched the converter to z.fromJSONSchema, kept a small object-root wrapper around it.

packages/core/src/plugins/types.ts
masonjames: Got it. I kept the public manifest type as our constrained subset, then cast through `z.core.JSONSchema.ObjectSchema` inside the wrapper so we don’t make the manifest contract broader or Zod-dependent.

packages/core/src/emdash-runtime.ts
masonjames: I can't take credit though - this was gpt-5.5 high!

packages/core/tests/integration/mcp/plugin-tools.test.ts
masonjames: Good catch. I moved the tool-name generation into a shared helper and updated the integration test to import the real implementation.

packages/core/src/mcp/server.ts
```diff
@@ -419,6 +445,66 @@ export function createMcpServer(): McpServer {
...
+
+		server.registerTool(
+			name,
+			{
+				title: tool.title,
+				description: tool.description,
+				inputSchema:
+					tool.input ??
+					(tool.inputSchema
+						? jsonSchemaObjectToZod(tool.inputSchema)
```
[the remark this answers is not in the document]
masonjames: Addressed in 84dbc425. Plugin MCP inputSchema conversion now goes through safeJsonSchemaObjectToZod(), which catches invalid trusted-plugin JSON schemas, logs a warning, and falls back to the permissive object schema so one bad plugin cannot take down the whole MCP endpoint. Added integration coverage for an invalid trusted-plugin inputSchema staying listed and callable.

packages/core/src/mcp/plugin-tool-name.ts
```diff
@@ -0,0 +1,25 @@
...
+const PLUGIN_ID_DASH_PATTERN = /-/g;
+const AMBIGUOUS_PLUGIN_SEGMENT_PATTERN = /__/;
+
+export function pluginToolName(pluginId: string, toolName: string): string {
+	const readableSegments = pluginId
+		.replace(PLUGIN_SCOPE_PREFIX_PATTERN, "")
+		.split(FORWARD_SLASH_PATTERN)
+		.map((segment) => segment.replace(PLUGIN_ID_DASH_PATTERN, "_"));
+	const readablePluginId = readableSegments.join("__");
+	if (readableSegments.some((segment) => AMBIGUOUS_PLUGIN_SEGMENT_PATTERN.test(segment))) {
```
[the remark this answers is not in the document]
masonjames: Addressed in 84dbc425. pluginToolName() now also treats the joined readable plugin id as ambiguous when it contains triple underscores, so boundary cases like @a-/b and @a/-b get distinct hash-suffixed names instead of collapsing.

packages/core/src/mcp/plugin-tool-name.ts
```diff
@@ -0,0 +1,25 @@
...
+}
+
+function stableToolNameHash(value: string): string {
+	let hash = 0x811c9dc5;
+	for (let index = 0; index < value.length; index++) {
+		hash ^= value.charCodeAt(index);
+		hash = Math.imul(hash, 0x01000193);
+	}
+	return (hash >>> 0).toString(16).padStart(8, "0");
+}
```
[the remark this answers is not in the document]
masonjames: Addressed in 84dbc425. Added packages/core/tests/unit/mcp/plugin-tool-name.test.ts covering readable scoped names, stable hash fallback for ambiguous dash normalization, and the @a-/b vs @a/-b boundary collision case.

masonjames:
Synced this with current `origin/main` and resolved the conflicts. I kept the upstream docs/sidebar reorg and placed the MCP Tools doc under Sandboxed Plugins, kept the MCP test runtime wired to the real plugin route/tool handlers, and added the missing `getPluginMcpTools` test-runtime stub surfaced by the merge. Also updated the changeset from the old registry CLI package name to `@emdash-cms/plugin-cli`.

The recent review points are still covered here: invalid trusted-plugin `inputSchema` falls back safely, the `@a-/b` vs `@a/-b` tool-name collision is covered, and the dedicated `pluginToolName` unit tests are in place. Local validation and CI are green.

masonjames:
Refreshed onto current `main` (`7f20c8327`) in `42a37dfde`.

Local validation is clean: full build, 52 focused MCP tests, package typecheck, zero-diagnostic type-aware lint, formatting, and `git diff --check`. Hosted CI has accepted the new head and is currently queued.

PR #1075, opened 2026-05-17
fix(import): narrow unknown error body in wordpress-plugin source (#1053)

### LINKED ISSUE #1053: Type errors in emdash 0.12.0 under strict TypeScript
### Problem
Two files in emdash's source fail `tsc --noEmit` when consumed by a strict TypeScript project. Since emdash ships `.ts` source (not compiled `.d.ts`), `skipLibCheck` doesn't help.

### Errors
**1. `src/database/repositories/byline.ts:204`**
error TS2345: Argument of type 'Kysely | Transaction' is not assignable to parameter of type 'Kysely'.
`listTablesLike(trx, "ec_%")` passes a `Transaction<Database>` to a param typed `Kysely<any>`. Kysely's `Transaction` is no longer assignable to `Kysely` due to a `mergeInto` variance change.

**Fix: `listTablesLike(trx as unknown as Kysely<any>, "ec_%")`**

**2. `src/import/sources/wordpress-plugin.ts:302`**
error TS18046: 'error' is of type 'unknown'.`response.json()` returns `unknown` under `@cloudflare/workers-types`, so `error.message` fails without narrowing.

**Fix: `const error: any = await response.json().catch(() => ({}));`**

### Workaround
Patching via `pnpm.patchedDependencies` for now.

### Steps to reproduce
```bash
mkdir repro && cd repro
npm init -y
npm install emdash@0.12.0 @types/sax @cloudflare/workers-types typescript
tsconfig.json:

{
  "compilerOptions": {
    "strict": true,
    "module": "ESNext",
    "moduleResolution": "bundler",
    "skipLibCheck": true,
    "noEmit": true
  },
  "include": ["index.ts"]
}

index.ts:
import "emdash";
npx tsc --noEmit
Produces the two errors above. skipLibCheck: true has no effect because emdash ships .ts source, not .d.ts

### Logs / error output

```shell

```

### Environment
- emdash: 0.12.0
- TypeScript: 5.9.3 (strict mode, `skipLibCheck: true`)
- Kysely: 0.27.x

ascorbic:
## What does this PR do?

Closes #1053.

Fixes the one real type error #1053 reported, and adds a regression guard so the whole class can't ship again.

**The reporter's two errors, re-triaged:**

- **`byline.ts:204` (TS2345 kysely `Transaction`→`Kysely`)** — a TypeScript **5.x** variance behaviour. It does **not** reproduce on TypeScript 6 and is already resolved by the TS6 upgrade in #1074 (merged). No code change needed.
- **`wordpress-plugin.ts` (TS18046 `'error' is of type 'unknown'`)** — real. The analyze-endpoint error body from `response.json()` is `unknown` under `@cloudflare/workers-types` and was read without narrowing. This file ships as **raw source** via the `emdash/routes/*` export, so it fails strict consumer typechecks. Fixed with assertion-free narrowing (clean under both the consumer's Workers-typed config *and* emdash's own `astro/client` typecheck; no `as`, no lint warning, runtime unchanged).

**The reporter's premise was right, their repro was wrong.** Bare `import "emdash"` resolves to compiled `dist/*.d.mts` and `skipLibCheck` covers it — verified clean. The errors only surface through the **source-exported subpaths** (`emdash/routes/*`, `emdash/ui`, `emdash/api/*`, `emdash/auth/providers/*`), which ship raw `.ts` that `skipLibCheck` cannot skip. Our own `pnpm typecheck` misses this: it runs `tsgo` against our lenient tsconfig (excludes `src/astro/**`, no Workers types). A strict consumer on real `tsc` deployed to Workers sees a different program.

**Regression guard (`scripts/typecheck-public-source.mjs` + `fixtures/public-types/`):** reconstructs the consumer view — real `tsc`, `@cloudflare/workers-types`, `emdash/locals` — over every `.ts`/`.tsx` source-exported subpath, and compares per-`(file, errorCode)` diagnostic counts to a committed baseline (same idea as the perf query-count snapshot). Wired into the CI typecheck job.

- The `.astro` / `virtual:emdash/*` (allowlisted) module-resolution surface is the **documented boundary** — a hermetic type-only harness can't resolve it; real sites do, via `astro check` (covered by the demo/template typecheck jobs). `node_modules/` diagnostics are dropped (third-party, unstable pnpm-store paths).
- **Hard sanity gates** make every degenerate `tsc` outcome fail loudly (unexpected exit code, unparseable error lines / format drift, exit-0-with-errors, path-rooting drift) — the guard can't silently pass.
- A **hardcoded #1053 sentinel** asserts `wordpress-plugin.ts` TS18046 never returns, independent of the baseline file. TDD-verified: reverting the fix makes the harness FAIL naming exactly that error; restored → green.
- The **23 baseline groups are pre-existing strict-consumer debt** (implicit-any, `ui.ts`/`astro-portabletext`, dual-package `Kysely` realm, `Locals.tokenScopes`, import payload shapes). Tracked separately and burned down incrementally; the harness fails on any *growth* or *new* shipped-source error, shrinking as debt is fixed.

Built on #1074 (TS6). Adversarially reviewed twice (the narrowing fix and the harness); review found four false-pass/flaky vectors in the harness, all fixed before this PR (exit/parse sanity gates, count-based baseline, node_modules exclusion, tightened virtual allowlist).

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests
- [x] Chore (dependencies, CI, tooling)

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes (PR adds 0 diagnostics; repo baseline of 38 unchanged)
- [x] `pnpm test` passes (or targeted tests for my change) — the harness itself is the regression test (TDD red/green proven via the #1053 sentinel); the one-line narrowing has no behavioural change to unit-test
- [x] I have added/updated tests for my changes — the public-source typecheck harness is the regression guard for #1053 and the whole shipped-source class
- [x] User-visible strings in the admin UI are wrapped for translation (if applicable) — N/A
- [x] I have added a changeset — `emdash` patch for the wordpress-plugin narrowing fix
- [ ] New features link to an approved Discussion — N/A (not a feature)


ascorbic:
Superseded by #1076, which resolves #1053 as one coherent change: the `wordpress-plugin` narrowing fix plus compiling the source-exported subpaths to dist (the structural root-cause fix). The fixture-based consumer-typecheck harness here was exploratory scaffolding the final architecture makes unnecessary — #1076 replaces it with a fast static guard. Closing in favour of #1076.

PR #1086, opened 2026-05-18
fix(core): preserve taxonomy + content_taxonomies data on D1 in migration 036

### LINKED ISSUE #1021: Migration 036 (0.12.0) silently destroys content_taxonomies data on Cloudflare D1
### Description

Migration 036_i18n_menus_and_taxonomies calls PRAGMA foreign_keys = OFF before executing DROP TABLE taxonomies as part of rebuildTaxonomies. On Cloudflare D1, this pragma is silently ignored — D1 always enforces foreign keys (PRAGMA foreign_keys always returns 1 and cannot be changed).

As a result, DROP TABLE taxonomies triggers the ON DELETE CASCADE declared in content_taxonomies:

-- From 001_initial migration:
  constraint "content_taxonomies_taxonomy_fk"
    foreign key ("taxonomy_id") references "taxonomies"("id")
    on delete cascade

All rows in content_taxonomies are deleted before the table is even rebuilt. Since rebuildContentTaxonomies runs after rebuildTaxonomies, it finds an empty source table and produces an empty result. The data loss is silent — the migration reports success.

 **Impact**

  - All post–taxonomy associations are destroyed
  - Category archive pages show no content
  - getEntryTerms() returns empty for all entries
  - Homepage links to articles use root category path (2 segments) instead of canonical leaf path, causing 404s

**Workaround**

  Before deploying the upgrade:

1.   D1 Time Travel restore to pre-upgrade state
2. Back up content_taxonomies: CREATE TABLE ct_backup AS SELECT * FROM content_taxonomies
3. Clear it: DELETE FROM content_taxonomies
4. Recreate without FK (dropping the child table does not trigger cascade): 
CREATE TABLE ct_new (collection text not null, entry_id text not null, taxonomy_id text not null, constraint content_taxonomies_pk primary key (collection, entry_id, taxonomy_id)); 
INSERT INTO ct_new SELECT * FROM content_taxonomies; 
DROP TABLE content_taxonomies; 
ALTER TABLE ct_new RENAME TO content_taxonomies;
5. Deploy — migration 036 detects no FK on content_taxonomies → skips rebuildContentTaxonomies → data preserved
6. Restore data: INSERT INTO content_taxonomies SELECT * FROM ct_backup; DROP TABLE ct_backup

**Suggested fix**

In rebuildTaxonomies, instead of DROP TABLE taxonomies, use the same table-rebuild pattern used elsewhere in migration 036: create a new table, copy data, drop old, rename. This avoids triggering any CASCADE on dependent tables, regardless of whether FK enforcement can be disabled.

Alternatively, before dropping taxonomies, explicitly save and restore content_taxonomies data within the same migration, making the operation idempotent on D1.

### Steps to reproduce

1. Deploy EmDash on Cloudflare D1 with a populated content_taxonomies table (e.g. after a WXR import or manual taxonomy tagging)
2. Upgrade to a version that includes migration 036 (≥ 0.10.0)
3. Migration runs → content_taxonomies is now empty

### Environment

- Cloudflare D1 (SQLite), compatibility_date: 2025-01-01
- PRAGMA foreign_keys always returns 1, cannot be set to 0
- EmDash 0.12.0, migrating from 0.6.0

### Logs / error output

```shell

```

ascorbic:
## What does this PR do?

Fixes silent data loss in migration 036 on Cloudflare D1.

D1 ignores `PRAGMA foreign_keys = OFF`, and its replacement `PRAGMA defer_foreign_keys = ON` only defers constraint *validation* — it doesn't suppress CASCADE *actions* (verified empirically). So when migration 036 ran the i18n rebuild of `taxonomies` on D1, dropping the old table fired:

1. The `ON DELETE CASCADE` on `content_taxonomies.taxonomy_id` → every post/taxonomy association wiped.
2. The `ON DELETE SET NULL` on the new `taxonomies_new.parent_id` self-FK → every hierarchical taxonomy's parent flattened to NULL.

Both were silent. The reporter (#1021) hit (1) upgrading 0.6.0 → 0.12.0 on a populated WP-imported site. (2) was caught by adversarial review of the initial fix — same bug class, same blast radius.

The migration now physically removes those FK relationships before any drop:

- **`content_taxonomies` FK**: rebuilt as the first step of `up()`, copying values verbatim. By the time `rebuildTaxonomies` runs, no FK points at the old `taxonomies` table.
- **`taxonomies` self-FK**: declared as `REFERENCES taxonomies_new(id)` during creation. SQLite's `ALTER TABLE ... RENAME TO` rebinds the FK target to the new name on rename. Confirmed against both better-sqlite3 and the D1-mimicking test harness.

The rollback path is restructured to match. `down()` now:

1. Asserts `content_taxonomies` has no rows pointing at translation_groups with no surviving `taxonomies` row (previously this would surface mid-rollback, after other tables were already irreversibly stripped — leaving the user unrecoverable).
2. Remaps `content_taxonomies.taxonomy_id` back to row ids while the locale/translation_group columns are still readable.
3. Rebuilds the four widened tables in their pre-036 shape.
4. Restores the `content_taxonomies` FK as the final step.

Also restores `idx_content_taxonomies_term` (from migration 015) after each rebuild — SQLite drops indexes with the table, and neither old nor new code recreated it.

**Forward-fix only.** Installs that already lost data running 036 will need D1 Time Travel to recover. Noted in the changeset.

Closes #1021

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Implementation notes

- `rebuildContentTaxonomies` (the existing helper) is unchanged in shape but moved to run first in `up()`. The value remap it used to do was load-bearing only after `rebuildTaxonomies` had seeded `translation_group = id`, which now happens later. The legacy `taxonomy_id` values are already row ids that resolve as translation_group references after the migration completes; no remap is needed. The comment in the file flags this coupling explicitly so future changes to the `translation_group` seed don't silently break things.
- `down()` splits the previous `rebuildContentTaxonomiesDown` into `remapContentTaxonomiesDown` (just the value UPDATE — runs against the FK-less post-up table) and `restoreContentTaxonomiesFk` (the FK-re-adding rebuild — runs after `taxonomies` is back to its pre-036 shape).
- `assertContentTaxonomiesResolve` is a new fail-fast check mirroring `assertSingleLocale`'s existing pattern. Surfaces dangling associations up front rather than at the moment the FK is restored.
- Removed `PRAGMA foreign_keys = OFF/ON` from both `up()` and `down()` (SQLite path). Confirmed by audit that no intermediate operation depends on FK enforcement being disabled: `_emdash_menus`, `_emdash_menu_items`, `_emdash_taxonomy_defs` have no FKs in the pre-036 schema; `taxonomies.parent_id` self-FK is handled by the temp-name trick above; `content_taxonomies` FK is removed in step 1.
- Postgres path (`pgWiden`, `pgRemapContentTaxonomies`) is unchanged.

## Tests

Three new regression tests run against a Kysely instance backed by better-sqlite3 with `PRAGMA foreign_keys = OFF` patched to a no-op — mimicking D1's locked-on FK enforcement. They fail against the pre-fix migration and pass against this one:

- `preserves taxonomy parent_id hierarchy on D1 (regression for #1021)` — seeds a hierarchical taxonomy, runs `up()`, verifies `parent_id` is preserved.
- `preserves content_taxonomies rows on D1 (regression for #1021)` — seeds a post→term association, runs `up()`, verifies the row survives.
- `refuses to rollback when content_taxonomies has dangling rows` — verifies the new fail-fast assertion fires before any destructive work.

The `createD1LikeDatabase` helper is local to this test file. A workspace-wide D1 dialect for `describeEachDialect` is a separate ambient task; this is the smallest thing that catches the specific bug class on existing infrastructure.

All 33 migration tests pass. 156 broader DB tests pass.

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes (0 diagnostics on changed files)
- [x] `pnpm test` passes for the migration test files; broader DB test suite (`tests/integration/database/`, `tests/unit/database/`) is also green
- [ ] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) — n/a

- [x] This PR includes AI-generated code — model/tool: Claude Opus 4.7 via OpenCode. Adversarial review pass run with the same model in a sub-agent; findings (self-FK cascade, dangling-row case, missing `idx_content_taxonomies_term`, regression test gap) all incorporated.

### DIFF
--- a/.changeset/fix-1021-d1-taxonomy-cascade.md
+++ b/.changeset/fix-1021-d1-taxonomy-cascade.md
@@ -1,0 +1,9 @@
+---
+"emdash": patch
+---
+
+Fixes silent data loss in migration 036 on Cloudflare D1 (#1021). D1 ignores `PRAGMA foreign_keys = OFF` and its replacement `defer_foreign_keys` only defers constraint validation — it doesn't suppress CASCADE actions — so dropping the old `taxonomies` table during the i18n rebuild fired both the `ON DELETE CASCADE` on `content_taxonomies` (wiping all post-taxonomy associations) and the `ON DELETE SET NULL` on the new table's own `parent_id` (flattening taxonomy hierarchies).
+
+The migration now physically removes those FK relationships before any drop: it rebuilds `content_taxonomies` without the FK as the first step of up(), and points the new `taxonomies` table's self-FK at its temporary name (`taxonomies_new`) which SQLite rebinds on RENAME. The rollback path is restructured to match and now refuses to run when `content_taxonomies` has rows referencing translation groups with no surviving `taxonomies` row, surfacing dangling data before any destructive work. The `idx_content_taxonomies_term` index from migration 015 is also restored after each rebuild (it was previously dropped with the table and never recreated).
+
+This is forward-fix only — installs that already lost data when running 036 will need to restore from D1 Time Travel.
--- a/.changeset/fix-1021-d1-taxonomy-cascade.md
+++ b/.changeset/fix-1021-d1-taxonomy-cascade.md
@@ -2,8 +2,12 @@
 "emdash": patch
 ---
 
-Fixes silent data loss in migration 036 on Cloudflare D1 (#1021). D1 ignores `PRAGMA foreign_keys = OFF` and its replacement `defer_foreign_keys` only defers constraint validation — it doesn't suppress CASCADE actions — so dropping the old `taxonomies` table during the i18n rebuild fired both the `ON DELETE CASCADE` on `content_taxonomies` (wiping all post-taxonomy associations) and the `ON DELETE SET NULL` on the new table's own `parent_id` (flattening taxonomy hierarchies).
+Fixes silent data loss in migration 036 on Cloudflare D1 (#1021). D1 ignores `PRAGMA foreign_keys = OFF` and its replacement `defer_foreign_keys` only defers constraint validation, it doesn't suppress CASCADE actions, so dropping any table during the i18n rebuild fired its child cascades. Three FK relationships were affected:
 
-The migration now physically removes those FK relationships before any drop: it rebuilds `content_taxonomies` without the FK as the first step of up(), and points the new `taxonomies` table's self-FK at its temporary name (`taxonomies_new`) which SQLite rebinds on RENAME. The rollback path is restructured to match and now refuses to run when `content_taxonomies` has rows referencing translation groups with no surviving `taxonomies` row, surfacing dangling data before any destructive work. The `idx_content_taxonomies_term` index from migration 015 is also restored after each rebuild (it was previously dropped with the table and never recreated).
+- `content_taxonomies.taxonomy_id -> taxonomies(id) ON DELETE CASCADE` wiped all post-taxonomy associations.
+- `taxonomies.parent_id -> taxonomies(id) ON DELETE SET NULL` flattened taxonomy hierarchies.
+- `_emdash_menu_items.menu_id -> _emdash_menus(id) ON DELETE CASCADE` wiped every menu item on the install (along with `parent_id -> _emdash_menu_items(id) ON DELETE CASCADE` mopping up nested items).
 
-This is forward-fix only — installs that already lost data when running 036 will need to restore from D1 Time Travel.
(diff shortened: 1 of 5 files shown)


ascorbic:
/review there has been a new commit since your review. Please re-check

PR #1142, opened 2026-05-22
fix(admin): emit nested-list level in prosemirrorToPortableText

OrangeManLi:
## What does this PR do?

Fixes nested-list serialization in the admin's Portable Text editor. `PortableTextEditor`'s local `convertList` (`packages/admin/src/components/PortableTextEditor.tsx`) walks each `listItem`'s content but only handles `paragraph` children — it silently skips any sibling `bulletList` / `orderedList`, and emits every block with `level: 1` regardless of nesting depth.

As a result, Tab-indenting a bullet in the editor (which produces a valid nested ProseMirror tree with `bulletList → listItem → [paragraph, bulletList → listItem → paragraph]`) saves portable-text where every list block has `level: 1` and the nested child items are flattened — or, in the bullet-with-only-a-nested-list shape, dropped entirely.

`packages/core/src/content/converters/prosemirror-to-portable-text.ts` already has the correct recursive form (`convertList → convertListItem → convertListItemNested`) — this PR mirrors that behavior in the admin's local copy so the editor's `onChange` output matches.

### Repro (before this fix)

ProseMirror doc:
```
bulletList
  listItem
    paragraph "Parent"
    bulletList
      listItem
        paragraph "Child"
```

Output from `_prosemirrorToPortableText`:
```js
[
  { _type: 'block', listItem: 'bullet', level: 1, children: [{ text: 'Parent' }] },
  // ❌ "Child" missing entirely
]
```

### After this fix

```js
[
  { _type: 'block', listItem: 'bullet', level: 1, children: [{ text: 'Parent' }] },
  { _type: 'block', listItem: 'bullet', level: 2, children: [{ text: 'Child' }] },
]
```

Closes #

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [ ] `pnpm typecheck` passes
- [ ] `pnpm lint` passes
- [x] `pnpm format` has been run (auto-format workflow already pushed a `style: format` commit on this branch)
- [x] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable) — N/A, no user-visible strings touched
- [x] I have added a [changeset](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#changesets) (if this PR changes a published package) — `.changeset/whole-buses-repair.md`, patch bump for `@emdash-cms/admin`
- [ ] New features link to an approved Discussion — N/A, this is a bug fix

> The three `pnpm typecheck` / `lint` / `test` boxes are unchecked because the npm registry was returning intermittent fetch failures during PR prep and `pnpm install` couldn't complete locally. The new test in `packages/admin/tests/components/PortableTextEditor.list.test.ts` is small and self-contained — CI will exercise it.

- [x] This PR includes AI-generated code — model/tool: Claude Opus 4.7 (via Claude Code)

## Screenshots / test output

The added test file (`packages/admin/tests/components/PortableTextEditor.list.test.ts`) covers:

- single-level bullet list (sanity, both items `level: 1`)
- 2-level nest (`level: 1` then `level: 2`)
- mixed bullet+number nesting (`bullet level: 1` then `number level: 2`)
- 3-level deep nesting (`level: 1, 2, 3`)

Each case builds a synthetic ProseMirror doc and asserts the `level`/`listItem`/text tuples returned by `_prosemirrorToPortableText`.


OrangeManLi:

OrangeManLi:
Heads-up on the **Translation Overview** check failure — I dug in and I don't think it's caused by this PR.

The job dies with:

\`\`\`
[ERR_PNPM_LOCKFILE_MISSING_DEPENDENCY] Broken lockfile: no entry for 'kysely@0.27.6' in pnpm-lock.yaml
This issue is probably caused by a badly resolved merge conflict.
\`\`\`

That constraint comes from `packages/workerd/package.json` (`"kysely": "^0.27.0"`, plus a peer `>=0.27.0`), which was introduced in 02ed8ba32ef1f4301d84465b934430eee08eef74 (Node.js plugin isolation via workerd sandbox, #426). But `pnpm-lock.yaml` on `main` only has `kysely@0.29.2`, no `0.27.x` entry — so any PR whose base includes that commit fails the `pnpm install --frozen-lockfile` step Translation Overview runs.

I checked the other currently-open PRs and the same Translation Overview failure is present on #1141 and #1143, which corroborates that the lockfile drift is base-side, not introduced here.

Happy to push a lockfile regeneration on this branch if maintainers prefer to fix it through here, but I figured it'd be cleaner to surface it for a `main` fix so the same failure doesn't keep landing on every PR. Just let me know.

OrangeManLi:
Pushed a follow-up addressing the inverse direction: `convertPTList` was also flattening nested lists (it ignored `block.level` and just emitted siblings), and the run-grouping in `portableTextToProsemirror` broke on `listItem`-type changes — so opening a saved nested list back in the editor would lose its hierarchy as soon as a different list type (e.g. an ordered child under a bullet) appeared.

Changes in e6ccb5b9:

- Rewrites `convertPTList` to recurse on `level`, mirroring `@emdash-cms/core/portable-text-to-prosemirror.convertList`.
- Loosens the outer run-grouping in `portableTextToProsemirror` to fold `level > 1` blocks into the current run regardless of their `listItem` type (the recursive call inside handles type switches per subtree).
- Adds PT → PM tests and a PT → PM → PT round-trip test covering the regression case (`[bullet L1, number L2, bullet L1]` must stay as one bullet list with a numbered child, not three lists at the document root).

Translation Overview should also clear now — rebased onto `main` which picked up #1144's kysely workspace-catalog fix.

ascorbic (approved):
Thanks!

PR #1143, opened 2026-05-22
fix(core): emdash types reads un-enveloped /schema response

nesmachny:
## What does this PR do?

`emdash types` crashes with `TypeError: Cannot read properties of undefined (reading 'collections')`.

The CLI's `types` command calls `client.schemaExport()`, which routed through the private `request()` helper. `request()` unwraps the standard `{ data }` API envelope and returns `json.data`. But `GET /_emdash/api/schema` returns a **bare** `{ collections, version }` object — there is no `data` wrapper — so `schemaExport()` returned `undefined`, and the CLI then threw on `schema.collections`.

`schemaExport()` now reads the raw response directly (`requestRaw()` + `assertOk()`), mirroring its sibling `schemaTypes()`. The `/schema` endpoint and its other consumers are unchanged — this is a non-breaking, client-side fix.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes (0 errors)
- [x] `pnpm test` passes (or targeted tests for my change) — `packages/core/tests/unit/client/client.test.ts`, 18/18
- [ ] User-visible strings in the admin UI are wrapped for translation (if applicable) — N/A, no admin UI changes
- [ ] New features link to an approved Discussion — N/A, bug fix

- [x] This PR includes AI-generated code — model/tool: Claude Opus 4.7 (Claude Code)

## Screenshots / test output

The new test reproduces the bug: against the un-enveloped `/schema` response it failed with the exact production error before the fix, and passes after it.

```
 Test Files  1 passed (1)
      Tests  18 passed (18)
```


nesmachny:

ascorbic:
Thanks for this, and for the failing-then-passing test, your diagnosis of the root cause was exactly right.
The same crash (emdash types reading collections off undefined) was fixed from the other side in #1458, which merged back in June. Rather than have the client read /schema raw, #1458 made the /schema route return the standard `{ data }` envelope like every other endpoint, so the normal request() unwrap works and the API surface stays consistent.
That means this PR is now superseded by #1458. We went with normalising the server rather than special-casing the client, but the report and the fix were spot on. Hope you'll send more.

PR #1226, opened 2026-05-29
fix: server-side content list search + locale-aware list indexes

### LINKED ISSUE #1219: Bug - Content entities - Navigation and search fail (or run extremely slowly) on large datasets
### Description

When you have a large set of content entities, the navigation and search of those content entities doesn't work as expected.

### Navigation
For navigation, large sets of content entities seem to hit some API limits (or latency) and take a long time to query -- in my testing around 20 seconds or more to load a new panel if you're trying to go back pretty far.

https://github.com/user-attachments/assets/31f81b4f-8678-4af5-a93e-55506afcf801

### Search
A similar problem happens when you're trying to search for a content entity that lives further back in the collection.

From the initial page, it isn't included in the index

Image

It's only once you navigate to that page directly (or at least near it, haven't tested fully), that the entity item will be truly searchable.

### Steps to reproduce

1. Create a content entity.
2. Add a **lot** of entries. Over 1000.
3. Try searching or navigating those entries.

### Environment

- emdash: 0.15.0
- Cloudflare Workers
- MacOS, Chrome

### Logs / error output

```shell

```

scottbuscemi:
## What does this PR do?

Closes #1219.

On collections with 1000+ entries:
- **Search** only matched rows already loaded on the current page. An entry far back in the collection couldn't be found from the first page — it became searchable only after you navigated near it. (`ContentList` filtered `data.pages.flatMap(...).filter(byTitle)` client-side; the API client never sent a search term; the list query/handler/repository had no search filter.)
- **Navigation** was slow on large, i18n-enabled collections partly because locale-filtered ordered lists weren't covered by an index.

**Fix:**

*Server-side search*
- `contentListQuery` schema gains an optional `q` (trimmed, 1–200 chars).
- `ContentRepository.findMany`/`count` apply a case-insensitive substring filter across handler-resolved `searchColumns`, OR'd. `lower()` on both sides for SQLite/Postgres parity; LIKE wildcards (`% _ \`) in the query are escaped with `ESCAPE '\'`.
- `handleContentList` resolves the searchable columns from the collection's fields — always `slug`, plus `title`/`name` when defined — so collections without those columns don't error.
- Admin: `fetchContentList` sends `q`; `ContentList` debounces the search box (300ms) and reports it up via a new `onSearchChange` prop. The search box now stays mounted on a zero-match query (so it can be cleared), the empty state distinguishes "no content" from "no matches", and in server mode the result count reflects the server `total` rather than the loaded page size. Client-side filtering is retained only as legacy behavior when `onSearchChange` isn't supplied.

*Navigation*
- New migration `041_content_locale_list_index` adds locale-aware composite list indexes across all `ec_*` tables (idempotent, forward-only), mirrored in `SchemaRegistry` for new collections. Index names are kept short enough that the `updated`/`created` discriminator survives Postgres's 63-byte identifier truncation for long collection slugs (a collision the first cut of these indexes would have hit at slug length ≥ 40).

### Deliberately out of scope
The UI's page-number pagination over forward-only cursors is the bigger driver of deep-navigation latency; that's a larger UX refactor, deferred. The per-page COUNT is also intentionally kept — there's a regression guard requiring `total` on every page.

**Testing:** `packages/core/tests/integration/content/content-list-search.test.ts` (`describeEachDialect`): deep entry, case-insensitive, slug match, wildcard-escape, unfiltered fallback. Admin `ContentList.test.tsx`: server-mode query reporting, empty-results search box, server-total count, plus a pure-string index-name distinctness test. Full core suite green; `pnpm lint:json` clean; `emdash` + `@emdash-cms/admin` typecheck pass. (Live-Postgres truncation not exercised locally — no `PG_CONNECTION_STRING`; modelled by the string test and runs against Postgres in CI.)

**Manual verification:**
1. Collection with 1000+ entries.
2. Search a term that only matches a deep entry → it appears (previously didn't until you paged near it).
3. With i18n enabled, confirm locale-filtered lists still page correctly.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

- [x] This PR includes AI-generated code — model/tool: Claude Opus 4.8 (Claude Code, review-fix commits)

---

### Try this PR

**[Open a fresh playground →](https://fix-1219-content-list-search-emdash-playground.emdash-cms.workers.dev)**

<sub>Tracks `fix/1219-content-list-search`. Updated automatically when the playground redeploys.</sub>


ascorbic:
@scottbuscemi tell ya boy to use the PR template!

PR #1246, opened 2026-05-31
search: don't assume a title column exists in searchable collections

### LINKED ISSUE #1178: Bug: search MCP tool returns D1_ERROR: no such column: c.title when collections argument is omitted
### Description

The `search` tool exposed by the emdash MCP server returns a D1 SQL error when called without the optional `collections` filter argument. Supplying `collections` makes the same query succeed, so the bug is specifically in the "search across all collections" code path.     This was discovered when I pointed my Anthropic AI to consume the MCP and it's constantly tripping over this error.

Error returned:

```
[SEARCH_ERROR] D1_ERROR: no such column: c.title at offset 47: SQLITE_ERROR
```

The error message suggests the un-filtered search SQL aliases a table as `c` and references a `c.title` column that doesn't exist at that level (entry titles live inside the per-entry `data` JSON blob, not as a top-level column on any joined table).

### Steps to reproduce

Reproduced via a direct MCP Streamable-HTTP client (no Anthropic in the loop) to isolate the bug to the emdash MCP server.

1. Provision emdash on Cloudflare Workers with D1 + R2 bindings; have multiple collections seeded with `search` enabled in their `supports` array.
2. Mint a personal API token with `content:read` scope (`EMDASH_MCP_TOKEN`).
3. Connect to `/_emdash/api/mcp` via the MCP Streamable-HTTP transport (`initialize` → `notifications/initialized` → `tools/call`).
4. Call the `search` tool twice in the same session — same query, only difference is the `collections` argument:

   **Call A — with `collections`:**

   ```json
   {
     "name": "search",
     "arguments": {
       "query": example software",
       "collections": ["pages", "posts"],
       "limit": 5
     }
   }
   ```

   **Call B — without `collections`:**

   ```json
   {
     "name": "search",
     "arguments": {
       "query": "example software",
       "limit": 5
     }
   }
   ```

## Expected behavior

Calling `search` without `collections` should run the query across all collections that have `search` enabled and return a (possibly empty) `items` array.

## Actual behavior

Un-filtered `search` calls return `is_error: true` with:

```
[SEARCH_ERROR] D1_ERROR: no such column: c.title at offset 47: SQLITE_ERROR
```

`search` calls *with* a `collections` argument (e.g. `["pages", "posts"]`) succeed and return results normally.

### Environment

- `emdash`: `0.14.0` (also reproduced on `0.10.0`)
- `@emdash-cms/cloudflare`: `0.14.0`
- `astro`: `6.1.2`
- `@astrojs/cloudflare`: `13.1.6`
- Runtime: Cloudflare Workers (paid plan), `compatibility_date: 2026-03-29`, `compatibility_flags: ["nodejs_compat"]`
- Bindings: D1 (`DB`), R2 (`MEDIA`), KV (`SESSION`), Worker Loaders (`LOADER`)
- Node: 22.x (for build)
- OS: Windows 11 (build host), Cloudflare Workers (runtime)

`astro.config.mjs`:

```js
emdash({
  database: d1({ binding: "DB", session: "auto" }),
  storage: r2({ binding: "MEDIA" }),
  sandboxRunner: sandbox(),
})
```

### Logs / error output

```shell
Output from a direct MCP client hitting the same `/_emdash/api/mcp` endpoint twice in one session, varying only the `collections` argument:


MCP endpoint: https://<site>/_emdash/api/mcp

✓ initialized (session <no session id>)
  protocol 2024-11-05

── search WITH collections=['pages','posts'] ────────────────────────────────
✓ is_error: false
{
  "items": []
}

── search WITHOUT collections (bug repro) ────────────────────────────────
✗ is_error: true
[SEARCH_ERROR] D1_ERROR: no such column: c.title at offset 47: SQLITE_ERROR


Notes:
- The server does not return an `Mcp-Session-Id` header on `initialize` (`<no session id>` above) but `tools/call` still works without it.
- The same behavior originally surfaced via the Anthropic Messages API with `mcp_servers` wired to this endpoint — Claude would issue `search` calls both with and without `collections`, and only the latter failed. The direct-client repro above isolates the bug to emdash.
```

mvanhorn:
## What does this PR do?

Search and suggestions hardcoded `c.title` in the generated SQL, so searching a collection whose schema has no `title` field failed with `D1_ERROR: no such column: title` and broke multi-collection search entirely.

The query now detects whether the collection actually defines a `title` field (via the fields registry, with identifier validation) and only selects it when present: `NULL` in the main search projection and `COALESCE(slug, id)` for suggestions, dropping the `title IS NOT NULL` filter when there is no title field.

Closes #1178

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [ ] `pnpm typecheck` passes 
- [x] `pnpm lint` passes (changed files)
- [x] `pnpm test` passes (targeted tests for my change: search + mcp/search, 13 passing)
- [x] User-visible strings: N/A (no admin UI strings)
- [x] I have added a changeset (`emdash: patch`)
- [x] New features link to an approved Discussion: N/A (bug fix)

- [x] This PR includes AI-generated code — model/tool: GPT-5.5 (Codex CLI)

## Screenshots / test output

```
tests/integration/mcp/search.test.ts + tests/integration/search/suggest.test.ts
Test Files  2 passed (2)
     Tests  13 passed (13)
```


ascorbic:
You could avoid the extra query by updating `getSearchableFields` to get the info you need in the same query. 

If you want to improve things further in a follow-up, you could put all of the metadata queries (getSearchConfig, getSearchableFields, ftsTableExists at least) into a new `getCollectionFieldInfo` function that's wrapped with `requestCached("collectionFields:" + slug, ...)` or something.

ascorbic (changes requested):
Avoid the extra hot-path query

mvanhorn:
Done in 50dbafc - added `getSearchableFieldInfo` on FTSManager, which joins collections+fields once and returns both the searchable slugs and whether a title field exists. That replaces the old two-query `getSearchableFields` AND the per-route `collectionHasField` probe, so search routes are now one query cheaper than main rather than one more. Also rebased onto main; the snapshot diff should reflect the reduction once CI regenerates it.

The `getCollectionFieldInfo` + `requestCached` consolidation sounds right as the follow-up - happy to pick that up in a separate PR.

ascorbic:
Thanks. This has been superseded by #1550

PR #1258, opened 2026-06-01
feat: extensible bylines

MohamedH1998:
## What does this PR do?

Adds support for site-specific byline metadata — Twitter handle, pronouns,
company, localised job title, etc. — via a new `/byline-schema` admin
screen. Sites can register custom fields with their own type, validation,
and per-locale storage policy without editing code; values surface on
`BylineSummary.customFields` for the frontend.

Custom-field values can be set at both create and update time through
the same `customFields` map on `POST` and `PUT /_emdash/api/admin/bylines`.
In the admin, registered fields render inline with Name, Bio, etc. — no
"Custom fields" section header — and are available in both the **New byline**
and **Edit byline** dialogs. The schema link sits at the top of the
Bylines page (admin-only) rather than the global sidebar so admins find
it in context.

### Design choices worth flagging

- **Schema in the database, not config.** `_emdash_byline_fields` is the
  source of truth, with `options.byline_fields_version` driving cache
  invalidation. Admins register fields through the admin UI; no code
  edits required.
- **Per-field `translatable` flag.** Translatable values are stored per
  locale (one row per locale in the byline's `translation_group`).
  Non-translatable values are stored once per `translation_group` and
  surface on every locale variant. The flag is locked once values
  reference the field — flipping would orphan rows in the wrong table.
- **`url` scheme allowlist.** URL fields require `http:` or `https:`;
  `javascript:`/`data:`/`mailto:`/`ftp:`/`file:` reject. Mirrors
  `httpUrl` in `api/schemas/common.ts` — custom URLs typically render
  as `<a href>` so this closes the XSS footgun.
- **Create + update share one transaction on Node/PG.** Byline row
  write and per-field writes roll back together on partial failure.
- **D1 retry-safe create via idempotent recovery.** D1 has no
  transactions, so a crash between the row insert and the field writes
  leaves a partial byline that the API would otherwise refuse to
  recover from. A retry POST is treated as completing the abandoned
  create iff the full fixed-column payload, the translation-group
  identity, and the existing custom-field subset all match the
  incoming request. Anything else collapses to a standard
  duplicate-slug `CONFLICT` — recovery only fires when the retry is
  provably the same request as the original.
- **Cache coherency via parity bit.** `options.byline_fields_version`
  carries meaning in its parity — odd = mutation in flight (or
  crashed), even = stable. The cache bypasses the global holder while
  odd. `markVersionDirty` is parity-aware idempotent; `markVersionClean`
  always advances to a **new** even value so two concurrent mutators
  can't collapse on the same key and pin a stale snapshot.
  Idempotent-retry exits also run `markVersionClean` — same code path
  doubles as crash recovery and false-clean recovery.

`BylineSummary` gains an optional
`customFields: Record<string, CustomFieldValue>`. Existing
object-literal consumers stay source-compatible — the property is
optional and runtime always returns `{}` when no fields are registered.
Builds on the bylines-i18n foundation from #1146.

### What's deferred

- **`required` flag enforcement.** The admin UI exposes a Required
  toggle and the value persists on the field definition, but the
  server accepts missing values and `null` for any field. Needs a
  design call on the enforcement model (reject on first set only?
  reject on every save? coerce vs reject?). TODOs are in
  `BylineRepository.coerceFieldValue` and `BylineFieldEditor`.
- **Full multi-isolate concurrency on schema mutations.** The
  parity-aware bookend has a residual race between two concurrent
  `markVersionClean` calls — bounded by the inter-clean duration
  (~ms). Schema mutations are admin-only and rare; acceptable for
  this PR. A CAS-on-bump or dialect-specific row lock would close
  it but raises dialect-divergence scope.
- **Idempotency keys for D1 retries.** The recovery branch carries
  multiple conjunctive guards approximating "same request as before"
  via state inspection. If similar D1 retry-safety problems recur on
  other create endpoints, an explicit `Idempotency-Key` header would
  replace the apparatus with one lookup. Worth a Discussion before
  the next D1 retry surface comes up.
- **`translationOf`-flavoured create recovery.** The sibling-locale
  guard returns `CONFLICT` before the recovery branch can fire on
  translation creates, so partial `translationOf` retries fall through
  to manual recovery via the edit flow. Uncommon enough that
  reordering the guard wasn't justified for this PR.

Closes #1174

Closes #

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/1174

- [x] This PR includes AI-generated code — model/tool: Claude Opus 4.7, GPT-5.5

## Screenshots / test output

https://github.com/user-attachments/assets/c0be1b30-9242-4821-8a9b-64b1a344ef63


packages/core/src/bylines/field-defs-cache.ts
```diff
@@ -0,0 +1,132 @@
...
+	const dirty = version % 2 !== 0;
+	return requestCached(`${REQUEST_CACHE_KEY_DEFS_PREFIX}${version}`, async () => {
+		if (isolated || dirty) {
+			return new BylineSchemaRegistry(db).listFields();
+		}
+		if (holder.cached !== null && holder.cachedVersion === version) {
+			return holder.cached;
+		}
+		const defs = await new BylineSchemaRegistry(db).listFields();
+		holder.cached = defs;
```
ascorbic: The jsdoc says that it caches the promise not the resolved value, but we're awaiting it before assigning. It's better to cache the promise because, as the comment says, it allows us to coalesce concurrent requests to use a single db query.

MohamedH1998: Fixed

ascorbic:
Pushed a merge of `main` into this branch (`e6d45e93`, with Mohamed's go-ahead) to recover a fix that the merge would otherwise silently drop.

Background: #1298 (byline avatar storage key) landed on `main` and adds a `LEFT JOIN media` to the byline hydration queries, surfacing `avatarStorageKey`/`avatarAlt` on `BylineSummary`. This branch's custom-field work reshapes the joined rows into a minimal `BylineRow` before `withCustomFields`, and that reshape dropped the two new avatar columns — so a plain merge compiles fine and all tests pass, but `avatarStorageKey` silently comes back `null` (the join runs, its values are discarded). 

The fix carries `avatar_storage_key`/`avatar_alt` through the reshape (typed `BylineRowWithAvatar[]`) in `getContentBylines` and `getContentBylinesMany`. `findByUserIds` passes the joined rows straight through, so it was already correct. Query-count snapshots regenerated for the merged tree. Byline tests (incl. the avatar hydration cases from #1298) pass.

Note: the merged tree shows a uniform +1 query/route from the per-request `byline_fields_version` read. That's expected and accepted for now (it buys immediate cross-isolate invalidation); consolidating cold-start option reads is a separate follow-up, not a blocker.

PR #1285, opened 2026-06-02
fix: surface per-entry SEO metadata on entry.data.seo

### LINKED ISSUE #1270: Bug - SEO - Various options (noindex toggle / canonical URLs) don't affect page outputs
#### Hide from search engines

On a content collection entry, toggling the `Hide from search engines` field (added by the **SEO** checkbox) doesn't appear to add the meta tag it says it does (`<meta name="robots" content="noindex">`).

Toggling this back and forth doesn't appear to have an effect, even if you make changes to the content as well to try and trigger a republish.

Image

Image

#### Canonical URLs

A similar pattern seems to happen with the `SEO` plugin's canonical URL field. Adding an entry to this field doesn't change the rendered `canonical` link tag on the page

### Steps to reproduce

1. Create a collection with *SEO* enabled.
2. Create a new content collection entity.
3. Toggle the `Hide from search engines` field or add a canonical URL value.
4. Look at the rendered output. There will be no effect.

### Environment

- emdash: 0.15.0
- Cloudflare Workers
- MacOS / Chrome

### Logs / error output

```shell

```

ascorbic:
## What does this PR do?

Fixes SEO fields (the **Hide from search engines** noindex toggle and **canonical URL**) having no effect on rendered pages.

SEO is stored in a side table (`_emdash_seo`) and the admin API hydrates it correctly, but the front-end query path never did: the content loader builds `entry.data` only from the content table's own columns, so `getSeoMeta(post, ...)` always read an absent `data.seo` and fell back to defaults (`noIndex: false`, canonical derived from path). Toggling anything in the SEO panel produced no change in the output.

`loadEntry` now `LEFT JOIN`s `_emdash_seo` and surfaces the result as a nested `entry.data.seo` object. Because SEO is 1:1 with content (PK on `collection + content_id`), it folds into the existing entry query with **zero extra database round-trips** — verified against the query-count snapshot (unchanged).

Only the single-entry path is hydrated, not collection lists: templates call `getSeoMeta` on the single rendered entry, while list pages render bylines/tags rather than per-item canonical/robots.

Closes #1270

## Type of change

- [x] Bug fix

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] User-visible strings in the admin UI are wrapped for translation (if applicable). N/A

## Screenshots / test output

New `tests/unit/loader-seo.test.ts` runs on both SQLite and Postgres (the JOIN is dialect-sensitive):

- attaches noindex + canonical to `entry.data.seo`
- no `data.seo` when no SEO row exists
- `noIndex:false` surfaced once a row exists (e.g. canonical-only)
- raw `seo_*` columns are not leaked as flat data fields

Full core suite: 3594 passing against Postgres (3557 SQLite). `pnpm query-counts` matches the baseline exactly.

---

### Try this PR

**[Open a fresh playground →](https://fix-seo-loader-hydration-emdash-playground.emdash-cms.workers.dev)**

<sub>Tracks `fix/seo-loader-hydration`. Updated automatically when the playground redeploys.</sub>

### DIFF
--- a/.changeset/fix-seo-loader-hydration.md
+++ b/.changeset/fix-seo-loader-hydration.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fix SEO fields (noindex toggle, canonical URL) not affecting rendered pages. The content loader now surfaces per-entry SEO metadata on `entry.data.seo`, so `getSeoMeta()` reflects values set in the admin SEO panel. SEO is folded into the existing entry query via a LEFT JOIN, adding no extra database round-trips.
--- a/packages/core/src/loader.ts
+++ b/packages/core/src/loader.ts
@@ -25,12 +25,27 @@ import { isMissingColumnError, isMissingTableError } from "./utils/db-errors.js"
 const FIELD_NAME_PATTERN = /^[a-zA-Z_][a-zA-Z0-9_]*$/;
 
 /**
- * SEO columns joined in from `_emdash_seo` on the single-entry path. Kept out
- * of the generic field mapping (see SYSTEM_COLUMNS) and surfaced as a nested
- * `data.seo` object instead. SEO lives in a side table, so a LEFT JOIN folds
- * it into the entry load at zero extra query cost.
+ * SEO columns joined in from `_emdash_seo` on the single-entry path, mapped to
+ * aliased result keys. SEO lives in a side table, so a LEFT JOIN folds it into
+ * the entry load at zero extra query cost; the result is surfaced as a nested
+ * `data.seo` object (see extractSeo) rather than flat fields.
+ *
+ * The `_emdash_` prefix on the aliases guarantees they can never collide with
+ * a content field. Field slugs must match `/^[a-z][a-z0-9_]*$/`, so a user can
+ * legitimately define a `seo_title` field; selecting the joined column under
+ * its bare name would shadow that field in the result set and drop the user's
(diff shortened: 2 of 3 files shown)


ascorbic:
Thanks @emdashbot — addressed both:

**1. Column collision (fixed in 32cc51ac).** Real bug. A collection can legitimately define a `seo_title` field (valid slug, not reserved), and the unaliased `SELECT c.*, s.seo_title` plus adding `seo_title` to `SYSTEM_COLUMNS` would shadow and drop it. The joined columns are now aliased with an `_emdash_` prefix (`_emdash_seo_title`, etc.), which can never collide since field slugs must match `/^[a-z][a-z0-9_]*$/` and so can't start with `_`. `extractSeo` reads the aliases, and only the aliases are excluded from field mapping. Added a regression test that defines a `seo_title` field and asserts the user's value survives while the panel value lands on `data.seo`.

**2. Missing `_emdash_seo` table.** Not addressing this one. `018_seo.ts` is a foundational migration; any database the loader runs against has it. Adding a non-JOIN fallback would mean a probe/branch to guard against a half-migrated database that's already broken, and it would undercut the zero-extra-query property that's the whole point of folding SEO into the existing query. Treating migrations as current is the right baseline here.

Both dialects (SQLite + Postgres) pass; query-count snapshot unchanged.

PR #1298, opened 2026-06-02
perf: resolve byline avatar storage key during hydration

ascorbic:
## What does this PR do?

Byline content-credit hydration (`getContentBylines` / `getContentBylinesMany`, the path that populates `entry.data.bylines` for `getEmDashCollection` / `getEmDashEntry`) now `LEFT JOIN`s the `media` table and surfaces the avatar's `avatarStorageKey` + `avatarAlt` on `BylineSummary`. List pages can build a direct avatar URL straight from the hydrated byline instead of doing a per-author `MediaRepository.findById`, which was an N+1: a page rendering 20 posts by distinct authors paid ~20 extra queries. The join adds no round-trip.

Author-inferred bylines (`findByUserIds`, used when an entry has no explicit credits) get the same join, so behaviour is consistent whether a byline is explicitly credited or inferred from the author.

Two supporting changes:

- Seed support for byline avatars (`bylines[].avatar`): an opt-in, hermetic way to attach an avatar to a seeded byline (`storageKey` of a file already in storage, plus optional `alt`/`filename`/`mimeType`/`width`/`height`). Nothing is downloaded; a `media` row is created and linked via `avatarMediaId`. Idempotent (reuses an existing row by `storage_key`) and validated.
- Perf fixture + query-count snapshots: a `/contributors` route (optimized, joined avatar) and `/contributors-naive` route (per-author `findById`) in `fixtures/perf-site`, so the snapshot encodes the win as a regression guard.

### Measured effect (committed snapshots)

| route | sqlite | d1 cold | d1 warm |
|---|---|---|---|
| `/contributors` (joined) | 5 | 13 | 4 |
| `/contributors-naive` (per-avatar lookup) | 12 | 20 | 11 |

Consistent 7 fewer queries on both engines for an 8-post / 7-author page, and the naive path scales with author count while the joined path stays flat.

## Type of change

- [x] Performance improvement

(Also adds a small opt-in seed feature, `bylines[].avatar`, described above.)

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (targeted: seed, bylines, repositories, integration)
- [x] I have added a changeset (two: the hydration fix and the seed avatar support, both `emdash` patch)

- [x] This PR includes AI-generated code. Model/tool: Claude Opus 4.8 (OpenCode), then put through a four-cycle iterative adversarial review with GPT-5.5 (a different model family) to converge.

## Notes

- `BylineSummary.avatarStorageKey` / `avatarAlt` are optional and null on the plain finders (`findById`, `findBySlug`), which do not join media. Rely on the content-credit hydration path for them.
- Deliberately out of scope: a `UNIQUE(media.storage_key)` migration. The seed avatar reuse-by-`storage_key` is deterministic but not race-safe under concurrent seed application, which isn't a supported scenario (seeding is single-shot init). Noted in a code comment.


ascorbic:
Thanks for the reviews. Addressed in 7911a344:

- **emdashbot** (catch-block masking): `deleteMediaRow` is now best-effort, it logs and swallows so a cleanup failure can't shadow the original byline-write error.
- **Copilot** (whitespace in `storageKey`/`filename`/`mimeType`): these are used verbatim in the media lookup, so validation now rejects leading/trailing whitespace, not just empty/blank.
- **Copilot** (misleading docs): dropped the `{ulid}{ext}` format claim from the `BylineSummary`/`SeedBylineAvatar` docstrings (keys can be path-like, as the fixtures show), and changed the API-schema comment from "absent" to "null" since `rowToByline` always materializes the fields.

Out of scope (noted in code): a `UNIQUE(media.storage_key)` migration; the seed reuse-by-storage_key is deterministic but not race-safe under concurrent seeding, which isn't a supported scenario.

PR #1305, opened 2026-06-02
fix(admin): resolve plugin admin pages opened at the plugin root

### LINKED ISSUE #281: Bug: Plugin admin React page resolution is brittle for settings routes
## Summary

When building a native EmDash plugin with a React admin page registered at `/settings`, the admin UI initially failed to load the intended page. It only became reliable after exporting the same page component under multiple keys: `/`, `/settings`, and `/settings/`.

## Environment

- Self-hosted EmDash workspace clone
- Ubuntu 24.04
- Node.js 22
- Native plugin with `adminEntry` React page and admin page path `/settings`

## Symptoms

- Plugin showed up in admin successfully
- Opening the plugin settings page initially showed a plugin error screen
- The UI appeared to fall back to a plugin route/Block Kit path that returned 404
- Exporting multiple path variants fixed the issue

## Local Fix

This worked consistently:

```ts
export const pages = {
  "/": SettingsPage,
  "/settings": SettingsPage,
  "/settings/": SettingsPage,
}
```

Using only `"/settings"` did not appear robust enough.

## Why This Feels Like A Product Bug

- One logical page path should not require multiple export aliases
- The failure mode is confusing because it looks like a plugin route problem instead of page matching
- It creates avoidable friction for plugin authors building admin UIs

## Suggested Improvements

1. Normalize plugin admin page paths in the admin router
2. Treat `/settings` and `/settings/` identically
3. Document the expected `pages` export contract more explicitly for plugin admin React pages

## Impact

- Slows down development of plugin admin settings pages
- Produces confusing runtime behavior for otherwise-correct plugin registrations

eyupcanakman:
## What does this PR do?

Closes #281.

Opening a plugin from the Plugin Manager showed a Plugin Error. The Settings gear links to the plugin root (`/plugins/<id>/`, an empty route splat), but a plugin that registers its admin page at `/settings` has no page at the root, so `usePluginPage` returned null and the admin fell through to the Block Kit route, which 404s for a React plugin. The lookup also compared paths by exact key, so `/settings` and `/settings/` were treated as different.

**Fix:**

- `resolvePluginPagePath` now resolves the plugin root to the first registered page, so opening a plugin lands on its admin page instead of a 404. It also matches a trailing-slash variant, so `/settings` and `/settings/` resolve to the same registration.
- `usePluginPage` (page rendering) and the sidebar nav-visibility check share this resolver.
- Update the plugin docs and skill reference, which said page paths must match exactly.

**Testing:** verified in a running admin (`demos/plugins-demo`, which wires the `api-test` plugin registered only at `/test`). Before, the Plugin Manager gear opened `/plugins/api-test` and rendered "Plugin Error: Plugin route not found (404)"; after, it renders the plugin's page. `resolvePluginPagePath` has unit coverage for the exact, trailing-slash (both directions), both-registered, root-fallback, and unregistered cases. `@emdash-cms/admin` typecheck, `oxlint`, and `oxfmt` pass.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable). No `messages.po` changes are included.

- [x] This PR includes AI-generated code (model: Claude Opus 4.8)

### DIFF
--- a/.changeset/fix-plugin-page-root-resolution.md
+++ b/.changeset/fix-plugin-page-root-resolution.md
@@ -1,0 +1,5 @@
+---
+"@emdash-cms/admin": patch
+---
+
+Fixes plugin admin pages showing a Plugin Error when opened from the Plugin Manager. The Settings gear opens a plugin at its root (`/plugins/<id>/`), but a plugin that registers its page at `/settings` rather than `/` had no page there, so the admin fell through to a 404. Plugin page resolution now resolves the plugin root to the first registered page and treats `/settings` and `/settings/` as the same path, so a single registration works from the Plugin Manager gear and the sidebar regardless of trailing slash.
--- a/docs/src/content/docs/plugins/creating-native-plugins/react-admin.mdx
+++ b/docs/src/content/docs/plugins/creating-native-plugins/react-admin.mdx
@@ -224,7 +224,7 @@ export const widgets = {
 ```
 
 <Aside type="caution">
-	Page paths in `pages` must match the `path` values in `admin.pages`. Widget keys must match the `id` values in `admin.widgets`.
+	Page paths in `pages` should match the `path` values in `admin.pages`. A trailing slash is treated as equivalent, so `/settings` and `/settings/` resolve to the same page. Opening a plugin at its root resolves to the first registered page. Widget keys must match the `id` values in `admin.widgets`.
 </Aside>
 
 ## Using admin components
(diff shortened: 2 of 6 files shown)


ascorbic (approved):
Thanks!

Issue #1281, opened 2026-06-02
Admin stylesheet bleeds into public pages in `astro dev`, overriding host design tokens (layer-ordering)


edrpls:
## Summary

As of **0.16.0**, running `astro dev` injects `@emdash-cms/admin/dist/styles.css` (the Kumo/Tailwind v4 admin theme) into the `<head>` of **public** (non-admin) pages. That sheet declares a large set of generic design-token custom properties inside `@layer theme` — including names a host site commonly defines itself (`--text-sm/base/lg/xl/2xl/3xl/4xl`, `--radius-sm/md/lg`, `--leading-snug/normal`, `--font-mono`, …).

Because of CSS **layer ordering** (named layers merge by the order each layer is *first declared* in the document, and Vite injects the admin sheet *after* the host's own token stylesheet in the dev `<head>`), `@layer theme` ends up **outranking** the host site's token layer. The admin theme's values then win on public pages — e.g. a host `--text-lg: 1.125rem` silently becomes the admin sheet's `--text-lg: 16px`, so the whole type scale (and radius/leading/mono) renders wrong on every dev page.

This is **dev-only**: in the production build the admin CSS is correctly isolated to its own route chunk (`admin@_@astro.*.css`) and never reaches public pages. So the symptom appears only under `astro dev` — which also means it can slip into visual-regression / computed-style tests that run against the dev server, and it degrades the local dev experience for any site whose token names overlap Tailwind's defaults.

## Environment
- emdash 0.16.0 / 0.16.1
- Astro 6.x SSR, `@astrojs/node`
- Reproduced in Chromium

## Steps to reproduce
1. Host Astro app using emdash, with its own CSS custom properties (e.g. `:root` / `@layer` defining `--text-lg`, `--radius-md`, `--font-mono`, etc.).
2. `astro dev`, open any **public** page (not `/_emdash/admin`).
3. Inspect `<head>`: `@emdash-cms/admin/dist/styles.css` is present.
4. `getComputedStyle(document.documentElement).getPropertyValue('--text-lg')` → the admin sheet's value (e.g. `16px`), not the host's.
5. Confirm the order-dependence: with the admin sheet injected *before* the host token sheet, the host wins; injected *after* (as Vite does), the admin theme wins.

## Expected
The admin UI's stylesheet should not affect public, non-admin routes — and should never be able to override a host site's design tokens.

## Likely fix (host can't control this from their side)
[...]

Issue #1250, opened 2026-05-31
Bug: Byline Avatar Upload Missing from Admin UI


marcusbellamyshaw-cell:
### Description

**Date:** 2026-05-31
**Affects:** Emdash 0.15.0 (confirmed)
**Priority:** Low–Medium
**Type:** Bug (data model and API support the field; UI omission makes it inaccessible)
**Submitted by:** Every Bit Texas

---

## Summary

The `avatarMediaId` field exists on the byline data model and is accepted by the `PUT /api/admin/bylines/:id` endpoint, but there is no avatar upload control on the byline create or edit page in the Emdash admin UI. The field can only be set programmatically via the API — there is no way for an editor to attach a photo to a byline through the UI.

## Expected Behaviour

The byline edit page should include a media picker or image upload control for the avatar, consistent with how other image fields work elsewhere in the admin (e.g. post featured image, media library picker).

## Actual Behaviour

No avatar field is rendered. `avatarMediaId` is always `null` for any byline created or edited through the UI, regardless of what is entered in other fields.

---

## Evidence from Type Definitions

The `bylineUpdateBody` schema (confirmed in `emdash@0.15.0`) includes the field:

```ts
declare const bylineUpdateBody: z.ZodObject<{
  slug:          z.ZodOptional<z.ZodString>;
  displayName:   z.ZodOptional<z.ZodString>;
  bio:           z.ZodOptional<z.ZodNullable<z.ZodString>>;
  avatarMediaId: z.ZodOptional<z.ZodNullable<z.ZodString>>;  // ← supported by API
  websiteUrl:    z.ZodOptional<z.ZodNullable<z.ZodString>>;
  userId:        z.ZodOptional<z.ZodNullable<z.ZodString>>;
  isGuest:       z.ZodOptional<z.ZodBoolean>;
}>;
```

The field is also present on the byline object returned in content responses:

```ts
byline: {
  id:            string;
  displayName:   string;
  bio:           string | null;
  avatarMediaId: string | null;  // ← always null; no UI to set it
  websiteUrl:    string | null;
  ...
}
```

---

## Impact

Site templates that render byline cards cannot display author avatar photos because the field is never populated. We built avatar rendering into our post template, then had to remove it as dead code once we confirmed the admin has no way to set the value.

---

## Suggested Fix

[...]

Issue #1242, opened 2026-05-31
Admin content edit route should preserve locale for slug-based i18n entries


WellDunDun:
## Summary

In an i18n site, the Admin UI edit route can open the wrong locale variant when the URL identifier is a slug-like ID and another locale shares the same slug.

We hit this in an Astro/Cloudflare EmDash site with `pages/about`:

- English row: `id: "about"`, `slug: "about"`, `locale: "en"`
- Arabic row: `id: "about-ar"`, `slug: "about"`, `locale: "ar"`, `translationOf: "about"`

Navigating directly to:

```text
/_emdash/admin/content/pages/about
```

opened the Arabic row and made it difficult/impossible to select/edit the English row.

## What appears to happen

The content list page is locale-aware and defaults to `i18n.defaultLocale`, but the content edit page does not validate/preserve a `locale` search param and calls:

```ts
fetchContent(collection, id)
```

which fetches:

```text
/_emdash/api/content/pages/about
```

without `?locale=en`.

Because EmDash permits duplicate slugs across locales, an unscoped slug lookup is ambiguous.

## Expected behavior

When i18n is enabled, the Admin content edit route should pass the active locale through to content reads and writes, likely defaulting to `manifest.i18n.defaultLocale` when no explicit `?locale=` is present.

Example:

```text
/_emdash/api/content/pages/about?locale=en
```

## Local workaround

We patched `@emdash-cms/admin@0.14.0` locally to:

- accept `locale` in the edit route search params
- compute `activeLocale = searchParams.locale ?? i18n.defaultLocale`
- include locale in the single-content query key
- call `fetchContent(collection, id, { locale: activeLocale })`
- pass `rawItem.locale ?? activeLocale` to update/autosave calls

We also patched the API/runtime side so PUT requests with `?locale=` use that locale when resolving slug-like IDs.

## Version

Observed on EmDash `0.14.0` / `@emdash-cms/admin@0.14.0`.

WellDunDun:
Additional confirmed repro path from downstream integration:

The content edit route can still lose locale even after direct edit links preserve/default a locale, because the **Translations** panel opens sibling entries by id without passing the sibling locale.

Observed on `@emdash-cms/admin@0.14.0`:

1. Open an i18n content entry, e.g. `/_emdash/admin/content/pages/about?locale=en`.
2. In the Translations panel, click Edit/Open on the Arabic sibling.
[...]

PR #1306, opened 2026-06-02
feat(triage): freeform maintainer directives on reproduced issues

ascorbic:
## What

Adds a **freeform maintainer-directive path** for the `triage/reproduced` and `triage/by-design` states — the gap where the investigation bot has reproduced a bug but deferred the fix (e.g. diagnose returned `needs-design-decision` with options, as in #1281) and there was previously no way for a maintainer to respond to the bot's comment.

## How it works

A maintainer comments `@emdashbot go with option A+B` (or any freeform instruction). Then:

```
triage/reproduced
  └─ @emdashbot implement A+B   (maintainer-reply.yml)
       ├─ authorize (permission API: admin/write/triage)
       ├─ classify intent (Flue/kimi → implement | close | takeover | unclear)
       └─ implement → repository_dispatch(maintainer-directive, {issueNumber, directive})
            └─ investigate.yml (directed run: overrides the fix gate)
                 └─ bot/fix-<n>  →  triage/awaiting-reporter   (normal ask comment)
                      ├─ reporter: "yes, fixed"        → PR → verified
                      └─ maintainer: @emdashbot confirm → PR → verified
```

The produced fix routes into the **existing** `awaiting-reporter` loop, reusing reporter-reply's confirm/reject verbatim — no new PR-opening code.

This complements `reporter-reply.yml` (which owns `awaiting-reporter`); the two gate on disjoint label states and never both fire on one comment.

## Files

- **`maintainer-reply.yml`** (new) — `issue_comment` trigger; gates on `reproduced`/`by-design` + non-bot + an `@emdashbot` wake word at line start; permission-API auth (not the spoof-prone `author_association`); freeform classify → `implement` (dispatch) / `close` (relabel `by-design`, **no auto-close**) / `takeover` (disengage) / `unclear` (clarify).
- **`classify-maintainer-reply.ts`** (new) — cheap kimi classifier → `{proceed|steer|close|takeover|unclear}` + extracted directive.
- **`classifier.ts`** — new `maintainerIntentSchema`.
- **`investigate.ts`** — `maintainerDirective` payload; a directive overrides the bot's *judgment* gates (is-it-a-bug at stage 0, is-it-intended at verify) **and** the fix gate, but not the *capability* gates (can't reproduce → can't verify a fix → bails honestly).
- **`investigate.yml`** — `directive` input + `maintainer-directive` dispatch type, threaded into the agent payload via `--rawfile`; directed-aware wording on the skipped / not-reproduced / reproduced-no-fix comments.

## Security

Mirrors the existing reporter-reply/investigate discipline: attacker-shaped text (directive, classifier reasoning, bot-context, issue title) goes to `/tmp` via `--rawfile`, never inlined into shell or an unescaped `$GITHUB_OUTPUT` heredoc; intent is enum-whitelisted before it reaches a step output; auth is an authoritative permission-API lookup; the directive travels JSON-escaped through `repository_dispatch` (the App token has `contents:write` but not `actions:write`); branch names derive from the validated integer issue number. Auth matches reporter-reply (`admin/write/triage`).

## Review

Ran an adversarial review pass over the diff before opening. Cleared injection, double-fire, the `--slurp | jq` CI footgun (#1291), and the dispatch wiring. Fixed one finding — a directive was being silently dropped at the stage-0/verify judgment gates with a misleading "declined to reproduce" comment on the by-design path; it now proceeds to a fix. Capability-gate bails got directed-aware wording.

## Testing notes

The triage bot runs only in CI (GitHub Actions + Flue agent against the live repo), so this can't be exercised locally. Validated: Flue typecheck + build (3 workflows discovered), all three workflow YAMLs parse, oxfmt clean. No changeset — neither `.flue` nor `.github` is a published package.

---

### Try this PR

**[Open a fresh playground →](https://feat-maintainer-reply-emdash-playground.emdash-cms.workers.dev)**

<sub>Tracks `feat/maintainer-reply`. Updated automatically when the playground redeploys.</sub>

---

## Also fixes: AI classifier silently stuck (live bug on #1242, #1250)

While testing, found the existing `reporter-reply` AI classifier was **broken in production** — reporters confirming "yes, fixed" got re-asked forever and no PR ever opened from the classifier path (only the deterministic `@emdashbot confirm` path worked, which is how recent issues actually reached `triage/verified`).

**Root cause:** the classifier scraped its result out of `flue run`'s stdout, but flue interleaves build-log lines and pretty-prints the JSON result — defeating both the line-by-line and slurp parses, so every classification silently defaulted to `unclear`. `investigate.ts` had already hit this and moved to a file handoff (`INVESTIGATE_RESULT_PATH`) with a comment calling stdout-scraping fragile; the classifiers never got that treatment — and the new `classify-maintainer-reply` had copied the same fragile scrape, so this feature would have been **broken on arrival**.

**Fix:** a shared `persistClassifierResult()` helper writes the result to `CLASSIFY_RESULT_PATH`; both `reporter-reply` and `maintainer-reply` now read that file instead of scraping stdout. Mirrors the proven investigate handoff exactly.

After merge, the stuck issues (#1242, #1250) can be unstuck by the reporter re-commenting or a maintainer `@emdashbot confirm`.

*Note: the classifier path can't be exercised locally (needs the CF AI Gateway + Actions); verification is by mirroring investigate's working pattern, and the next reporter reply on any awaiting-reporter issue will confirm it end-to-end.*


ascorbic:
Thanks — all findings addressed in b4bcbaff.

**Logic bugs (both reviewers):**
- **Empty `botContext` fallback** — the orchestrator passes `""` (not `undefined`) when an issue has no bot comments, so `??` didn't fall through. Switched to `payload.botContext?.trim() || "(unavailable…)"` so the classifier keeps its prompt cue.
- **Directed run misrouted to by-design** — a directed run whose fix is abandoned could carry `verdict: intended-behavior` and hit the `intended-behavior` outcome branch, flipping to `triage/by-design` against the maintainer's explicit override. The parse step now gates that branch on `DIRECTED != true`, so a directed run falls through to the `reproduced` branch and its directed-aware "couldn't produce a verified fix" wording.

**Copilot's correctness/doc finds:**
- **Whitespace-only directive** — a blank directive (possible via manual `workflow_dispatch`) now normalizes to empty, so `directed` and the payload reflect only a real instruction.
- **Multi-line reasoning in `close`** — block-quotes every line via `sed 's/^/> /'` instead of prefixing only the first.
- **`classify-maintainer-reply` header** — corrected to "permission-API admin/write/triage on `reproduced`/`by-design`", not "OWNER/MEMBER/COLLABORATOR on any `triage/*`".

PR #1331, opened 2026-06-04
fix(cli): make export-seed i18n-aware via data-driven locale detection

### LINKED ISSUE #1330: export-seed CLI is not i18n-aware: collapses locale variants into duplicate seed ids
## Summary

`emdash export-seed` (the CLI command) is not i18n-aware: it never loads the i18n config, so `isI18nEnabled()` always returns `false` when run from the CLI. As a result, for any i18n-enabled project the exported seed **strips the `:locale` suffix from seed-local ids**, collapsing every locale variant of a taxonomy / menu / content entry into a single bare id. When two or more locale variants exist, this produces **duplicate ids in the seed**, which `validateSeed` then rejects — making the seed un-appliable by every consumer (`emdash seed`, the runtime auto-seed, and the setup route).

## Environment

- emdash `0.17.0` (`packages/core`)
- Project with i18n enabled (`astro.config.mjs` → `i18n: { defaultLocale: "ar", locales: ["ar", "en"], ... }`)

## Root cause

`isI18nEnabled()` reads a module-level `_config` that is only populated by `setI18nConfig()`, which is **only called from `packages/core/src/astro/middleware.ts`** (server runtime). The `export-seed` CLI never initializes it, so in the CLI `isI18nEnabled()` is always `false`.

`exportTaxonomies` / `exportMenus` / content export then gate the locale suffix on that flag:

```ts
// packages/core/src/cli/commands/export-seed.ts:291
const defSeedId =
    i18nEnabled && def.locale ? `tax:${def.name}:${def.locale}` : `tax:${def.name}`;
```

With `i18nEnabled === false`, the `ar` and `en` rows of the same taxonomy both serialize to `tax:category`, producing byte-identical duplicates. The same pattern affects:

- taxonomy defs — line 291
- taxonomy terms — lines 306, 319
- menus — line 384
- menu items — lines 666, 682
- content entries — lines 640, 719

## Reproduction

1. An i18n project with a taxonomy that has both `ar` and `en` locale rows (distinct `locale`, shared `translation_group`), e.g. the default `category` / `tag`.
2. Run `emdash export-seed --with-content --database <db>` (i.e. via the CLI, not the runtime).
3. Inspect the output: the taxonomy appears twice with identical `id: "tax:category"`.
4. Apply it with `emdash seed`:

```
ERROR  Seed validation failed:
ERROR    - taxonomies[1].name: duplicate taxonomy name "category"
ERROR    - taxonomies[10].name: duplicate taxonomy name "tag"
```

## Impact

- `export-seed` silently produces a **lossy and invalid** seed for any i18n project — locale variants are merged, and the seed fails validation when ≥2 variants collide.
- Because the runtime auto-seed (`emdash-runtime.ts`) guards `applySeed` behind `if (validation.valid)` with the failure swallowed by a surrounding `try/catch`, an invalid exported seed **silently no-ops** — the project never seeds, with no error surfaced.

## Suggested fix

Make `export-seed` derive locale-awareness from the data rather than the runtime-only flag. The DB rows already carry a `locale` column, so gating the suffix on the per-row value (`def.locale ? \`tax:${def.name}:${def.locale}\` : \`tax:${def.name}\``) yields correct ids for both single-locale (null `locale` → bare id) and multi-locale (suffixed) projects, without needing the global flag. This would need to stay consistent across the export *and* import id contract for taxonomies, menus, and content.

(Alternatively: load/initialize the i18n config in the CLI before exporting, e.g. from the project's Astro config, so `isI18nEnabled()` reflects reality.)

Happy to send a PR if you'd like — I wanted to confirm the preferred direction (data-driven per-row vs. loading config in the CLI) before touching the export/import id contract.

MA2153:
## What does this PR do?

`emdash export-seed` runs in the CLI, which never calls `setI18nConfig()`, so `isI18nEnabled()` is always `false` there. For any i18n-enabled project the exporter stripped the `:locale` suffix from taxonomy / menu / content seed ids, collapsing every locale variant into a single bare id. When ≥2 locale variants exist this produced byte-identical duplicate ids, which `validateSeed` then rejected — making the exported seed un-appliable by every consumer (`emdash seed`, runtime auto-seed, setup route).

The fix derives locale-awareness from the **data** rather than the runtime-only flag. A new `detectI18nEnabled()` helper still honors `isI18nEnabled()` when it's set (runtime), then falls back to counting **distinct locales** across the i18n-aware tables (`_emdash_taxonomy_defs`, `_emdash_menus`, each `ec_*`). A project is multi-locale only when more than one distinct locale exists.

Note: the issue suggested gating on `def.locale ? … : …`, but that doesn't work — `locale` is `NOT NULL DEFAULT <defaultLocale>` and so is never null, which would suffix every single-locale project and break the existing bare-id contract. The distinct-locale count is what actually separates the two cases. Single-locale projects keep **bare ids** (backwards-compatible); multi-locale projects keep their per-locale suffixes and `translationOf` links.

The flag is computed once and threaded through `exportTaxonomies` / `exportMenus` / `exportContent`, replacing the three runtime-only `isI18nEnabled()` calls.

Closes #1330

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes (`packages/core` clean; pre-existing unrelated failure in `packages/plugin-cli` `oauth.ts` re: `registry-lexicons` exports)
- [x] `pnpm lint` passes (only 8 pre-existing diagnostics in unrelated demo/template `astro.config.mjs` files)
- [x] `pnpm test` passes (targeted: full `tests/unit/seed`, `tests/unit/cli`, `tests/integration/seed` — 275 tests green)
- [x] I have added/updated tests for my changes — new `tests/unit/seed/export-seed-cli-i18n.test.ts` reproduces the bug (TDD: written failing first) with i18n config unset, covering taxonomies, menus, content, and the single-locale bare-id case
- [ ] User-visible strings in the admin UI are wrapped for translation — n/a, no admin UI strings touched
- [x] I have added a [changeset](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#changesets) (`emdash: patch`)

## Screenshots / test output

```
 RUN  v4.1.5 /home/malloo/projects/emdash/packages/core
 ✓ tests/unit/seed/export-seed-cli-i18n.test.ts (4 tests)
   ✓ suffixes taxonomy def + term ids per locale instead of colliding
   ✓ suffixes menu ids per locale instead of colliding
   ✓ suffixes content entry ids per locale instead of colliding
   ✓ keeps bare ids for genuinely single-locale projects

 Test Files  16 passed (16)
      Tests  275 passed (275)
```

### DIFF
--- a/.changeset/fix-export-seed-cli-i18n.md
+++ b/.changeset/fix-export-seed-cli-i18n.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fix `emdash export-seed` collapsing locale variants into duplicate seed ids on i18n projects. The CLI never initializes the runtime i18n config, so `isI18nEnabled()` was always `false` and the exporter stripped the `:locale` suffix from taxonomy, menu, and content seed ids — merging every locale variant into one bare id and producing duplicates that `validateSeed` rejected. The exporter now derives locale-awareness from the data (a project is multi-locale when its i18n-aware tables hold more than one distinct locale), so multi-locale exports keep their per-locale suffixes and `translationOf` links while genuinely single-locale projects still export bare ids.
--- a/packages/core/src/cli/commands/export-seed.ts
+++ b/packages/core/src/cli/commands/export-seed.ts
@@ -9,6 +9,7 @@ import { resolve } from "node:path";
 import { defineCommand } from "citty";
 import consola from "consola";
 import type { Kysely } from "kysely";
+import { sql } from "kysely";
 
 import { createDatabase } from "../../database/connection.js";
 import { runMigrations } from "../../database/migrations/runner.js";
@@ -17,6 +18,7 @@ import { MediaRepository } from "../../database/repositories/media.js";
 import { OptionsRepository } from "../../database/repositories/options.js";
 import { TaxonomyRepository } from "../../database/repositories/taxonomy.js";
 import type { Database } from "../../database/types.js";
+import { validateIdentifier } from "../../database/validate.js";
 import { isI18nEnabled } from "../../i18n/config.js";
 import { SchemaRegistry } from "../../schema/registry.js";
 import type { FieldType } from "../../schema/types.js";
@@ -118,11 +120,17 @@ export async function exportSeed(db: Kysely<Database>, withContent?: string): Pr
 	// 2. Export collections and fields
 	seed.collections = await exportCollections(db);
 
+	// Decide locale-awareness from the data. The runtime sets the i18n config via
+	// middleware, but the CLI never does, so `isI18nEnabled()` is always false
+	// under `emdash export-seed` (#1330). Detecting multiple locales in the data
+	// keeps the export locale-aware without the runtime flag.
+	const i18nEnabled = await detectI18nEnabled(db, seed.collections);
+
 	// 3. Export taxonomy definitions and terms
-	seed.taxonomies = await exportTaxonomies(db);
+	seed.taxonomies = await exportTaxonomies(db, i18nEnabled);
 
 	// 4. Export menus
-	seed.menus = await exportMenus(db);
+	seed.menus = await exportMenus(db, i18nEnabled);
 
 	// 5. Export widget areas
 	seed.widgetAreas = await exportWidgetAreas(db);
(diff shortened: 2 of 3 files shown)


MA2153:
Addressed in 0fe9138e. Wrapped the per-collection `ec_*` locale probe in `detectI18nEnabled` with `isMissingTableError` so a collection row that outlives its table (non-atomic `deleteCollection` on D1) is skipped instead of crashing the export. Added a regression test that drops `ec_post` while keeping its collection row and asserts `export-seed` still succeeds — verified it fails with `no such table: ec_post` without the guard.

ascorbic (approved):
Thanks

PR #1339, opened 2026-06-04
test: harden silently-passing tests, prune framework-only and duplicate ones

ascorbic:
## What does this PR do?

Audit-driven cleanup of low-value and misleading tests. The audit surfaced a long list of suspect tests; this PR acts only on the ones that are genuinely broken or worthless, after verifying each against the code under test. (Many flagged tests turned out to be legitimate, mocks at a network/dependency boundary that still assert real logic, and were left untouched.)

**Fixed tests that passed even when the asserted behavior was broken:**
- `integration/auth/device-flow.test.ts` — the Token Revoke suite used `if (!result.success) return;` with no preceding `expect`, so a failed device-code/token exchange returned green. Added `expect(...success).toBe(true)` before the narrowing guard (matching the rest of the file).
- `integration/wordpress-migration/theme-unit-test.test.ts` — ~7 conversions gated all assertions on `if (post)`, silently passing if the fixture lookup missed. Now assert the post exists (fixture verified to contain the blocks) and convert; the embed test (fixture has none) feeds synthetic markup; the content-integrity loop asserts it verified at least one post.
- `e2e/content-crud.spec.ts` "publish action changes status" — clicked Publish behind `if (isVisible())` and never checked the result. Now publishes unconditionally and asserts the action flips to "Unpublish".
- `e2e/plugins.spec.ts` / `e2e/wordpress-import.spec.ts` "does not show error" — negative-only assertions that passed on a blank page. Anchored each with a positive content assertion.

**Made registration-only tests actually exercise behavior:**
- `unit/plugins/hooks.test.ts` — the "hook sorting" and "content:beforeSave" tests only checked `getHookCount`. Rewritten to run the pipeline against a real in-memory DB context and assert priority order, dependency ordering, and content transformation/chaining. (The capability-enforcement tests were kept, they verify real registration gating.)

**Strengthened accept-only / no-assertion tests:**
- `unit/fields/all-fields.test.ts` — image/file/portableText "validate structure" tests only asserted valid input did not throw; added reject cases.
- `cloudflare/db/do-dialect.test.ts` — the no-op transaction test had no assertion; now asserts the read-only path issues no SQL to the stub.

**Removed framework-only / tautological / duplicate tests:**
- Deleted `admin/tests/editor/input-rules.test.ts` — it built a vanilla TipTap editor from bare `StarterKit` and tested the framework, not any EmDash code.
- `auth-atproto/tests/auth.test.ts` — removed the "descriptor shape invariants" block (asserted that hardcoded constants are non-empty strings).
- `cloudflare/tests/do-config.test.ts` — removed passthrough tests redundant with the full-descriptor assertions.
- `SignupPage`, `MediaDetailPanel`, `invite-flow`, `settings-pages` — removed/merged exact or same-name duplicate tests.

No production code changed; this is test-only.

Closes #

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [x] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (or targeted tests for my change) — ran the changed suites (core plugins/fields/auth, auth-atproto, cloudflare, admin SignupPage/MediaDetailPanel); e2e specs use standard Playwright APIs and run in CI
- [x] I have added/updated tests for my changes (if applicable) — this PR is the test change
- [ ] User-visible strings wrapped for translation — n/a, no admin UI strings touched
- [ ] I have added a changeset — n/a, test-only, no published package source changed
- [ ] New features link to an approved Discussion — n/a, not a feature

## Screenshots / test output

Targeted runs after changes: core plugins dir 592 passed, device-flow + theme-unit-test 51 passed, fields 41 passed, auth-atproto 13 passed, cloudflare do-config/do-dialect 11 passed, admin SignupPage + MediaDetailPanel 25 passed. `pnpm lint:quick` clean, `pnpm typecheck` clean across all packages.


ascorbic:
Thanks @copilot — addressed all three:

1. **sqlite handle not closed** (`hooks.test.ts`) — fixed; `afterEach` now keeps a reference to the better-sqlite3 instance and calls `sqliteDb.close()` after `db.destroy()`, matching the pattern in `exclusive-hooks.test.ts`/`email-pipeline.test.ts`.
2. **classic-editor selector** (`theme-unit-test.ts`) — fixed; the test now selects a classic-category post that also contains no `<!-- wp:` markers, so it genuinely exercises the classic-HTML fallback path instead of passing on whatever is in the category.
3. **vite-config / changeset in a "test-only" PR** — good catch, but this was a stale-merge-base artifact, not an intended change. That work is the already-merged #1336; my branch's copies were byte-identical to `main`, so they only appeared because the branch was cut before #1336 landed. I've merged latest `main` (they now drop out of the diff) and reverted an unrelated regenerated `demos/cloudflare/emdash-env.d.ts` that got swept in. The PR is now genuinely test-only, so the description/checklist stand.

PR #1361, opened 2026-06-05
fix(core): pre-bundle runtime-reached SSR deps for cold Cloudflare dev cache

MA2153:
## What does this PR do?

On a **cold** Vite cache, `pnpm dev` in a Cloudflare-adapter EmDash site could throw repeatedly on the first SSR requests:

```
[vite] An error happened during full reload
The file does not exist at ".../node_modules/.vite/deps_ssr/chunk-XXXX.js?v=HASH" ...
```

Root cause: the `cloudflare` branch of `ssr.optimizeDeps.include` in `packages/core/src/astro/integration/vite-config.ts` pre-bundles SSR deps so workerd loads them in one cold-startup pass, but several deps are only reached **at request time** — the middleware chain (behind the excluded `virtual:emdash` boundary), lazy adapter entrypoints, first-render imports, and re-export specifiers like `astro/zod`. Vite's startup scan misses them, discovers each at request time, and re-optimizes; overlapping re-optimizations race on `deps_ssr/` chunk hashes and crash in-flight requests.

This PR:
- **Adds a cold-cache guard test** (`packages/core/tests/integration/smoke/dep-optimizer-cold-cache.test.ts`) that boots `templates/starter-cloudflare` with a wiped `.vite` cache, drives the dev-bypass / frontend / admin routes, and asserts Vite logs **zero post-startup dependency re-optimizations**. This fails automatically whenever a request-time SSR dep is missing from the include list — including future ones — instead of someone chasing the crash by hand. (The crash itself is a non-deterministic race; the test asserts on the deterministic upstream re-optimize signal.)
- **Seeds the missing specifiers** into the cloudflare include branch (middleware chain, `@emdash-cms/cloudflare/db/d1` + `storage/r2`, `emdash/ui` + `runtime` + `media/local-runtime`, `astro/zod`, plus bare forms of a few admin/auth deps the test caught re-optimizing).
- **Serializes smoke test files** (`fileParallelism: false`) so the new test never races the existing site-matrix smoke suite on a shared site's `.vite` cache.

These additions are dev-server-only (`ssr.optimizeDeps.include`) and Cloudflare-only — no effect on production `astro build` output or the Node adapter branch.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [ ] `pnpm lint` passes — clean for all files changed in this PR; the repo has 8 pre-existing diagnostics in `demos/*` / `templates/*` `astro.config.mjs` files unrelated to this change.
- [x] `pnpm test` passes (or targeted tests for my change) — the cold-cache guard test was verified RED→GREEN locally; CI's `test-smoke` job runs it as the authoritative check.
- [ ] User-visible strings wrapped for translation — n/a, no admin UI strings.
- [x] I have added a changeset (`emdash` patch)
- [ ] New features link to an approved Discussion — n/a, bug fix.

## Screenshots / test output

RED (before the fix), the guard reported re-optimize lines such as:
```
✨ new dependencies optimized: @lingui/react, @oslojs/crypto/hmac, @oslojs/crypto/subtle, @oslojs/crypto/rsa
✨ optimized dependencies changed. reloading
✨ new dependencies optimized: @cloudflare/kumo/primitives
```
GREEN after seeding the specifiers: 0 re-optimizations, 0 crash lines.

### DIFF
--- a/.changeset/cold-cache-ssr-deps.md
+++ b/.changeset/cold-cache-ssr-deps.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Pre-bundle runtime-reached SSR deps (middleware chain, lazy Cloudflare adapter entrypoints, `emdash/ui`, `emdash/runtime`, `astro/zod`) in the Cloudflare branch of `ssr.optimizeDeps.include`. Without these the cold Vite dev cache discovered them at request time and could cascade a `deps_ssr` re-optimize race that crashed the first SSR requests.
--- a/packages/core/src/astro/integration/vite-config.ts
+++ b/packages/core/src/astro/integration/vite-config.ts
@@ -432,6 +432,41 @@ export function createViteConfig(
 							"astro/assets/fonts/runtime.js",
 							"astro/assets/services/noop",
 							"@astrojs/cloudflare/image-service",
+							// Runtime-reached SSR deps the startup scan can't see: the
+							// middleware chain sits behind the excluded `virtual:emdash`
+							// boundary, adapter entrypoints load lazily on first query, and
+							// ui/runtime/portabletext are imported during first render.
+							// Without these, the cold dev cache discovers them at request
+							// time and cascades a deps_ssr re-optimize race. Guarded by
+							// tests/integration/smoke/dep-optimizer-cold-cache.test.ts.
+							"emdash/ui",
+							"emdash/runtime",
+							"emdash/media/local-runtime",
+							"emdash/middleware",
+							"emdash/middleware/redirect",
+							"emdash/middleware/setup",
+							"emdash/middleware/auth",
+							"emdash/middleware/request-context",
+							// Astro's public re-export -- a different specifier than the
+							// `astro > zod/v4` entry above, so it must be listed separately.
+							"astro/zod",
+							// Cloudflare adapter entrypoints, loaded lazily on first
+							// content query / media op.
+							"@emdash-cms/cloudflare/db/d1",
+							"@emdash-cms/cloudflare/storage/r2",
+							// Bare specifiers for deps already listed above in their
+							// `emdash > ...` / `emdash > @emdash-cms/admin > ...` chained
+							// form. The chained form pre-bundles the copy resolved through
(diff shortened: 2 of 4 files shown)


ascorbic:
This should have been fixed already by #1359

MA2153:
Thanks for the reviews. Addressed the test reliability feedback and reconciled with the #1359 revert.

**Test fixes** (`dep-optimizer-cold-cache.test.ts`):
- `fetchWithRetry` now throws `lastError` after exhausting retries instead of returning silently, matching `site-matrix-smoke.test.ts`. A persistently-5xx/unreachable server now fails the test rather than producing a false green with the routes never driven.
- The SIGKILL fallback now checks `server.exitCode === null` instead of `server.killed` (which flips true the instant `kill()` is called, not on exit), so a server that ignores SIGTERM is actually force-killed.

**Re: @ascorbic — "already fixed by #1359":** #1359 was reverted on `main` in #1363, so that fix is no longer present. I've merged post-revert `main` into this branch, dropping the `exclude`/`noExternal` code that had been pulled in via an earlier merge. The diff against `main` is now just the `ssr.optimizeDeps.include` additions, the regression smoke test, `fileParallelism: false`, and the changeset.

But the include-list change isn't really the point of this PR — the **regression test is**. Whichever mechanism we land for the immediate cold-cache race, the durable value here is the guard: it boots a cold-cache Cloudflare site, drives the routes, and asserts **zero post-startup re-optimizations**. That fails automatically the next time *any* request-time SSR dep goes unoptimized — including future deps that don't exist yet — instead of someone rediscovering the `deps_ssr` crash by hand. #1359 fixed the instances that existed at the time but shipped no guard against the next one. This test is approach-agnostic; it would equally protect a re-landed exclude/noExternal fix.

Happy to keep just the test on top of whatever fix you prefer.

ascorbic:
I think this is now finally fixed by #1710. Are you still seeing it?

MA2153:
Will open a new PR if I re-encounter the issue

PR #1378, opened 2026-06-08
feat(core): optional distributed object cache for query results

scottbuscemi:
## What does this PR do?

Adds an **optional, opt-in distributed object cache** for query results. Content reads (`getEmDashCollection`, `getEmDashEntry`, `resolveEmDashPath`) and chrome reads (site settings, menus, taxonomies, per-entry terms, collection info, public comments) can be served from a fast key/value store instead of hitting the database on every request. It sits **beneath the per-request cache and above the database**, dramatically reducing read pressure on D1/SQLite — especially on Cloudflare, where KV absorbs far more requests than D1.

The cache is **off by default and fully opt-in**. Configure a backend in `astro.config.mjs`:

```ts
import { kvCache } from "@emdash-cms/cloudflare"; // Workers KV (distributed)
import { memoryCache } from "emdash/astro";       // in-isolate (Node / local dev)

emdash({
	database: d1({ binding: "DB" }),
	objectCache: kvCache({ binding: "CACHE" }),
});
```

Invalidation is **epoch-based and automatic**: content, byline, taxonomy, menu, and settings writes bump a per-namespace version, instantly orphaning stale entries (no key enumeration). Value and epoch are fetched in one parallel round-trip via a stable-key envelope, so there are no orphaned keys to accumulate. Preview and visual-edit requests bypass the cache, so editors previewing see live content; other reads are served from the cache, which only ever stores published content. After an edit, anonymous visitors may see stale content until isolates pick up the bumped epoch — immediate on the memory backend, and on KV bounded by KV's edge-cache propagation (eventually consistent, up to ~60s) plus the `revalidate` window (default 1s, configurable). Backend reads are bounded by a timeout so a slow/unavailable cache can never hang a render.

Existing sites are unaffected until they opt in.

Closes #

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (or targeted tests for my change) — object-cache unit/content/comments/schema/entry-terms + cloudflare kv-timeout suites
- [ ] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/... _Draft — Discussion to be linked before review._

## Screenshots / test output

```
object-cache.test.ts / object-cache-content.test.ts / object-cache-comments-schema.test.ts / entry-terms-object-cache.test.ts — 25 passed
cloudflare kv-timeout.test.ts — 2 passed
pnpm lint — 0 diagnostics; pnpm typecheck — clean
```

---

### Try this PR

**[Open a fresh playground →](https://feat-object-cache-emdash-playground.emdash-cms.workers.dev)**

<sub>Tracks `feat/object-cache`. Updated automatically when the playground redeploys.</sub>


packages/core/src/object-cache/index.ts
```diff
@@ -0,0 +1,493 @@
...
+/**
+ * Invalidate every cached value in `namespace` by bumping its epoch.
+ *
+ * Sync and non-blocking: the local epoch is stamped immediately (so the
+ * writing isolate is instantly consistent) and the backend write is deferred
+ * via `after`. Other isolates pick up the new epoch within their `revalidate`
+ * window. No-ops when the cache is disabled.
+ */
+export function invalidateObjectCache(namespace: string): void {
+	const stamp = Date.now();
```
ascorbic: This could cause a race condition if there are two writes in the same millisecond, meaning the new one doesnt get a new key, losing the update. You can make it monotonic to solve that:
```suggestion
	const prev = epochCache.get(namespace)?.value ?? 0;
	const stamp = Math.max(prev + 1, Date.now());
```

ascorbic (commented):
This looks great. I've not tested it yet, but it does seem like something we should be doing. Could you update it and resolve conflicts?

PR #1379, opened 2026-06-08
perf(core): lazily load sandboxed plugins

scottbuscemi:
## What does this PR do?

Lazily loads sandboxed plugins so they no longer block the first request of a cold isolate.

Previously, every build-time sandboxed plugin was provisioned on the Worker Loader during runtime init, so the first request of each cold isolate — **including anonymous reads** — waited on plugins it never exercised. A write-only plugin (declaring only `content:afterSave`/`afterDelete`, for example) could add over a second to content reads that never trigger it.

Plugins are now **registered** (cheap, synchronous) at init and **loaded on first use** of an extension point they actually declare:

- **Build-time:** the integration reads each plugin's `dist/manifest.json` and emits its declared hooks/routes onto the entry (`SandboxedPluginEntry` gains `hooks?`/`routes?`). Plugins without a manifest are treated as "unknown" and loaded eagerly to stay correct.
- **`create()`** registers build-time entries instead of loading them, and seeds route-auth metadata from declared routes so a route can be authorized without provisioning the isolate.
- **`ensureSandboxedPluginLoaded()`** loads a plugin on first use, memoizes per isolate, dedupes concurrent first-loads, and re-checks enabled state after load.
- **Hook/route dispatch** iterates only candidates whose declared metadata includes the event and loads only those, preserving config order. A public render loads no plugin that doesn't declare `page:metadata`.

Marketplace/registry plugins remain eagerly loaded for now (deferring them needs an R2 manifest-only fetch — follow-up). No public API or configuration change.

Closes #

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [x] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (or targeted tests for my change) — sandbox-lazy-load + virtual-modules-sandbox suites (13 passed)
- [ ] New features link to an approved Discussion. _n/a — performance improvement, no API change._

## Screenshots / test output

```
sandbox-lazy-load.test.ts / virtual-modules-sandbox.test.ts — 13 passed
pnpm lint — 0 diagnostics; pnpm typecheck — clean
```

### DIFF
--- a/.changeset/lazy-sandboxed-plugins.md
+++ b/.changeset/lazy-sandboxed-plugins.md
@@ -1,0 +1,11 @@
+---
+"emdash": patch
+---
+
+Lazily load sandboxed plugins so they no longer block the first request of a cold isolate.
+
+Previously, every build-time sandboxed plugin was provisioned on the Worker Loader during runtime init, so the first request of each cold isolate — including anonymous reads — waited on plugins it never exercised. A write-only plugin (declaring only `content:afterSave`/`afterDelete`, for example) could add over a second to content reads that never trigger it.
+
+Plugins are now registered (cheap, synchronous) at init and loaded on first use of an extension point they actually declare. At build time the integration reads each plugin's `dist/manifest.json` and emits its declared hooks/routes; hook and route dispatch load only the plugins relevant to the current event, preserving config order. Plugins without a manifest are treated as "unknown" and loaded eagerly to stay correct. Route authorization is seeded from declared routes, so a route can be authorized without provisioning the isolate. A public render loads no plugin that doesn't declare `page:metadata`.
+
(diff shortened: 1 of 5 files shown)


ascorbic:
Over a second to load a plugin seems pretty crazy. We should really investigate why that's taking so long, and see how it could be optimised too, because we want pages that use plugins to be fast too.

ascorbic:
If we can I'\d rather fix this at the source, so that hot path plugins are fast too. Do you have any numbers on how long plugin loads take and where the time is spent? Dynamic workers should be pretty much instant, so it's surprising if it's taking a long time.

PR #1405, opened 2026-06-11
fix(core): reclaim abandoned init locks to prevent isolate-poisoning deadlock

### LINKED ISSUE #1274: Session KV lookup hangs indefinitely, blocking entire page response for logged-in users
**Product:** EmDash CMS  
**Version:** 0.15.0 (first observed in 0.14.0, confirmed still present in 0.15.0)  
**Adapter:** `@astrojs/cloudflare` (Cloudflare Workers + KV)  
**Date:** 2026-05-29

---

## Summary

When a user has an active EmDash admin session cookie, page loads occasionally hang indefinitely — the browser tab spinner spins forever and no HTML is delivered. The issue is intermittent and clears on its own after some time. It never occurs in private/incognito windows (no cookie).

The issue affects **any browser** (confirmed on Chrome desktop and Chrome iOS).

---

## Steps to reproduce

1. Log into the EmDash admin panel in a browser (establishes a session cookie)
2. Navigate to any public-facing page on the site (e.g. the homepage)
3. Observe: the browser tab spinner runs indefinitely — no HTML is received, the page never renders
4. Open the same URL in a private/incognito window (no session cookie) → page loads immediately and correctly
5. Once the hang starts, all pages on the site hang for that browser — retrying does not help. The issue persists until the session cookie is cleared (e.g. via browser cache wipe). It can occur after hours of normal use with no obvious trigger.

---

## Expected behaviour

If the session KV lookup is slow or unavailable, EmDash's auth middleware should either:
- Time out after a reasonable threshold (e.g. 3–5 seconds) and continue the request with `locals.user = null` (degrade gracefully — treat the user as unauthenticated for this request), or
- Return a partial/fallback response rather than holding the response stream open indefinitely

The public page should always load. Admin toolbar features can safely be absent when the session cannot be confirmed quickly.

---

## Actual behaviour

The response stream is never completed. The Worker holds the connection open indefinitely (until Cloudflare's 30-second subrequest/CPU limit forces a termination). The user sees an empty browser tab with a spinning loading indicator.

---

## Secondary Error: sandboxed plugin fires on every request

A separate error log reveals a second issue firing on **every page load** across the entire site:

```
EmDash: Sandboxed plugin atproto page:metadata error: Error: Storage collection not declared: records
```

This occurs on every `GET` request to any post or page, continuously throughout the day. The `outcome: ok` means it is non-fatal and pages load successfully. However:

1. The atproto marketplace plugin is invoking a Worker-to-Worker call (via the `LOADER` sandboxed plugin binding) on every single page request
2. The plugin is misconfigured — it requires a storage collection named `records` which has not been declared
3. Normally this fails fast and is swallowed, but if the sandboxed `LOADER` Worker is under load, this call could stall instead of failing — which may be a contributing factor to the intermittent hangs

**Immediate fix:** either configure the atproto plugin with the required `records` storage collection, or uninstall it from the EmDash marketplace if it is not in use. Eliminating this unnecessary Worker-to-Worker call on every request reduces overhead and removes a potential hang vector.

---

## Log Evidence

Cloudflare Workers observability traces confirm two hang windows, both exclusively affecting a single device: **iPhone running Chrome (CriOS/149, Houston TX)**. All other devices — iPhone Safari, Windows Chrome, bots, uptime monitors — load normally throughout both windows.

**Window 1: ~10:08–10:10pm CDT (03:08–03:10 UTC)**

| Time (UTC) | URL | Wall time | CPU time | Outcome | UA |
|---|---|---|---|---|---|
| 03:08:49 | `GET /` | 60,550 ms | 88 ms | canceled | iPhone Chrome |
| 03:09:16 | `GET /` | 4,100 ms | 6 ms | canceled | iPhone Chrome |
| 03:09:21 | `GET /_emdash/admin` | 60,149 ms | 67 ms | canceled | iPhone Chrome |
| 03:09:54 | `GET /` | 36,350 ms | 43 ms | canceled | iPhone Chrome |
| 03:10:28 | `GET /` | 1,508 ms | 88 ms | ✓ ok | iPhone Safari |
| 03:10:40 | `GET /` | 1,535 ms | 116 ms | ✓ ok | iPhone Chrome |

The iPhone Chrome load at 03:10:40 succeeded immediately after the user cleared their browser cache (removing the session cookie).

**Window 2: ~11:05–11:13pm CDT (04:05–04:13 UTC)**

| Time (UTC) | URL | Wall time | CPU time | Outcome | UA |
|---|---|---|---|---|---|
| 04:05:01 | `GET /` | 30,000 ms | 36 ms | canceled | iPhone Chrome |
| 04:05:34 | `GET /` | 60,050 ms | 48 ms | canceled | iPhone Chrome |
| 04:08:32 | `GET /_emdash/admin/logout` | 5,800 ms | 5 ms | canceled | iPhone Chrome |
| 04:08:38 | `GET /_emdash/admin/logout` | 60,850 ms | 56 ms | canceled | iPhone Chrome |
| 04:10:11 | `GET /_emdash/admin/logout` | 50,600 ms | 48 ms | canceled | iPhone Chrome |
| 04:11:02 | `GET /_emdash/admin/logout` | 170,400 ms | 158 ms | canceled | iPhone Chrome |
| 04:11:08 | `GET /` | 60,100 ms | 54 ms | canceled | iPhone Chrome |
| 04:13:34 | `GET /` | 17,849 ms | 16 ms | canceled | iPhone Chrome |

**Critical pattern across all hung requests:** CPU time is 5–158 ms; wall time is 4,000–170,400 ms. The Worker executes a tiny amount of code then waits on I/O indefinitely. This is not CPU exhaustion — it is a blocking I/O call with no timeout.

All hung requests have **no child spans** in the trace data — no D1 queries, no KV reads, no R2 operations were recorded. This is consistent with either (a) the Worker stalling before reaching any binding, or (b) a binding call (e.g. KV session read) hanging mid-execution — in which case its span would never be emitted. A normal successful `GET /` completes in ~1.2 seconds with ~28 child spans.

## Diagnosis

The symptom — browser tab spinner, not an in-page spinner — indicates the HTTP response itself never starts arriving, not a client-side JavaScript failure. Because the issue is:

- **Consistent when a session cookie is present**
- **Never occurs without a session cookie** (private window, other browsers without a session, external uptime monitors)
- **Affects any browser** (confirmed on Chrome desktop and Chrome iOS)
- **Intermittent** (sometimes fast, sometimes hangs)

The trace evidence rules out KV latency or slow D1 queries as the cause — the Worker stalls **before any binding is accessed**. This points to something in EmDash's request initialization path (module-level code, middleware setup, or sandboxed plugin bootstrap) that deadlocks or blocks indefinitely for certain requests, likely triggered by the presence of a session cookie.

---

## Environment

- **Runtime:** Cloudflare Workers (zone-routed, not `workers.dev`)
- **Session storage:** Cloudflare KV (`SESSION` binding)
- **Database:** Cloudflare D1
- **Astro output:** `server` (full SSR, no static generation)

---

## Impact

- Affects only users with active admin sessions — public visitors are unaffected
- During a hang window, the admin cannot load any page on the site in any browser while logged in, without switching to private mode or logging out
- Workaround: clear browser cookies/cache for the site, which removes the session cookie and restores normal loading
- No data loss; the site is fully functional for visitors without a session cookie

---

## Suggested fix

Since the hang occurs before any binding is accessed, the fix likely needs to be in the request initialization path rather than in a specific binding call. A global request timeout guard would prevent indefinite hangs regardless of where in initialization the stall occurs:

```ts
// Pseudocode — exact implementation depends on EmDash internals
const INIT_TIMEOUT_MS = 5_000;

async function handleRequest(request: Request, env: Env, ctx: ExecutionContext) {
  const timeout = new Promise<Response>((resolve) =>
    setTimeout(() => resolve(new Response("Service unavailable", { status: 503 })), INIT_TIMEOUT_MS)
  );
  return Promise.race([processRequest(request, env, ctx), timeout]);
}
```

Alternatively, if the stall is in session validation specifically, a timeout with graceful fallback to unauthenticated state would prevent the hang while keeping the page functional for the visitor.

ascorbic:
## What does this PR do?

Fixes an isolate-poisoning deadlock in runtime and database initialization on workerd.

**Mechanism:** `getRuntime()` (middleware) and `getDatabase()` (runtime) guard cold-isolate initialization with single-owner state — a `runtimeInitializing` boolean and a shared `dbInitPromise` respectively. If the request that owns the init is cancelled mid-`await` (client disconnects during cold migrations on contended D1), workerd tears down its continuation and the `finally` that releases the guard **never runs**. The guard is stuck forever: every subsequent request in that isolate spins in the wait path (or awaits a never-settling promise) until the platform kills it at the wall limit — observed in production as 524s at exactly ~100s, with `cf-cache-status: BYPASS`, zero bytes, and no Server-Timing. The isolate stays poisoned until eviction, and the CDN cache's request coalescing amplifies each poisoned fill to every concurrent visitor of that URL.

External reproduction during diagnosis: probes that aborted in-flight cold requests drove the hang rate on a test deployment from ~17% to 44% of uncached requests — each abort during a cold init manufactured another poisoned isolate. This is self-reinforcing in production: slow cold init → user abandons → poisoned isolate → more 524s.

**Fix:** both sites now use a reclaimable init lock (`packages/core/src/utils/init-lock.ts`):

- Waiters poll instead of awaiting a cross-request promise (which workerd flags and which can never settle if the owner dies).
- The owner records a claim timestamp; a waiter that finds the lock held past a 15s deadline assumes the owner is dead, reclaims the lock, and runs init itself — the isolate self-heals.
- Waiters give up with an error after 30s, so the worst case degrades to an error response instead of a 100-second hang.

TDD: the test simulating the abandoned owner (init never settles, cleanup never runs) was written first and fails against the old boolean/shared-promise shape.

Closes #1274

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes (`packages/core`; the repo-wide run fails on a pre-existing unbuilt-workspace error in `registry-client`, unrelated to this change)
- [x] `pnpm lint` passes (zero new diagnostics vs. the pre-existing baseline)
- [x] `pnpm test` passes — core unit (174 files / 2,639 tests) and integration (70 files / 1,039 tests) suites
- [ ] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/... _n/a — bug fix._

## Screenshots / test output

```
 Test Files  174 passed (174)        # tests/unit
      Tests  2639 passed (2639)

 Test Files  70 passed (70)          # tests/integration
      Tests  1039 passed (1039)
```

New suite: `packages/core/tests/unit/utils/init-lock.test.ts` (6 tests) — cached short-circuit, single init shared by concurrent waiters, **stale-owner reclaim after deadline (the production deadlock)**, lock release on init failure, bounded max wait, waiter picks up owner's cached value.

Production fingerprint this fixes (test deployment, cache-busted requests): 524 at ~100,0xx ms, `cf-cache-status: BYPASS`, 0 bytes, no Server-Timing; bimodal isolates (healthy ones serve ~300–500ms misses throughout).

---

### Try this PR

**[Open a fresh playground →](https://fix-init-lock-poisoned-isolate-emdash-playground.emdash-cms.workers.dev)**

<sub>Tracks `fix/init-lock-poisoned-isolate`. Updated automatically when the playground redeploys.</sub>

### DIFF
--- a/.changeset/init-lock-reclaim.md
+++ b/.changeset/init-lock-reclaim.md
@@ -2,4 +2,4 @@
 "emdash": patch
 ---
 
-Fix isolate-poisoning deadlock in runtime and database initialization. If the request that owned a cold-isolate init was cancelled mid-await (e.g. the client disconnected during migrations), the release in its `finally` never ran, leaving the init guard stuck forever — every subsequent request in that isolate hung until the platform killed it at the wall limit (observed as 524s after 100s) and the isolate was poisoned until eviction. Initialization is now guarded by a reclaimable lock: waiters poll (never awaiting a cross-request promise), a stale owner is reclaimed after a deadline, and waiters give up with an error rather than hanging indefinitely.
+Fixed a bug where a visitor disconnecting at the wrong moment during a cold start could leave that server instance permanently broken: every subsequent request to it would hang until the platform timed it out (a 524 error on Cloudflare, after 100 seconds), and the instance stayed broken until it was recycled. Sites no longer get stuck — startup now recovers automatically, and in the worst case a request fails fast with an error instead of hanging.
--- a/.changeset/init-lock-reclaim.md
+++ b/.changeset/init-lock-reclaim.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fix isolate-poisoning deadlock in runtime and database initialization. If the request that owned a cold-isolate init was cancelled mid-await (e.g. the client disconnected during migrations), the release in its `finally` never ran, leaving the init guard stuck forever — every subsequent request in that isolate hung until the platform killed it at the wall limit (observed as 524s after 100s) and the isolate was poisoned until eviction. Initialization is now guarded by a reclaimable lock: waiters poll (never awaiting a cross-request promise), a stale owner is reclaimed after a deadline, and waiters give up with an error rather than hanging indefinitely.
--- a/packages/core/src/astro/middleware.ts
+++ b/packages/core/src/astro/middleware.ts
@@ -68,32 +68,6 @@ import type { EmDashHandlers } from "./types.js";
 const RUNTIME_INIT_DEADLINE_MS = DB_INIT_DEADLINE_MS + 15_000;
 
 /**
- * The runtime singleton and its init lock live on globalThis behind a
- * Symbol — same reasoning as SETUP_VERIFIED_KEY below: the bundler can
- * duplicate this module across SSR chunks, and a duplicated instance/lock
- * would mean multiple runtimes (each with its own cron scheduler) per
- * isolate, initializing and reclaiming independently.
- */
-const RUNTIME_HOLDER_KEY = Symbol.for("emdash:runtime-holder");
-interface RuntimeHolder {
-	instance: EmDashRuntime | null;
-	lock: InitLock;
-}
-
-function getRuntimeHolder(): RuntimeHolder {
-	// eslint-disable-next-line typescript/no-unsafe-type-assertion -- globalThis symbol slot, written only below
-	let holder = setupFlagStore[RUNTIME_HOLDER_KEY] as RuntimeHolder | undefined;
-	if (!holder) {
-		holder = { instance: null, lock: createInitLock() };
-		setupFlagStore[RUNTIME_HOLDER_KEY] = holder;
-	}
-	return holder;
-}
-
-/** Whether i18n config has been initialized from the virtual module */
-let i18nInitialized = false;
-
-/**
  * Whether we've verified the database has been set up.
  * On a fresh deployment the first request may hit a public page, bypassing
  * runtime init. Without this check, template helpers like getSiteSettings()
@@ -118,6 +92,32 @@ function isSetupVerified(): boolean {
 function markSetupVerified(): void {
 	setupFlagStore[SETUP_VERIFIED_KEY] = true;
 }
+
+/**
+ * The runtime singleton and its init lock live on globalThis behind a
+ * Symbol — same reasoning as SETUP_VERIFIED_KEY above: the bundler can
+ * duplicate this module across SSR chunks, and a duplicated instance/lock
+ * would mean multiple runtimes (each with its own cron scheduler) per
+ * isolate, initializing and reclaiming independently.
+ */
+const RUNTIME_HOLDER_KEY = Symbol.for("emdash:runtime-holder");
+interface RuntimeHolder {
+	instance: EmDashRuntime | null;
+	lock: InitLock;
+}
+
+function getRuntimeHolder(): RuntimeHolder {
+	// eslint-disable-next-line typescript/no-unsafe-type-assertion -- globalThis symbol slot, written only below
+	let holder = setupFlagStore[RUNTIME_HOLDER_KEY] as RuntimeHolder | undefined;
+	if (!holder) {
+		holder = { instance: null, lock: createInitLock() };
+		setupFlagStore[RUNTIME_HOLDER_KEY] = holder;
+	}
+	return holder;
+}
+
+/** Whether i18n config has been initialized from the virtual module */
+let i18nInitialized = false;
 
 /**
  * Get EmDash configuration from virtual module
--- a/packages/core/src/astro/middleware.ts
+++ b/packages/core/src/astro/middleware.ts
@@ -51,14 +51,16 @@ import {
 	runWithContext,
 } from "../request-context.js";
 import { isMissingTableError } from "../utils/db-errors.js";
+import { createInitLock, initWithLock } from "../utils/init-lock.js";
 import type { EmDashConfig } from "./integration/runtime.js";
 import { createPublicPluginApiRouteHandler } from "./public-plugin-api-routes.js";
 import type { EmDashHandlers } from "./types.js";
 
 // Cached runtime instance (persists across requests within worker)
 let runtimeInstance: EmDashRuntime | null = null;
-// Whether initialization is in progress (prevents concurrent init attempts)
-let runtimeInitializing = false;
+// Guards concurrent init attempts; reclaimable so a request cancelled
+// mid-init can't poison the isolate (see utils/init-lock.ts)
+const runtimeInitLock = createInitLock();
 
 /** Whether i18n config has been initialized from the virtual module */
 let i18nInitialized = false;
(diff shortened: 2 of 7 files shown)


ascorbic:
Review addressed in 85bc8b4c:

**(a) Deadline composition** — `MIGRATION_RACE_WAIT_MS` is now exported from the runner and the db lock deadline derives from it (`DB_INIT_DEADLINE_MS = MIGRATION_RACE_WAIT_MS + 20_000` = 30s), so they can't drift. The outer runtime lock uses `DB_INIT_DEADLINE_MS + 15_000` (45s) — inner strictly less than outer. Waiter `maxWaitMs` now defaults to `deadlineMs + 15s` so a waiter always survives long enough to reclaim a dead owner.

**(b) waitUntil anchoring** — `initWithLock` gained an `anchor` option; both call sites pass `(p) => after(() => p)` (the existing `after()` helper resolves the host's waitUntil via the virtual module). A cancelled owner's init now completes under waitUntil, populates the cache, and releases the lock — prevention first, reclaim as the net. The anchored copy is rejection-swallowed so a failed init can't surface as an unhandled rejection in the host.

**(c) globalThis singletons** — runtime instance + lock and db cache + lock now live behind `Symbol.for` keys (`emdash:runtime-holder`, `emdash:db-cache`), same reasoning as the adjacent setup-verified flag. Duplicated SSR chunks share one lock/cache; no more double-runtime/double-cron hazard.

**(d) Slow-live-owner test** — added: owner slower than the deadline, waiter reclaims; asserts init runs at most twice, both callers resolve, cache converges, and a third caller is served from cache. Plus an anchor test.

**(e)** Date.now() guard comment added (clock only advances across I/O in workerd — the sleep is load-bearing). Confirmed the doInit path degrades gracefully on a maxWait throw: the existing `catch` at the runtime-init site logs and continues, leaving `locals.emdash` unset — admin/API routes return the structured `NOT_CONFIGURED` error rather than a raw 500.

Also worth stating for the record (per review): reclaim-induced concurrent migration lands in the already-handled `waitForConcurrentMigrator` path — the runner detects the UNIQUE-constraint race and waits it out, which is what makes polling-reclaim safe here.

Suites: 2,641 unit + 1,039 integration pass; core typecheck and lint clean.

ascorbic:
Both review findings addressed in 90711fdd:

**Clobber race (emdashbot #1, Copilot both comments)** — confirmed real, and the suggested fix is what shipped: `InitLock` gains a monotonic `generation` claim counter, each claim bumps it, and the `finally` only clears `ownerStartedAt` while still the current owner. A slow stale owner finishing (or failing) after a reclaim can no longer release the reclaimer's claim.

**Test gap (emdashbot #2)** — you were right that the "at most twice" test passed by timing accident: the genuinely dangerous interleaving needs the stale owner to *fail* slowly (success populates the cache, which already shields later callers). Added the regression test with the full overlap: owner A rejects at t≈400 after waiter B reclaimed at t≈300, third caller C arrives at t≈440 in the clobber window while B is mid-init. Before the fix it fails with 3 inits; after, C waits for B and `initCalls === 2`.

**Also picked up Copilot's secondary suggestion** — init is now normalized via `Promise.resolve().then(init)` so a synchronous throw still attaches the waitUntil anchor before rejecting, and **the ordering nit** — `setupFlagStore`/`SETUP_VERIFIED_KEY` moved above `getRuntimeHolder`.

9/9 lock tests, 2,642 unit + 1,039 integration pass; typecheck/lint/format clean.

ascorbic:
Follow-up round addressed in 6f8ab6d8 — all three reviewers converged on the publication side of the same race, and you're right that the cache write needed the same gating as the release:

**Generation-gated publication** — `initWithLock` now passes an `isCurrentClaim` predicate into `init`. Both call sites publish only while still the current owner, so a reclaimed-but-live init finishing late can no longer flip the singleton back to a stale instance.

**The orphan-scheduler leak (ask-bonk)** — went one better than suppressing the write: a superseded runtime now calls its own `runtime.stopCron()` (public API, documented for exactly this shutdown case), so the loser's scheduler is stopped rather than leaked — on either side of the race. The superseded instance is still returned to the request that built it, since it's fully functional for that request.

**Db cache identity flip (ask-bonk/Copilot)** — same gating on `holder.cache.set`; the concurrent-second-migrator degradation is unchanged and remains covered by the runner's UNIQUE-race handling.

**Test readability (ask-bonk)** — added the requested comment: the existing convergence test's inits write the cache ungated, so it demonstrates last-writer-wins explicitly (`expect(cache).toBe("slow-owner")` is now asserted, not implied), with a pointer to the new gated-publication test that shows real-caller behavior (reclaimer's value survives, each caller still gets its own return value, claim checks observed as `[true, false]`).

10/10 lock tests, 2,643 unit + 1,039 integration, typecheck/lint/format clean.

PR #1410, opened 2026-06-11
perf(cloudflare): opt-in coalescing of same-turn SELECTs into one D1 batch() round trip

ascorbic:
## What does this PR do?

Adds an experimental, opt-in coalescing driver for D1: SELECT queries issued in the same event-loop turn are buffered and executed as a **single `D1Database.batch()` call** — one HTTP round trip instead of N. Production pages currently issue 5–7 fully serialized queries at 15–40ms each; with coalescing, queries that are issued concurrently (e.g. `Promise.all` in templates, parallel component renders) collapse into one round trip.

```ts
emdash({
	database: d1({ binding: "DB", session: "auto", coalesce: true }),
});
```

Design constraints (deliberately conservative):
- **SELECT-only.** Writes, `WITH` (SQLite allows CTEs on writes), PRAGMA, EXPLAIN — everything else takes the direct path immediately, with kysely-d1's exact result mapping.
- **Per-request session only.** The coalescing Kysely (and its buffer) is created per request in `createRequestScopedDb`, so cross-request buffering is impossible by construction. The shared singleton never coalesces. `session` must be enabled.
- **Atomic-batch error fallback.** D1 batches are atomic — one bad statement rejects the whole call. On batch failure every buffered statement is re-executed individually (safe: all SELECTs), so innocent queries resolve and only the failing one rejects with its own error. Per-query error semantics preserved.
- Flush window is `setTimeout(0)` (macrotask) — Kysely awaits internally between connection acquisition and execution, so a microtask window closes before sibling queries arrive. A lone query skips `batch()` and executes directly. Queries arriving mid-batch open a fresh window.

**Notable finding while implementing:** kysely's `SqliteAdapter` reports `supportsMultipleConnections: false`, which makes Kysely's `RuntimeDriver` serialize *every* query behind a connection mutex — even `Promise.all`'d queries never run concurrently against D1 today. The coalescing dialect overrides this (its shared connection is concurrent-safe by design; transactions throw regardless). This mutex is plausibly a contributor to the fully-serialized query waterfalls observed in production even where templates already parallelize.

Refactor note: `EmDashD1Dialect` moved from `d1.ts` to a new `d1-dialect.ts` to avoid a circular ESM import (the coalescing dialect extends it while `d1.ts` imports the coalescing dialect); behavior unchanged.

Closes #

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [x] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes (cloudflare + core)
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (or targeted tests for my change) — 7 new tests; full @emdash-cms/cloudflare suite 162/162
- [x] I have added a [changeset](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#changesets) (if this PR changes a published package) — minor, @emdash-cms/cloudflare
- [ ] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/... _Opt-in experimental flag on existing d1() config — flagged in the description for a maintainer call on whether this needs a Discussion._

## Screenshots / test output

New mock-D1 test suite covers: same-turn coalescing in issue order with per-caller rows and preserved bind params; sequential awaits never batch; writes + WITH execute immediately (order-asserted) while sibling SELECTs still coalesce; batch rejection → individual fallback with the underlying error; direct-path result mapping (rows / numAffectedRows / insertId, 0 changes → undefined); new window after each flush; transactions rejected.

---

### Try this PR

**[Open a fresh playground →](https://perf-d1-query-coalescing-emdash-playground.emdash-cms.workers.dev)**

<sub>Tracks `perf/d1-query-coalescing`. Updated automatically when the playground redeploys.</sub>

### DIFF
--- a/.changeset/dirty-pugs-attack.md
+++ b/.changeset/dirty-pugs-attack.md
@@ -1,0 +1,5 @@
+---
+"@emdash-cms/cloudflare": minor
+---
+
+Adds an experimental opt-in `coalesce` option to the `d1()` adapter. When enabled (alongside `session`), SELECT queries issued in the same event-loop turn on the per-request session database are buffered and executed as a single D1 `batch()` call — one HTTP round trip instead of N fully-serialized ones. Writes, CTEs and other statements always execute immediately, and if a batch fails the buffered queries are retried individually so each keeps its own error semantics.
--- a/.changeset/dirty-pugs-attack.md
+++ b/.changeset/dirty-pugs-attack.md
@@ -2,4 +2,12 @@
 "@emdash-cms/cloudflare": minor
 ---
 
-Adds an experimental opt-in `coalesce` option to the `d1()` adapter. When enabled (alongside `session`), SELECT queries issued in the same event-loop turn on the per-request session database are buffered and executed as a single D1 `batch()` call — one HTTP round trip instead of N fully-serialized ones. Writes, CTEs and other statements always execute immediately, and if a batch fails the buffered queries are retried individually so each keeps its own error semantics.
+New experimental `coalesce` option for the `d1()` adapter, for much faster uncached page loads:
+
+```ts
+emdash({
+	database: d1({ binding: "DB", session: "auto", coalesce: true }),
+});
+```
+
+When enabled, read queries that a page issues at the same time are sent to D1 as a single round trip instead of one at a time. A page that runs half a dozen queries — settings, menus, the entry, related posts — pays for one trip to the database instead of six, which can cut uncached render time by more than half. Each query still gets its own results and its own errors, and writes are unaffected. Requires `session` to be enabled; off by default while experimental.
--- a/infra/blog-demo/astro.config.mjs
+++ b/infra/blog-demo/astro.config.mjs
@@ -16,7 +16,7 @@ export default defineConfig({
 	integrations: [
 		react(),
 		emdash({
-			database: d1({ binding: "DB", session: "auto" }),
+			database: d1({ binding: "DB", session: "auto", coalesce: true }),
 			storage: r2({ binding: "MEDIA" }),
 			plugins: [formsPlugin()],
 			sandboxed: [webhookNotifier],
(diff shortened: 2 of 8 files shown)


ascorbic:
Added `coalesce: true` to the two benchmark sites (infra/cache-demo, infra/blog-demo) in f9791b10, so merging this PR turns coalescing on for both deployments and the before/after comparison is direct. Pre-merge baseline for cache-demo (from London, worker remote-LHR, ~20ms/query): warm-isolate MISS render 148–250ms with 6–10 fully serialized queries (db span = entire middleware window), cold-isolate MISS 945–1075ms total. Expected post-merge: db span collapses to 1–2 batch windows; warm-MISS render roughly 60–100ms. I'll re-run the identical battery after deploy.

ascorbic:
All three review items addressed in 4c42244c:

- **Silent no-op foot-gun (emdashbot)** — `createDialect` now warns at startup when `coalesce: true` is set without sessions enabled: `[emdash] d1({ coalesce: true }) has no effect without sessions — set session: "auto" (or "primary-first") to enable query coalescing.` Chose a warning over a hard error since the config is otherwise valid and functional. (No unit test for the warn itself — it lives in the `cloudflare:workers`-importing module that's deliberately kept out of the test import graph.)
- **Fallback determinism (Copilot)** — the batch-failure fallback now re-executes statements sequentially in issue order rather than via `Promise.all`; it's an error path where determinism beats latency, and it avoids piling concurrent retries onto a database that just failed.
- **Debuggability (Copilot)** — `PendingQuery` carries the SQL text and the missing-result rejection now names the statement index and SQL.
- **Ordering caveat (Copilot)** — documented on the `coalesce` option: buffered reads execute at the flush window while writes execute immediately, so a read and write issued concurrently in the same turn may run write-first; sequential `await` code is unaffected.

Cloudflare suite 162/162, typecheck/lint/format clean.

PR #1426, opened 2026-06-12
fix(seed): honor non-en default locale across export-seed → seed round-trip

### LINKED ISSUE #1421: CLI hardcodes 'en' as default locale: export-seed → seed round-trip rewrites a non-'en' default locale to 'en'
### Summary

The seed CLI assumes the default locale is `en`. For any project whose configured `defaultLocale` is **not** `en`, an `export-seed` → `emdash seed` round-trip silently rewrites locale-bearing rows (menus, taxonomies, content) from their real default locale to `en`. The text is preserved but filed under the wrong locale, so the runtime's locale-scoped lookups (`getMenu`, taxonomy/content queries) no longer resolve it.

This is a follow-up to #1330: that fix made `export-seed` infer i18n *from the data*, but two independent `en` assumptions remain.

### Repro

1. Project configured multi-locale with a non-`en` default, e.g. `i18n: { defaultLocale: "de", locales: ["de", "en"] }`.
2. The data currently holds a **single distinct locale** — the default, `de`. (Normal: a bilingual site that has only authored its primary locale so far, or any site that only ever uses its one default locale.)
3. `emdash export-seed --with-content` → the emitted `menus`/`taxonomies`/`content` carry **no `locale`** field.
4. `emdash seed <that file>` into a fresh DB.
5. **Result:** every row lands at `locale = 'en'`. `getMenu("primary")` under the `de` locale returns nothing.

### Root cause — two `en` assumptions

**1. Export omits `locale` for single-locale data — `detectI18nEnabled` (`packages/core/src/cli/commands/export-seed.ts`).** The CLI never initializes the i18n config, so locale-awareness is inferred from the data and is true only when **>1 distinct locale** is present. A project that currently has data in just one locale is treated as non-i18n, so `exportMenus` (L469-L476) and the taxonomy/content equivalents omit `locale` entirely. Note this triggers even for projects *configured* multi-locale — it keys off how many locales the data happens to contain, not the config.

**2. Apply backfills the wrong default — `packages/core/src/seed/apply.ts:548` (and L227 for taxonomies):**
\`\`\`ts
const fallbackLocale = getI18nConfig()?.defaultLocale ?? "en";
const locale = menu.locale ?? fallbackLocale;
\`\`\`
`emdash seed` also runs outside the Astro runtime, and the CLI does not load `astro.config`, so `getI18nConfig()` is null and `fallbackLocale` collapses to `"en"` — never the project's real default. (The `_emdash_menu_items.locale` column default of `'en'` is the same assumption baked into the schema.)

Together: rows in a non-`en` default locale are exported without a locale, then re-applied as `en`. Only projects whose default is literally `en` survive the round-trip.

### Why it can't simply be fixed at apply-time

The CLI has no access to `defaultLocale` — it does not load `astro.config`. So `apply` can't "just read the configured default." The locale has to either be **carried in the seed** or **passed explicitly**.

### Suggested fixes (either/both)

- **Make the seed self-describing.** Have `export-seed` emit the actual `locale` even for single-locale data (don't treat "one distinct locale" as "no i18n"), **or** add a top-level `defaultLocale` to `SeedFile` that `apply` honors as the fallback.
- **Let `apply` be told the default.** Add a `--default-locale` flag (and/or have the runtime-driven seed path pass `getI18nConfig().defaultLocale`) instead of hardcoding `"en"`.

### Workaround (consumer side)

Pin `locale` explicitly in the seed after exporting — re-attach the data's actual locale to each menu/taxonomy/content entry that `export-seed` left bare.

MA2153:
## What does this PR do?

`emdash export-seed` now emits a top-level `defaultLocale` for single-locale projects, and `applySeed` honors it when backfilling the locale of menus, taxonomies, and content rows that omit an explicit `locale`.

Previously an `export-seed` → `seed` round-trip silently rewrote a non-`en` default locale (e.g. `de`) to `en`: the CLI runs outside the Astro runtime, so the locale fallback collapsed to `en`. Projects whose default locale is already `en` are unaffected.

Closes #1421

## Type of change

- [x] Bug fix

## Checklist

- [ ] `pnpm typecheck` passes 
- [ ] `pnpm lint` passes 
- [x] `pnpm test` passes (or targeted tests for my change) — 8/8 in the two seed test files below
- [ ] User-visible strings in the admin UI are wrapped for translation — n/a, no admin UI strings

## Screenshots / test output

```
✓ tests/unit/seed/export-seed-cli-i18n.test.ts
✓ tests/integration/seed-i18n-default-locale.test.ts
Test Files  2 passed (2)
     Tests  8 passed (8)
```

> Note: `pnpm typecheck`, `pnpm lint`, and `pnpm format` were not run locally — the type-aware tooling OOM-killed the dev machine. Targeted seed tests pass; CI will cover the rest.

### DIFF
--- a/.changeset/fix-seed-default-locale.md
+++ b/.changeset/fix-seed-default-locale.md
@@ -1,0 +1,12 @@
+---
+"emdash": patch
+---
+
+Fix seed CLI hardcoding `en` as the default locale (#1421)
+
+`emdash export-seed` now emits a top-level `defaultLocale` for single-locale
+projects, and `emdash seed` (`applySeed`) honors it when backfilling the locale
+of menus, taxonomies, and content rows that omit an explicit `locale`. Previously
+an `export-seed` → `seed` round-trip silently rewrote a non-`en` default locale
+(e.g. `de`) to `en`, since the CLI runs outside the Astro runtime and the
+fallback collapsed to `en`. Projects whose default locale is `en` are unaffected.
--- a/packages/core/src/cli/commands/export-seed.ts
+++ b/packages/core/src/cli/commands/export-seed.ts
@@ -20,7 +20,7 @@ import { OptionsRepository } from "../../database/repositories/options.js";
 import { TaxonomyRepository } from "../../database/repositories/taxonomy.js";
 import type { Database } from "../../database/types.js";
 import { validateIdentifier } from "../../database/validate.js";
-import { isI18nEnabled } from "../../i18n/config.js";
+import { getI18nConfig, isI18nEnabled } from "../../i18n/config.js";
 import { SchemaRegistry } from "../../schema/registry.js";
 import type { FieldType } from "../../schema/types.js";
 import type {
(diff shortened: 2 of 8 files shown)


MA2153:
Good call — added `defaultLocale` validation in `validateSeed` (70926d5d). A blank or whitespace-padded value is now rejected as `defaultLocale: must be a non-empty string with no leading or trailing whitespace`, mirroring the existing `avatar.storageKey` guard. Covered by 5 new cases in `validate.test.ts` (omitted, non-en, non-string, empty, whitespace-padded); all 56 validate tests + 121 seed tests pass.

ascorbic (approved):
Thanks

PR #1439, opened 2026-06-12
feat: native content filtering by byline credit

### LINKED ISSUE #1358: Add native content filtering by byline credits
### Description

EmDash does not currently provide a native server-side way to query content entries credited to a byline through the `_emdash_content_bylines` junction table.

For author archive pages, site code needs to answer: "which posts is byline X credited on?" Today, `getEmDashCollection()` can filter content-table columns such as `primary_byline_id`, but that only finds entries where the byline is the first/primary credit. It misses co-authored posts where the same byline is a secondary credit, because those relationships only exist in `_emdash_content_bylines`.

Expected behavior: EmDash should expose a first-class content query/filter for byline credits, similar to taxonomy term filtering, so sites can query all entries credited to a byline without raw DB access.

Actual behavior: sites must query `_emdash_content_bylines` directly via `getDb()` to collect content IDs, then pass those IDs back into `getEmDashCollection()` for normal status, locale, ordering, and hydration behavior.

### Steps to reproduce

1. Create a `posts` collection.
2. Create two bylines, for example `alice` and `bob`.
3. Create/publish a post with both bylines in order: `alice` first, `bob` second.
4. Try to build Bob's author archive using only the public content query API.
```ts
import { getBylineBySlug, getEmDashCollection } from "emdash";

const byline = await getBylineBySlug("bob");
if (!byline) throw new Error("Missing byline");

const bylineGroup = byline.translationGroup ?? byline.id;

const result = await getEmDashCollection("posts", {
  where: { primary_byline_id: bylineGroup },
  orderBy: { published_at: "desc" },
});
```

5. The post is not returned, because Bob is a secondary credit and `primary_byline_id` stores only Alice's byline group.
6. The only reliable workaround is a raw query against `_emdash_content_bylines`, then a second content query by ID.

```ts
import { getEmDashCollection } from "emdash";
import { getDb } from "emdash/runtime";

const db = await getDb();
const rows = await db
  .selectFrom("_emdash_content_bylines as cb")
  .where("cb.collection_slug", "=", "posts")
  .where("cb.byline_id", "=", bylineGroup)
  .select("cb.content_id")
  .execute();

const postIds = rows.map((row) => row.content_id);

const result = await getEmDashCollection("posts", {
  where: { id: postIds },
  orderBy: { published_at: "desc" },
  limit: postIds.length,
});
```

Possible API shape:

```ts
const result = await getEmDashCollection("posts", {
  where: { byline: bylineGroup },
  orderBy: { published_at: "desc" },
});
```

This should join `_emdash_content_bylines`, match `byline_id` against the byline `translation_group`, and return entries where the byline is any explicit credit, not just the primary credit.

### Environment

- emdash version: 0.17.0
- Latest checked version: 0.17.1 also appears to lack this API
- Node.js version: v22.22.3
- Runtime: Cloudflare Workers
- OS: macOS

### Logs / error output

```shell
No runtime exception is thrown. The issue is silent incomplete results: filtering by `primary_byline_id` omits valid co-authored posts where the target byline is not first.
```

ascorbic:
## What does this PR do?

Adds a first-class way to query content by byline credit. `getEmDashCollection` (and the underlying `loadCollection`) now accept a reserved `byline` key in `where` that INNER JOINs the `_emdash_content_bylines` junction table and matches entries credited to one or more byline translation groups.

Unlike filtering on the content table's `primary_byline_id` column, this matches **every explicit credit**, so co-authored entries where the byline is a secondary credit are included. This makes author archive pages correct without dropping to raw `getDb()` access, which was the only prior workaround.

- Reserved `byline` where-key in the loader; accepts a single translation group or an array (OR). Composes with taxonomy, field, locale, status, and ordering filters via a shared `SELECT DISTINCT` joined path.
- `byline_id` is matched against the byline's `translation_group` (the value credits store since migration 040), so a credit spans every locale of a byline.
- Empty-array filter short-circuits to an empty result; range operators on `byline` are rejected with a warning.
- New `getEntriesByByline(collection, byline, options)` runtime helper, exported from the package entry point, mirroring `getEntriesByTerm`.

`byline` was already in `RESERVED_FIELD_SLUGS`, so no user-defined field can shadow the key.

Closes #1358

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] User-visible strings in the admin UI are wrapped for translation (n/a — no admin UI strings; server-side runtime query API only)
- [x] New features link to an approved Discussion: filed against issue #1358 (enhancement); opening at maintainer direction.

- [x] This PR includes AI-generated code — model/tool: Claude Opus 4.8 (OpenCode). Adversarial review pass run via GPT-5.5.

## Screenshots / test output

New `loader-byline-filter.test.ts`: 9 tests under `describeEachDialect` (SQLite + Postgres-in-CI) covering primary credit, secondary/co-authored credit, single-byline isolation, array OR, DISTINCT dedup, byline+field, byline+taxonomy, empty array, and range rejection. Full core suite: 3819 tests pass. Lint: 0 diagnostics. Typecheck: clean.

---

### Try this PR

**[Open a fresh playground →](https://feat-byline-content-filter-emdash-playground.emdash-cms.workers.dev)**

<sub>Tracks `feat/byline-content-filter`. Updated automatically when the playground redeploys.</sub>


ascorbic:
Pushed `ea2acac9` addressing the failing **Tests** check and the review comments.

### Root cause of the Postgres test failures
The `Tests` job failed only on the `[postgres]` dialect with `could not identify an equality operator for type json`. The `JOIN + SELECT DISTINCT table.*` approach is invalid on Postgres when a selected column is `json` (Postgres has no equality operator for `json`, so it can't dedupe such rows). SQLite is permissive, so it passed locally and in the SQLite suite. This was a **latent bug in the existing taxonomy filter path** too, only never exercised on Postgres (the taxonomy tests were SQLite-only).

Fix: both taxonomy and byline filters are now applied as correlated `EXISTS` semi-joins instead of `INNER JOIN ... DISTINCT`. No row fan-out, no `DISTINCT`, so the `json` equality problem disappears and it works on both dialects. Verified the full byline + taxonomy + field-filter suites pass on SQLite **and** a local Postgres 17.

### Copilot review comments
1. **Empty taxonomy array → `t.slug IN ()`** (loader.ts) — valid, and still latent after the refactor. Empty taxonomy filter arrays now short-circuit to an empty result, matching the byline filter's existing behavior. Added a regression test.
2. **Weak range-rejection test** — valid. The test now seeds a second *uncredited* published entry and asserts both rows return, proving the `byline` range filter is actually dropped rather than coincidentally passing with a single row.

PR #1442, opened 2026-06-12
feat(admin): status, author, and date-range filters for content list (#1288)

### LINKED ISSUE #1288: Bug - Content Collections - Better Admin Filtering options + Faster Page loads for large collections
### Description

The changes in #1064 should be brought into the admin UI, specifically for better filtering options for Content Collections.

Similar problem state to #1219.

A common use case would be:
- Looking for entities within a specific timeframe / changing the sort order
- Looking for entities with a specific, custom boolean attribute checked
- Looking for entities associated with specific authors

This problem is compounded by the slow load speeds when you navigate through a lot of content entities.

https://github.com/user-attachments/assets/121e7832-d579-4e79-975f-8b954d10adeb

### Steps to reproduce

1. Create a content collection. Add some posts.
2. Try to search and filter.

### Environment

- emdash: 0.17.0
- Cloudflare Workers
- MacOS / Chrome

ascorbic:
## What does this PR do?

Brings richer filtering into the admin content list (#1288). Editors can now narrow a collection by:

- **Status** — draft / published / scheduled / archived (exposes the existing API `status` param in the UI).
- **Author** — the user who authored the entry (`author_id`).
- **Date range** — an inclusive range over a chosen timestamp column (created / updated / published).

Filtering runs server-side, so it works across the whole collection rather than only the loaded page (the same class of problem as #1219).

### Core

- `ContentRepository.findMany`/`count` gain an inclusive date-range filter over a whitelisted timestamp column (`createdAt`/`updatedAt`/`publishedAt`). Bounds compare lexicographically against the stored ISO 8601 timestamps, which is correct because every timestamp is written via `toISOString()`. Date-only (`YYYY-MM-DD`) bounds are normalized to UTC day boundaries server-side so a date-only upper bound stays inclusive of the whole day. Author filtering reuses the existing `where.authorId`.
- New `findDistinctAuthorIds` on the content repo and a chunked `UserRepository.findByIds`.
- New `handleContentAuthors` handler and `GET /_emdash/api/content/{collection}/authors` route, gated on `content:read_drafts` (editorial tier) since the response carries author emails and reveals authors of unpublished entries. Returns only users who have authored at least one live entry.
- `contentListQuery` extended with `authorId` / `dateField` / `dateFrom` / `dateTo` (all additive and optional); `contentAuthorsResponseSchema` and the OpenAPI path added.

### Admin

- A filter bar in `ContentList` (status, author, date-field + from/to date inputs) built on Kumo `Select`/`Input`, with a "Clear filters" control. The bar is opt-in and the component stays backward-compatible.
- The author dropdown is hidden when no entry in the collection has an `author_id`.
- Date inputs widen to UTC day boundaries so the inclusive upper bound covers the whole selected day.

Scope note: the issue also mentions filtering by a custom boolean attribute. That's intentionally deferred to keep this PR focused; happy to follow up.

Closes #1288

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] User-visible strings in the admin UI are wrapped for translation. No `messages.po` changes included.
- [ ] New features link to an approved Discussion — none yet. Implemented directly against maintainer-filed #1288; a maintainer should open/approve a Discussion or waive this before merge.

## Screenshots / test output

```
Test Files  258 passed (258)
     Tests  3850 passed (3850)
```

Manually verified in the running admin (Node/SQLite blog demo, 9 seeded posts):

- Status filter narrowed to the single draft.
- Date filter: a 2020 range returned no results; a same-day 2026 range returned content (confirming the inclusive end-of-day upper bound includes noon-stamped rows).
- Author dropdown correctly hidden when seeded content has bylines but no `author_id`.
- Arabic (RTL) renders correctly — sidebar and filter bar flip, no layout breakage.


ascorbic:
Thanks for the review. Addressed in fdf0fa8e:

1. **Date-only bound footgun (`schemas/content.ts`, `lib/api/content.ts`)** — instead of restricting to datetime-only, the handler now normalizes date-only (`YYYY-MM-DD`) bounds to UTC day boundaries (`from → T00:00:00.000Z`, `to → T23:59:59.999Z`). Date-only stays a friendly API input and the upper bound is inclusive of the whole day; full datetimes pass through unchanged. The "ISO date or datetime" docs are now accurate. Added a regression test (`treats a date-only upper bound as inclusive of the whole day`).

2 & 3. **PII / unpublished-metadata leak on the authors endpoint** — good catch. Gated the route on `content:read_drafts` (CONTRIBUTOR) instead of `content:read` (SUBSCRIBER). The author filter is editorial tooling and this tier matches where the list route stops forcing `status=published`, so subscribers can no longer reach the emails or learn authors of unpublished entries.

4. **Locale in the authors query key** — removed. Authors are collection-wide (the endpoint doesn't scope by locale), so the key no longer includes locale and the selection stays valid across locale switches without refetching.

5. **Client doc wording** — resolved by the normalization in (1); date-only and datetime bounds both behave correctly now, so the "ISO date or datetime" description is accurate.

PR #1458, opened 2026-06-13
fix: npx emdash types crashes due to schema envelope mismatch

### LINKED ISSUE #1188: npx emdash types is broken against deployed (non-dev) instances
### Description

**Summary**

Running npx emdash types --url <deployed-instance> --token <token> fails with:

_ERROR  Failed to fetch schema: Cannot read properties of undefined (reading 'collections')
Tracing through the source, there are two separate defects that combine to make production type generation unusable._

**Defect 1 — client unwrap mismatch**

EmDashClient.request() ([src/client/index.ts:775-780](https://github.com/emdash-cms/emdash/blob/main/src/client/index.ts)) returns json.data for every response:

_const json = (await response.json()) as { data: T };
return json.data;_

But the /_emdash/api/schema route ([src/astro/routes/api/schema/index.ts:92](https://github.com/emdash-cms/emdash/blob/main/src/astro/routes/api/schema/index.ts)) returns the schema un-enveloped:

_return new Response(JSON.stringify({ ...schemaExport, version }), ...);_

So client.schemaExport() returns undefined, and the CLI then dereferences schema.collections.length → crash. Other schema endpoints (e.g. /schema/collections) do use the { success, data } envelope — the inconsistency is the root cause.

Verified with curl: GET /_emdash/api/schema returns {"collections":[...], "version":"..."} with no data wrapper.

**Defect 2 — dev/prod typegen divergence**

Even with defect 1 fixed, the CLI calls /schema?format=typescript which produces a slim output:

```
// Generated by EmDash CLI
import type { PortableTextBlock } from "emdash";

export interface Page { ...; bylines?: ContentBylineCredit[]; }
export interface Post { ...; bylines?: ContentBylineCredit[]; }
```

Compared to what the dev server writes during astro:server:setup, this output is missing:

<reference types="emdash/locals" />
ContentBylineCredit in the import (file references it without importing → TS compile error)
declare module "emdash" { interface EmDashCollections { pages: Page; posts: Post; } } — without this, getEmDashCollection("posts") loses its typed return

The "rich" generator lives at POST /_emdash/api/typegen ([src/astro/routes/api/typegen.ts:95](https://github.com/emdash-cms/emdash/blob/main/src/astro/routes/api/typegen.ts)) but is gated by import.meta.env.DEV, so it's unreachable on a deployed Worker.

**Expected**

npx emdash types --url <deployed-instance> --token <token> should produce the same emdash-env.d.ts content the dev server writes locally — usable as-is.

**Use case**

CI workflow that regenerates emdash-env.d.ts from the deployed staging schema and opens a PR when the types drift, so devs don't have to remember to run npx emdash dev after a schema change in staging.

### Steps to reproduce

1. npx emdash types --url <deployed-instance> --token <token>
2. fails with exit code 1
3. seems to be an issue with the CLI expecting data as envelope
4. whilst /_emdash/api/schema yields collections as the top element, unenveloped

### Environment

Version: emdash@0.14.0 (server commit 102535f)

### Logs / error output

```shell
ERROR  Failed to fetch schema: Cannot read properties of undefined (reading 'collections')                                                                                                               10:33:41 PM

=> no data property exists in the output.
```

mvanhorn:
## What does this PR do?

`npx emdash types` crashes with `Cannot read properties of undefined (reading 'collections')` against any instance (deployed and local dev — reproduced on 0.15.0 and 0.17.2). Root cause is an envelope inconsistency. `EmDashClient.request()` (packages/core/src/client/index.ts:778-782) returns `json.data` for every response, and `schemaExport()` (client/index.ts:389-391) goes through `request`. But the `/_emdash/api/schema` route (packages/core/src/astro/routes/api/schema/index.ts:92-95) returns the schema un-enveloped as `{ ...schemaExport, version }` with no `data` wrapper. So `client.schemaExport()` returns `undefined`, and the CLI `types` command (packages/core/src/cli/commands/types.ts:42-43) dereferences `schema.collections.length` and throws. Sibling schema endpoints (e.g. `/schema/collections`) do use the `{ success, data }` envelope, confirming the JSON-schema route is the outlier.

Make the JSON `/schema` route consistent with the rest of the API by wrapping its body in the standard `{ success: true, data: { ...schemaExport, version } }` envelope (keep the `?format=typescript` branch and `X-Schema-Version` header unchanged, since `schemaTypes()` uses `requestRaw` and reads text directly). This restores the contract `request()` expects so `schemaExport()` returns the schema object. Verify no other caller depends on the un-enveloped JSON shape; the only consumers are `schemaExport()` (CLI) and any direct `curl`, and enveloping is the documented norm. If enveloping the route risks an external contract, the fallback is to switch `schemaExport()` to `requestRaw` + manual parse, but route-side enveloping is preferred for cross-endpoint consistency.

Closes #1188

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [ ] `pnpm typecheck` passes
- [ ] `pnpm lint` passes

- [x] This PR includes AI-generated code — model/tool: Codex (GPT-5.x)

## Screenshots / test output

- happy path: `schemaExport()` returns an object with a populated `collections` array against the enveloped route.
- happy path: `version` field is preserved in the returned schema object.
- edge: `schemaTypes()` (`?format=typescript`) still returns raw TS text, unaffected by the envelope change.
- error path: a non-2xx schema response surfaces a clear error instead of an undefined dereference.

AI was used for assistance.

### DIFF
--- a/.changeset/emdash-types-schema-unwrap.md
+++ b/.changeset/emdash-types-schema-unwrap.md
@@ -1,0 +1,7 @@
+---
+"emdash": patch
+---
+
+Fix `npx emdash types` crash caused by the schema endpoint envelope (#1188)
+
+The `/schema` route returned an enveloped JSON body while `client.request()` already unwraps the `.data` field, so `emdash types` received `undefined` and crashed. The route now returns the un-enveloped shape the client expects.
--- a/packages/core/src/astro/routes/api/schema/index.ts
+++ b/packages/core/src/astro/routes/api/schema/index.ts
@@ -13,7 +13,7 @@ import type { APIRoute } from "astro";
 import { hashString } from "emdash";
 
 import { requirePerm } from "#api/authorize.js";
-import { handleError, requireDb } from "#api/error.js";
+import { apiSuccess, handleError, requireDb } from "#api/error.js";
 import { SchemaRegistry } from "#schema/registry.js";
 import { generateTypeScript } from "#schema/zod-generator.js";
 
@@ -89,20 +89,12 @@ import type { PortableTextBlock } from "emdash";
 
 		const version = await hashString(JSON.stringify(schemaExport));
 
-		return new Response(
-			JSON.stringify({
-				...schemaExport,
-				version,
-			}),
-			{
-				status: 200,
-				headers: {
-					"Content-Type": "application/json",
-					"Cache-Control": "private, no-store",
-					"X-Schema-Version": version,
-				},
-			},
-		);
+		const response = apiSuccess({
+			...schemaExport,
+			version,
+		});
+		response.headers.set("X-Schema-Version", version);
+		return response;
 	} catch (error) {
 		return handleError(error, "Schema export failed", "SCHEMA_EXPORT_ERROR");
 	}
--- a/packages/core/tests/unit/client/client.test.ts
+++ b/packages/core/tests/unit/client/client.test.ts
@@ -1,5 +1,5 @@
-import { describe, it, expect } from "vitest";
 import { Role } from "@emdash-cms/auth";
+import { describe, it, expect } from "vitest";
 
 import { GET as schemaGET } from "../../../src/astro/routes/api/schema/index.js";
 import { EmDashClient, EmDashApiError } from "../../../src/client/index.js";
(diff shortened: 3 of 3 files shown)


ascorbic (approved):
Thanks!

mvanhorn:
Thanks a lot for taking this too, @ascorbic. Fixing the schema envelope mismatch should keep npx emdash types from crashing.

mvanhorn:
Thanks @ascorbic - fixing the schema envelope mismatch so npx emdash types stops crashing on the collections lookup is a solid catch.

PR #1459, opened 2026-06-13
fix: scheduled entries stay hidden after scheduledAt passes

### LINKED ISSUE #1402: Scheduled entries stay hidden after scheduledAt passes
## Bug Description

Scheduled content can remain unavailable on public single-entry routes after its scheduled time has passed. In `emdash@0.17.2`, the loader maps `scheduled_at` date columns into `Date` objects on `entry.data.scheduledAt`, but `getEmDashEntry()` checks visibility with `dataStr(data, "scheduledAt")`, which only returns string values. As a result, a loaded row with `status = "scheduled"` and a past `scheduledAt` can still be treated as not visible.

## Steps to Reproduce

1. Create an unpublished entry and schedule it for a future time.
2. Wait until the scheduled time has passed.
3. Request the public single-entry route that uses `getEmDashEntry(collection, slug)`.

## Expected Behavior

The entry should be publicly visible once `scheduledAt <= now`, or the row should be promoted to `published` by the runtime.

## Actual Behavior

The row is loaded with `status = "scheduled"`, but `scheduledAt` is a `Date`, so `dataStr()` treats it as missing and `isVisible()` returns false. The site returns a 404 for a due scheduled entry.

## Observed Environment

- `emdash@0.17.2`
- Astro + Cloudflare Worker/D1
- D1 row had `status = scheduled`, `scheduled_at` in the past, and public slug route returned 404.

## Notes

Likely fix: make the visibility check accept `Date` values for `scheduledAt`, or keep scheduled date fields as strings for this check. A related site-level mitigation is to promote due scheduled rows to `published` before route resolution.

mvanhorn:
## What does this PR do?

Single-entry public routes return 404 for due scheduled content. `getEmDashEntry()` gates visibility through `isVisible()` in `packages/core/src/query.ts` (around L602-608), which reads `scheduledAt` via `dataStr(data, "scheduledAt")`. But `mapRowToData()` in `packages/core/src/loader.ts` maps the `scheduled_at` column (a member of `DATE_COLUMNS`, loader.ts:174) into a `Date` object. `dataStr` (query.ts:245) only returns string values and falls back to `""` for a `Date`, so `scheduledAt` resolves to `undefined`, the `status === "scheduled" && scheduledAt && new Date(scheduledAt) <= new Date()` test short-circuits to false, and the entry stays hidden after its scheduled time passes. Confirmed on emdash@0.17.2 against both Cloudflare D1 and Postgres (Railway). The collection-level SQL path (`buildStatusCondition`, loader.ts:375) is correct and unaffected; only the post-load JS visibility check on the single-entry route is wrong.

In `isVisible()` (query.ts), make the `scheduledAt` read tolerate a `Date` value as well as an ISO string. The minimal fix is to coerce the raw `data.scheduledAt` field directly (e.g. accept a `Date` instance, or an ISO string, or a parseable value) instead of routing it through `dataStr`, then compare `new Date(scheduledAt) <= new Date()`. Keep `status` reading through `dataStr` as-is. Do not touch the SQL filter in loader.ts; it already handles scheduled rows correctly. Guard against invalid/NaN dates so a malformed `scheduledAt` does not unexpectedly publish or crash.

Closes #1402

## Type of change

- [x] Bug fix

## Checklist

- [x] This PR includes AI-generated code — model/tool: Codex (GPT-5.x)

## Screenshots / test output

Covered by tests: happy path: entry with `status="scheduled"` and a past `scheduledAt` (supplied as a `Date`) is visible from `getEmDashEntry`. happy path: entry with `status="scheduled"` and a future `scheduledAt` remains hidden (404). edge: `scheduledAt` supplied as an ISO string still works (back-compat with any string-valued path). edge: `status="published"` entry is visible regardless of `scheduledAt`. error path: `status="scheduled"` with missing/invalid `scheduledAt` stays hidden rather than throwing.

AI was used for assistance.


mvanhorn:
Huge thanks for the merge, @ascorbic. Scheduled entries showing up once scheduledAt passes fixes that 404 on due content.

mvanhorn:
Thanks @ascorbic. Keeping scheduled entries hidden until scheduledAt passes fixes those 404s on due content cleanly.

PR #1461, opened 2026-06-14
fix(registry): declaredAccess as the plugin trust contract

ascorbic:
## What does this PR do?

Makes `declaredAccess` the authoritative plugin trust contract end-to-end, fixing registry installs that failed with **"Plugin manifest has changed since you consented"** for any plugin declaring a hook-registration capability (`hooks.email-transport:register`, `hooks.email-events:register`, `hooks.page-fragments:register`) or `users:read`.

Root cause: the release record only carried the structured `declaredAccess`, whose vocabulary couldn't express hook registrations, so the publish step (`buildDeclaredAccess`) dropped them. The admin reconstructed the consent list from that lossy record, so the acknowledged permissions never matched the bundle the install handler enforced — a guaranteed `DECLARED_ACCESS_DRIFT` no user could clear.

The fix implements Path B from the design discussion: `declaredAccess` is the contract at every boundary (bundle manifest, registry record, install consent), while the runtime keeps enforcing the legacy `capabilities` strings, now **derived** from `declaredAccess` at the parse boundary via a total, isomorphic converter. The security-critical enforcement code is untouched.

- Completes the `declaredAccess` vocabulary in the lexicon (`email.events`/`email.transport`, `page.fragments`, `users`) — this is the RFC update in #694.
- Adds `capabilitiesToDeclaredAccess` / `declaredAccessToCapabilities` to `@emdash-cms/plugin-types`, proven isomorphic by a 1440-state exhaustive round-trip test.
- Bundler emits `declaredAccess`; publish carries it verbatim and deletes the lossy `buildDeclaredAccess`/`findUnmappedCapabilities`.
- `reconcileManifestAccess` re-derives `capabilities`/`allowedHosts` from `declaredAccess` at the two bundle-parse chokepoints both marketplace and registry installs share.
- Admin consent derives the list via the shared converter (one source of truth), replacing a component-local flattener.

An adversarial security pass found and this PR fixes two HIGH issues: (1) an empty `allowedHosts` array decoding to *unrestricted* network (a privilege inversion), and (2) the capability-string consent gate being blind to host scope, letting a record advertising one `allowedHosts` install a bundle enforcing another. The install handler now refuses any bundle whose enforced access (capabilities **and** hosts) differs from its signed record.

Design: RFC #694. Umbrella: Discussion #296.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm test` passes (plugin-types 32, plugin-cli 393, core 3940, admin 8, aggregator 88)
- [x] I have added/updated tests for my changes (round-trip property, publish/admin/consumer regressions, the two security-fix regressions)
- [x] User-visible strings in the admin UI are wrapped for translation — n/a: the admin change is capability-derivation logic; the only rendered values are capability identifiers (data, shown as Badges), not new translatable strings. No `messages.po` changes.
- [x] Linked to the registry umbrella Discussion #296 and RFC #694

## Screenshots / test output

```
plugin-types  32 passed
plugin-cli   393 passed
core        3940 passed (incl. 6 new install-integrity regressions)
admin          8 passed (RegistryPluginDetail)
aggregator    88 passed (extension re-validation)
typecheck    all 25 packages clean
lint         0 diagnostics
```

Migration note (in the changeset): re-publish affected plugins to adopt the new bundle format; existing installs are unaffected.

---

### Try this PR

**[Open a fresh playground →](https://feat-registry-declared-access-contract-emdash-playground.emdash-cms.workers.dev)**

<sub>Tracks `feat/registry-declared-access-contract`. Updated automatically when the playground redeploys.</sub>

### DIFF
--- a/.changeset/registry-declared-access-contract.md
+++ b/.changeset/registry-declared-access-contract.md
@@ -1,0 +1,9 @@
+---
+"@emdash-cms/plugin-types": minor
+"@emdash-cms/plugin-cli": minor
+"@emdash-cms/registry-lexicons": patch
+"@emdash-cms/admin": patch
+"emdash": minor
+---
+
+Fixes registry installs failing with "Plugin manifest has changed since you consented" for plugins that declare hook-registration capabilities (email transport, email events, page fragments) or read user records. Plugin bundles now declare their access as a structured `declaredAccess` contract that the registry record, the install-consent dialog, and the sandbox all read consistently, so every capability a plugin declares is shown for consent and enforced — no capability is silently dropped. Re-publish affected plugins to adopt the new bundle format; existing installs are unaffected.
--- a/packages/admin/package.json
+++ b/packages/admin/package.json
@@ -38,6 +38,7 @@
     "@dnd-kit/sortable": "^10.0.0",
     "@dnd-kit/utilities": "^3.2.2",
     "@emdash-cms/blocks": "workspace:*",
+    "@emdash-cms/plugin-types": "workspace:*",
     "@emdash-cms/registry-client": "workspace:*",
     "@emdash-cms/registry-lexicons": "workspace:*",
     "@floating-ui/react": "^0.27.16",
--- a/packages/admin/src/components/RegistryPluginDetail.tsx
+++ b/packages/admin/src/components/RegistryPluginDetail.tsx
@@ -15,6 +15,7 @@
 
 import { Badge, Button, LinkButton, Select, Tabs, Tooltip } from "@cloudflare/kumo";
 import type { TabsItem } from "@cloudflare/kumo";
+import { declaredAccessToCapabilities, type DeclaredAccess } from "@emdash-cms/plugin-types";
 import { checkEnvCompatibility } from "@emdash-cms/registry-client/env";
 import type { MessageDescriptor } from "@lingui/core";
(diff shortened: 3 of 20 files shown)


ascorbic:
Good catch — fixed in a1db32a8.

`handleRegistryUpdate` now applies the same `enforcedAccessEqual` gate as `handleRegistryInstall`: the new bundle's `declaredAccess` must match the new signed record's before the update proceeds. An update that changes only the host scope (`api.good.com` → `evil.com`) keeps the capability set identical, so it would previously sail through `diffCapabilities` and the escalation prompt — exactly the gap you describe. It now fails closed with `DECLARED_ACCESS_DRIFT`.

On the test point: the host-scope rejection logic lives in the shared `enforcedAccessEqual` predicate, which is unit-tested in `registry-access-integrity.test.ts` (host divergence, host widening, extra-capability, ordering-insensitivity). Both handlers call it identically; there's no full handler-level harness for either the install or update path in the current suite, so coverage for both is at the predicate level plus the verified wiring. Happy to add a handler-integration harness if you'd prefer it over the predicate-level guard.

Issue #1480, opened 2026-06-14
Collection `where` taxonomy filter joins on `t.id` (term id) instead of `translation_group`, so it only matches default-locale term slugs


MA2153:
## Summary

The collection `where` taxonomy filter joins the term table on the wrong key, so it only ever matches **default-locale** term slugs. Filtering a non-default locale by that locale's translated term slug returns nothing.

In `packages/core/src/loader.ts`, the taxonomy `EXISTS` semi-join is:

```ts
AND EXISTS (
  SELECT 1 FROM content_taxonomies ct
  INNER JOIN taxonomies t ON t.id = ct.taxonomy_id   // <-- wrong key
  WHERE ct.collection = ${type}
    AND ct.entry_id = ${tableName}.id
    AND t.name = ${name}
    AND t.slug IN (${slugs})
)
```

Since migration `036_i18n_menus_and_taxonomies.ts`, `content_taxonomies.taxonomy_id` stores the term's **`translation_group`**, not a row `id`. The two coincide only for the default-locale term (where `id == translation_group`). So `t.id = ct.taxonomy_id` resolves exclusively to the default-locale term row, and `t.slug` is therefore always the default-locale slug.

## Evidence

Every other taxonomy join in the codebase uses `translation_group` (plus a `locale` filter to pick the right per-locale term):

- `packages/core/src/database/repositories/taxonomy.ts:263,294` — `.innerJoin("taxonomies", "taxonomies.translation_group", "content_taxonomies.taxonomy_id")`
- `packages/core/src/taxonomies/index.ts:256,301,364` — same join
- `packages/core/src/taxonomies/index.ts:8-12` documents the contract explicitly: *"Because `content_taxonomies.taxonomy_id` stores the translation_group (not a specific term id), the joins here are `taxonomies.translation_group = content_taxonomies.taxonomy_id` + filter by `taxonomies.locale`."*

The loader's filter join is the lone `t.id = ct.taxonomy_id` site, and it carries no `locale` predicate.

## Impact

- Filtering content in a non-default locale by a **localized** term slug (e.g. `where: { category: "actualites" }` on `fr-fr`, where the default-locale slug is `news`) matches nothing.
- Filtering by the **default-locale** slug currently "works" only incidentally, because the join always lands on the default-locale term — it ignores the query's `locale` entirely.

## Suggested direction (not in scope here)

[...]

PR #1481, opened 2026-06-14
fix(loader): apply every taxonomy key in collection `where` filters

### LINKED ISSUE #1479: Collection `where` filter silently drops all but the first taxonomy key
## Summary

When a collection query's `where` filter contains **more than one taxonomy key**, only the first is applied. Every subsequent taxonomy key is silently dropped with a warning, so the query returns rows filtered by just one taxonomy instead of the intersection the caller asked for.

```ts
// Intended: entries tagged in BOTH taxonomies
getEmDashCollection("posts", {
  where: { category: ["news"], region: ["emea"] },
});
// Actual: filtered by `category` only; `region` ignored.
// Console: [emdash] where filter: only one taxonomy is supported per query, "region" ignored
```

This is a correctness bug, not just log noise: any UI that exposes facets from two or more taxonomies at once (a typical library/catalog browse page) returns results that ignore all but one facet group, with no error surfaced to the caller.

## Root cause

The single-taxonomy restriction is a deferred TODO, not an intentional semantic limit. The original taxonomy-JOIN implementation took only the first entry:

```ts
// For now, support single taxonomy filter (can extend later for multiple)
const [taxName, termSlugs] = Object.entries(taxonomyFilters)[0];
```

PR #1064 generalized the `where` map to also carry field, range, and `byline` filters — where **multiple keys already AND together** — and converted that silent `[0]` pickup into the explicit `only one taxonomy is supported per query` warning in `loader.ts`, but left taxonomy filtering otherwise single. So taxonomy is the one filter type in the map that never got generalized, even though the map's other keys already compose.

## Why this is now straightforward

#1064 also rewrote taxonomy filtering from `INNER JOIN ... DISTINCT` into a correlated `EXISTS` semi-join (to avoid Postgres's inability to `DISTINCT` a row containing a `json` column). `EXISTS` composes cleanly: N taxonomies = N `EXISTS` clauses `AND`'d together, each independently constraining the same base row — which is exactly the intended "match a term in *each* taxonomy" semantics, with no fan-out and no `DISTINCT`.

Proposed shape:

- Replace the scalar `taxonomyFilter: { name, slugs } | null` with a list `taxonomyFilters: { name, slugs }[]`.
- Drop the second-taxonomy warning/`continue`; collect every taxonomy key.
- Emit one `EXISTS (...)` per filter, `AND`'d. Each clause keeps its own `t.name = <name> AND t.slug IN (<that name's slugs>)` scoping — important because term slugs are unique only **within** a taxonomy, so each `EXISTS` must pin its own `t.name`; slugs must not be pooled into a single `IN`.
- The existing empty-slugs short-circuit generalizes: any one filter with zero slugs ⇒ empty result.

Semantics: **AND across taxonomies, OR (slug `IN`) within a taxonomy** — consistent with how field/byline keys in the same `where` map already behave.

## Indexing / cost

No schema change required — this is an API-shape fix on an already-wired path:

- `content_taxonomies` PK `(collection, entry_id, taxonomy_id)` serves the `EXISTS` correlation `ct.collection = ? AND ct.entry_id = <table>.id` as a left-prefix.
- The `taxonomies` join is by PK (`id`); `t.name` is backed by `idx_taxonomies_name` and the `(name, slug, locale)` unique constraint.

So each added `EXISTS` repeats the same fully-indexed pattern; cost scales linearly with the (small) number of selected taxonomies.

## Secondary question (pre-existing, please confirm)

Independent of this fix: the `EXISTS` join is `taxonomies t ON t.id = ct.taxonomy_id`, but since the i18n migration `content_taxonomies.taxonomy_id` stores a term's `translation_group`. Those only coincide for the default-locale term (where `id == translation_group`). Is the join key intended to be locale-correct here, or does taxonomy filtering currently only match default-locale terms? Worth confirming while this query is being touched — but it's separate from the multi-taxonomy bug above.

MA2153:
## What does this PR do?

A collection query whose `where` map contains more than one taxonomy key only applied the first; every subsequent taxonomy key was silently dropped with a `only one taxonomy is supported per query` warning. So a browse page exposing facets from two or more taxonomies (`{ category: ["news"], region: ["emea"] }`) returned rows filtered by a single taxonomy instead of the intersection, with no error surfaced to the caller.

This collects every taxonomy filter and emits one correlated `EXISTS` semi-join per taxonomy, AND'd together. Each clause keeps its own `t.name` scoping, so slugs are never pooled across taxonomies (term slugs are unique only *within* a taxonomy). The empty-slugs short-circuit is generalized to fire if any one taxonomy filter has zero slugs.

Semantics: **AND across taxonomies, OR (slug `IN`) within a taxonomy** — consistent with how field and byline keys in the same `where` map already compose. No schema or index change: the `EXISTS` correlation is a left-prefix of `content_taxonomies`' PK, so each added clause repeats the same fully-indexed pattern.

The secondary question raised in the issue (the `EXISTS` join uses `t.id = ct.taxonomy_id`, but `taxonomy_id` stores a `translation_group` since migration 036, so filtering only matches default-locale term slugs) is confirmed real, pre-existing, and out of scope here — filed separately as #1480.

Closes #1479

## Type of change

- [x] Bug fix

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] User-visible strings in the admin UI are wrapped for translation (if applicable). — n/a, no admin UI strings; this is a server-side query path

## Screenshots / test output

New test `loader-taxonomy-filter.test.ts` (written failing-first, TDD): single-taxonomy filter, AND across two taxonomies (#1479), OR-within / AND-across composition, and empty-array short-circuit.

```
Test Files  1 passed (1)
     Tests  4 passed (4)
```

Regression sweep (loader byline/field/cursor + i18n): `79 passed (79)`. Core typecheck clean; repo lint diagnostics unchanged (8 pre-existing, none in touched files).


ascorbic (approved):
Thanks!

PR #1492, opened 2026-06-15
feat(cloudflare): Durable Objects SQL database backend with read replication

ascorbic:
## What does this PR do?

Adds a first-class **Durable Objects SQL** database backend for EmDash on Cloudflare, as an alternative to D1, with experimental read replication. The goal is to cut read round-trip latency by serving reads from a Durable Object replica near the visitor instead of a single-region D1 primary.

A new `durableObjects()` config adapter stores the whole CMS in one Durable Object's SQLite (`EmDashDB`). With `session: "auto"` plus the `experimental` + `replica_routing` compatibility flags:

- **Reads** route to the nearest replica (`namespace.get(id)`).
- **Writes** are proxied to the primary by the DO itself (replicas are read-only).
- **Read-your-writes** is provided by the bookmarks API: a write records its bookmark, and later reads in the same request (and, for authenticated users, subsequent requests via a cookie) wait for the replica to catch up before serving.

Key implementation notes:

- **One DO stub per request.** The request-scoped path resolves a single stub and reuses it for every query in the request. (D1's per-query binding overhead is the thing we're trying to beat; the driver must never cache a stub across requests, since a DO stub is a per-request I/O object.)
- **Transactions match D1.** The driver rejects `beginTransaction`, so EmDash's existing `withTransaction` helper degrades to direct execution — the same atomicity profile the codebase already runs under on D1.
- **DO replication APIs are feature-detected** (they aren't in `workers-types` yet), so the class also works as a plain single-DO database when the flag is off.

Also adds two deployed perf fixtures for head-to-head measurement against the existing D1 baseline:

- `infra/do-demo` — DO with read replicas (`do-demo.emdashcms.com`), registered in perf-monitor.
- `infra/do-solo-demo` — DO single primary, no `replica_routing` — isolates the DO/RPC-architecture cost from the replica-routing win. Deployed but not yet registered in perf-monitor (pending content seed).

The measured perf-monitor Single Post slug was repointed to `notes-on-simplicity` (present in the repo seed) after reseeding the blog/cache fixtures, so all sites are comparable.

This is a maintainer-directed prototype (no prior Discussion, per the maintainer's instruction to open the PR directly).

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (or targeted tests for my change) — `@emdash-cms/cloudflare` suite (174 tests) passes; new dialect tests added
- [ ] User-visible strings in the admin UI are wrapped for translation — n/a, no admin UI strings
- [x] I have added a [changeset](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#changesets) (if this PR changes a published package) — `@emdash-cms/cloudflare` minor
- [ ] New features link to an approved Discussion — n/a, maintainer-directed prototype

- [x] This PR includes AI-generated code — model/tool: claude-opus-4-8 (OpenCode)

## Screenshots / test output

Validated end-to-end on a local `EmDashDB`: setup runs all migrations, the seed applies, and content reads (homepage, single post, content API with bylines/revisions) all serve correctly. An adversarial review pass (review → fix → re-review) closed four critical/major read-your-writes bugs before this PR; remaining items are documented limitations (connection-scoped PRAGMAs, best-effort singleton-sink under concurrent maintenance writes) and a follow-up (a `vitest-pool-workers` harness to unit-test the DO class directly — a gap shared with the existing preview DO class).

> [!NOTE]
> Replica routing only activates on deployed workers — local `pnpm dev` (miniflare) rejects the `replica_routing` flag, so the do-demo fixture's `wrangler.jsonc` documents dropping it for local runs. The latency win shows up across perf-monitor's regional probes, not locally.

---

### Try this PR

**[Open a fresh playground →](https://feat-do-sql-driver-emdash-playground.emdash-cms.workers.dev)**

<sub>Tracks `feat/do-sql-driver`. Updated automatically when the playground redeploys.</sub>


ascorbic:
Addressed the review (commit 91c7ceb0):

- **MED — `waitForBookmark` stall:** now bounded by a 250ms `Promise.race` timeout in both `query()` and `batchQuery()`. A stale/cross-DO-id bookmark no longer makes every read pay the full wait budget for the cookie's 24h life; on timeout we serve a possibly-stale read (self-heals on the next fresh bookmark).
- **LOW — `commit()` not in `finally`:** moved both call sites (anon + main paths) into `finally` so a write-then-failed-render still persists the bookmark. Also closes the same gap on the D1 session path; added an `emdash` changeset.
- **LOW — PRAGMA bookmark dropped when `rowsWritten===0`:** PRAGMA statements now capture a bookmark regardless of row count.
- **LOW — `isReadStatement` upper-cases whole SQL:** now an anchored leading-keyword regex; no full-body allocation.
- **LOW — anonymous writes lose RYW:** respectfully disputing — the `commit()` `!isAuthenticated` guard is **identical to the D1 adapter** (`d1.ts`), so this isn't a divergence from D1. Changing it is a cross-cutting decision for both adapters and out of scope for this PR; happy to take it up separately if we want anon-write RYW across the board.

PR #1438, opened 2026-06-12
feat: responsive srcset for media via astro:assets

ascorbic:
## What does this PR do?

Makes EmDash media render responsive `srcset`s automatically. The `Image` and Portable Text image components now route locally/R2-stored media through Astro's configured image service (`astro:assets`) — the Cloudflare Images binding on Workers, sharp on Node — so they emit width-appropriate candidates and modern formats (e.g. WebP) instead of a single full-size ``.

Astro's image services only optimize **absolute** URLs whose host is authorized via `image.remotePatterns`; relative same-origin proxy URLs are passed through unchanged. So this PR:

- Resolves media URLs to absolute (via the public origin) before handing them to `getImage`.
- Auto-registers `image.remotePatterns` for: the storage `publicUrl` host (R2 custom domain, S3/CDN); the site origin scoped to the media route (`/_emdash/api/media/file/**`) when `siteUrl` is set; and a **dev-only**, host-agnostic media-route pattern so `astro dev` works with no configuration (the dev origin/port isn't known at build time). The dev pattern is gated to `command === "dev"` and never ships in a production build.
- Detects passthrough (the service returned the URL unchanged) and falls back to a plain ``, so an unauthorized host / missing image service / unknown dimensions never produces a useless same-URL srcset.

No template changes are required, and existing sites render exactly as before when optimization isn't possible.

Closes #

## Type of change

- [ ] Bug fix
- [x] Feature (requires [maintainer-approved Discussion](https://github.com/emdash-cms/emdash/discussions/categories/ideas))
- [ ] Refactor (no behavior change)
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests
- [ ] Chore (dependencies, CI, tooling)

> Note: opened at the maintainer's request without a prior Discussion.

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md)
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (or targeted tests for my change)
- [x] `pnpm format` has been run
- [x] I have added/updated tests for my changes (if applicable)
- [ ] User-visible strings in the admin UI are wrapped for translation (if applicable) — n/a, no admin UI strings
- [x] I have added a [changeset](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#changesets) (if this PR changes a published package)
[...]

PR #1494, opened 2026-06-15
feat(media): binding-based image transforms for same-origin media (Access-safe)

scottbuscemi:
## What does this PR do?

Adds a binding-based image transform service so responsive images work for same-origin media behind Cloudflare Access (and anywhere loopback fetches are restricted, e.g. `global_fetch_strictly_public`).

**Background.** The responsive-srcset feature (#1438) hands an **absolute** media URL to Astro's image service, which makes Astro's `/_image` endpoint `fetch()` that URL to load the source bytes before transforming. For same-origin R2/local media that absolute URL is the Worker's **own** origin, so the load is a self-referential subrequest. When the site is behind Cloudflare Access, that subrequest carries no Access cookie → Access returns a login redirect instead of image bytes → `/_image` returns **404**. It "works locally" only because `astro dev` fetches a *different* (public) origin for the source.

**Fix.** Serve transforms from an EmDash-owned route, `GET /_emdash/api/media/transform/{key}`, that reads the source bytes straight from the storage adapter (the R2 binding) and resizes them with a configured transformer (the Cloudflare `IMAGES` binding). The browser requests this route directly — with its own Access cookie — and the Worker never re-fetches the media URL, so there is no loopback subrequest.

The `Image` / Portable Text image components prefer the transform route for **same-origin** media when a transformer is configured, and fall back to the existing `astro:assets` path otherwise (external CDN/`publicUrl` media, Node + sharp, or no transformer). Behavior is unchanged when `images` is not configured.

Opt-in wiring on Cloudflare:

```ts
import { imageBinding } from "@emdash-cms/cloudflare";

emdash({
  storage: r2({ binding: "MEDIA" }),
  images: imageBinding({ binding: "IMAGES" }), // requires { "images": { "binding": "IMAGES" } } in wrangler
});
```

### Implementation
- **core**: `ImageServiceDescriptor` + `virtual:emdash/images` (mirrors the storage descriptor pattern so core stays portable); transform route handler; portable URL/srcset builders + param validation in `media/image-transform.ts`; the transformer is built in middleware and exposed as `locals.emdash.transformImage`.
- **cloudflare**: `imageBinding()` config helper + `media/transform-runtime.ts` that uses the `IMAGES` binding.

Closes # 

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

> Additive and opt-in (no behavior change unless `images` is configured). Opened at a maintainer's direction; prior Discussion waived.

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] User-visible strings in the admin UI are wrapped for translation — n/a, no admin UI strings
- [x] New features link to an approved Discussion — n/a, Discussion waived by maintainer

## Screenshots / test output

```
Test Files  4 passed (4)
     Tests  57 passed (57)   # media/image-transform, astro/routes, middleware-prerender, media/responsive
```
New tests: `tests/unit/media/image-transform.test.ts` (URL/srcset builders, key safety, param validation) and transform-route handler + injection coverage in `tests/unit/astro/routes.test.ts`.

---

### Try this PR

**[Open a fresh playground →](https://fix-image-transform-binding-access-emdash-playground.emdash-cms.workers.dev)**

<sub>Tracks `fix/image-transform-binding-access`. Updated automatically when the playground redeploys.</sub>

### DIFF
--- a/.changeset/image-transform-binding.md
+++ b/.changeset/image-transform-binding.md
@@ -13,8 +13,8 @@ To enable it on Cloudflare, add an `IMAGES` binding to your wrangler config and 
 import { imageBinding } from "@emdash-cms/cloudflare";
 
 emdash({
-  storage: r2({ binding: "MEDIA" }),
-  images: imageBinding({ binding: "IMAGES" }),
+	storage: r2({ binding: "MEDIA" }),
+	images: imageBinding({ binding: "IMAGES" }),
 });
 ```
 
--- a/.changeset/image-transform-binding.md
+++ b/.changeset/image-transform-binding.md
@@ -1,0 +1,21 @@
+---
+"emdash": minor
+"@emdash-cms/cloudflare": minor
+---
+
+Adds a binding-based image transform service so responsive images work behind Cloudflare Access.
+
+Previously, resizing R2/local media routed through Astro's `/_image` endpoint, which made the server fetch the media's own URL to load the source bytes — a self-referential request that fails when the site is behind Cloudflare Access or has loopback fetches disabled, surfacing as 404s on transformed images. EmDash now serves transforms from `/_emdash/api/media/transform/{key}`, reading bytes straight from the storage adapter and resizing them with a configured transformer, so no server-side fetch of the media URL is made.
+
+To enable it on Cloudflare, add an `IMAGES` binding to your wrangler config and wire it up:
+
+```ts
+import { imageBinding } from "@emdash-cms/cloudflare";
+
+emdash({
+  storage: r2({ binding: "MEDIA" }),
+  images: imageBinding({ binding: "IMAGES" }),
+});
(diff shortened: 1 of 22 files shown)


ascorbic (changes requested):
This feels like it should be being handled a lot more narrowly. Right now it's basically reproducing the whole Astro image pipeline and Cloudflare Astro image service just to switch it so it loads from R2 rather than fetch. It gives us two services running next to each other, each with their own implementation. A better approach could be to implement an Astro image service that works like the existing Astro cloudflare-binding service but reads from R2 rather than fetch.

ascorbic:
Thanks. Closing this as superseded by #1549

PR #1505, opened 2026-06-16
fix(admin): use the configured Site Icon as the admin favicon

swissky:
## What does this PR do?

The admin shell (`admin.astro`) sets its favicon only from the build-time `admin.favicon` config, falling back to the bundled EmDash mark. The **Site Icon** users configure in **Settings → General** (the `favicon` site setting) was ignored by the backend — even though the public site already uses it via `EmDashHead` (`renderSiteIdentity({ favicon: siteSettings.favicon })`).

So setting a Site Icon brands the public site but **not** the admin — unlike WordPress, where the Site Icon also brands `wp-admin`.

This loads the site settings at request time and prefers the configured Site Icon for the admin `<link rel="icon">`:

**Site Icon (Settings → General) → build-time `admin.favicon` → bundled EmDash mark**

```ts
const settings = await getSiteSettingsWithDb(emdash.db, emdash.storage ?? null);
const favicon = settings.favicon?.url ?? adminConfig?.favicon; // then <link rel="icon" href={favicon}>
```

No behaviour change when no Site Icon is set (the existing config/default fallback still applies). Wrapped in try/catch so a pre-setup or settings error can't break the admin shell.

Closes # 

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes (`emdash` package — ran clean)
- [ ] `pnpm lint` passes (left to CI)
- [ ] `pnpm test` passes — n/a: change is in an Astro route `<head>` template (no unit harness); behaviour verified manually
- [ ] `pnpm format` has been run (left to CI)
- [ ] I have added/updated tests for my changes — see above (Astro template head)
- [x] User-visible strings in the admin UI are wrapped for translation (if applicable) — n/a, no strings added

- [x] This PR includes AI-generated code — model/tool: Cursor + Claude Opus 4.8

## Screenshots / test output

Single-file change in `packages/core/src/astro/routes/admin.astro` (+ changeset). Reuses the existing `getSiteSettingsWithDb` loader (same import the sitemap route uses). `pnpm --filter emdash typecheck` is clean.


packages/core/src/astro/routes/admin.astro
```diff
@@ -14,13 +14,30 @@ import { Font } from "astro:assets";
...
+let siteFavicon: string | undefined;
+if (emdash?.db) {
+	try {
+		const settings = await getSiteSettingsWithDb(emdash.db, emdash.storage ?? null);
+		siteFavicon = settings.favicon?.url ?? undefined;
+	} catch {
+		// Settings unavailable (e.g. pre-setup) — fall back to the config/default.
+	}
+}
+const favicon = siteFavicon ?? adminConfig?.favicon;
```
ascorbic: Shouldn't it default to the admin favicon? That's more specific than the site one.

swissky: Good call — agreed. An explicitly configured `admin.favicon` is admin-specific and shouldn't be silently overridden by the more general Site Icon. I've flipped the precedence to `adminConfig?.favicon ?? siteFavicon` (admin.favicon → Site Icon → bundled mark). The Site Icon still brands the admin in the common case where no `admin.favicon` is set, which was the goal of the PR. Comment + changeset updated, and rebased on latest `main`.

packages/core/src/astro/routes/admin.astro
```diff
@@ -17,13 +17,31 @@ import { Font } from "astro:assets";
...
+let siteFavicon: string | undefined;
+if (emdash?.db) {
+	try {
+		const settings = await getSiteSettingsWithDb(emdash.db, emdash.storage ?? null);
+		siteFavicon = settings.favicon?.url ?? undefined;
+	} catch {
+		// Settings unavailable (e.g. pre-setup) — fall back to the config/default.
+	}
+}
+const favicon = adminConfig?.favicon ?? siteFavicon;
```
ascorbic, line 55: As we're defaulting to `adminConfig?.favicon` we could skip getting `siteFavicon` if it's present

swissky: Done — the settings read is now skipped entirely when an explicit `admin.favicon` is configured (013c8ad), so the common self-hosted case with a build-time favicon no longer pays the per-request settings lookup.

packages/core/src/astro/routes/admin.astro
```diff
@@ -42,8 +60,8 @@ Astro.response.headers.set("Cache-Control", "private, no-store");
 		<meta name="viewport" content="width=device-width, initial-scale=1.0" />
 		<link rel="stylesheet" href={adminStylesUrl} />
 		<Font cssVariable="--font-emdash" />
-		{adminConfig?.favicon ? (
-			<link rel="icon" href={adminConfig.favicon} />
+		{favicon ? (
+			<link rel="icon" href={favicon} />
```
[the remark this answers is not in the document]
swissky: Fixed in 6eb89fe1 — the `<link>` now carries `type={faviconType}`, sourced from `settings.favicon.contentType` and only set when the Site Icon drives the favicon; a build-time `admin.favicon` string keeps its current attribute-less output (Astro omits the attribute for `undefined`).

packages/core/src/astro/routes/admin.astro
```diff
@@ -33,6 +34,23 @@ const pageTitle = adminConfig?.siteName ? `${adminConfig.siteName} Admin` : "EmD
...
+// Favicon precedence: an explicit build-time `admin.favicon` is admin-specific
+// and always wins — in that case the settings read is skipped entirely.
+// Otherwise fall back to the Site Icon configured in Settings → General, so
+// the admin shell is branded like the public site (WordPress-style Site Icon,
+// which also brands wp-admin), and finally the bundled EmDash mark below.
+const emdash = Astro.locals.emdash;
+let siteFavicon: string | undefined;
+if (!adminConfig?.favicon && emdash?.db) {
+	try {
+		const settings = await getSiteSettingsWithDb(emdash.db, emdash.storage ?? null);
```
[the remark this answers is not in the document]
swissky: Added in 6eb89fe1 — new e2e case in `admin-fixes.spec.ts`: uploads an SVG via the media API, sets it as the Site Icon through `POST /_emdash/api/settings`, then loads `/_emdash/admin` and asserts `link[rel="icon"]` points at the media-file URL with `type="image/svg+xml"`. Verified green locally.

ascorbic (approved):
Thanks!

PR #1444, opened 2026-06-12
feat(core): RelationRepository (content-references data layer)

MA2153:
## What does this PR do?

Adds `RelationRepository`, the application data-access layer over the content-references database schema that landed in #1367 (migration `043`). It is the next incremental slice in the content-references line of work — schema-only previously, now a typed, tested repository on top.

The class deliberately mirrors `TaxonomyRepository` and owns both halves of the feature, exactly as taxonomy owns terms + `content_taxonomies`:

- **Relation definitions** (`_emdash_relations`, row-per-locale, `translation_group`): `create` (with `translationOf` that inherits structural fields and joins the group), `findById`, `findByName`, `findTranslations`, `list`, `findForCollection`, `update` (labels only), `delete` (purges edges when the last translation of a group is removed).
- **Directed reference edges** (`_emdash_content_references`, locale-agnostic, keyed by `translation_group`): `addReference` (append/dedupe), `removeReference`, `getChildren` (forward), `getParents` (backlink), `setChildren` (replace + positional `sort_order`), `clearReferencesForGroup` (entry-deletion cleanup primitive), and `countChildren` / `countParents` / `countChildrenForParents` (batched at `SQL_BATCH_SIZE`).

**Scope (deliberately thin):** this is a data layer only. There is **no Zod, no request validation, and no API/route** — those are the next slice. The repository trusts its typed inputs; collection-agreement and relation-existence validation belong to the future handler/API layer. `clearReferencesForGroup` is provided as the cleanup primitive, but **wiring it into the content-delete path is also deferred** to that slice.

Backwards-compatible and additive — a new exported class only; no existing behavior changes.

Part of the content-references effort (Discussion #386; schema groundwork in #1367).

## Type of change

- [x] Feature (requires [maintainer-approved Discussion](https://github.com/emdash-cms/emdash/discussions/categories/ideas))

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md)
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes 
- [x] `pnpm test` passes (or targeted tests for my change) 
- [x] `pnpm format` has been run 
- [x] I have added/updated tests for my changes (if applicable)
- [ ] User-visible strings in the admin UI are wrapped for translation (n/a — no admin UI strings; pure data layer)
[...]

PR #1535, opened 2026-06-18
feat(core): content-references HTTP API (relation defs + entry edges)

MA2153:
## What does this PR do?

Adds the HTTP API layer over the content-references repository — the third bottom-up slice after the schema (migration `043`) and `RelationRepository`. Two surfaces:

- **Relation-definition CRUD** under `/_emdash/api/relations` — manage relationship *types* (editor-readable via `schema:read`, admin-writable via `schema:manage`), treated as schema-like structural config.
- **Content-entry reference edges** under `/_emdash/api/content/:collection/:id/references/:relation/{children,parents}` — attach/read directed references on a specific entry, ownership-aware like the taxonomy edge endpoint.

This is the trust boundary: Zod at the route, and the write-path invariants (relation-exists, collection-agreement, directed storage) the repository deliberately skipped. References are stored only in `_emdash_content_references`, keyed by `translation_group` — no `ec_*` column.

The final commit hardens the slice against a code review:

- Create-path failures map precisely instead of flattening to 500 — bad `translationOf` → `NOT_FOUND`, name/locale collisions → `CONFLICT`, missing collections → `COLLECTION_NOT_FOUND`.
- Edge reads resolve entries in one batched `IN` query (chunked) instead of N+1 per edge.
- Edge reads are cursor-paginated (`?cursor`/`?limit`, default 50, max 100) returning `nextCursor`; the backlink (parents) side is genuinely unbounded, so this matters. Bad cursors → `INVALID_CURSOR`.
- Each resolved entry ref carries the `locale` of the variant returned, so the variant fallback is explicit rather than a silent wrong-locale substitution.
- `content:edit_own` is gated before the existence lookup on `POST children` to close an existence-oracle.
- Empty `PATCH` relation bodies are rejected with `VALIDATION_ERROR`.

Closes #

## Type of change

Approved feature Discussion: https://github.com/emdash-cms/emdash/discussions/386

Part of the planned content-references effort tracked under the schema (#1367) and repository (#1444) slices.

## Checklist

- [ ] `pnpm typecheck` passes — repo-wide typecheck has 5 pre-existing failures in `packages/plugin-cli` / `src/plugins/*` from an unbuilt `@emdash-cms/plugin-types` (build-order, unrelated to this PR); **0 new** errors introduced by these changes.
- [ ] `pnpm lint` passes — `pnpm lint:json` is clean for all files in this PR; the only remaining diagnostics are pre-existing in unrelated demo/template `astro.config.mjs` files.
- [x] `pnpm test` passes (targeted): relation/reference handler, route, and repository suites — 60 tests green.
- [x] User-visible strings in the admin UI are wrapped for translation — n/a, no admin UI in this slice; server error messages are English-only and mapped to localized strings client-side by code.
- [x] I have added a [changeset](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#changesets) (`emdash`, minor)
- [x] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/386

## Screenshots / test output

```
Test Files  4 passed (4)
     Tests  60 passed (60)
```


MA2153:
Addressed in `0aa217d6`.

**Findings 1 & 2 (relation/reference error codes)** — Added a `Relations (content references)` block to the `ErrorCode` enum in `api/errors.ts` with all eight codes (`RELATION_LIST/CREATE/GET/UPDATE/DELETE/TRANSLATIONS_ERROR`, `REFERENCES_GET_ERROR`, `REFERENCES_SET_ERROR`). All end in `_ERROR`, so they keep the 500 default mapping — no `mapErrorStatus` change. Handlers keep their raw-string literals as you noted is acceptable; the union is now complete.

**Finding 4 (missing child-table guard)** — Confirmed reachable: `deleteCollection` drops the content table but doesn't cascade to relation defs. Wrapped the batch queries in `findManyByIdOrSlug` to return an empty map on `isMissingTableError`, mirroring `findTranslationsForGroups`. `POST …/references/:relation/children` against a dropped child collection now returns a structured `NOT_FOUND` instead of a 500. Added a failing-first integration test (`a child collection dropped after the relation is created yields NOT_FOUND, not a 500`).

**Finding 3 (OpenAPI docs)** — Deferred out of this PR's scope.

**Finding 5 (`requireDb` vs authz ordering)** — Leaving as-is. Both GET and POST already check the permission first, and the POST has an explicit comment justifying auth-before-lookup. Moving `requireDb` above the perm check would make an unauthenticated caller against an unconfigured DB get a 500 instead of a 403 — the current order is marginally better on info-leak, so the consistency change is net-negative here.

MA2153:
Addressed in `1af449ec`.

**Findings 1 & 2 (parent/child anchor collection dropped → 500)** — Hardened the shared `_findByIdOrSlug` to return `null` on `isMissingTableError`, matching its sibling methods `findManyByIdOrSlug` and `findTranslationsForGroups`. This covers all three anchor lookups (children GET, children SET, parents GET) and also the route-level ownership lookup in `children.ts`, which a handler-only guard would have missed. Added a reproducing test that drops `ec_post` (and `ec_page` for the parents side) and asserts `NOT_FOUND` rather than 500. Confirmed no existing test relied on the throw.

**Finding 4 (schema requires ignored fields for translations)** — `createRelationBody` now makes `parentCollection`/`childCollection` optional with a refine that requires them only when `translationOf` is absent. Threaded through `CreateRelationInput` (both now optional), with a narrowing guard in `handleRelationCreate` and an explicit else-branch in `RelationRepository.create` that throws if a base relation omits them (the columns are NOT NULL). Added repo tests for both the omit-on-translation success path and the missing-on-base failure path.

Note: `name` has the same characteristic — it's required by the schema but inherited from the source when `translationOf` is set. I left it as-is to keep this change scoped to the fields you flagged; happy to relax it too if you'd like consistency.

**Finding 3 (OpenAPI docs)** — Still deferred out of this PR's scope.

ascorbic (approved):
Thanks!

Issue #1462, opened 2026-06-14
Intermittent "EmDash is not initialized" (NOT_CONFIGURED 500 / 503) on Cloudflare Workers admin under load — per-request `emdash.db` missing on some isolates


swissky:
## Summary

On Cloudflare Workers (Astro adapter, D1), authenticated **admin** requests intermittently fail with:

```json
{"error":{"code":"NOT_CONFIGURED","message":"EmDash is not initialized"}}
```

returned as **500**, plus **503** on the admin content pages themselves (`/_emdash/admin/content/<collection>`). It affects **all collections at once** (clinics, events, pages, posts, products) — i.e. it is not data-specific. Public/front-end pages that read the same D1 are unaffected throughout.

The 500 originates from the `requireDb` guard:

```js
function requireDb(db) {
  if (!db) return apiError("NOT_CONFIGURED", "EmDash is not initialized", 500);
  return null;
}
// called as: const err = requireDb(emdash?.db); if (err) return err;
```

So for the failing requests, the **per-request `emdash` context has no `.db`** — the runtime/context isn't initialized for that request, even though D1 is healthy.

## Environment

- `emdash@0.19.0`, `@emdash-cms/cloudflare@0.19.0`
- `astro@6.4.4`, `@astrojs/cloudflare@13.6.1`
- Worker entry: `@astrojs/cloudflare/entrypoints/server` (the EmDash 0.19 worker wrapper was previously rolled back here due to an unrelated admin SSR hang)
- DB: D1 via `d1({ binding: 'DB' })` (reproduced both with and without the experimental `coalesce: true` / `session: 'auto'` options — removing them did **not** fix it)
- One in-process plugin registered via the integration (`format: 'standard'`)

## Symptoms / evidence

- Intermittent and **per-isolate**: the same endpoint returns 500 `NOT_CONFIGURED` on some requests and works on others. Reloading sometimes works, sometimes not (\"works in incognito, not in the normal browser\" was just hitting a warm vs. cold isolate).
- **Worst immediately after every `wrangler deploy`** (all isolates cold), then partially settles as isolates warm — strongly suggesting an init/cold-start race rather than bad config.
- `wrangler tail` shows the admin **page** request itself as `outcome: canceled, cpuTime: 0ms, wallTime: ~21000ms` in some cases (awaiting something that never resolves), and fast `500` on the `/authors` API in others.
- D1 is healthy the whole time (`SELECT 1` returns immediately; front-end collection reads work).
[...]

Issue #733, opened 2026-04-23
Avoid Astro session KV reads for anonymous public requests without session cookie


huckabarry:
### Description

On Cloudflare Workers, public SSR requests can trigger an Astro session lookup even when the request has no `astro-session` cookie. Because Astro's Cloudflare adapter stores sessions in Workers KV, this turns anonymous public traffic into KV read traffic / read misses.

In my deployed site this showed up as repeated Cloudflare Workers KV free-tier warnings. The active generated session namespace had ~159k KV operations over Apr 21-23 UTC, almost entirely reads, while listing keys returned an empty namespace.

```text
afterword-emdash-lab-session
159,078 total KV operations
159,074 reads
4 writes
keys: []
```

This appears to come from the EmDash middleware reading the session user before it knows whether the request actually has an Astro session cookie:

```ts
const sessionUser = context.isPrerendered ? null : await context.session?.get("user");
```

That means anonymous public requests still call into the configured session backend. On Cloudflare, that backend is KV via `@astrojs/cloudflare`'s generated `SESSION` binding.

### Expected behavior

Anonymous public SSR requests without an `astro-session` cookie should not read from session storage.

EmDash should still read the session when needed for admin routes, edit mode, preview flows, or requests that actually carry the session cookie.

### Actual behavior

Anonymous public SSR requests read from Astro session storage, producing Cloudflare Workers KV read misses.

### Local patch that resolved it

This local patch stopped the KV read volume on anonymous public traffic while preserving logged-in/editor behavior:

```ts
const hasSessionCookie = cookies.get("astro-session") !== undefined;

const sessionUser =
  context.isPrerendered || !hasSessionCookie
    ? null
    : await context.session?.get("user");
```

For the built output the equivalent is:

```js
const hasSessionCookie = cookies.get("astro-session") !== void 0;
const sessionUser = context.isPrerendered || !hasSessionCookie ? null : await context.session?.get("user");
```

### Environment

- `emdash`: `0.6.0`
- `@emdash-cms/cloudflare`: `0.6.0`
- `astro`: `6.1.4`
- `@astrojs/cloudflare`: `13.1.7`
- Runtime: Cloudflare Workers
- Database: D1 with `session: "auto"`
- Astro sessions: Cloudflare KV `SESSION` binding generated by `@astrojs/cloudflare`

### Related context

[...]

PR #1538, opened 2026-06-18
fix(core): prevent isolate-wide hang from a stalled session.get on Workers (#1274)

### LINKED ISSUE #1274: Session KV lookup hangs indefinitely, blocking entire page response for logged-in users

swissky:
## What does this PR do?

Fixes logged-in pages (admin **and** public) hanging indefinitely on Cloudflare Workers (#1274).

A request cancelled mid-`session.get("user")` (client disconnect / context teardown) can leave the underlying session-store read as a promise that **never settles** — it neither resolves nor rejects. Awaiting it directly hangs the request, and because the stalled promise is shared at the isolate level, **every later session-bearing request on that isolate hangs too** — the exact 0-CPU / multi-minute-wall / `canceled` / no-child-spans signature reported in the issue. It's reliably reproducible on **fresh isolates right after a deploy**, and a surrounding `try/catch` can't help because the promise never rejects.

`handlePluginRouteAuth`, `handlePublicRouteAuth` and `handlePasskeyAuth` now resolve the session user through a shared `resolveSessionUser()` helper that:

1. **Anchors** the read with `after()` so a cancelled request still drives it to completion — the promise settles and the isolate is **not poisoned** for subsequent requests (prevents the hang rather than merely surviving it); and
2. Applies a **fail-closed timeout backstop** — a still-stalled read resolves to `undefined`, and every caller already treats the absence of a session user as unauthenticated (anonymous on public routes, `401`/redirect on protected ones). It can only ever **drop** privileges for that one request, never grant them.

This mirrors the reclaimable-cache / `after()` pattern already used in core for settings/secrets.

Closes #1274.

## Type of change

- [x] Bug fix

## Checklist

- [x] `pnpm typecheck` passes (`emdash` package, workspace deps built)
- [x] `pnpm lint` passes (`oxlint --type-aware --deny-warnings`)
- [x] `pnpm format` has been run (`oxfmt` + `prettier`)
- [ ] Tests: no new tests — the trigger is a cancellation/timing race in the workerd session layer that's hard to reproduce deterministically in a unit test; verified instead with live before/after Workers traces (below). Happy to add a test in whatever shape you prefer.
- [x] User-visible strings: none changed
- [x] Changeset added (`emdash` patch)

- [x] This PR includes AI-generated code — model/tool: Cursor + Claude Opus 4.8

## Verification — production, Cloudflare Workers, EmDash 0.21.0

Captured with `wrangler tail` (data anonymised — paths/timings/outcomes only).

**Before** — logged-in requests wedge:

| request | wall | cpu | outcome |
| --- | --- | --- | --- |
| `GET /_emdash/admin` | 7,551 ms | 0 ms | `canceled` |
| `GET /_emdash/admin/content/…` | 7,710 ms | 0 ms | `canceled` |

**After this change** — fresh post-deploy isolate, same conditions, **26/26** admin+API requests `ok`:

| request | wall | cpu | outcome |
| --- | --- | --- | --- |
| `GET /_emdash/admin` | 84 ms | 15 ms | `ok` |
| `GET /_emdash/api/dashboard` | 452 ms | 21 ms | `ok` |
| `GET /_emdash/api/content/…` | ~200 ms | 2–14 ms | `ok` |
| login · passkey verify · manifest · plugins · users · trash · authors | <500 ms | <100 ms | `ok` |

**0 `canceled`, 0 hangs.**


swissky:
Can i get my own Emdashbot 😅 he is great

swissky:
Great review — you're right, the guard was under-applied. Pushed a follow-up (8dd4180):

1. **`middleware.ts:419`** — agreed, this is the *first* session read and runs before the auth middleware, so the original fix wasn't actually covering the path it set out to. Now uses the shared helper, keeping the `!hasSessionCookie` short-circuit (the #733 KV-read-miss optimization).
2. **`snapshot.ts:34`** — same treatment, switched to the helper.
3. **Extraction + tests** — pulled `resolveSessionUser` into a shared `astro/session-user.ts` so all three sites import it, and added `tests/unit/astro/session-user.test.ts` for the two testable behaviors: a never-settling read resolves to `undefined` within the timeout instead of hanging, and a rejecting read fails closed to `undefined`.

Local checks pass: `oxlint --type-aware --deny-warnings`, `tsgo` typecheck, and the new tests (4/4).

ascorbic:
btw, this is emdashbot if you _do_ want one! https://github.com/emdash-cms/emdash/tree/main/infra/flue-review

swissky:
I took #1538 for a proper spin on throwaway, isolated infra to check whether it actually clears the admin hang we keep hitting on Cloudflare Workers — and wanted to share what came out, including one gap and one unrelated bug.

## Setup

- A throwaway Worker with its **own** D1 + KV (sessions) + R2 — nothing shared with prod.
- `emdash@1538` / `@emdash-cms/cloudflare@1538` from pkg.pr.new (reports as `0.21.0`), `@astrojs/cloudflare`, `compatibility_date 2026-06-10`, `nodejs_compat`.
- **Smart Placement `mode: smart`** — this matters, prod runs with it on. `cf-placement: local-ZRH` confirmed it was active for every request below.
- To force the bad case I redeployed repeatedly so requests kept landing on **cold** isolates — which, per #1462, is exactly when this bites ("worst immediately after every deploy").

## What the cold path looks like

A cold isolate spends real time in `getRuntime()` before `locals.emdash` (incl. `.db`) is attached:

```
server-timing: rt;dur≈500 "Runtime init", rt.db;dur≈466 "DB init + migrations"
```

Warm, the same header is `rt;dur=0`. When several authenticated admin requests hit the same cold isolate, they race that init, and I saw two failure modes:

1. **The #1274 hang.** `wrangler tail`: `outcome: canceled, cpuTime: 0ms, wallTime 100–680s`, on both public pages and admin/API routes. 34 such `canceled` requests in a single window.
2. **`NOT_CONFIGURED` 500.** `getRuntime()` throws transiently during the race, the middleware's catch leaves `locals.emdash` unset, and `requireDb(emdash?.db)` returns 500. Fast: `cpuTime 2–4ms, wallTime 20–350ms`. This is #1462.

The #1538 timeout helps with (1) — it stops the *unbounded* hang — but on a fully wedged cold isolate (`cpuTime: 0`, no JS scheduled) the `Promise.race` can't always fire, and when the request does get through it then lands on (2). So with **#1538 alone, under Smart Placement + cold isolates, the admin still doesn't come up** — the symptom just shifts from "hangs forever" to "fast 500 / NOT_CONFIGURED".

Sanity checks that line up with that reading:
- Smart Placement **off**, or once the isolate is **warm** (`rt;dur=0`) → admin loads fine. Prod mostly survives because steady traffic + a keep-warm cron keep isolates warm; the incident that made us disable placement was a cold burst.

## What actually cleared it

Wrapping the `getRuntime()` call in the middleware in a small retry/wait — i.e. *wait for init to finish* instead of failing the request — removed both symptoms:

```js
let runtime;
for (let attempt = 0; attempt < 4; attempt++) {
  try {
    runtime = await getRuntime(config, initSubTimings);
    if (runtime?.db) break;
  } catch (err) {
    if (attempt === 3) throw err;
  }
  await new Promise((r) => setTimeout(r, 75 * (attempt + 1)));
}
```

With this **plus** the #1538 session timeout, Smart Placement **on**, on a freshly deployed (cold) Worker:

- ~150-request concurrent burst across `/`, `/_emdash/admin/`, `/_emdash/api/{manifest,dashboard,auth/me}`: **30×200, 30×302, 90×401, 0×500, 0 canceled**, slowest 2.9s (the cold init — but it completes).
- Authenticated admin loaded cleanly straight after the cold deploy; reloading the content list repeatedly → all 200/302, no `NOT_CONFIGURED`.
- Full **logout → clear cookies → fresh login** cycle works (this was the exact thing that used to wedge).
- `cf-placement: local-ZRH` throughout, so this is genuinely with Smart Placement on, not by turning it off.

So my takeaway: **#1538 is necessary (it kills the unbounded hang) but on Workers it isn't sufficient on its own** — the per-request `emdash.db` cold-init race (#1462) needs the runtime init to be awaited/retried. This also matches the "Expected" in #1462 ("the request should wait for initialization to complete"). Happy to open a separate PR for the `getRuntime()` retry against #1462, or fold it in here — whatever you prefer.

## Separate bug: `/authors` 500

Independent of all the above: `GET /_emdash/api/content/<collection>/authors` returns 500 **deterministically**, even when `emdash.db` is healthy and the admin is otherwise fully usable (it only breaks the author-filter dropdown). Notes:

- Happens for collections **with** authors and **without** them (e.g. an empty `pages` collection), so it's not the NULL-author data case.
- The underlying `SELECT DISTINCT author_id ... WHERE deleted_at IS NULL AND author_id IS NOT NULL` runs fine by hand against the same D1.
- `console.error("Content authors error:", …)` never surfaced in `wrangler tail`, so I couldn't capture the actual exception remotely.

Flagging it here because it shows up in the same traces and might be worth a look alongside #1462. Glad to provide tail JSON or test against a patch.

ascorbic:
Great work. Let's get this one in, then follow up with the `getRuntime` fix.

PR #1542, opened 2026-06-19
fix(plugins): hoist id/version regexes into defineNativePlugin to avoid TDZ on Workers (#1370)

### LINKED ISSUE #1370: definePlugin TDZ: "Cannot access 'SIMPLE_ID' before initialization" — every route 500s on Cloudflare Workers (0.17.1 & 0.17.2)
## Summary

On a Cloudflare Workers (D1 + R2) Astro SSR site that registers **native plugins** via `definePlugin`, upgrading `emdash` from **0.15.0 → 0.17.1 or 0.17.2** makes **every route return HTTP 500** at runtime. `astro check` and `astro build` pass cleanly — the failure only appears at runtime under `workerd`:

```
ReferenceError: Cannot access 'SIMPLE_ID' before initialization
    at defineNativePlugin (dist/server/chunks/adapt-sandbox-entry_*.mjs:796:3)
    at definePlugin        (dist/server/chunks/adapt-sandbox-entry_*.mjs:791:10)
    at createPlugin        (dist/server/chunks/wait-until_*.mjs:758:10)
    at <module init>       (dist/server/chunks/wait-until_*.mjs:1014:3)
    at <module init>       (dist/server/chunks/adapt-sandbox-entry_*.mjs:770:5)
```

`SIMPLE_ID` is a `const` declared *later* in the same emitted chunk than a top-level `createPlugin` → `definePlugin` → `defineNativePlugin` call that reads it — i.e. a temporal-dead-zone / circular-init ordering problem in the bundled output.

## Environment

- emdash **0.17.2** (also reproduced on **0.17.1**; **0.15.0 is unaffected**)
- @emdash-cms/cloudflare 0.17.2, @emdash-cms/admin 0.17.2
- astro 6.4.4, @astrojs/cloudflare 13.6.1
- wrangler 4.98.0 / workerd, Node 24, pnpm 11
- `output: 'server'`, `emdash({ database: d1({binding:'DB'}), storage: r2({binding:'MEDIA'}) })`
- 3 **native** plugins registered via `definePlugin` (entrypoint module paths)

## Reproduction

1. `pnpm build`
2. `npx wrangler dev --config dist/server/wrangler.json`
3. `curl http://127.0.0.1:8799/` → **500**, log shows the trace above (every route).

The adapter build emits `no_bundle: true`, so a real `wrangler deploy` behaves identically.

## Notes / what was tried

- Reproduces with the **docs-minimal** Cloudflare config too (no custom `vite.optimizeDeps` / `ssr.noExternal`). The integration itself sets `ssr.noExternal: ["emdash", "@emdash-cms/admin"]`, so emdash is re-bundled into the worker regardless — which appears to be where Rollup flattens the circular init into a TDZ.
- 0.17.1 and 0.17.2 both affected; 0.15.0 works fine on the identical setup.

Is this a known issue, and is there a recommended config or a fix version? Happy to provide a minimal reproduction repo if useful.

marcusbellamyshaw-cell:
## What does this PR do?

Fixes a crash that returns **HTTP 500 on every route** on Cloudflare Workers (`workerd`) when a site registers native plugins via `definePlugin`.

The plugin id/version validation regexes (`SIMPLE_ID`, `SCOPED_ID`, `SEMVER_PATTERN`) were module-scope `const`s read inside `defineNativePlugin`. Under the integration's `ssr.noExternal: ["emdash", …]`, Rollup re-bundles emdash into the worker and flattens a circular import between the worker entry and this chunk. When a top-level `definePlugin(...)` (a native plugin's default export, mounted via `createPlugin`) runs at module-init time mid-cycle, it reaches `defineNativePlugin` before those `const`s have initialized — and `const` bindings throw inside their temporal dead zone:

```
ReferenceError: Cannot access 'SIMPLE_ID' before initialization
    at defineNativePlugin
    at definePlugin
    at createPlugin
    at <module init>
```

`astro build` / `astro check` pass because it's purely a bundled init-order issue; 0.15.0 was unaffected only because its bundle ordering happened to differ.

Moving the three regexes into the function body makes them evaluate at call time (after the literals are parsed), so the TDZ cannot occur regardless of how the bundler orders the circular init. Validation behavior is byte-identical — they're used only inside `defineNativePlugin`. The `e18e/prefer-static-regex` lint is suppressed on these three lines with a justification: module scope is exactly what triggers the bug, and the per-call compile cost is negligible (once per plugin registration). A leaf-module or `var` alternative would not survive Rollup's chunk flattening — only call-time evaluation is robust.

Closes #1370

## Type of change

- [x] Bug fix

## Checklist

- [ ] `pnpm typecheck` passes — *not run cleanly locally: branching off `main` left the installed workspace deps (`@emdash-cms/plugin-types`) out of sync, producing unrelated pre-existing errors in `manifest-schema.ts` / `types.ts`. This change is type-neutral and adds no diagnostics in the touched file.*
- [x] `pnpm lint` passes (oxlint, 0 errors repo-wide)
- [x] `pnpm test` passes (targeted: `define-plugin.test.ts` — 40 passed)
- [x] `pnpm format` has been run (oxfmt clean)
- [x] I have added/updated tests for my change
- [ ] User-visible strings wrapped for translation — n/a (no admin UI strings)

- [x] This PR includes AI-generated code — model/tool: **Claude Opus 4.8 (Claude Code)**

## Screenshots / test output

The literal TDZ depends on bundler output, so it isn't reproducible in a plain vitest. The existing suite already covers the validation semantics; this PR adds an issue-anchored regression test, and the manual repro below covers the runtime crash.

Manual repro (Cloudflare adapter site that registers a native plugin):

```
pnpm build
npx wrangler dev --config dist/server/wrangler.json
curl -s -o /dev/null -w "%{http_code}\n" http://127.0.0.1:8799/   # 500 before, 200 after
```

Unit test (`packages/core/tests/unit/plugins/define-plugin.test.ts`):

```
 Test Files  1 passed (1)
      Tests  40 passed (40)
```

### DIFF
--- a/.changeset/fix-define-plugin-tdz.md
+++ b/.changeset/fix-define-plugin-tdz.md
@@ -2,4 +2,4 @@
 "emdash": patch
 ---
 
-Fixes a crash that returned HTTP 500 on every route on Cloudflare Workers when native plugins are registered with `definePlugin`. The plugin id/version validation regexes were module-scoped and could be read before initialization during the worker's circular module init (`Cannot access 'SIMPLE_ID' before initialization`). They are now function-local, so plugin registration is safe regardless of bundle ordering.
+Fixes a crash that returned HTTP 500 on every route on Cloudflare Workers when native plugins are registered with `definePlugin` (`Cannot access 'SIMPLE_ID' before initialization`). Plugin registration is now safe regardless of bundle ordering.
--- a/.changeset/fix-define-plugin-tdz.md
+++ b/.changeset/fix-define-plugin-tdz.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes a crash that returned HTTP 500 on every route on Cloudflare Workers when native plugins are registered with `definePlugin`. The plugin id/version validation regexes were module-scoped and could be read before initialization during the worker's circular module init (`Cannot access 'SIMPLE_ID' before initialization`). They are now function-local, so plugin registration is safe regardless of bundle ordering.
--- a/packages/core/src/plugins/define-plugin.ts
+++ b/packages/core/src/plugins/define-plugin.ts
@@ -22,11 +22,6 @@ import type {
 	PluginStorageConfig,
 } from "./types.js";
 
-// Plugin ID validation patterns
-const SIMPLE_ID = /^[a-z0-9-]+$/;
-const SCOPED_ID = /^@[a-z0-9-]+\/[a-z0-9-]+$/;
-const SEMVER_PATTERN = /^\d+\.\d+\.\d+/;
-
 /**
  * Define a native EmDash plugin.
  *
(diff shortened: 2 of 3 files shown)


ascorbic (approved):
Thanks!

Issue #1282, opened 2026-06-02
Expose an outer user middleware hook around EmDash initialization and response mutation


sam-fakhreddine:
## Summary

EmDash's integration middleware runs with Astro's `'pre'` order, which means it executes **before** user middleware in `src/middleware.ts`. This makes it impossible to short-circuit the D1 initialization from user-level code, even when serving a full cached HTML response.

## Observed behaviour

Every page render runs ~34 D1 queries and takes ~400ms server-side regardless of whether user middleware returns early. `server-timing` confirms EmDash's DB work happens before user middleware can intercept:

```
server-timing: render;dur=435, mw;dur=435, db.total;dur=435, db.count;dur=34
```

Attempting KV-based HTML caching in `src/middleware.ts`:

```ts
export const onRequest = defineMiddleware(async (context, next) => {
  const cached = await kv.get(`html::${pathname}`);
  if (typeof cached === 'string' && cached.length > 0) {
    return new Response(cached, { status: 200, headers }); // early return
  }
  const response = await next();
  await kv.put(`html::${pathname}`, await response.text(), { expirationTtl: 300 });
  // ...
});
```

The early return skips page rendering, but EmDash's pre-middleware has already run all 34 D1 queries. TTFB on a KV HIT equals TTFB on a MISS.

## Environment

- EmDash: 0.15.0  
- Astro: 6.4.2  
- `@astrojs/cloudflare`: 13.6.0  
- Cloudflare Pages (D1 binding)

## What would help

**Option A:** Expose a way for users to register middleware at `'pre'` order (before EmDash init), so a caching layer can short-circuit the entire stack.

**Option B:** EmDash implements its own page-level HTML cache (KV-backed) with a configurable TTL — keeps caching internal where it can run before D1 init.

The 34-query / 400ms baseline is reasonable for a cold CMS render. The issue is there is no provided path to avoid it on subsequent requests.

swissky:
Adding context from working on the caching side of this. The middleware-order point is confirmed and still current on `main` — every EmDash middleware (runtime init, redirect, setup, auth, request-context) registers with `order: "pre"` in `packages/core/src/astro/integration/index.ts`, so user middleware can never run before D1 init. Astro also gives integrations no "post-user" slot they could opt into per-request, so any fix here is an EmDash API decision (e.g. a config flag that lets a user-supplied gate run before runtime init).

[...]

PR #1558, opened 2026-06-20
fix(core): apply baseline security headers set-if-absent so host-set values win (#1393)

### LINKED ISSUE #1393: Middleware overwrites host-app security headers on public (non-/_emdash) routes
**EmDash version:** 0.15.0

`finalizeResponse()` in `packages/core/src/astro/middleware.ts` unconditionally sets these headers on **every** response — including anonymous public routes (the `runAnon` path), not just `/_emdash`:

- `X-Content-Type-Options: nosniff`
- `Referrer-Policy: strict-origin-when-cross-origin`
- `Permissions-Policy: camera=(), microphone=(), geolocation=(), payment=()`

Because the middleware registers with `order: 'pre'` (cf. #1282), on the response path it runs *after* the host app's own middleware — so it **overwrites** any of these headers the host already set. CSP is handled defensively (`if (!res.headers.has('Content-Security-Policy'))`), but `Permissions-Policy` / `Referrer-Policy` / `X-Content-Type-Options` are not, and there's no config flag to disable or customize them.

**Impact:** the CMS dictates the security headers of the entire host site. In our case the host sets a stricter `Permissions-Policy` (additionally disabling `interest-cohort`, `browsing-topics`, `attribution-reporting`, `private-state-token-*`); EmDash silently replaces it with its shorter value on every public marketing page. A host can't tighten these on its own public routes.

**Repro:** integrate EmDash, set `Permissions-Policy` in host middleware, request any public route, observe EmDash's value on the wire.

**Proposed fix (any one):**
1. Scope `finalizeResponse`'s baseline headers to `/_emdash` routes (the surface EmDash owns);
2. Set-if-absent (`if (!res.headers.has(name))`), matching the existing CSP pattern, so the host wins; or
3. A config option to disable EmDash's header injection.

marcusbellamyshaw-cell:
## What does this PR do?

`finalizeResponse()` unconditionally set `X-Content-Type-Options`, `Referrer-Policy`, and `Permissions-Policy` on every response. Because the middleware registers with `order: 'pre'` (#1282), on the response path it runs *after* the host app's own middleware, so it overwrote any of these headers the host had already set on its public routes — letting the CMS dictate the security headers of the entire host site (the reporter's host sets a stricter `Permissions-Policy`).

These three headers are now applied set-if-absent (`if (!res.headers.has(name))`), matching the existing `Content-Security-Policy` guard in the same function, so a host that sets a stricter value on its own routes wins. EmDash still provides its baseline defaults when the host sets nothing.

Closes #1393

## Type of change

- [x] Bug fix

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (targeted: new + existing astro middleware suites)

### DIFF
--- a/.changeset/fix-1393-security-headers-set-if-absent.md
+++ b/.changeset/fix-1393-security-headers-set-if-absent.md
@@ -2,8 +2,4 @@
 "emdash": patch
 ---
 
-fix(core): apply baseline security headers set-if-absent so host-set values win (#1393)
-
-`finalizeResponse()` unconditionally set `X-Content-Type-Options`, `Referrer-Policy`, and `Permissions-Policy` on every response. Because the middleware registers with `order: 'pre'` (#1282), on the response path it runs after the host app's own middleware, so it overwrote any of these headers the host had already set on its public routes — letting the CMS dictate the security headers of the entire host site.
-
-These three headers are now applied set-if-absent (`if (!res.headers.has(name))`), matching the existing `Content-Security-Policy` guard, so a host that sets a stricter value (e.g. an extended `Permissions-Policy`) on its own routes wins. EmDash still provides its baseline defaults when the host sets nothing.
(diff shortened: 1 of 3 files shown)


ascorbic (approved):
Thanks

PR #1560, opened 2026-06-20
fix(core): fail fast with an actionable error for plugins without an entrypoint (#1416)

### LINKED ISSUE #1416: virtual:emdash/plugins generates 'import undefined' for in-process plugins under env-replacement scenarios
## Summary

EmDash's `generatePluginsModule` in `dist/astro/index.mjs` emits `import pluginDefN from "undefined";` for in-process plugins (those returned by `definePlugin({...})` rather than a file path) when the build runs under certain env-replacement scenarios. Rollup then fails with:

```
[vite]: Rollup failed to resolve import "undefined" from " virtual:emdash/plugins".
```

The same build succeeds when run directly from a shell that loads `.env` through Vite's dotenv loader. It fails when an external process (e.g., Playwright's `webServer.env`) replaces `process.env` wholesale before invoking `pnpm build`.

## Reproduction

Project: Astro 6 SSR with `@astrojs/node@^10` standalone adapter, EmDash 0.13.0 exact-pinned.

```js
// astro.config.mjs (excerpt)
import emdash from 'emdash/astro';
import { resendEmailPlugin } from './src/lib/my-in-process-plugin.ts';

export default defineConfig({
  integrations: [
    emdash({
      database: sqlite({ url: `file:${process.env.EMDASH_DB_PATH}` }),
      plugins: [resendEmailPlugin()].filter((p) => p !== null),
    }),
  ],
});
```

```ts
// src/lib/my-in-process-plugin.ts (excerpt)
import { definePlugin } from 'emdash';

export function resendEmailPlugin() {
  if (!isUsable(process.env.RESEND_API_KEY)) return null;
  return definePlugin({
    id: 'misarico-resend-email',
    version: '0.0.0',
    capabilities: ['hooks.email-transport:register'],
    hooks: { 'email:deliver': { exclusive: true, async handler(event) { /* ... */ } } },
  });
}
```

**Works:** Direct shell invocation where `.env` is loaded by Vite at config time → `RESEND_API_KEY` is a real key → plugin returns a real definition → build succeeds.

**Fails:** Any invocation where `process.env` is replaced before `pnpm build` runs, with `RESEND_API_KEY` set to a non-placeholder value. Example: Playwright's `webServer.env` block. The Vite dotenv loader still runs but doesn't override the parent-provided env, so the plugin sees the env-block value and returns a real definition. EmDash then attempts to generate the import statement and fails.

## Root cause (from reading `dist/astro/index.mjs`)

`generatePluginsModule(descriptors)` iterates plugin descriptors and emits import statements:

```js
descriptors.forEach((descriptor, index) => {
  if (descriptor.format === 'standard') {
    needsAdapter = true;
    const varName = `pluginDef${index}`;
    imports.push(`import ${varName} from "${descriptor.entrypoint}";`);  // <-- entrypoint may be undefined
    // ...
  } else {
    const varName = `createPlugin${index}`;
    imports.push(`import { createPlugin as ${varName} } from "${descriptor.entrypoint}";`);
    // ...
  }
});
```

Both branches reference `descriptor.entrypoint`. For in-process plugins (defined inline via `definePlugin(...)` and passed directly to the integration's `plugins:` array), `entrypoint` is `undefined`, producing `import pluginDef0 from "undefined";`.

In-process plugins are documented and supported (see README's `plugins: [trustedPlugin()]` example), but the virtual-module generator assumes every plugin has a file entrypoint.

## Why it sometimes works

When `pnpm build` is invoked with the user's normal shell env + Vite's dotenv loader runs, some combination of vars makes the plugin loader take a different code path. We could not isolate the exact triggering var (tried matching shell PATH, NODE_OPTIONS, PNPM_HOME individually — none reproduced). The reliable difference is: shell-invoked build with `.env` loaded → succeeds; env-replaced subprocess invoking the same build → fails.

## Suggested fix

Branch on plugin format/source before emitting the import:

```js
descriptors.forEach((descriptor, index) => {
  if (descriptor.format === 'inline' || !descriptor.entrypoint) {
    // In-process plugin already in memory; reference the runtime
    // plugin registry rather than emitting an import.
    const varName = `pluginDef${index}`;
    instantiations.push(`runtimePlugins[${index}]`);  // or whatever the registry shape is
    return;
  }
  // Existing standard-format branch (unchanged)
  // ...
});
```

OR explicitly reject in-process plugins at config validation time with a clear error message ("EmDash's astro integration requires file-path plugins; in-process `definePlugin(...)` results are not supported in build mode").

## Workaround we shipped

In `playwright.config.ts`, skip the build step in the test webServer command and rely on a prior `pnpm build` from `pnpm check`:

```ts
{
  command: 'pnpm preview',  // not 'pnpm build && pnpm preview'
  // ...
}
```

Plus added `pnpm test:e2e:full = pnpm build && pnpm test:e2e` so fresh-clone developers aren't surprised.

## Environment

- EmDash: 0.13.0 (exact-pinned)
- Astro: ^6.3.5
- @astrojs/node: ^10
- @astrojs/react: ^5.0.5
- Node: 22 LTS
- pnpm: 9.15.9
- Playwright: ^1.x

Happy to share more context or a minimal reproduction repo if useful.

marcusbellamyshaw-cell:
## What does this PR do?

`generatePluginsModule` assumed every plugin descriptor had a file entrypoint and emitted `import pluginDefN from "${descriptor.entrypoint}";`. For an in-process plugin — an inline `definePlugin({...})` result passed directly to `plugins: []` — `entrypoint` is `undefined`, so the generated virtual module contained `import pluginDef0 from "undefined";` and the build failed deep in Rollup with the cryptic `Rollup failed to resolve import "undefined" from "virtual:emdash/plugins"`.

The generator now throws a clear, actionable error that names the offending plugin and explains that `plugins: []` entries must resolve to a file/package entrypoint (bundled at build time), so an inline `definePlugin({...})` result is not supported — move the plugin into its own module and reference it via a factory that returns a descriptor with an `entrypoint`.

Closes #1416

## Type of change

- [x] Bug fix

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (targeted: tests/unit/plugins — 600 tests)

### DIFF
--- a/.changeset/fix-1416-plugins-missing-entrypoint.md
+++ b/.changeset/fix-1416-plugins-missing-entrypoint.md
@@ -2,8 +2,4 @@
 "emdash": patch
 ---
 
-fix(core): fail fast with an actionable error for plugins without an entrypoint (#1416)
-
-`generatePluginsModule` assumed every plugin descriptor had a file entrypoint and emitted `import pluginDefN from "${descriptor.entrypoint}";`. For an in-process plugin — an inline `definePlugin({...})` result passed directly to `plugins: []` — `entrypoint` is `undefined`, so the generated virtual module contained `import pluginDef0 from "undefined";` and the build failed deep in Rollup with the cryptic `Rollup failed to resolve import "undefined" from "virtual:emdash/plugins"`.
-
-The generator now throws a clear, actionable error that names the offending plugin and explains that `plugins: []` entries must resolve to a file/package entrypoint (bundled at build time), so an inline `definePlugin({...})` result is not supported — move the plugin into its own module and reference it via a factory that returns a descriptor with an `entrypoint`.
+Fixes a cryptic build failure when a configured plugin has no resolvable entrypoint (#1416). Passing an inline `definePlugin({...})` result directly to `plugins: []` previously failed deep in the bundler with an unhelpful "failed to resolve import" error. The build now fails fast with a clear message that names the offending plugin and explains that `plugins: []` entries must resolve to a file or package entrypoint — move the plugin into its own module and reference it from there.
--- a/.changeset/fix-1416-plugins-missing-entrypoint.md
+++ b/.changeset/fix-1416-plugins-missing-entrypoint.md
@@ -1,0 +1,9 @@
+---
+"emdash": patch
+---
+
+fix(core): fail fast with an actionable error for plugins without an entrypoint (#1416)
+
+`generatePluginsModule` assumed every plugin descriptor had a file entrypoint and emitted `import pluginDefN from "${descriptor.entrypoint}";`. For an in-process plugin — an inline `definePlugin({...})` result passed directly to `plugins: []` — `entrypoint` is `undefined`, so the generated virtual module contained `import pluginDef0 from "undefined";` and the build failed deep in Rollup with the cryptic `Rollup failed to resolve import "undefined" from "virtual:emdash/plugins"`.
+
+The generator now throws a clear, actionable error that names the offending plugin and explains that `plugins: []` entries must resolve to a file/package entrypoint (bundled at build time), so an inline `definePlugin({...})` result is not supported — move the plugin into its own module and reference it via a factory that returns a descriptor with an `entrypoint`.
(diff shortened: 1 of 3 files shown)


ascorbic (approved):
Thanks!

Issue #867, opened 2026-04-30
Auto save validation errors


dcardosods:
### Description

While modifying a collection item, I kept getting auto save validation errors.

### Steps to reproduce

1. Got to edit a collection item (e.g. a project form the portfolio template)
2. Make a change (e.g. add a featured image) 
3. Unable to save

<img width="812" height="545" alt="Image" src="https://github.com/user-attachments/assets/180bf89d-8b7f-4df6-a79d-855666d17a95" />

### Environment

emdash version 0.8.0
portifolio template
node v22.22.2
 Mac OS

### Logs / error output

```shell

```

ascorbic:
@ask-bonk try to reproduce this issue. The error messages from the image are below:

`{error: {code: "VALIDATION_ERROR",..*} ›error: {code: "VALIDATION ERROR" code: "VALIDATION_ERROR" message: "content.0._key: Invalid inpl`

`Autosave failed content.o._key: Invalid input: expected string, received undefined; content.1._key: Invalid input: expected string, received undefined; content.2._key: Invalid input: expected string, received undefined`

PR #1561, opened 2026-06-20
fix(core): accept naive datetime-local values for datetime fields so they round-trip (#1368)

### LINKED ISSUE #1368: datetime field cannot round-trip through admin (autosave: Invalid input)
## Reproduction

1. Define a collection with a `datetime` field, e.g. `{ slug: \"event_at\", label: \"Event at\", type: \"datetime\" }`.
2. Seed an entry with `event_at: \"2026-06-04T18:30:00\"` (naive ISO, no offset).
3. Open the entry in the admin editor.
4. Try to autosave any change.

**Expected:** save succeeds.

**Actual:** Autosave fails with `event_at: Invalid input`.

## Root cause

The Zod validator for `datetime` in `node_modules/emdash/dist/zod-generator-*.mjs` is:

\`\`\`js
case \"datetime\": return z.string().datetime().or(z.string().date());
\`\`\`

- `z.string().datetime()` requires RFC 3339 with `Z` or a TZ offset.
- `z.string().date()` requires bare `YYYY-MM-DD`.

But the admin form renders `<input type=\"datetime-local\">`, which emits naive `YYYY-MM-DDTHH:mm` (no `Z`, no offset). Same shape as the seeded value above. Neither branch of the union accepts it, so the field is unsavable through its own editor.

Versions checked: emdash 0.14.0.

## Suggested fixes

In order of preference:

1. **Loosen the validator** to accept naive datetimes too — they're what the form already produces:
   \`\`\`js
   case \"datetime\":
     return z.string().regex(/^\\d{4}-\\d{2}-\\d{2}(T\\d{2}:\\d{2}(:\\d{2})?(Z|[+-]\\d{2}:\\d{2})?)?$/);
   \`\`\`
2. **Normalize on submit**: have the admin form append a `Z` (or the browser-local offset) before POSTing the value. Backwards-compatible with stored naive values too, if the field's data setter normalizes.
3. **Add a distinct `datetime-local` field type** (UTC vs local-naive) and have the admin form use the right one per field.

(1) gets every existing deployment unstuck with one line.

## Workaround we used downstream

We switched our affected fields from `datetime` to `date` to keep working: [BHS-P-C/bhspandc-cms#3](https://github.com/BHS-P-C/bhspandc-cms/issues/3). We'd happily go back to `datetime` once this is fixed.

marcusbellamyshaw-cell:
## What does this PR do?

A `datetime` field could not be saved through its own admin editor. The generated validator was `z.string().datetime().or(z.string().date())`, which only accepts ISO with a `Z` suffix or a bare `YYYY-MM-DD` date. But `<input type="datetime-local">` (and many seeds) produce a naive datetime such as `2026-06-04T18:30:00` (no `Z`, no offset). Because the admin re-sends every loaded field on autosave, a stored naive datetime failed validation and the entry became unsavable — the same class of round-trip bug as #867.

The validator now uses `z.iso.datetime({ offset: true, local: true }).or(z.iso.date())`, which accepts ISO with `Z`, ISO with a timezone offset, naive datetimes (with or without seconds), and date-only values, while still rejecting non-dates and impossible dates. This also replaces the deprecated `z.string().datetime()` API.

Closes #1368

## Type of change

- [x] Bug fix

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (targeted: tests/unit/schema — 115 tests)


ascorbic (approved):
Thanks!

PR #1562, opened 2026-06-20
fix(core): grant plugin media write access on storage-only backends (#1313)

### LINKED ISSUE #1313: ctx.media.upload() unavailable for in-process plugins when using R2 Worker binding
## Summary

`ctx.media.upload()` is silently absent from the plugin context when the site uses `r2()` storage (Worker binding). The capability docs state that `media:write` grants `ctx.media.getUploadUrl()`, `ctx.media.upload()`, and `ctx.media.delete()`, but with R2 binding only the read methods (`get`, `list`) are available.

## Version

- emdash 0.16.1
- @emdash-cms/cloudflare 0.16.1

## Root cause

In `src/plugins/context.ts`, `createMediaAccessWithWrite` (which includes `upload()`) is only created when `this.getUploadUrl` is truthy:

```typescript
// context.ts:927
if (capabilities.has("media:write") && this.getUploadUrl) {
    media = createMediaAccessWithWrite(this.db, this.getUploadUrl, this.storage);
} else if (capabilities.has("media:read")) {
    media = createMediaAccess(this.db);  // read-only — no upload()
}
```

However, `upload()` itself doesn't use `getUploadUrl` at all — it only needs `storage`:

```typescript
// context.ts:458
async upload(filename, contentType, bytes) {
    if (!storage) {
        throw new Error("Media upload() requires a storage backend...");
    }
    await storage.upload({ key: storageKey, body: new Uint8Array(bytes), contentType });
    // ...
}
```

The R2 Worker binding adapter intentionally does not provide `getUploadUrl` (presigned URLs are unsupported via Worker binding), so `this.getUploadUrl` is never set and the condition fails — despite `this.storage` being fully functional.

## Evidence this is unintentional

`@emdash-cms/cloudflare`'s sandbox wrapper (`src/sandbox/wrapper.ts:111`) works around the same issue by setting `getUploadUrl` to a stub that immediately throws:

```typescript
getUploadUrl: () => { throw new Error("getUploadUrl is not available in sandbox mode. Use media.upload(...) instead."); }
```

The stub has no functional purpose other than making `this.getUploadUrl` truthy so the gate passes and `upload()` becomes accessible. This indicates the intent is for `upload()` to work with R2 binding — the fix just wasn't applied to in-process (native) plugins.

## Suggested fix

```typescript
// context.ts:927
if (capabilities.has("media:write") && (this.getUploadUrl || this.storage)) {
    media = createMediaAccessWithWrite(this.db, this.getUploadUrl!, this.storage);
```

`createMediaAccessWithWrite` already handles a missing `getUploadUrl` gracefully — `upload()` just needs to be callable when `storage` is available.

## Workaround

In-process plugins using R2 Worker binding can access `env.MEDIA` directly via a lazy dynamic import at request time (the same approach the bridge uses internally for sandboxed plugins).

marcusbellamyshaw-cell:
## What does this PR do?

Fixes `ctx.media.upload()` and `ctx.media.delete()` being silently unavailable to in-process (native) plugins on storage backends that don't support presigned URLs — notably the Cloudflare R2 Worker binding.

In `PluginContextFactory.createContext`, media write access was gated on `this.getUploadUrl` being truthy:

```ts
if (capabilities.has("media:write") && this.getUploadUrl) {
  media = createMediaAccessWithWrite(this.db, this.getUploadUrl, this.storage);
}
```

But `upload()` never uses `getUploadUrl` — it only needs `storage`. The R2 Worker binding intentionally provides `storage` but no presigned-URL support, so `getUploadUrl` is never set and a plugin granted `media:write` loses **all** write access despite having a fully functional storage backend. The only workaround today is to bypass the media API and write to the bucket binding directly.

This is the same gap the `@emdash-cms/cloudflare` sandbox wrapper already papers over by setting `getUploadUrl` to a stub that throws — confirming the intent is for `upload()` to work on the binding.

### The fix

- Gate media write access on `(this.getUploadUrl || this.storage)`, so a plugin can write whenever it actually can — via a presign provider **or** a direct storage backend.
- Make `getUploadUrlFn` optional in `createMediaAccessWithWrite`. When no presign provider is configured, `media.getUploadUrl()` now throws a clear "not supported — use media.upload() instead" error rather than silently stripping `upload()`/`delete()` from the context.
- No change to behavior when `getUploadUrl` **is** configured.

Closes #1313

## Type of change

- [x] Bug fix

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (or targeted tests for my change) — `packages/core` plugin suites: 706 passed

## Screenshots / test output

New tests in `packages/core/tests/integration/plugins/capabilities.test.ts` (describe: "Media Access write gate (#1313)") cover:

- `media:write` + storage (no `getUploadUrl`) → `upload()`/`delete()` available (the regression)
- `upload()` round-trips bytes through the storage backend
- `getUploadUrl()` rejects with a clear error when no presign provider exists
- existing `getUploadUrl`-configured path still grants write
- `media:read` stays read-only; `media:write` with neither storage nor presign nor read → no media access

```
Test Files  30 passed (30)
     Tests  706 passed (706)
```

### DIFF
--- a/.changeset/media-write-r2-binding.md
+++ b/.changeset/media-write-r2-binding.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fix `ctx.media.upload()` and `ctx.media.delete()` being unavailable to in-process plugins on storage backends that don't support presigned URLs, such as the Cloudflare R2 Worker binding (#1313). A plugin granted `media:write` previously lost all write access whenever `getUploadUrl` was unset — even though `upload()` only needs a storage backend — forcing plugins to bypass the media API and write to the bucket directly. Write access is now granted when either a presigned-URL provider or a direct storage backend is configured. Calling `media.getUploadUrl()` on a backend without presigned-URL support now throws a clear "not supported" error instead of silently stripping `upload()` from the context.
--- a/packages/core/src/plugins/context.ts
+++ b/packages/core/src/plugins/context.ts
@@ -432,14 +432,20 @@ export function createMediaAccess(db: Kysely<Database>): MediaAccess {
 
 /**
  * Create full media access with write operations.
- * If storage is not provided, upload() will throw at call time.
+ *
+ * `getUploadUrlFn` is optional: when omitted (e.g. the R2 Worker binding has no
+ * presigned-URL support), getUploadUrl() throws at call time but upload() and
+ * delete() remain available. If `storage` is not provided, upload() throws at
+ * call time. See #1313.
  */
 export function createMediaAccessWithWrite(
 	db: Kysely<Database>,
-	getUploadUrlFn: (
-		filename: string,
-		contentType: string,
-	) => Promise<{ uploadUrl: string; mediaId: string }>,
+	getUploadUrlFn:
+		| ((
+				filename: string,
+				contentType: string,
+		  ) => Promise<{ uploadUrl: string; mediaId: string }>)
+		| undefined,
 	storage?: Storage,
 ): MediaAccessWithWrite {
 	const mediaRepo = new MediaRepository(db);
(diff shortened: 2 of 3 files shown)


ascorbic:
Please fix the formatting issue on this. The bot can't push to your rep.

marcusbellamyshaw-cell:
Closing this as superseded by #1616 (merged), which fixes the same root cause — `media:write` silently degrading to read-only because the runtime threads `storage` but never `getUploadUrl`. The merged change landed an equivalent implementation in `createMediaAccessWithWrite`/`PluginContextFactory` (grants write when `storage` *or* `getUploadUrl` is present, derives `getUploadUrl` from storage, warns once when neither is configured) plus the same integration coverage in `capabilities.test.ts`. Nothing here remains to add. Thanks for the review @ascorbic.

PR #1570, opened 2026-06-21
fix(admin): keep hook order stable on content list when a refetch errors (#1415)

### LINKED ISSUE #1415: Admin: ContentListPage still throws React #300 (hook-order) — inline useCallback for onLoadMore sits below early returns
First off — thank you for EmDash. We're running it in production for a furniture studio site and it's been a genuine pleasure to work with; the admin UX and the Astro/D1 integration are excellent. 🙏

Ran into a small hook-order crash that looks like a follow-up to #439 / #467 — same class of bug, but a different hook the previous fix didn't cover.

## Symptom

On any content collection list (`/_emdash/admin/content/<collection>`), an interaction that triggers a refetch — e.g. clicking a column header to change sort, or fast navigation — can crash the page with:

```
Rendered fewer hooks than expected. This may be caused by an accidental early return statement.
```

(Minified React error #300.) A reload recovers; content is never affected.

## Root cause

`packages/admin/src/router.tsx`, `ContentListPage`. #467 correctly lifted `React.useMemo` for `items` above the early returns — but there's still a hook below them: the inline `React.useCallback` passed to `onLoadMore`.

```tsx
const items = React.useMemo(...);              // ✅ fixed in #467, above guards

if (!manifest)         return <LoadingScreen />;      // ~L405
if (!collectionConfig) return <NotFoundPage ... />;   // ~L411
if (error)             return <ErrorScreen ... />;    // ~L415

return (
  <ContentList
    onLoadMore={React.useCallback(() => void fetchNextPage(), [fetchNextPage])}  // ~L437 ← hook below early returns
  />
);
```

When sort changes, `sort` is part of the infinite-query key, so React Query refetches. On the transitional render where a guard fires (e.g. `error` or a brief loading state), the component returns early and never reaches the `useCallback` — so that render runs one fewer hook than the previous one, tripping React #300.

It reproduces on **any** collection (the trigger is query/manifest state, not collection data). Confirmed on `@emdash-cms/admin@0.17.2` and current `main`.

## Proposed fix

Lift the callback above the early returns, exactly as #467 did for `items`:

```tsx
const items = React.useMemo(...);
const handleLoadMore = React.useCallback(() => void fetchNextPage(), [fetchNextPage]);

if (!manifest)         return <LoadingScreen />;
if (!collectionConfig) return <NotFoundPage ... />;
if (error)             return <ErrorScreen ... />;

return <ContentList ... onLoadMore={handleLoadMore} ... />;
```

Might also be worth grepping for other inline `use*` hooks inside JSX returns that sit below guards, in case this one has siblings.

Happy to open a PR if useful. Thanks again for the project! 🙏

---

**Environment:** `emdash` / `@emdash-cms/admin` 0.17.2 · Astro 6 · Cloudflare adapter (D1)

marcusbellamyshaw-cell:
## What does this PR do?

Fixes a crash (React #300 "Rendered fewer hooks than expected") on the content list page when an action that triggers a refetch errors.

`ContentListPage` declared the `onLoadMore` `React.useCallback` inline in JSX, *below* three early returns (`!manifest`, `!collectionConfig`, `error`). On the render that takes the `error` guard, that hook never runs, so the render executes one fewer hook than a successful render — which is exactly what React #300 forbids. Any refetch that fails (e.g. a transient API error after the list has loaded) crashed the whole page instead of showing the error state.

Fix: lift the callback above the early returns so the hook count is identical on every render path. Added a comment documenting the invariant.

Closes #1415.

## Type of change
- [x] Bug fix

## Checklist
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

## Screenshots / test output
New test in `packages/admin/tests/router.test.tsx`: renders the page, swaps the content handler to a 500, refetches, and asserts the Retry button renders (the page no longer crashes). Targeted admin + core suites green; typecheck/lint/format clean.


ascorbic (approved):
Thanks!

PR #1572, opened 2026-06-21
fix(core): preserve locale casing in API locale filter (#1551)

### LINKED ISSUE #1551: `localeCode` schema lowercases the `?locale=` filter, returning zero rows for locales with uppercase subtags
### Description

Filtering content by a locale that has an uppercase subtag (like `zh-TW` or `en-US`) returns nothing on SQLite/D1, even though the rows exist.

My site's default locale is `zh-TW`. The admin content list came up completely blank, while the public site rendered all the posts fine. The API was returning `{"items":[],"total":0}` but the DB clearly had the data.

Tracked it down to `localeCode` in `packages/core/src/api/schemas/common.ts`:

```ts
export const localeCode = z
    .string()
    .regex(/^[a-z]{2,3}(-[a-z0-9]{2,8})*$/i, "Invalid locale code")
    .transform((v) => v.toLowerCase());
```

That `.toLowerCase()` rewrites `?locale=zh-TW` to `zh-tw` before the query runs. But stored locales keep the configured casing — the column DEFAULT comes straight from `defaultLocale` (`database/migrations/040_byline_i18n.ts:121`) and the read fallback uses it raw (`query.ts:513`), so rows are stored as `zh-TW`. SQLite and D1 compare strings case-sensitively, so `WHERE locale = 'zh-tw'` matches zero rows. The public site dodges this because it never goes through the query-param schema.

So the filter lowercases but storage doesn't — the two sides disagree on casing. Affects any locale with an uppercase subtag (`zh-TW`, `zh-Hant`, `en-US`, `pt-BR`, `fr-CA`, ...).

### Steps to reproduce

1. Set up i18n with an uppercase subtag, e.g. `defaultLocale: "zh-TW"`.
2. Create or import some content (stored as `locale = 'zh-TW'`).
3. Run on SQLite (or D1).
4. `GET /_emdash/api/content/<collection>?locale=zh-TW` → `{"items":[],"total":0}`.
5. Meanwhile `SELECT count(*) FROM ... WHERE locale = 'zh-TW'` returns the real count, and the admin content list shows up blank.

### Environment

- emdash version: 0.21.0 (source unchanged on `main`, HEAD `cb1c689`)
- Node.js version: 22
- Runtime: Node
- OS: Linux (container)

### Logs / error output

```shell

```

marcusbellamyshaw-cell:
## What does this PR do?

Fixes the content and search APIs returning zero results when filtering by a locale with an uppercase region or script subtag (e.g. `?locale=zh-TW`, `pt-BR`, `zh-Hant`).

The `localeCode` validator applied `.transform(v => v.toLowerCase())`, but config `locales`/`defaultLocale`, the stored `locale` column, and the public query path all preserve the original BCP-47 casing. SQLite/D1 string comparison is case-sensitive, so the lowercased filter matched nothing. The sibling `localeFilterQuery` (taxonomies/menus) never lowercased, so this was also an internal inconsistency.

Fix: drop the transform so the value is preserved verbatim. Validation stays case-insensitive (the `/i` regex is unchanged).

Closes #1551.

## Type of change
- [x] Bug fix

## Checklist
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

## Screenshots / test output
New `localeCode validator (#1551)` describe in `packages/core/tests/unit/api/schemas.test.ts`: preserves `zh-TW`/`pt-BR`/`zh-Hant`, still validates case-insensitively, rejects malformed codes, and verifies `contentListQuery`/`contentCreateBody` keep the casing.


marcusbellamyshaw-cell:
Thanks for the thorough review. Addressed the two completeness points:

**✅ Suggestion #2 — `localeFilterQuery` tightening (done in e288db7):** Switched the shared `?locale=` filter to reuse `localeCode`, so the taxonomy/menu endpoints now get the same casing-preserving BCP-47 validation as `contentListQuery` and search. Confirmed all 25 call sites use it purely as an optional query filter, so the only behavior change is that a malformed locale (e.g. `_invalid_`) is now rejected with a clear error instead of silently matching zero rows. Added a unit test and a sentence to the changeset.

**⏭️ Suggestion #1 — forward-only migration for pre-existing lowercase rows:** Deliberately deferring this out of #1572. It's a real point, but rewriting the `locale` column across every `ec_*` table is a data migration with meaningfully larger blast radius than this API-layer fix, and it deserves its own PR + changeset + test coverage (including the case where a site legitimately uses lowercase locales). Keeping #1572 scoped to the query-path fix per the repo's scope-discipline guidance. I'll open a follow-up issue/PR to canonicalize stored locale casing against `getI18nConfig().locales`.

marcusbellamyshaw-cell:
Filed the deferred locale-canonicalization migration as #1599 for tracking.

ascorbic:
Superseded by #2174. The conflict resolution is unchanged, but GitHub rejected maintainer pushes to this organization-owned fork despite `maintainerCanModify` being enabled.

~ Sol

PR #1406, opened 2026-06-11
feat(image): honor alignment + surface display size for migrated images (#1404)

SailingGreg:
## What does this PR do?

The Gutenberg to Portable Text importer captures WordPress image `alignment` on the image node, but it was dropped at render and not editable. This wires it through end to end and surfaces the existing display-size controls for migrated images.

- **Render** (`packages/core/src/components/Image.astro`): emit `emdash-image--align-{left,right,center,wide,full}` from the node`s `alignment`, with default CSS (left/right float and wrap text; center centers; wide/full span the column).
- **Editor** (`packages/admin`): thread `alignment` through the PortableText/TipTap serializer (both directions) and the TipTap image node; add an Alignment selector to `ImageDetailPanel`; reflect alignment in the image NodeView so the editor preview matches the published output.
- **Display size**: relax the gate from `width && height` to `src` so WordPress-imported images (which carry only `displayWidth`/`displayHeight`, no originals) are resizable; the aspect-ratio lock and reset-to-original degrade gracefully when originals are unknown.

Validated end to end on a real WordPress-migrated site: editor -> save -> published page all agree on the canonical `alignment` field.

Closes #1404

## Type of change

- [x] Bug fix
- [ ] Feature (requires [maintainer-approved Discussion](https://github.com/emdash-cms/emdash/discussions/categories/ideas))
- [ ] Refactor (no behavior change)
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests
- [ ] Chore (dependencies, CI, tooling)

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md)
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (or targeted tests for my change)
- [x] `pnpm format` has been run
- [x] I have added/updated tests for my changes (if applicable)
- [x] User-visible strings in the admin UI are wrapped for translation (if applicable)
- [x] I have added a changeset (this PR changes published packages)
- [ ] New features link to an approved Discussion

## AI-generated code disclosure

- [x] This PR includes AI-generated code — model/tool: Claude Opus 4.8 (via Claude Code)

## Screenshots / test output

Container-render repro asserts the alignment class is emitted (default render and via a delegating component override); validated end to end on a live WordPress-migrated site (editor -> save -> published render).

PR #1576, opened 2026-06-22
fix: preserve image alignment through editor round-trip and render it (#1404)

### LINKED ISSUE #1404: Image alignment from WordPress import is dropped at render and not editable in the admin editor
## Summary

The Gutenberg → Portable Text importer captures WordPress image alignment and writes an `alignment` attribute onto image nodes, but:

1. the Portable Text image **renderer** ignores it, so alignment is silently dropped on the rendered page; and
2. the admin editor **image panel** has no alignment control, so it cannot be set or corrected after import.

Net effect: left/center/right/wide/full alignment from WordPress survives in the stored content but is invisible on the site and uneditable in the editor.

*(Display sizing is fine — `displayWidth`/`displayHeight` are honored in both render and editor. This issue is specifically about alignment.)*

## Pointers

- **Emitted on import:** `packages/gutenberg-to-portable-text/src/transformers/core.ts:399` — `alignment: mapAlignment(align)` on the `image` block; alignment types at `packages/gutenberg-to-portable-text/src/types.ts:66,193`.
- **Dropped at render:** `packages/core/src/components/Image.astro` — `Props.node` declares `asset/alt/caption/width/height/displayWidth/displayHeight` but no `alignment`; the `<figure class="emdash-image">` output applies no alignment.
- **Not editable:** `packages/admin/src/components/editor/ImageDetailPanel.tsx` — `ImageAttributes` has no `alignment`; the panel exposes alt/caption/title + width/height only.

## Proposed fix

1. Add `alignment?: "left" | "center" | "right" | "wide" | "full"` to the image node `Props` in `Image.astro` and map it to an alignment class/style.
2. Add `alignment` to `ImageAttributes` and an alignment selector in `ImageDetailPanel` (persist via the editor `ImageNode`).
3. Ship default CSS for the alignment classes so renderer output is styled out of the box.

Happy to open a PR for this if useful.

## Environment

emdash 0.17.2, @astrojs/node, SQLite, single-locale.

marcusbellamyshaw-cell:
## What does this PR do?

Image blocks imported from WordPress/Gutenberg carry a horizontal alignment (`left`/`right`/`center`/`wide`/`full`), but the Portable Text ↔ ProseMirror round-trip dropped it — so opening an imported post in the editor and re-saving flattened every aligned image to the default. This carries `alignment` through both the core and admin converters and the TipTap image-node schema, and renders it in `Image.astro` with matching CSS. Round-trip regression tests added in both packages.

Closes #1404

## Type of change

- [x] Bug fix

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (targeted: 3 core + 2 admin round-trip tests)
- [ ] User-visible strings wrapped for translation — n/a, no new admin UI strings
- [x] I have added a changeset (`emdash` + `@emdash-cms/admin`, both patch)


marcusbellamyshaw-cell:
Thanks for the review. Addressed the `[needs fixing]` item: rewrote `.changeset/image-alignment-round-trip.md` as user-facing release notes, dropping the internal mechanics (converters, image node schema, `Image.astro`) per AGENTS.md guidance (59dba7f).

On the two suggestions:

- **Admin converter validation symmetry** — leaving as-is for now. The Gutenberg importer (`mapAlignment`) only ever emits the valid set, so there's no live path that persists an out-of-set value; mirroring `isImageAlignment` in the admin would be defense-in-depth on a value that can't currently occur, and I'd rather keep this PR scoped to the round-trip fix.
- **Editor preview not reflecting alignment** — agreed this is a worthwhile follow-up (WYSIWYG + a control to view/change alignment), but it's a new editor feature beyond this import-preservation fix, so I'll track it separately rather than expand scope here.

ascorbic:
Thanks for this, @marcusbellamyshaw-cell! Closing because the same end-to-end image alignment fix landed in #1406, which has merged. Much appreciated.

PR #1577, opened 2026-06-22
fix(core): add offset pagination to getEmDashCollection for numbered archives (#1364)

### LINKED ISSUE #1364: getEmDashCollection only supports cursor pagination, making /page/{number} archive routes inefficient
### Description

`getEmDashCollection()` currently supports cursor/keyset pagination via `limit` and `cursor`, but does not support numeric page or offset pagination.

This works well for cursor URLs like:

```ts
const { entries, nextCursor } = await getEmDashCollection("posts", {
  limit: 20,
  cursor,
  orderBy: { published_at: "desc" },
});
```

But it makes canonical numbered archive URLs like `/page/2`, `/page/3`, etc. inefficient or awkward to implement.

For a server-rendered site that needs public URLs like:

```text
/page/2
/page/3
/tag/security/page/2
```

there is no supported way to ask EmDash for “page 5 with 20 posts per page” or “skip the first 80 posts.”

Expected:

EmDash should provide an API for numbered archive pagination, for example:

```ts
await getEmDashCollection("posts", {
  limit: 20,
  offset: 80,
  orderBy: { published_at: "desc" },
});
```

or:

```ts
await getEmDashCollection("posts", {
  page: 5,
  perPage: 20,
  orderBy: { published_at: "desc" },
});
```

Actual:

The only supported mechanism is cursor pagination. To render `/page/5`, a site must either walk cursors page by page or over-fetch from the beginning and slice:

```ts
const pageSize = 20;
const page = 5;

const { entries, nextCursor } = await getEmDashCollection("posts", {
  limit: page * pageSize,
  orderBy: { published_at: "desc" },
});

const pageEntries = entries.slice((page - 1) * pageSize, page * pageSize);
```

This avoids sequential cursor walking, but it fetches and hydrates all previous posts too. For `/page/50` with page size 20, that means fetching up to 1,000 posts to render 20.

Suggested fix:

Add first-class numeric pagination support to `getEmDashCollection()` and the underlying EmDash loader/repository layer.

Possible API options:

```ts
getEmDashCollection("posts", {
  limit: 20,
  offset: 80,
  orderBy: { published_at: "desc" },
});
```

or:

```ts
getEmDashCollection("posts", {
  page: 5,
  perPage: 20,
  orderBy: { published_at: "desc" },
});
```

The result could include pagination metadata:

```ts
{
  entries,
  nextPage,
  previousPage,
  hasMore,
  total?
}
```

A `total` field could be optional because it may require an extra count query.

### Steps to reproduce

1. Create an EmDash site with more posts than fit on one archive page.
2. Add a server-rendered numbered archive route:

```astro
---
// src/pages/page/[page].astro
import { getEmDashCollection } from "emdash";

const page = Number(Astro.params.page);
const pageSize = 20;

// No native offset/page API exists, so this has to over-fetch.
const { entries, nextCursor } = await getEmDashCollection("posts", {
  limit: page * pageSize,
  orderBy: { published_at: "desc" },
});

const posts = entries.slice((page - 1) * pageSize, page * pageSize);
---

<ul>
  {posts.map((post) => <li>{post.data.title}</li>)}
</ul>
```

3. Visit `/page/10`.
4. Observe that the route must fetch all posts from pages 1 through 10 just to render page 10.
5. Try to use an offset-style API:

```ts
await getEmDashCollection("posts", {
  limit: 20,
  offset: 180,
});
```

6. Observe that `offset` is not part of `CollectionFilter` and is not supported by the loader.

### Environment

- emdash version: `0.17.1`
- Astro version: `6.3.7` via `https://pkg.pr.new/astro@16335`
- Node.js version: `v22.22.3`
- Runtime: Cloudflare Workers
- OS: macOS

### Logs / error output

```shell
No runtime error is thrown. This is an API limitation: numbered archive routes require inefficient cursor-walking or fetch-and-slice workarounds because EmDash only exposes cursor pagination.
```

marcusbellamyshaw-cell:
## What does this PR do?

`getEmDashCollection` only supported cursor/keyset pagination, so numbered archive routes (`/page/2`, `/tag/security/page/3`) had no efficient option — they had to either walk cursors page-by-page or over-fetch from the start and slice (fetching up to 1,000 rows to render 20 on `/page/50`).

This adds an additive `offset` option to the public `CollectionFilter` (and the underlying loader filter), so numbered archives become a one-liner:

```ts
const perPage = 20;
const { entries, hasMore } = await getEmDashCollection("posts", {
  limit: perPage,
  offset: (page - 1) * perPage,
  orderBy: { published_at: "desc" },
});
```

Details:

- **`offset` is applied as a SQL `OFFSET`**, cross-dialect: SQLite only accepts `OFFSET` with a `LIMIT`, so a bare offset uses `LIMIT -1` (unbounded); Postgres takes a standalone `OFFSET`.
- **`offset` and `cursor` are mutually exclusive** — cursor (keyset) wins when both are set, so the two don't stack into a double skip. Only a positive integer applies (0 / negative / fractional are no-ops).
- **New `hasMore` on `CollectionResult`**, set whenever `limit` is provided, computed from the existing over-fetch-by-one — so archive routes can render a "next page" link without an extra count query.
- Offset calls **bypass limit bucketing** (their limit is part of the pagination contract), and `offset` is part of the request-cache key so distinct pages don't collide.

Closes #1364

## Type of change

- [x] Bug fix

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] User-visible strings in the admin UI are wrapped for translation — n/a, no admin UI strings (site-side query API + docs)
- [ ] New features link to an approved Discussion — n/a, additive API closing a triaged bug report

## Screenshots / test output

```
Test Files  5 passed (5)
     Tests  31 passed (31)
```

New tests: `loader-offset-pagination.test.ts` (skip/disjoint pages/past-end/cursor-wins), `query-collection-offset.test.ts` (offset forwarding, `hasMore` semantics, cache-key separation), plus a bucketing-exemption case in `query-collection-bucketing.test.ts`.

### DIFF
--- a/packages/core/src/loader.ts
+++ b/packages/core/src/loader.ts
@@ -615,6 +615,13 @@ export interface CollectionFilter {
 	 */
 	cursor?: string;
 	/**
+	 * Skip this many rows before returning results (offset pagination).
+	 * Use with `limit` for numbered archive routes (`/page/2`):
+	 * `offset = (page - 1) * perPage`. Ignored unless it is a positive
+	 * integer, and ignored entirely when `cursor` is set (cursor wins).
+	 */
+	offset?: number;
+	/**
 	 * Filter by field values, taxonomy terms, byline credits, or ranges.
 	 *
 	 * Taxonomy names are detected automatically and filtered via JOIN.
@@ -750,6 +757,17 @@ export function emdashLoader(): LiveLoader<EntryData, EntryFilter, CollectionFil
 				// Cursor pagination: over-fetch by 1 to detect next page
 				const fetchLimit = limit ? limit + 1 : undefined;
 
+				// Offset pagination (numbered archive routes). Keyset (cursor)
+				// and offset are mutually exclusive ways to express "the next
+				// page" — when both are supplied, cursor wins and offset is
+				// dropped so the two don't stack into a double skip. Only a
+				// positive integer applies; 0 / negative / fractional are no-ops.
+				const rawOffset = cursor ? undefined : filter?.offset;
+				const offset =
+					typeof rawOffset === "number" && Number.isInteger(rawOffset) && rawOffset > 0
+						? rawOffset
+						: undefined;
+
 				// Build cursor condition if cursor is provided
 				const cursorCondition = cursor ? buildCursorCondition(cursor, orderBy) : null;
 
@@ -866,6 +884,20 @@ export function emdashLoader(): LiveLoader<EntryData, EntryFilter, CollectionFil
 						type,
 						tableName,
 					);
+
+					// LIMIT/OFFSET clause. SQLite only accepts OFFSET when a
+					// LIMIT is present, so a bare offset uses `LIMIT -1`
+					// (unbounded); Postgres takes a standalone OFFSET.
+					let limitOffsetClause = sql``;
+					if (fetchLimit != null && offset != null) {
+						limitOffsetClause = sql`LIMIT ${fetchLimit} OFFSET ${offset}`;
+					} else if (fetchLimit != null) {
+						limitOffsetClause = sql`LIMIT ${fetchLimit}`;
+					} else if (offset != null) {
+						limitOffsetClause = isPostgres(db)
+							? sql`OFFSET ${offset}`
+							: sql`LIMIT -1 OFFSET ${offset}`;
+					}
 					result = await sql<Record<string, unknown>>`
 						SELECT *, ${termsSelect}, ${bylinesSelect} FROM ${sql.ref(tableName)}
 						WHERE deleted_at IS NULL
@@ -876,7 +908,7 @@ export function emdashLoader(): LiveLoader<EntryData, EntryFilter, CollectionFil
 						${bylineCond}
 						${fieldCondsSQL ? sql`AND ${fieldCondsSQL}` : sql``}
 						${orderByClause}
-						${fetchLimit ? sql`LIMIT ${fetchLimit}` : sql``}
+						${limitOffsetClause}
 					`.execute(db);
 				}
 
--- a/packages/core/src/loader.ts
+++ b/packages/core/src/loader.ts
@@ -603,24 +603,17 @@ export interface WhereRange {
 export type WhereValue = string | string[] | WhereRange;
 
 /**
- * Filter for loadCollection - type is required
+ * Fields shared by every collection filter, independent of pagination mode.
+ *
+ * Cursor and offset pagination are mutually exclusive, so they live on the
+ * `CursorCollectionFilter` / `OffsetCollectionFilter` variants rather than
+ * here. Use the {@link CollectionFilter} union for any value that may be
+ * either.
  */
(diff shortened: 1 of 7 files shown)


packages/core/src/loader.ts
```diff
@@ -539,6 +539,13 @@ export interface CollectionFilter {
...
 	 */
 	cursor?: string;
+	/**
+	 * Skip this many rows before returning results (offset pagination).
+	 * Use with `limit` for numbered archive routes (`/page/2`):
+	 * `offset = (page - 1) * perPage`. Ignored unless it is a positive
+	 * integer, and ignored entirely when `cursor` is set (cursor wins).
+	 */
+	offset?: number;
 	/**
```
ascorbic: Instead of hving two mutually exclusive, optional properties on this, could you change it to a union type. Use a base type, with two different subtypes for the offset and cursor variants and export CollectionFilter as the union of these. This will mean it will show up as a type error if a user passes both.

marcusbellamyshaw-cell: Done in b57ac46a. Split into a shared `CollectionFilterBase` plus `CursorCollectionFilter` (`offset?: never`) and `OffsetCollectionFilter` (`cursor?: never`) variants, with `CollectionFilter` exported as their union — applied to both the public query filter and the loader filter, so supplying both is now a compile-time error. The internal query→loader mapping spreads only the active key (cursor wins) so it still satisfies the union, and there's a `@ts-expect-error` test locking in the both-at-once rejection.

packages/core/tests/unit/query-collection-offset.test.ts
```diff
@@ -0,0 +1,105 @@
...
+			entries: makeEntries(5),
+			cacheHint: {},
+		} as any);
+
+		const result = await run(() => getEmDashCollection("post"));
+
+		expect(result.hasMore).toBeUndefined();
+	});
+
+	it("keeps offset out of the request-cache key so different pages don't collide", async () => {
```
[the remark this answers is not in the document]
marcusbellamyshaw-cell: Renamed in f3e1ff97 to "includes offset in the request-cache key so distinct pages don't collide", matching what the test asserts.

ascorbic:
@marcusbellamyshaw-cell  Can you run `pnpm format`

marcusbellamyshaw-cell:
Branch is already formatted — `pnpm format` produces no changes, and `pnpm format:check` (oxfmt + prettier) passes locally on `5464f241`; the green **Format** (`pull_request`) check confirms the same in CI.

The red check is **Auto Format** (`pull_request_target`), which only runs `oxfmt` and pushes a `style: format` commit. It's failing at the checkout step, not on a format diff — `actions/checkout@v7` now refuses to check out fork code under `pull_request_target` (the "pwn request" guard), so it will fail on any fork PR until the workflow is updated (e.g. `allow-unsafe-pr-checkout: true` or run oxfmt against the base ref). Happy to send a small PR for that if useful.

PR #1578, opened 2026-06-22
fix(core): implement pagination for search() and searchCollection() (#1463)

### LINKED ISSUE #1463: search() ignores the cursor option and never returns nextCursor — pagination is impossible
## Summary

The public `search()` API advertises keyset pagination through its types but does not implement it. `SearchOptions` declares a `cursor?: string` field and `SearchResponse` declares `nextCursor?: string`, but `searchWithDb` neither reads the incoming `cursor` nor ever populates `nextCursor`. As a result, search results are silently capped at `limit` and there is no way to fetch a second page — callers that wire up "load more" against the documented shape get a button that never appears (and a `cursor` that's silently discarded).

## Where

`packages/core/src/search/query.ts`, `searchWithDb`:

```ts
const limit = options.limit ?? 20;
// ...options.cursor is never read...
allResults.sort((a, b) => b.score - a.score);
const items = allResults.slice(0, limit);
return { items };          // no nextCursor
```

Compare with the declared types in `packages/core/src/search/types.ts`:

```ts
export interface SearchOptions {
  // ...
  /** Pagination cursor */
  cursor?: string;
}

export interface SearchResponse {
  items: SearchResult[];
  /** Cursor for next page of results */
  nextCursor?: string;
}
```

## Expected

One of:

1. Implement keyset pagination: honour `options.cursor` and return a `nextCursor` when more results exist (consistent with how `getEmDashCollection` paginates), or
2. If cross-collection cursor pagination isn't feasible (results are merged from per-collection FTS queries and re-sorted by score), drop `cursor`/`nextCursor` from the public types so the API doesn't advertise pagination it can't deliver, and document the `limit` cap.

## Impact

Any consumer rendering paginated search results is silently limited to the first `limit` hits, with no indication that more matches exist. The mismatch is invisible at compile time because the types declare the fields.

## Version

Observed in `emdash@0.19.0` (runtime bundle and `packages/core` source agree).

marcusbellamyshaw-cell:
## What does this PR do?

The search API advertised keyset pagination through its types — `SearchOptions.cursor` and `SearchResponse.nextCursor` — but `searchWithDb` never read the incoming cursor nor populated `nextCursor`. Results were silently capped at `limit` with no way to fetch a second page, so any "load more" wired against the documented shape got a button that never appears (and a `cursor` that's discarded).

This implements real pagination for both `search()` and `searchCollection()`.

**Why offset, not pure keyset:** search results are merged from per-collection FTS queries and re-sorted by `bm25` score, so there's no single stable keyset column to encode the way `getEmDashCollection` encodes `(orderValue, id)`. Instead we page the merged, score-sorted set by **offset**, carried opaquely in the cursor's `orderValue` — reusing `encodeCursor`/`decodeCursor` so the cursor keeps the same base64-JSON shape and `InvalidCursorError` handling as the rest of the API.

Each collection fetches its top `(offset + limit + 1)` rows, which is exactly what's needed for the merged window `[offset, offset + limit)` to rank correctly even when the whole window comes from one collection, plus one row to detect a further page. A `nextCursor` is issued only when at least one result exists past the page.

- `search()` / `searchWithDb` and `searchCollection()` both honour `cursor` and return `nextCursor`.
- The `/_emdash/api/search` endpoint now accepts a `cursor` query param and returns `nextCursor`; a malformed cursor surfaces as `400 INVALID_CURSOR` via the existing `handleError` mapping.

Closes #1463

## Type of change

- [x] Bug fix

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] User-visible strings in the admin UI are wrapped for translation — n/a, no admin UI strings (server-side search API + docs)
- [ ] New features link to an approved Discussion — n/a, makes an already-advertised API actually work

## Screenshots / test output

New `tests/integration/search/pagination.test.ts` covers: nextCursor presence past the limit, omission on the final page, a cursor walk over disjoint pages covering every match exactly once, malformed-cursor rejection, and single-collection pagination.

```
Test Files  8 passed (8)
     Tests  45 passed (45)
```

### DIFF
--- a/.changeset/search-pagination.md
+++ b/.changeset/search-pagination.md
@@ -11,8 +11,8 @@ Both `search()` and `searchCollection()` now honour `cursor` and return a `nextC
 ```ts
 let cursor: string | undefined;
 do {
-  const { items, nextCursor } = await search("quarterly report", { limit: 20, cursor });
-  render(items);
-  cursor = nextCursor;
+	const { items, nextCursor } = await search("quarterly report", { limit: 20, cursor });
+	render(items);
+	cursor = nextCursor;
 } while (cursor);
 ```
--- a/.changeset/search-pagination.md
+++ b/.changeset/search-pagination.md
@@ -1,0 +1,18 @@
+---
+"emdash": minor
+---
+
+Implements pagination for `search()` and `searchCollection()`
+
+`search()` advertised keyset pagination through its types (`options.cursor`, `SearchResponse.nextCursor`) but never read the cursor or returned one, so results were silently capped at `limit` with no way to load a second page.
+
+Both `search()` and `searchCollection()` now honour `cursor` and return a `nextCursor` whenever more matches exist. Pass the previous `nextCursor` back as `cursor` to walk subsequent pages; it becomes `undefined` on the last page. The `/_emdash/api/search` endpoint accepts a `cursor` query parameter and returns `nextCursor` in its response (a malformed cursor returns a 400 `INVALID_CURSOR`).
+
+```ts
+let cursor: string | undefined;
+do {
+  const { items, nextCursor } = await search("quarterly report", { limit: 20, cursor });
+  render(items);
+  cursor = nextCursor;
+} while (cursor);
+```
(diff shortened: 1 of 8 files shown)


ascorbic (approved):
Thanks!

Issue #1042, opened 2026-05-14
Setup probe redirects public pages to `/_emdash/admin/setup` on D1 transient errors / cold starts


takustaqu:
### Description

When the EmDash setup probe fails due to a transient D1 error (e.g., on cold start), **all requests** — including fully prerendered public pages like `/about` and `/privacy` — are redirected to `/_emdash/admin/setup`, regardless of the request path.

- **What happened**: Navigating to a public page shows "Redirecting from /about/ to /_emdash/admin/setup" for ~2 seconds, then the browser is taken to the admin setup wizard.
- **What I expected**: Public pages should render normally. The setup redirect should only occur for requests to `/_emdash/*` paths.

### Steps to reproduce

1. Deploy an Astro hybrid site (mix of prerendered + SSR pages) with EmDash to Cloudflare Workers.
2. Configure a Cloudflare Zone Route (`example.com/*` → Worker) so all requests go through the Worker — a common production setup.
3. Wait a few minutes for the Worker isolate to be evicted (triggering a cold start).
4. Navigate to any public page (e.g., `/about`) in a browser.
5. Observe: after ~2 seconds, the page redirects to `/_emdash/admin/setup`.

**Quick verification via curl:**

```bash
curl -sL https://example.com/about | head -5
```

Returns:

```html
<!doctype html>
<title>Redirecting to: /_emdash/admin/setup</title>
<meta http-equiv="refresh" content="2;url=/_emdash/admin/setup">
```

> **Note:** The response status is `200 OK`, not `302`. This means any middleware that inspects `response.status` to intercept setup redirects will not catch it.

### Environment

- emdash version: 0.12.0
- Node.js version: 22.x
- Runtime: Cloudflare Workers (`@astrojs/cloudflare` adapter)
- Astro: 6.x (`output: "hybrid"`)
- Database: Cloudflare D1 with Sessions API
- OS: macOS (dev) / Cloudflare edge (production)

### Logs / error output

```shell
From `wrangler tail` at the time of the incident:


D1_ERROR: no such table: _emdash_migrations


> **Note:** The `_emdash_migrations` table **does exist** in the database. This is a transient D1 Sessions API initialization error on cold start, not a genuine "not set up" state.

**Problematic code** (`dist/astro/middleware.mjs`):


} catch {
  // ❌ Redirects ALL paths to setup — including public pages
  return context.redirect("/_emdash/admin/setup");
}


**Proposed fix:**


-} catch {
[...]

PR #1581, opened 2026-06-22
fix(core): skip setup probe during prerender so it isn't baked into static pages

naota70:
## What does this PR do?

The anonymous setup probe redirects to `/_emdash/admin/setup` when the migrations
table is missing. During **prerender** the build database is legitimately empty
(CI / first deploy), so the probe reports a missing table and returns
`context.redirect(...)` — which a prerendered route serializes into static HTML,
shipping a redirect to the setup wizard on **every prerendered page** in
production. There is no live visitor to send to the wizard at build time, so this
skips the probe entirely when `context.isPrerendered` is true (session reads are
already skipped for prerender for the same reason). Live SSR requests are
unaffected.

Reproduced deterministically on a Cloudflare Workers build (Workers Builds) with
an empty CI D1: prerendered pages were emitted as setup-redirect HTML.

Related: #1042 (the runtime/transient variant of this probe redirecting
incorrectly, addressed by #1345). This PR covers the build-time/prerender case.

## Type of change

- [x] Bug fix

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (packages/core: 4023 passed; added a failing-without-fix regression test)
- [ ] User-visible strings are wrapped for translation (N/A — no admin UI strings)
- [x] I have added a changeset (`emdash` patch)

- [x] This PR includes AI-generated code — model/tool: Claude Opus 4.8 (1M context)

## Screenshots / test output

`pnpm --filter ./packages/core test` → 274 files, 4023 passed.
New regression test fails on `main` (probe redirects during prerender) and passes
with this change.

### DIFF
--- a/.changeset/fix-setup-probe-prerender-redirect.md
+++ b/.changeset/fix-setup-probe-prerender-redirect.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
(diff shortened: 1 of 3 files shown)


naota70:

ascorbic (approved):
Thanks

Issue #1446, opened 2026-06-12
bug(0.18): catch-all routes silently return null from getEmDashEntry - 'route cache overwritten' warning


Vallhalen:
## Summary

After upgrading from `0.16.1` → `0.18.0`, all CMS-driven pages served via catch-all routes (`src/pages/[slug].astro`, `src/pages/en/[slug].astro`) silently return `null` from `getEmDashEntry`, causing them to fall through to a 404 / redirect. Pages with dedicated `.astro` files (e.g. `regulamin.astro`) continue to work fine.

Reproduced **twice** with deploy → rollback → re-deploy cycle (versions `f3ce6777` and `5d84583b`). Same affected pages, same warning in `wrangler tail`. Same site, same content - only the EmDash version differs.

## Repro

1. Site uses two catch-all routes that resolve CMS entries via `getEmDashEntry`:
 - `src/pages/[slug].astro` - Polish (default locale)
 - `src/pages/en/[slug].astro` - English (locale `en`)
2. Upgrade `emdash`, `@emdash-cms/cloudflare`, `@emdash-cms/admin` from `0.16.1` to `0.18.0` (cloudflare adapter, D1).
3. `npm run build && wrangler deploy`.
4. Hit any CMS-driven page (e.g. `/konsultacje`, `/en/terms`).

Result: `getEmDashEntry("pages", "konsultacje")` returns `{ entry: null }`. Rendered routes fall through to `Astro.redirect("/404")` (302) or `Astro.rewrite("/404")` (404).

## Warning observed via `wrangler tail --format pretty`

Every request to an affected page logs:

```
[WARN] Internal Warning: route cache overwritten. (/[slug]_src/pages/[slug].astro)
[WARN] Internal Warning: route cache overwritten. (/en/[slug]_src/pages/en/[slug].astro)
[WARN] [cache] `cache.set()` was called but caching is not enabled.
 Configure a cache provider in your Astro config under `experimental.cache` to enable caching.
```

The `(.../path/file.astro)` suffix is unusual - the leading `/` and missing separator before `_src` suggest some kind of internal cache-key derivation that's clobbering itself. Both PL and EN catch-alls produce the warning, and both are affected.

## Working vs broken in production

Site: chcedointernetu.pl (long-migrated EmDash instance, 0.5 → … → 0.16.1 chain).

| URL | Type | 0.16.1 | 0.18.0 |
|---|---|---|---|
[...]

PR #1603, opened 2026-06-24
fix(core): split SEO fetch from loadEntry to avoid D1 result-set column limit on wide collections

### LINKED ISSUE #1600: Loader 'SELECT c.*' + SEO LEFT JOIN crosses D1 result-set column limit on wide collection schemas (silent null entry)
## Summary

`getEmDashEntry()` and `getEmDashCollection()` fail silently with `D1_ERROR: too many columns in result set` when a collection's `ec_<name>` table has roughly 95+ columns, because the loader always selects `c.*` plus 5 SEO alias columns from the `LEFT JOIN _emdash_seo`. D1's per-query result set column limit (~100) makes wide flat-schema collections unreadable.

Surfaces as: entries the admin can see and edit just fine, but `getEmDashEntry("pages", "home")` returns `{ entry: null, error }` on the site. The error message is real but easy to miss because `JSON.stringify(error)` returns `"{}"` for `Error` objects (`.message` is non-enumerable).

## Repro

1. Create or migrate into a collection with ~95+ scalar fields (common when porting from WordPress / ACF / similar where every section is a top-level field, e.g. `hero_headline`, `sec_dlaczego_headline`, `sec_about_headline`, `whyus_1_title`, `whyus_2_title`, etc.).
2. Deploy on Cloudflare D1.
3. Call `getEmDashEntry("<collection>", "<slug>")` from a page template.
4. Returns `null` with an error message buried inside an `Error` instance.

Direct SQL `SELECT * FROM ec_<collection> WHERE slug = ?` works fine. The failure is specifically the loader's `SELECT c.*, <5 SEO aliases>` shape.

## Diagnosis

From [`loader-*.mjs` `loadEntry`](https://github.com/emdash-cms/emdash/blob/main/packages/emdash/src/loader.ts) (paraphrased, line numbers shift across versions):

```sql
SELECT c.*,
       s.seo_title AS _emdash_seo_title,
       s.seo_description AS _emdash_seo_description,
       s.seo_image AS _emdash_seo_image,
       s.seo_canonical AS _emdash_seo_canonical,
       s.seo_no_index AS _emdash_seo_no_index
FROM <ec_table> AS c
LEFT JOIN _emdash_seo AS s ON s.collection = ? AND s.content_id = c.id
WHERE c.deleted_at IS NULL
  AND ((c.slug = ? AND c.locale = ?) OR c.id = ?)
LIMIT 1
```

For an `ec_pages` table with N scalar columns, the result set has **N + 5** columns. D1 throws `too many columns in result set: SQLITE_ERROR` once N + 5 crosses ~100. The loader's try/catch wraps it as `{ error: new Error("Failed to load entry: ...") }` and the call site returns `null + error`.

In our case (long-migrated production instance, verified just now via `wrangler d1 execute --remote`):

- `ec_pages` has **98 columns** (14 system + 79 user-defined; the 5 SEO fields live in the separate `_emdash_seo` table and are only joined into the result set)
- `ec_pages.*` + 5 SEO LEFT JOIN aliases = **103 columns in the result set**, D1 errors out
- Other collections (`ec_services` 66 cols, `ec_posts` 19 cols, etc.) are fine
- Caught the real error by logging `error.message` instead of `JSON.stringify(error)`:
  ```
  Failed to load entry: D1_ERROR: too many columns in result set: SQLITE_ERROR
  ```

## Workarounds we considered (and why they don't help)

The natural first instinct is to drop a few unused columns to get the result set under the limit. For our `ec_pages` that would mean dropping at least 4 fields, which isn't viable on a long-migrated production schema where templates and content depend on those fields.

Worse, this is the bigger problem. **D1 also enforces a ~100 column limit on `ALTER TABLE ADD COLUMN`**. Trying to add the 101st column on a wide collection fails outright. So even if we accepted schema refactor as a workaround for *this specific bug*, the same underlying constraint comes back the moment we try to extend the collection with new fields. There is no in-place escape; the loader query shape is the real issue.

## Why this matters

- Common path for users coming from WordPress / ACF / page builders where every section is a top-level field.
- Per-collection `ec_*` tables (introduced in 0.20) make this surface visible in a way the previous JSON `data` blob did not.
- The failure is silent at the call site (`null` entry, error swallowed into a non-enumerable `.message`) so it looks like a data or migration problem rather than a query problem.
- D1 has the strictest result-set column limit among the supported backends, so this hits Cloudflare deployments first.

## Suggested fixes

Two non-exclusive options:

**1. Select only the columns the loader actually consumes.**
`loadEntry` / `loadCollection` read a known small set of system columns plus whatever `mapRowToData` reflects over. Selecting a stable allow-list (`id`, `slug`, `status`, `locale`, `published_at`, `scheduled_at`, `deleted_at`, `version`, `live_revision_id`, `draft_revision_id`, `translation_group`, `created_at`, `updated_at`, `primary_byline_id`, `author_id`) plus the JSON `data` column plus the SEO aliases would keep the result set width O(1) regardless of how wide `ec_*` gets.

This is the cleanest fix and probably backward-compatible since `mapRowToData` already maps by known keys.

**2. Fetch SEO with a separate query when the collection table is wide.**
Detect `ec_*` table column count once (cached) and switch to two queries (`SELECT c.*`, then `SELECT FROM _emdash_seo`) when it exceeds a threshold (say 85). Slightly more code, no behavior change at the public API.

Either approach protects long-lived flat-schema sites without requiring users to refactor their schemas into JSON repeaters.

## Adjacent quality-of-life

While in the area, consider rethrowing or re-wrapping the D1 error message so callers see something better than a generic `null`. Specifically, calling `error.message` (or `String(error)`) in the docs / examples instead of `JSON.stringify` would help users diagnose this class of issue in minutes instead of weeks. `Error` is famously non-enumerable.

## Repro instance

Happy to share a sanitized D1 dump or set up a repro repo if useful. The exact schema width and the surrounding tail of `wrangler tail` should be enough to repro deterministically.

Happy to PR option (1) if there's interest.

Diagnosed during work on #1446 (which was about a different bug, wedged-isolate cache, fixed by #1489 / #1530, but exposed this one when those fixes landed).

Vallhalen:
## What does this PR do?

Fixes a silent `null` entry from `getEmDashEntry()` on Cloudflare D1 when the collection's `ec_*` table is wide. The loader's single-query `LEFT JOIN _emdash_seo` added 5 alias columns to every result set, which pushed collections with ~95+ flat user fields past D1's per-query column limit (~100). D1 threw `D1_ERROR: too many columns in result set`, the loader's try/catch wrapped it as a generic `Failed to load entry` error, and the call site surfaced `null` (with the real message hidden inside a non-enumerable `Error.message`, so `JSON.stringify(error)` looked empty and made the failure mode hard to spot).

SEO is now fetched as a separate follow-up query and folded onto the row using the same alias names `extractSeo()` reads. Result-set width is bounded regardless of how many fields the collection has.

- One extra round trip per `loadEntry`. `loadCollection` was already join-free, so this brings the two paths into the same shape.
- No public-API change, `entry.data.seo` is identical before/after.
- `extractSeo()` still returns `null` for entries with no `_emdash_seo` row, preserving prior behavior.

Closes #1600

Background context (different bug, exposed this one once it was fixed): #1446.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (or targeted tests for my change), all 55 loader tests pass, including the 3 new ones in `loader-wide-collection.test.ts`
- [x] I have added/updated tests for my changes (if applicable), new file `packages/core/tests/unit/loader-wide-collection.test.ts` covers a 95-user-field collection across both dialects, with and without a SEO row
- [ ] User-visible strings in the admin UI are wrapped for translation, N/A, no admin UI changes
- [x] I have added a [changeset](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#changesets), `.changeset/fix-d1-result-set-column-limit.md` (patch)
- [ ] New features link to an approved Discussion, N/A, this is a bug fix

- [x] This PR includes AI-generated code, model/tool: Claude Opus 4.7 (via Claude Code), implementing the user-proposed fix from the linked issue

## Screenshots / test output

```
$ pnpm test loader

 RUN  v4.1.5 C:/tmp/emdash/packages/core

 Test Files  9 passed (9)
      Tests  55 passed (55)
   Start at  12:23:09
   Duration  2.44s
```

The new tests in `loader-wide-collection.test.ts`:

- `loads an entry from a collection with 95+ flat user columns`
- `still attaches data.seo on wide collections (SEO follow-up query)`
- `omits data.seo when no SEO row exists, even on wide collections`

Both dialects (SQLite + Postgres) covered via `describeEachDialect`.

## Notes for reviewers

The diff in `loader.ts` is the body of `loadEntry`'s table read: drops the `LEFT JOIN _emdash_seo` + `seoSelect` columns from the main query, then issues a small `SELECT seo_title, seo_description, seo_image, seo_canonical, seo_no_index FROM _emdash_seo WHERE collection = ? AND content_id = ?` after the row is found. The retrieved SEO values are copied onto the row under the same `_emdash_seo_*` keys the join used, so `extractSeo()` keeps working without changes.

The comment above the query has been updated to explain why we don't join (D1 column limit) so the next person who finds it doesn't undo this for "one less round trip".


Vallhalen:
## Verified on production with pkg.pr.new snapshot 🎯

Installed the `@1603` snapshot on our long-migrated production instance (`chcedointernetu.pl`) and ran a full stress test to confirm the fix holds end-to-end against Cloudflare D1:

```
npm i https://pkg.pr.new/emdash@1603 \
      https://pkg.pr.new/@emdash-cms/cloudflare@1603 \
      https://pkg.pr.new/@emdash-cms/admin@1603
```

**Test methodology:**

1. PITR'd D1 to a known-good bookmark.
2. Installed the PR snapshot, ran build + deploy.
3. **Restored the 8 unused columns** we had originally dropped to work around the bug (the workaround documented in the linked issue), bringing `ec_pages` back to **98 columns**, the exact width that previously failed in 0.20+ stock.
4. Hit homepage and all 9 originally-failing CMS routes.

**Result with the fix (98 columns):**

```
=== homepage ===
len: 221002 (full content, 0.16.1 baseline)
Sklepu internetowego: True   (hero typed strings)
Chcesz do internetu : True   (hero headline)
Konrad Krawczuk     : True   (about section)
inaczej?            : True   (sec_dlaczego section)
FAQ                 : True
h2 sections: 11

=== 9 CMS pages ===
200  /uslugi/strony-www
200  /uslugi/sklepy-internetowe
200  /portfolio
200  /blog
200  /o-mnie
200  /konsultacje
200  /uslugi/wdrozenia-ai
200  /uslugi/reklamy
200  /uslugi/optymalizacja
```

All correct. With the prior `LEFT JOIN _emdash_seo` shape, the same 98-column table would have produced a 103-column result set and failed with `D1_ERROR: too many columns in result set`, returning a silent `null` entry (this was exactly the production behavior before the fix). Cleaned up the stress columns afterwards.

## Bonus finding while stress-testing: D1's column limit applies to `ALTER TABLE` too

When I tried to push past 100 columns to verify the result-set width doesn't matter even at the extreme, D1 refused the `ALTER TABLE ADD COLUMN` itself:

```
$ wrangler d1 execute chcedointernetu --remote \
    --command "ALTER TABLE ec_pages ADD COLUMN stress_test_3 TEXT"
ERROR: too many columns on sqlite_altertab_ec_pages: SQLITE_ERROR [code: 7500]
```

So D1's per-table column ceiling sits around **100** (not the 2000 SQLite default). That means once a collection's `ec_*` table reaches ~95 user fields, two things happen simultaneously:

1. The old `SELECT c.*, <5 SEO aliases>` query crosses the result-set limit and `getEmDashEntry` starts returning silent nulls (what this PR fixes).
2. Adding any further field via the admin would fail with `too many columns on sqlite_altertab_<table>` even before this PR.

This PR keeps `loadEntry` working right up to D1's hard ceiling. Beyond that, the editor experience will break before the loader does, useful to know for migration guidance, but it doesn't change anything about this fix.

## Side effect: +1 round trip per `loadEntry`

Saw the `Query-count snapshot changes` workflow flag +8 queries across 8 routes (1 extra per `loadEntry` for the SEO follow-up). That's the intended trade-off, `loadCollection` was already SEO-join-free, so this brings the two paths into the same shape. Happy to switch to an adaptive variant (`SELECT c.*` + `LEFT JOIN` when narrow, separate query when ≥ N columns) if maintainers prefer to preserve the per-row round-trip count for typical schemas. Just let me know which direction you'd prefer and I can push another commit.

Vallhalen:
Rebased onto `main` after #1619 landed (which folded byline + taxonomy hydration into `loadEntry`'s query). The two changes compose cleanly: folded hydration adds two aggregated JSON columns (`_emdash_terms`, `_emdash_bylines`), and this PR drops the five SEO alias columns by moving SEO into a follow-up query. Net result for the single-entry path is `c.* + 2 aggregated columns`, which stays bounded regardless of how wide `ec_*` gets.

All 59 loader tests pass on the rebased branch (10 files, including the 3 new ones in `loader-wide-collection.test.ts`). `pnpm typecheck` and `pnpm lint` clean. `query-counts.snapshot.{d1,sqlite}.json` are unchanged from `main` (the SEO removal is offset by the folded hydration savings on the routes that hit `loadEntry`).

The bot also flagged overlap with #1399. Looked at it: that PR touches `loader.ts` for taxonomy-defs caching (different region of the file), so it should rebase cleanly on top of this one or vice versa. Happy to defer to whichever order the maintainers prefer.

Vallhalen:
Thanks for the careful read. Both stale comments addressed in 4484337e:

- `SEO_COLUMN_ALIASES` JSDoc rewritten to describe the follow-up-query design and explain the D1 column-limit rationale (so the next reader does not undo the split for "one less round trip").
- `loader-seo.test.ts:14-21` comment updated to match: the loader now fetches SEO in a follow-up query, and the dialect-sensitivity note shifted from "the LEFT JOIN SQL" to "the follow-up SEO query."

Tests still pass (59/59 across 10 loader files), typecheck and lint clean.

ascorbic:
I am trying to avoid these additional per-entry queries because they can really balloon the rendering time. Is there any way we can get around this?

Vallhalen:
Good push back, the extra round trip is the obvious cost and I was hoping someone would call it before this landed.

A cleaner direction: keep SEO joined in the single query, but collapse the 5 alias columns into one JSON column so the result set width stays bounded. D1's per-result-set cap is on column count, not bytes, so this stays safe on any schema width, and we drop the second round trip entirely.

Sketch (in the same spot in `loader.ts`):

```ts
const seoSelect = sql`(
  SELECT json_object(
    'seo_title', s.seo_title,
    'seo_description', s.seo_description,
    'seo_image', s.seo_image,
    'seo_canonical', s.seo_canonical,
    'seo_no_index', s.seo_no_index
  )
  FROM ${sql.ref("_emdash_seo")} AS s
  WHERE s.collection = ${type} AND s.content_id = c.id
  LIMIT 1
) AS _emdash_seo_json`;

const result = await sql<Record<string, unknown>>`
  SELECT c.*, ${seoSelect}, ${termsSelect}, ${bylinesSelect}
  FROM ${sql.ref(tableName)} AS c
  WHERE c.deleted_at IS NULL
  AND ((c.slug = ${id} AND c.locale = ${locale}) OR c.id = ${id})
  LIMIT 1
`.execute(db);

const row = result.rows[0];
if (row?._emdash_seo_json) {
  const parsed = JSON.parse(row._emdash_seo_json as string) as Record<string, unknown>;
  for (const [col, alias] of Object.entries(SEO_COLUMN_ALIASES)) {
    row[alias] = parsed[col] ?? null;
  }
  delete row._emdash_seo_json;
}
```

This mirrors what `foldedHydrationSelects` already does for bylines and terms (each is a single JSON-aggregated column in the SELECT, then expanded in JS), so the SEO path lines up with the same shape rather than being a special case.

Two things worth verifying before I push it up:

1. **D1 SQLite version on `json_object()`.** D1 has JSON1 enabled (it ships with SQLite >= 3.38), so `json_object()` and `json_extract()` are available. Worth confirming on whatever D1 baseline you're targeting.
2. **`null` row case.** When there's no `_emdash_seo` row for the content, the correlated subquery returns NULL (not an empty object), so `row._emdash_seo_json` is null and `extractSeo()` keeps returning null exactly as it does today.

Happy to push this onto the same branch and rerun the wide-collection tests if you're good with the direction.

ascorbic:
Yeah, a JSON column is probably the best plan. it's what I've used a lot to reduce round-trips.

ascorbic (approved):
Thanks!

PR #1616, opened 2026-06-24
fix(plugins): wire media:write upload() via storage

MA2153:
## What does this PR do?

Plugins declaring the `media:write` capability received a **read-only** `ctx.media` with no `upload()`. The plugin context factory only built a writable media surface (`MediaAccessWithWrite`) when an internal upload-URL provider (`getUploadUrl`) was wired into the context factory — but the runtime only ever threads `storage`, never `getUploadUrl`. As a result `ctx.media.upload()` was always `undefined`, and the `media:write` capability silently degraded to read-only with no signal at activation time. Any plugin producing derived media (thumbnails, transcodes, optimized variants, exports) was blocked.

The fix moves the gate to what `upload()` actually requires — `storage`, which the runtime already provides:

- Grant a writable media surface whenever `media:write` is declared **and** storage (or an explicit upload-URL provider) is configured.
- Derive `getUploadUrl()` from `storage` when no explicit provider is wired (pending record + signed PUT URL, mirroring the REST `/_emdash/api/media/upload-url` endpoint).
- When `media:write` is declared on a site with **no** storage, log a warning instead of silently handing back a read-only object.

No runtime wiring change was needed — the factory already receives `storage`, so fixing the gate corrects every call path.

Closes #

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] User-visible strings in the admin UI are wrapped for translation — n/a, no admin UI strings (the new server-side warning is an English-only console log, consistent with existing plugin logging)

## Screenshots / test output

Added 3 tests to `packages/core/tests/integration/plugins/capabilities.test.ts` (written test-first, confirmed failing before the fix):

- `media:write` + storage → `ctx.media.upload` / `getUploadUrl` defined
- `upload()` persists bytes to storage and creates a retrievable media record
- `media:write` with no storage → warns and stays read-only

```
Test Files  1 passed (1)
     Tests  67 passed (67)
```

Full `packages/core` suite: 4108 passed. Lint clean (0 diagnostics), typecheck passes.

### DIFF
--- a/.changeset/fix-plugin-media-write-upload.md
+++ b/.changeset/fix-plugin-media-write-upload.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes plugins declaring the `media:write` capability receiving a read-only `ctx.media` with no `upload()`. The plugin context only granted a writable media surface when an internal upload-URL provider was wired, which the runtime never supplied — so `ctx.media.upload()` was always `undefined`. Write access is now granted whenever storage is configured (all `upload()` needs), and `getUploadUrl()` is derived from storage when no provider is set. A `media:write` plugin on a site without storage now logs a warning instead of silently degrading to read-only.
--- a/packages/core/src/emdash-runtime.ts
+++ b/packages/core/src/emdash-runtime.ts
@@ -2965,6 +2965,7 @@ export class EmDashRuntime {
 		if (trustedPlugin && this.enabledPlugins.has(trustedPlugin.id)) {
 			const routeRegistry = new PluginRouteRegistry({
 				db: this.db,
+				storage: this.storage ?? undefined,
 				emailPipeline: this.email ?? undefined,
 				trustedProxyHeaders: getTrustedProxyHeaders(this.config),
 			});
--- a/packages/core/src/plugins/context.ts
+++ b/packages/core/src/plugins/context.ts
@@ -432,23 +432,55 @@ export function createMediaAccess(db: Kysely<Database>): MediaAccess {
 
 /**
  * Create full media access with write operations.
- * If storage is not provided, upload() will throw at call time.
+ *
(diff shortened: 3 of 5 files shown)


MA2153:
Addressed in 2f2610e3.

**[needs fixing] `emdash-runtime.ts` trusted route path** — Confirmed: `handlePluginApiRoute` built the `PluginRouteRegistry` without `storage`, so trusted `media:write` plugins invoked via `/_emdash/api/plugin/{id}/{route}` got a read-only/undefined `ctx.media`. Now passes `storage: this.storage ?? undefined`. Added a regression test (`tests/integration/runtime/plugin-media-route.test.ts`) that boots a real runtime with storage + a trusted `media:write` plugin whose route calls `ctx.media.upload()`, and asserts the bytes land in storage — it failed before the one-line change and passes after.

**[suggestion] stale `getUploadUrl` JSDoc** — Updated to describe that `getUploadUrl`/`upload` are derived from `storage` when no explicit provider is set, and that media writes are only unavailable when neither is present.

**[suggestion] warn spam + message accuracy** — The warning now fires once per factory instance (tracked by plugin id), so hook-participating plugins no longer log on every invocation. Also corrected the message: it no longer claims "media access is read-only" (which was only true when `media:read` is also declared) and instead states `upload() is unavailable`. Test updated to create the context twice and assert exactly one warning.

I kept the read-only fallback conditional on `media:read` rather than making it unconditional, to avoid granting a read surface a plugin didn't ask for; the message is now accurate for both cases.

All core tests pass (4109), lint clean, typecheck passes.

ascorbic (commented):
Thanks!

PR #1617, opened 2026-06-25
perf(core): batch cold-start init reads into one round trip

ascorbic:
## What does this PR do?

Reduces cold-start runtime-init latency by batching the reads `EmDashRuntime.create()` issues on a fresh isolate (the auto-seed gate, plugin-state load, and site-info load).

Previously these serialized on the singleton DB connection: its plain `SqliteAdapter` reports `supportsMultipleConnections: false`, so Kysely's connection mutex runs concurrent reads one at a time. This PR runs the read phase through a short-lived **coalescing** connection, so the same-tick reads collapse into a single `batch()` (D1) / `batchQuery` RPC (Durable Objects). Migrations still run on the non-coalescing singleton; the read connection is built per-init and discarded.

Measured on the demo sites via `infra/perf-monitor` (server-side `Server-Timing` `rt.*` phases, cold isolates across global probes):

| backend | baseline cold init (`rt`) | this PR |
| --- | --- | --- |
| D1 (`blog-demo`) | ~135ms | ~67ms (~2x) |
| Durable Objects (`do-demo`) | ~94ms / ~926ms (near/far region) | ~56ms / ~529ms (~40-45%) |

The tell: with the change, `rt.seedcheck`, `rt.plugins`, and `rt.site` report identical per-region durations (one batched round trip); on `main` `rt.site` is ~2x `rt.plugins` (serialized) and the seed gate sits inside a larger `rt.db`.

Backends without a coalescing dialect (Node/SQLite) transparently fall back to the singleton — synchronous, so batching is moot there.

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [x] Performance improvement
- [ ] Tests

## Implementation notes

- `packages/cloudflare/src/db/d1.ts` and `db/do-sql.ts` export `createCoalescingDialect`. On DO the singleton and the cold-start read connection **share one bookmark sink per database**, so reads issued right after a migration keep read-your-writes across replicas (the original relied on the singleton's sink). D1 needs no sink — its singleton and coalescing dialects both use the raw binding with no session, so reads hit the primary.
- `generateDialectModule` re-exports `createCoalescingDialect` via a namespace import, yielding `undefined` for backends that don't provide it (core then falls back to the singleton).
- Auto-seed is single-flighted per isolate via the existing `initWithLock` primitive (poll-based, workerd-safe), keyed by the configured database, so a reclaimed-and-rerun `create()` cannot apply the seed a second time.

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (4101 core + 185 cloudflare; plus 3 new runtime tests)
- [ ] User-visible strings wrapped for translation — n/a (no admin UI strings; no `messages.po` changes)
- [x] I have added a changeset (bumps `emdash` and `@emdash-cms/cloudflare`)
- [ ] New features link to an approved Discussion — n/a (performance improvement, not a feature)

- [x] This PR includes AI-generated code — model/tool: Claude Opus 4.8 (opencode). Authored and adversarially self-reviewed across multiple model passes; measured on real D1/DO deployments before submitting.

## Screenshots / test output

Cold `rt.*` from `perf-monitor` showed `rt.seedcheck == rt.plugins == rt.site` per region with this change (one batched round trip) vs. serialized phases on `main`.

### DIFF
--- a/.changeset/faster-cold-start-init-reads.md
+++ b/.changeset/faster-cold-start-init-reads.md
@@ -1,5 +1,6 @@
 ---
 "emdash": patch
+"@emdash-cms/cloudflare": patch
 ---
 
 Cuts cold-start latency by batching runtime initialization reads into a single round trip. On a fresh isolate (or Worker), the auto-seed check, plugin-state load, and site-info load now run as one batched query on backends that support it (Cloudflare D1, Durable Objects) instead of several serialized round trips. On D1 this roughly halves runtime init time on cold start. Other backends (Node/SQLite) are unaffected.
--- a/.changeset/faster-cold-start-init-reads.md
+++ b/.changeset/faster-cold-start-init-reads.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Cuts cold-start latency by batching runtime initialization reads into a single round trip. On a fresh isolate (or Worker), the auto-seed check, plugin-state load, and site-info load now run as one batched query on backends that support it (Cloudflare D1, Durable Objects) instead of several serialized round trips. On D1 this roughly halves runtime init time on cold start. Other backends (Node/SQLite) are unaffected.
--- a/packages/cloudflare/src/db/d1.ts
+++ b/packages/cloudflare/src/db/d1.ts
@@ -87,6 +87,22 @@ export function createDialect(config: D1Config): Dialect {
 	return new EmDashD1Dialect({ database: db });
 }
 
+/**
+ * Coalescing D1 dialect for the runtime's cold-start read phase, where the
+ * core runtime batches its init reads into one `batch()` round trip. Carries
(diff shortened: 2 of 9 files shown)


ascorbic:
Addressed the review finding in bbe121cc: the DO singleton bookmark sink map now lives on `globalThis` behind a `Symbol.for("emdash:do-singleton-bookmark-sinks")` key, matching core's `request-cache.ts` / `settings/index.ts` pattern. This guarantees the singleton dialect and the cold-start coalescing dialect resolve the same `BookmarkSink` even if Vite duplicates the module across SSR chunks, preserving read-your-writes between them.

The Integration Tests failure on the prior run was the Astro dev-server lock collision flake ("Another astro dev server is already running" on port 4398), unrelated to this change. The fresh push re-runs it.

Issue #1441, opened 2026-06-12
getEmDashEntry/getEmDashCollection hydrate taxonomy terms using request/default locale instead of explicit query locale


diogoascarneiro:
### Description

In EmDash `0.18.0`, entries returned by `getEmDashEntry()` / `getEmDashCollection()` now include hydrated taxonomy terms on `entry.data.terms`.

However, when calling these helpers with an explicit `locale`, the returned content entry is correctly localized, but `entry.data.terms` can contain taxonomy terms from the default locale.

Expected behavior: taxonomy term hydration should use the same resolved locale as the content query.

Actual behavior: term hydration appears to call `getAllTermsForEntries(type, ids)` without passing the resolved query locale, so term lookup falls back to `requestContext.locale` or default locale. On normal public routes, the EmDash request context often does not include `locale`, so terms hydrate in the default locale.

### Steps to reproduce

```ts
const { entry } = await getEmDashEntry("posts", "some-slug", {
  locale: "fr-fr",
});

// or

const { entries } = await getEmDashCollection("posts", {
  locale: "fr-fr",
});
```

6. Inspect hydrated terms:

```ts
console.log(entry?.data.terms?.tag);
```

Expected result:

```ts
[
  {
    label: "Securite",
    locale: "fr-fr",
    // ...
  },
]
```

Actual result:

```ts
[
  {
    label: "Security",
    locale: "en-us",
    // ...
  },
]
```

The issue appears to be that `getEmDashEntry()` / `getEmDashCollection()` correctly resolve the content query locale, but `hydrateEntryTerms()` calls:

```ts
getAllTermsForEntries(type, ids);
```

instead of something like:

```ts
getAllTermsForEntries(type, ids, { locale: resolvedLocale });
```

### Environment

- emdash version: `0.18.0`
- Node.js version: `22.22.2`
- Runtime: Cloudflare Workers
- OS: macOS

### Logs / error output

```shell
No thrown error. The bug is behavioral: localized entries are returned with default-locale hydrated taxonomy terms.
```

PR #1619, opened 2026-06-25
perf(core): fold byline + taxonomy hydration into the content query

ascorbic:
## What does this PR do?

Folds byline and taxonomy-term hydration into the content query. `loadEntry`/`loadCollection` previously issued three round trips per fetch — the content row, then a `_emdash_content_bylines` join, then a `content_taxonomies` join. This adds the hydration as correlated JSON-array subqueries on the content query (generalizing the existing `_emdash_seo` join), so a fetch is **one** round trip.

Each subquery correlates on `<entry>.id`, so the base query stays one row per entry — no join fan-out and no duplicated content payload (a naive `LEFT JOIN` of two many-to-many relations would Cartesian-explode and repeat the portable-text body on every row).

Measured per-page statement counts (sqlite query-counts harness), `main` vs this PR:

| route | before | after |
| --- | --- | --- |
| `/` (index) | 11 | 6 |
| `/posts/[slug]` | 24 | 17 |
| `/pages/about` | 10 | 5 |
| `/category/[slug]` | 15 | 10 |
| `/rss.xml` | 5 | 1 |
| `/contributors-naive` | 18 | 5 |

Each `getEmDashEntry`/`getEmDashCollection` drops its byline round trip, its taxonomy round trip, and the taxonomy-defs applicability read.

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [x] Performance improvement
- [ ] Tests

## Implementation notes

- **Dialect-aware aggregation** in `foldedHydrationSelects` (loader.ts): SQLite `json_group_array`/`json_object` (returns a JSON string) vs Postgres `json_agg`/`json_build_object` coalesced to `[]` (returns parsed JSON). `stashFolded` handles both. Ordering is done in the consumer (terms by label, credits by sortOrder), not in the aggregate, since the dialects differ there.
- **Locale correctness (#1441):** the terms subquery correlates on the entry's own locale (`t.locale = <entry>.locale`), so an `fr` entry gets its `fr` term variants — covered by a test on both dialects.
- **Byline fallbacks:** the fold expresses the common case (explicit credits). The consumer falls back to the existing query path when an entry has a byline reference (`authorId`/`primaryBylineId`) but no folded credit (author-fallback, or a locale-mismatched credit), or when byline custom fields are registered (which the fold can't carry).
- New `tests/integration/loader-fold.test.ts` runs under `describeEachDialect` (entry + list, bylines + terms, locale variant, empty case) — validated on SQLite **and** a local Postgres.

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes — 4109 sqlite + full Postgres parity (4231 with `EMDASH_TEST_PG`), 185 cloudflare
- [ ] User-visible strings wrapped for translation — n/a (no admin UI strings)
- [x] I have added a changeset (`emdash`, patch)
- [ ] New features link to an approved Discussion — n/a (performance improvement)

- [x] This PR includes AI-generated code — model/tool: Claude Opus 4.8 (opencode). Validated on both SQLite and a local Postgres; a self-review caught and fixed an i18n term-locale gap and a null-locale byline fallback gap before submission. An adversarial cross-model review pass is recommended.

## Notes

- The sqlite query-counts snapshot is updated here; the **d1** snapshot will be regenerated by CI (the fold applies equally to D1, which is SQLite-backed).
- Follow-up opportunities not in this PR: fold the author-fallback into SQL (currently a query-path fallback), and chrome cross-request caching.

---

### Try this PR

**[Open a fresh playground →](https://perf-fold-hydration-into-content-emdash-playground.emdash-cms.workers.dev)**

<sub>Tracks `perf/fold-hydration-into-content`. Updated automatically when the playground redeploys.</sub>


ascorbic:
Addressed all four review findings in 12c6e950:

1. **avatarMediaId dropped** — added `'avatarMediaId', b.avatar_media_id` to the folded byline JSON (loader.ts). Templates read `credit.byline.avatarMediaId` to render avatars, so this restores the public shape on the fast path.

2. **Stale SQLite query-counts snapshot** — regenerated with `node scripts/query-counts.mjs --target sqlite --update`. Both `query-counts.snapshot.sqlite.json` and `.queries.sqlite.json` now match the emitted SQL (no `ORDER BY` in the aggregate, term-locale correlation present). `pnpm query-counts` passes. Net vs main is a clear reduction everywhere (article 24->16, index 11->7); the previously committed snapshot was the earlier-revision artifact you flagged. Changeset numbers updated to match (16/7).

3. **Folded terms didn't prime the request cache** — the fast path now calls a new `primeFoldedEntryTerms` helper (exported from taxonomies/index.ts) that reuses `primeEntryTermsCache`, so it seeds the exact same wildcard + per-name keys (including `[]` for applicable taxonomies absent from the entry) as `getAllTermsForEntries`. Subsequent `getEntryTerms` calls hit the cache. Keeping the key logic in the taxonomies module avoids drift.

4. **Byline probe swallowed non-missing-table errors** — `needsQueryPath` is now `let`; on any error other than a missing table the probe falls back to the query path rather than serving `customFields: {}`.

Plus the requested **public-shape coverage**: `query-fold-public-shape.test.ts` asserts `data.bylines[0].byline.avatarMediaId` via `getEmDashEntry`, and that folded terms prime the cache (it seeds a term absent from `content_taxonomies` and confirms `getEntryTerms` still returns it). The loader-fold integration test also now seeds an avatar and asserts it survives the fold.

Lint clean, typecheck passes, all 4111 core tests pass, formatted.

ascorbic:
Follow-up on the query-count diff (thanks for catching this):

**`/contributors-naive` going 6→13 (SQLite) / 6→14 (D1) is correct, not a regression.** That fixture route does a per-avatar `mediaRepo.findById(credit.byline.avatarMediaId)`. The original fold dropped `avatarMediaId`, so `mediaId` was `undefined` and every avatar lookup silently no-op'd — the route wasn't doing the work it exists to measure. Restoring `avatarMediaId` means those lookups run again, so the count legitimately rises. (That's exactly the cost the `/contributors` JOIN is meant to remove.)

**The +1 my priming added to `/contributors`, `/pages/about`, `/rss`, `/search` was a real regression, now fixed** (030a15ba). The priming was seeding `[]` for taxonomies applicable-but-absent on the entry, which needed a `getTaxonomyDefs` query every fold render. I dropped that: priming now seeds only the wildcard + present-taxonomy keys straight from the folded data, no DB lookup. `getEntryTerms(id, absentTaxonomy)` falls through to its own cached query (the rarer case).

Net vs `main` is back to a clean reduction everywhere — index 11→6, article 24→17, `/contributors` 11→6, etc. The only route that rises vs the prior fold snapshot is `contributors-naive`, for the avatarMediaId reason above. CI will regenerate the D1 snapshot on this push.

PR #1636, opened 2026-06-26
fix(core): keep plugin cron alive in astro dev under Cloudflare adapter

### LINKED ISSUE #1635: Plugin ctx.cron is undefined under @astrojs/cloudflare in `astro dev` — plugin cron never schedules or fires
Under the `@astrojs/cloudflare` adapter running `astro dev`, a plugin's context `ctx.cron` is `undefined`. As a result plugins cannot schedule cron tasks (`ctx.cron.schedule(...)` is unreachable), the `_emdash_cron_tasks` table stays empty, and the plugin `cron` hook never fires — neither for any periodic work the plugin wants to do.

This contradicts the framework's documented intent for local dev.

### Environment

- `emdash` `0.23.0`
- adapter: `@astrojs/cloudflare`
- command: `astro dev` (local dev, workerd runtime)

### Expected

Per `generateSchedulerModule` (packages … astro integration), local dev — *including `astro dev` under the Cloudflare adapter* — is supposed to get a `NodeCronScheduler` so plugin cron still runs. Its own comment says:

> A *production* Cloudflare build has no persistent timers, so the Worker's `scheduled()` handler drives `runScheduledTasks()` instead and this exports `null`. Every other case — any other adapter (Node, Bun), and crucially local `astro dev` even under the Cloudflare adapter (no Cron Trigger fires in dev) — gets a `NodeCronScheduler` factory so plugin cron, scheduled publishing, and cleanup still run.

So `ctx.cron` should be defined in plugin route/hook contexts and scheduled tasks should fire on the in-process timer.

### Actual

In a plugin route/hook handler under `astro dev` + `@astrojs/cloudflare`:

- `ctx.cron === undefined`
- A call guarded by `if (ctx.cron) await ctx.cron.schedule(name, { schedule })` silently no-ops; the `_emdash_cron_tasks` table remains empty.
- The plugin's `cron` hook is never invoked, so periodic work never runs.

`ctx.content`, `ctx.media`, `ctx.kv`, and `ctx.storage` all work normally in the same handler — only `ctx.cron` is missing.

### Minimal reproduction

1. A trivial native plugin with (a) a `cron` hook and (b) a route or `plugin:activate` hook that calls `ctx.cron?.schedule("tick", { schedule: "*/1 * * * *" })`.
2. Register it in an Astro site using `@astrojs/cloudflare`.
3. `astro dev`, then trigger the route/activation.
4. Observe `ctx.cron` is `undefined`; the task is never persisted; the `cron` hook never fires.

(Confirmed by temporarily returning `{ hasCron: ctx.cron != null }` from a route handler → `hasCron: false`.)

### Diagnosis (pointers, not a fix)

- `PluginContextFactory.createContext` attaches `cron` only when `this.cronReschedule` is set:
  `if (this.cronReschedule) cron = new CronAccessImpl(db, plugin.id, this.cronReschedule);`
- The middleware's cron-init phase sets it only inside `if (deps.createScheduler) { … pipeline.setContextFactory({ cronReschedule }); scheduler.start(); }`.
- So `ctx.cron` is undefined whenever `deps.createScheduler` is falsy at runtime.
- `deps.createScheduler` comes from the `virtual:emdash/scheduler` module, whose body is decided by `generateSchedulerModule(adapterName, command)` — it returns `createScheduler = null` when `adapterName === "@astrojs/cloudflare" && command !== "serve"`.

The observable result is consistent with `createScheduler` resolving to `null` at runtime under the Cloudflare adapter's `astro dev` pipeline (e.g. if the command seen by `generateSchedulerModule` is `"build"` rather than `"serve"` when the adapter pre-bundles the worker for dev, or if the virtual module otherwise resolves to the production/null branch). That would skip the `setContextFactory({ cronReschedule })` call and leave `ctx.cron` undefined — exactly the symptom — even though the timer-firing behaviour is a separate concern.

### Impact

Any plugin that relies on cron (queue draining, periodic backfill/discovery, cleanup) is silently inert during `astro dev` under the Cloudflare adapter: it cannot schedule, and its `cron` hook never runs. The failure is silent (no warning), so it's easy to mistake for a plugin bug.

MA2153:
## What does this PR do?

Fixes plugin cron being silently inert during `astro dev` under the `@astrojs/cloudflare` adapter.

The scheduler virtual module (`virtual:emdash/scheduler`) decides whether to inject a `NodeCronScheduler` based on a command. It was reading Vite's `config.command`, which the Cloudflare adapter reports as `"build"` even during `astro dev` — the adapter produces the worker bundle via a nested Vite *build* pass. As a result `generateSchedulerModule` returned `createScheduler = null`, the middleware skipped `setContextFactory({ cronReschedule })`, and `ctx.cron` was left `undefined`. Plugins could not call `ctx.cron.schedule(...)`, `_emdash_cron_tasks` stayed empty, and the plugin `cron` hook never fired.

The fix threads Astro's authoritative `command` into the virtual-modules plugin and bases the scheduler decision on it (`dev` → in-process timer, everything else → null under Cloudflare). `astro dev` now keeps the `NodeCronScheduler` regardless of how the Cloudflare adapter pre-bundles the worker, so plugin cron, scheduled publishing, and cleanup run again in local dev.

Closes #1635

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] User-visible strings in the admin UI are wrapped for translation (if applicable) — n/a, no admin UI strings in this change

## Screenshots / test output

Added `createVirtualModulesPlugin scheduler wiring` tests reproducing the bug (Cloudflare adapter + Astro command `dev` while Vite reports `build`). Written test-first: the reproducing test failed (scheduler resolved to `null`) before the fix, passes after.

```
 Test Files  1 passed (1)
      Tests  15 passed (15)
```

### DIFF
--- a/.changeset/plugin-cron-undefined-in-dev.md
+++ b/.changeset/plugin-cron-undefined-in-dev.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes `ctx.cron` being undefined during `astro dev` under the `@astrojs/cloudflare` adapter, which silently prevented plugins from scheduling cron tasks and stopped the `cron` hook from ever firing in local dev. The in-process scheduler now keys off Astro's command rather than Vite's, so plugin cron, scheduled publishing, and cleanup run again.
--- a/packages/core/src/astro/integration/vite-config.ts
+++ b/packages/core/src/astro/integration/vite-config.ts
@@ -159,7 +159,10 @@ export interface VitePluginOptions {
 /**
  * Creates the EmDash virtual modules Vite plugin.
  */
-export function createVirtualModulesPlugin(options: VitePluginOptions): Plugin {
+export function createVirtualModulesPlugin(
+	options: VitePluginOptions,
+	astroCommand: "dev" | "build" | "preview" | "sync",
+): Plugin {
 	const { serializableConfig, resolvedConfig, pluginDescriptors, astroConfig } = options;
 
 	let viteCommand: "build" | "serve" | undefined;
@@ -294,8 +297,16 @@ export function createVirtualModulesPlugin(options: VitePluginOptions): Plugin {
 			// Generate scheduler module — a NodeCronScheduler factory on
 			// long-lived runtimes, or null under the Cloudflare adapter where
 			// a Cron Trigger drives scheduled work instead.
+			//
+			// Decide from Astro's command, not Vite's config.command: the
+			// Cloudflare adapter builds the worker bundle via a nested Vite
+			// *build* pass even during `astro dev`, so viteCommand reports
+			// "build" and would wrongly suppress the in-process timer (#1635).
+			// Astro's command stays "dev", which is the only case that should
+			// fall through to a NodeCronScheduler.
 			if (id === RESOLVED_VIRTUAL_SCHEDULER_ID) {
-				return generateSchedulerModule(astroConfig.adapter?.name, viteCommand);
+				const schedulerCommand = astroCommand === "dev" ? "serve" : "build";
+				return generateSchedulerModule(astroConfig.adapter?.name, schedulerCommand);
 			}
 		},
 	};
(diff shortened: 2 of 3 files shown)


ascorbic:
Thanks!

PR #1645, opened 2026-06-28
fix: include runtime manual collections in admin manifest so their content routes resolve

### LINKED ISSUE #1641: Collections created via Admin UI ("manual" source) return Page Not Found in admin — dashboard shows collection but content route is inaccessible
### Description

When a new collection is created through Admin > Content Types after initial deployment, the admin UI shows the collection on the dashboard (with correct entry counts) but navigating to /_emdash/admin/content/{slug} returns a "Page Not Found" error. The collection and its table are fully present in the database — the issue is isolated to admin routing.

The collection appears correctly on the frontend (the site's public pages query it successfully), so this is purely an admin UI routing issue.

### Steps to reproduce

1. Deploy an EmDash site to Cloudflare Workers with an initial seed (e.g. posts, pages)
2. After deployment, go to Admin > Content Types and create a new collection (e.g. currents, slug: currents, label singular: Current)
3. Add fields to the collection via the field editor
4. Save the collection — it appears on the admin dashboard with an arrow link
5. Click the arrow on the dashboard card, or navigate directly to /_emdash/admin/content/currents
6. Result: "Page Not Found" is displayed
7. Navigate to /_emdash/admin/content/currents/new — same result

Additional investigation:

Verified via wrangler d1 execute that the collection is fully registered:

- _emdash_collections contains the correct row with all fields populated
- ec_currents table exists with correct schema
- _emdash_fields contains all defined fields

Changing source from 'manual' to 'seed' directly in _emdash_collections via D1 did not resolve the issue. Re-deploying did not resolve it. The dashboard continues to show the collection and reflects correct entry counts when rows are inserted directly into the table.

By contrast, collections present in the seed at initial deploy time (posts, pages) are fully accessible in the admin — including when they have zero entries.

### Environment

emdash version: 0.6.0
Runtime: Cloudflare Workers (via @astrojs/cloudflare + @emdash-cms/cloudflare)
Database: Cloudflare D1
Storage: Cloudflare R2
Operating system: macOS
Browser: Safari 18

### Logs / error output

```shell
No server-side error logs visible. The admin UI renders a "Page Not Found" message client-side. D1 queries confirm the collection exists and is correctly structured — no database errors.
```

mvanhorn:
## Summary
Collections created through the Admin UI (`source: "manual"`) were absent from the admin manifest, so their content routes failed to resolve at runtime. This builds the manifest from the runtime manual collections via the `SchemaRegistry`, mapping schema field types to editor field kinds, so manual collections appear alongside file-defined ones.

Adds `manifest-build.test.ts` covering the manual-collection manifest path.

Fixes #1641

AI was used for assistance.


ascorbic:
@mvanhorn could you cook at the review feedback. It looks all valid to me

mvanhorn:
Addressed the review:

1. Added a changeset for the `emdash` package.
2. `ManifestResponse.collections` now uses `ManifestCollectionMap`, so the public type declares the extra fields (`hasSeo`, `urlPattern`, `id`, `widget`, `validation`) instead of omitting them. The types are centralized in `api/types.ts` and re-exported from the package root.

On the third point — I dug into `_buildManifest` and the runtime doesn't carry a config collection map at manifest-build time (collections are DB-derived, which already includes the manual ones), so `buildManifestCollections({}, db)` is correct there and the merge path isn't needed in production. Happy to revisit if you'd rather thread config through.

mvanhorn:
Rebased onto `main`; the branch is `MERGEABLE` again at `6e1a81ac`.

On the review feedback: I think it is already addressed. The three findings from the first review (missing changeset, `generateManifest()` returning objects wider than `ManifestResponse`, and the merge helper not wired to the production admin manifest) were fixed in July, and @emdashbot's re-review on 2026-07-21 approved and confirmed all three. Your comment predates those fixes, so if there is something else you wanted picked up, point me at it and I will.

The merge itself needed a real decision, so flagging it. All three conflicts were in files this PR owns, and `main` had meanwhile grown the inline manifest builder in `_buildManifest()` that this PR extracts. Resolving straightforwardly in favour of the extraction would have silently dropped five fields `main` had added since: `routable`, `titleField`, `dateField`, `hidden`, and `listColumns`.

So instead I brought `buildManifestCollections` up to parity: `routable` on both the config and DB paths, and `titleField` / `dateField` / `hidden` / `listColumns` on the DB path, including the `LIST_COLUMN_FIELD_TYPES` validation and the `MAX_COLLECTION_LIST_COLUMNS` cap with the same warnings. Those two constants moved with the logic, so they are gone from `emdash-runtime.ts`. `ManifestCollectionDescriptor` gained the corresponding optional fields.

Worth a look during review, since it means the shared helper is now the single definition of manifest shape rather than a subset of what the runtime used to build inline.

On checking: `tsc --noEmit` on `packages/core` is clean, `oxlint --type-aware --deny-warnings` on the three changed files is clean, and `manifest-build` plus `manifest-route` pass 16/16.

ascorbic (approved):
Thanks!

PR #1646, opened 2026-06-29
fix: store translation_group in taxonomy parent_id to stop cross-locale parent leak

### LINKED ISSUE #1347: Taxonomy term hierarchy uses locale-bound parent_id, causing cross-locale parent leak and silent tree flattening
## Description

Taxonomy term hierarchy (`taxonomies.parent_id`) is encoded as a **row id**, which is locale-bound, while every other relationship in the taxonomy i18n model keys off a locale-agnostic value (the taxonomy `name` string, or a `translation_group`). As a result, a child term can permanently end up parented to a term row in a *different* locale, which silently flattens the per-locale term tree with no error.

### The asymmetry

| Relationship | Stored as | Locale semantics |
|---|---|---|
| Term → taxonomy def | `taxonomies.name` (string) | locale-agnostic |
| Term → content entry | `content_taxonomies.taxonomy_id` = `translation_group` | locale-agnostic |
| Term → parent term | `taxonomies.parent_id` = **row id** | locale-**bound** |

A `parent_id` points at one term row, which lives in one locale. The hierarchy only holds together if every child's `parent_id` points at a parent row in the *same* locale, but nothing enforces that. `validateParentTerm` checks `parent.name === taxonomyName` (same taxonomy name), not same locale, so a cross-locale parent passes validation.

The only thing aligning locales is one best-effort block in `handleTermCreate` (`packages/core/src/api/handlers/taxonomies.ts`, ~L526–540): when creating a term translation, it resolves the source's parent `translation_group` and tries to swap in the same-locale sibling. It only fires when the translated parent **already exists**, and it only fixes the row being created — never existing children.

### Impact

- The per-locale tree silently flattens: `handleTermList` filters terms by locale, then `buildTree` only links a child if its `parentId` is in *that locale's* set (`taxonomies.ts` ~L101–107). A child whose `parent_id` references another locale's row is pushed to `roots` and appears top-level, with no error.
- The broken edge is permanent: nothing re-points existing children when a parent translation is created later. No backfill exists.

## Steps to reproduce

Assume a hierarchical taxonomy `category` with default locale `en` and a second locale `es`.

1. Create parent term `computing` (en) and child term `laptops` (en) with `parentId` = computing(en). Tree in `en` is correct.
2. Translate the **child** first: `POST /_emdash/api/taxonomies/category/terms/laptops/translations` with `{ "locale": "es", "slug": "portatiles" }`. Because the `es` translation of `computing` does not exist yet, the re-pointing guard finds no same-locale parent, so `portatiles`(es).`parent_id` is set to **`computing`(en)** — a cross-locale edge.
3. Now translate the **parent**: create `computacion` (es) as a translation of `computing`(en). This does **not** retroactively re-point `portatiles`(es).
4. `GET /_emdash/api/taxonomies/category/terms?locale=es`. Expected: `computacion` with child `portatiles`. Actual: `portatiles` appears as a **root** (its `parent_id` still references the `en` row, which is absent from the `es`-filtered set), so the Spanish tree is flattened.

There is no error or warning at any step.

### Suggested fix (root cause)

Store the parent's `translation_group` in `parent_id` instead of a row id — mirroring how `content_taxonomies.taxonomy_id` already works. Then `findChildren` / `buildTree` resolve `parent_group + locale → row`, translating a parent automatically adopts existing children in every locale, and the imperative re-pointing block plus its ordering hazard can be removed. This is a forward-only schema migration (change `parent_id` semantics + backfill existing values through `translation_group`) with backwards-compat implications, so it likely warrants a Discussion on the migration path.

## Environment

- emdash version: latest `main` (observed on commit `a6e8a918`)
- Runtime: Node / SQLite and Cloudflare D1 (logic is dialect-independent; affects both)
- Area: `packages/core` — taxonomy i18n

Relevant files:
- `packages/core/src/api/handlers/taxonomies.ts` (`validateParentTerm`, `handleTermCreate` re-pointing, `buildTree`)
- `packages/core/src/database/repositories/taxonomy.ts` (`findByName`, `findChildren`, `parent_id` handling)
- `packages/core/src/database/migrations/036_i18n_menus_and_taxonomies.ts` (i18n schema)

mvanhorn:
## What does this PR do?

Hierarchical taxonomy terms encoded a term's parent as a locale-bound row id, but every other taxonomy translation relationship keys off the locale-agnostic translation group. Because of that mismatch, a child term translated into a second locale could end up pointing at a parent row from a different locale. The per-locale tree then quietly collapsed: the child rendered at the root instead of under its parent, with no error to signal anything was wrong. A best-effort patch in the create path only helped when the translated parent already existed, and it never repaired children that were created earlier.

The fix is to treat parentage the same way the rest of the taxonomy code already treats cross-locale identity, by storing the parent's translation group rather than one locale's row. A child now links to the parent concept, so translating that parent later automatically re-nests the children that were waiting on it, in whichever locales they live. Reads resolve a parent within the child's own locale, so an unfiltered list that mixes locales keeps each child under the right parent instead of colliding on the shared group. Validation now compares groups too, which means a translation can no longer be parented to its own group, and a forward-only migration rewrites the parent links that already exist so current trees render correctly after upgrade.

This is backwards compatible: the migration is additive and forward-only, and the stored value stays a real row id (the group's anchor), so the existing foreign key on the column still holds.

Closes #1347

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] User-visible admin UI strings wrapped for translation (n/a: server-side handlers and a migration only, no admin UI strings)
- [x] I have added a changeset (this changes a published package)
- [ ] New features link to an approved Discussion (n/a: bug fix)

## Screenshots / test output

Lint, typecheck, and the taxonomy, migration, and seed suites pass (342 tests). New cases cover a child nesting under a parent that is translated after it, a mixed-locale list keeping each child under its own parent, the migration converting a legacy cross-locale link, and the rejection of invalid or self-referential parents.

AI was used for assistance.

### DIFF
--- a/.changeset/fix-1347-taxonomy-parent-locale.md
+++ b/.changeset/fix-1347-taxonomy-parent-locale.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes hierarchical taxonomy terms losing their parent in translated locales. A child term now stores its parent's translation group instead of a locale-bound row id, so translating a parent automatically re-nests its existing children in every locale instead of flattening them to the root. A forward-only migration backfills existing parent links.
--- a/packages/core/src/api/handlers/taxonomies.ts
+++ b/packages/core/src/api/handlers/taxonomies.ts
@@ -95,12 +95,16 @@ export interface TaxonomyDefTranslationsResponse {
  * Build tree structure from flat terms
  */
 function buildTree(flatTerms: TermWithCount[]): TermWithCount[] {
-	const map = new Map<string, TermWithCount>();
+	// `parentId` holds the parent's translation_group, so resolve links by group
+	// within the (already locale-filtered) set. Keying by row id would drop the
+	// link for every non-default locale, flattening the translated tree.
+	const byGroup = new Map<string, TermWithCount>();
 	const roots: TermWithCount[] = [];
-	for (const term of flatTerms) map.set(term.id, term);
+	for (const term of flatTerms) byGroup.set(term.translationGroup ?? term.id, term);
 	for (const term of flatTerms) {
-		if (term.parentId && map.has(term.parentId)) {
-			map.get(term.parentId)!.children.push(term);
+		const parent = term.parentId ? byGroup.get(term.parentId) : undefined;
+		if (parent) {
+			parent.children.push(term);
 		} else {
 			roots.push(term);
 		}
@@ -422,13 +426,6 @@ async function validateParentTerm(
 ): Promise<{ code: "VALIDATION_ERROR"; message: string } | null> {
 	if (parentId === undefined || parentId === null) return null;
 
-	if (termId !== undefined && parentId === termId) {
-		return {
-			code: "VALIDATION_ERROR",
-			message: "A term cannot be its own parent",
-		};
-	}
-
 	const parent = await repo.findById(parentId);
 	if (!parent) {
 		return {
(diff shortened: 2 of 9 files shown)


ascorbic (approved):
Nice! Solid fix, thanks

PR #1648, opened 2026-06-29
Add subtree (descendant) matching to taxonomy `where` filters

MA2153:
## What does this PR do?

Adds a `subtree` operator to collection `where` taxonomy filters so that selecting a parent term matches that term **and all of its descendants**, resolved in SQL:

```ts
where: { region: { subtree: "europe" } } // matches europe + every descendant region
```

Design and rationale: [Discussion #1647](https://github.com/emdash-cms/emdash/discussions/1647).

### Motivation

The taxonomy `where` filter currently matches by **exact slug only**, so "this term or anything filed under it" — choosing a parent category in a faceted browse UI — is inexpressible. The only workaround is to enumerate the subtree client-side and pass every descendant slug, which expands to one bound parameter per slug and overflows **D1's 100-bind-parameter cap** (`D1_ERROR: too many SQL variables`) on deep hierarchies. It also can't be chunked without breaking keyset pagination.

### Approach

- **`where: { <taxonomy>: { subtree: slug | slug[] } }`** — a new `WhereSubtree` operator added to the `WhereValue` union. The loader resolves the subtree **server-side from the root slug(s)** via a `WITH RECURSIVE` walk of `taxonomies.parent_id`, matching `content_taxonomies.taxonomy_id` through an `EXISTS` clause. Only the root slugs are bound, so the parameter count is independent of subtree size. After #1646 both columns live in translation_group space, so the walk is locale-correct and needs no extra join.
- **Guarded and additive.** The operator is inert unless `subtree:` is present. It is ignored (with a `console.warn`) on non-taxonomy keys and on `byline`; empty roots short-circuit to an empty result rather than emitting invalid SQL. The default exact-slug path is unchanged.
- **Opt-in `rollup` counts.** `getTaxonomyTerms(name, { rollup })` and the admin terms endpoint (`?rollup=1`) return **distinct-entry** subtree counts, so a facet badge equals what selecting that facet returns. Counts are computed with a single query and an in-memory fold up each term's ancestor chain (unioning entry ids to preserve DISTINCT), so the query count does not grow with the number of terms. Default counts (exact-term) are unchanged.
- **Portable.** The recursive CTE is standard SQL on both dialects; all new tests run under `describeEachDialect` (SQLite locally, Postgres in CI).

## Type of change

- [x] Feature (requires [maintainer-approved Discussion](https://github.com/emdash-cms/emdash/discussions/categories/ideas)) — see [#1647](https://github.com/emdash-cms/emdash/discussions/1647)

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm test` passes
- [x] No admin UI strings added — no `messages.po` changes included (n/a for i18n)
- [x] I have added a [changeset](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#changesets) (`emdash: minor`)
- [x] New feature links to a Discussion: https://github.com/emdash-cms/emdash/discussions/1647

## Screenshots / test output

Dialect-parity tests run under `describeEachDialect` (SQLite locally; Postgres in CI via `PG_CONNECTION_STRING`).

- `loader-taxonomy-subtree-filter.test.ts` (7) — single-root and multi-root union, a **>999-descendant overflow guard** (would exceed SQLite's bind limit if descendants were enumerated instead of walked in SQL), mixed exact + subtree across taxonomies, empty-roots short-circuit, cross-locale match by translation_group, and keyset pagination.
- `taxonomy-subtree-counts.test.ts` (5) — **distinct-entry rollup** (an entry tagged at both a parent and its child counts once), shared-ancestor counting across sibling subtrees, empty-subtree handling, `getTaxonomyTerms({ rollup })`, and `handleTermList({ rollup })`.
- `schemas.test.ts` — `?rollup` query-param coercion (`1`/`true`/`0`/`false`).

```
 Test Files  2 passed (2)
      Tests  12 passed (12)
```


MA2153 (commented):
RETRACTED

MA2153 (commented):
RETRACTED

MA2153:
Closing in favour of #2293.

This was opened on 2026-06-29. #1962 landed on 2026-07-12 and changed the premise: migration 051 made a taxonomy-filtered listing a single equality seek on

```
idx_content_taxonomies_pub (taxonomy_id, collection, deleted_at, published_at DESC, entry_id DESC)
```

That index requires an **equality** on `taxonomy_id`. Resolving descendants with a recursive CTE and feeding the filter from it turns that into a set match, so the planner drops the index and falls onto `sqlite_autoindex_content_taxonomies_1 (collection=?)`, walking the collection's entire pivot partition and temp-sorting the matches.

Measured on a schema-faithful repro (1,491-term taxonomy, 5 levels, 200k entries, no `ANALYZE`): a top-level subtree is 213 terms and reads **28,656 rows to render 10**. It also cannot be expressed as a single query at all, since one bound parameter per term exceeds D1's cap of 100.

#2293 proposes resolving the hierarchy at write time instead, which leaves the read path completely unchanged and makes a branch filter the same ~11-row seek as a leaf filter, at any depth. I'll open fresh PRs against that if the direction is approved.

PR #1662, opened 2026-06-29
feat(cloudflare): optional cachedBinding for Hyperdrive split caching

scottbuscemi:
## What does this PR do?

Adds an optional **`cachedBinding`** to the `hyperdrive()` database adapter so EmDash can serve **anonymous reads from a caching-enabled Hyperdrive configuration** while keeping every authenticated request and every write on the primary, caching-disabled binding.

```ts
// astro.config.mjs
database: hyperdrive({ binding: "HYPERDRIVE", cachedBinding: "HYPERDRIVE_CACHED" });
```

```jsonc
// wrangler.jsonc — two configs over the same database
{
  "hyperdrive": [
    { "binding": "HYPERDRIVE", "id": "<caching-disabled-id>" },
    { "binding": "HYPERDRIVE_CACHED", "id": "<caching-enabled-id>" }
  ]
}
```

### Background

PR #1614 shipped the Hyperdrive adapter and (correctly) requires Hyperdrive query caching to be **disabled**, because EmDash relies on read-after-write consistency: the admin/setup write a row and immediately read it back, and a default-on query cache can serve the pre-write result within its TTL (corrupting setup, showing editors stale content).

Cloudflare documents a [two-configuration pattern](https://developers.cloudflare.com/hyperdrive/configuration/query-caching/#disable-caching) — one cached connection for popular reads, one uncached — for exactly this situation. This PR makes that pattern a first-class opt-in for the adapter.

### How it works

The middleware already threads `isAuthenticated` and `isWrite` into `createRequestScopedDb` (added in #1614 for D1 replica routing). This PR uses those signals: a request routes to `cachedBinding` only when it is **anonymous *and* read-only** (`!isAuthenticated && !isWrite`); everything else uses the primary `binding`. So:

- Logged-in editors → always uncached → read-after-write stays correct.
- Writes (including anonymous writes, e.g. a public form plugin) → always uncached.
- Migrations and the cold-start singleton → always the primary binding.
- A `cachedBinding` that's missing at runtime → safe fallback to the primary binding.

The routing rule is a small pure function (`selectBindingName`) so it's unit-testable without a live DB or `cloudflare:workers`.

**Fully backwards-compatible:** omit `cachedBinding` and the adapter behaves exactly as before — the descriptor doesn't even carry the key.

### Caveats (documented in the JSDoc)

Anonymous reads of just-published content can be up to the cache's `max_age` stale (Hyperdrive default 60s, max 1h), and this cache is independent of EmDash's own cache invalidation. It's strictly opt-in for sites where a short public-read staleness window is acceptable.

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes — `@emdash-cms/cloudflare` clean
- [x] `pnpm lint` passes — `oxlint --type-aware --deny-warnings`: 0 warnings, 0 errors
- [x] `pnpm test` passes (or targeted tests for my change) — 200 cloudflare tests pass (9 new)
- [x] I have added/updated tests for my changes (if applicable) — pure `selectBindingName` routing tests + mocked `createRequestScopedDb` binding-selection tests + config-time descriptor tests
- [ ] User-visible strings in the admin UI are wrapped for translation (if applicable) — **n/a**, no admin UI strings
- [x] I have added a [changeset](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#changesets) — `@emdash-cms/cloudflare` minor
- [ ] New features link to an approved Discussion — **n/a**, opened by a maintainer (maintainer approval stands in for the Discussion).

- [x] This PR includes AI-generated code — model/tool: **Claude Opus 4.8 (via OpenCode)**

## Screenshots / test output

```
@emdash-cms/cloudflare  Test Files  14 passed (14)   Tests  200 passed (200)
oxlint --type-aware --deny-warnings   Found 0 warnings and 0 errors.
tsgo --noEmit (packages/cloudflare)   clean
```

---

### Try this PR

**[Open a fresh playground →](https://feat-hyperdrive-cached-binding-emdash-playground.emdash-cms.workers.dev)**

<sub>Tracks `feat/hyperdrive-cached-binding`. Updated automatically when the playground redeploys.</sub>


scottbuscemi:
Thanks @emdashbot — both points addressed. The review landed on commit `5c0d21ef`; the fixes are in `cc61a85d` and `0f769a95`.

**1. Internal anonymous GETs (`/_emdash/api/setup/status`, admin setup-gate) cached.** Fixed by gating the cached binding on the request being a **public, non-`/_emdash` path** in addition to anonymous + read-only. `selectBindingName` now takes the request `url`; any request under `/_emdash` (admin, setup, auth, internal APIs) uses the primary uncached binding even when anonymous. This was confirmed live: after completing setup, `GET /_emdash/api/setup/status` now returns `{"needsSetup":false}` and `/_emdash/admin` redirects to login instead of looping the wizard. (`packages/cloudflare/src/db/hyperdrive.ts`, `EMDASH_BASE_PATH` / `isEmDashInternalPath`, + routing tests for `/_emdash` paths and a "path merely contains `_emdash`" guard.)

**2. `isAuthenticated` too narrow (API-token requests treated as anonymous).** Took your preferred option and broadened the signal. The auth middleware that resolves tokens runs *after* the main middleware, so `locals.user` isn't populated at the adapter call — instead the main middleware now detects the `Authorization: Bearer` credential directly on the request and ORs it into `isAuthenticated`. API-token (`ec_pat_*`) and OAuth (`ec_oat_*`) reads therefore stay on the primary/uncached connection. This also fixes a latent read-your-writes gap in the **D1** replica adapter, which uses the same signal. Conservative (errs toward primary). (`packages/core/src/astro/middleware.ts` + a middleware test asserting a Bearer request without a session cookie yields `isAuthenticated: true`; new `emdash` changeset.)

Docs/JSDoc updated to describe the actual routing rule, and the deployed split-caching test site verifies public reads still serve from the cached binding while admin/setup/auth stay consistent.

docs/src/content/docs/deployment/database.mdx
```diff
@@ -219,6 +220,140 @@ database: postgres({
...
+	uncached primary.
+</Aside>
+
+<Aside type="caution">
+	Both configurations must point at the **same** database, or anonymous visitors see a different dataset than editors. Anonymous reads of just-published content can be up to the cache's `max_age` stale (Hyperdrive default 60s, max 1h), and this cache is independent of EmDash's own cache invalidation. Only opt in if a short public-read staleness window is acceptable; otherwise omit `cachedBinding` and keep caching disabled.
+</Aside>
+
+<Aside>
+	**Sandboxed plugins are D1-only.** The sandbox plugin bridge talks to a D1 binding directly, independent of the configured adapter, so sandboxed plugins aren't available on a Hyperdrive deployment.
+</Aside>
```
ascorbic, line 355: Not a blocker, but having multiple Asides next to each other is harder to read, and defeats the purpose of calling-out important points.

ascorbic (approved):

PR #1213, opened 2026-05-29
fix(admin): keep code block language picker open when using its dropdown

ascorbic:
## What does this PR do?

Fixes the code block language picker closing as you type, making it
impossible to choose a language (#1200).

### Root cause (confirmed from a DOM-mutation trace)

The bug only reproduces with a browser extension active (password manager /
autofill) -- it is fine in Incognito and in clean automation. A capture in
a real profile showed: typing a character that opens the suggestion
dropdown causes the **entire `react-renderer node-codeBlock` node view to
be removed and re-added** -- i.e. ProseMirror *recreates* the React node
view. The recreation resets the node view's local `isEditing` state, so the
picker disappears (the `blur`/`focusout` we saw was just the focused input
being torn out of the DOM).

Why the extension matters: the extension mutates the DOM around the
picker's input as you type. ProseMirror's node view cannot reconcile that
mutation and falls back to recreating the node view. No extension -> no
overlay mutation -> no recreation, which is why it never reproduced in a
clean browser.

### Fix

- **`ignoreMutation` on the node view** (the real fix): ignore any DOM
  mutation that is not inside the editable code content
  (`pre.emdash-code-block`). The picker overlay -- and anything an extension
  injects into it -- no longer triggers a node view recreation. Selection
  changes and real code edits are still observed normally.
- **Outside-click hardening** (a separate, latent path): the suggestion
  list is portalled to `document.body`, outside `popoverRef`, so a genuine
  click on a suggestion was treated as an outside click and dismissed the
  picker. Tag the portalled popup and treat it as part of the picker, and
  ignore untrusted (synthetic) pointer events.
- Both decisions extracted into pure helpers (`ignoreCodeBlockMutation`,
  `shouldDismissPicker`) with unit coverage.

Closes #1200

## Type of change

- [x] Bug fix

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md)
- [x] `pnpm typecheck` passes (`@emdash-cms/admin`)
- [x] `pnpm lint` passes (`pnpm lint:json` -> 0 diagnostics)
- [x] `pnpm test` passes (`tests/editor/CodeBlockNode.test.ts`, 12 passed)
- [x] `pnpm format` has been run
- [x] I have added/updated tests for my changes
- [ ] User-visible strings in the admin UI are wrapped for translation -- no new strings
[...]

PR #1673, opened 2026-06-30
fix(admin): keep code block language picker open while typing

### LINKED ISSUE #1200: Code block language picker loses focus when typing
### Description

The language picker in the code block editor loses focus and closes if you attempt to type, makign it impossible to choose a language for a block

### Steps to reproduce

1. open a post in the editor
2. add a code block using backticks or the slash menu
3. hover the block and click on the "language" button
4. type in the field
5. observe that it loses focus and closes

### Environment

- emdash 0.15.0

### Logs / error output

```shell

```

ascorbic:
## What does this PR do?

Fixes the code block language picker losing focus and closing the moment you type in it, making it impossible to choose a language.

Closes #1200

### Root cause

The code block is a **non-atom** ProseMirror node with live editable content (`NodeViewContent`). The picker's text input lived inside the node view, i.e. inside the editor's `contentEditable` DOM. Typing moved the DOM selection into the input, so ProseMirror read that selection, dispatched a selection-correcting transaction (`docChanged: false, selectionSet: true`, no steps), and the resulting node-view redraw — amplified by the `Focus` extension's per-node decorations (`mode: "all"`) — recreated the React node view mid-edit. The component remounted with `isEditing` reset to false, unmounting the input.

This reproduces in plain Safari with no extensions, which is why the earlier "password manager / synthetic event" theory (#1213) didn't fix it. Confirmed live by instrumenting the node view (MOUNT/UNMOUNT on every keystroke) and capturing the spurious selection transaction.

### Fix

Render the picker through Kumo's `Popover` so `Popover.Content` is portalled out of the editor's `contentEditable`. The caret never enters the editor, no spurious selection transaction fires, and the node view is not recreated. Popover also provides positioning, RTL, outside-click and Escape handling, and focuses the input on open (which also fixes a latent "input wasn't autofocused on open" annoyance).

Supersedes #1213, whose diagnosis (extension-injected DOM mutations) was incorrect — that PR should be closed.

## Type of change

- [x] Bug fix

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] I have added/updated tests for my changes (if applicable) — the bug is very hard to reproduce in Playwright, so any e2e test would have too many false negatives to be useful. Verified manually in Chrome and Safari instead.
- [x] User-visible strings in the admin UI are wrapped for translation (no new strings; existing `t`/aria labels preserved).

## Screenshots / test output

Verified in-browser against a real post: typing keeps focus and filters the list, selecting from the dropdown fills the input, Apply and Enter both commit (curated and free-form values), Escape and outside-click close without committing, and editing the code content itself is unaffected. Lint clean, build + typecheck pass, 1021 admin tests pass.


ascorbic:
Thanks for the review. Triage of the three findings:

**1. `Popover.Trigger` `render` vs `asChild` — no change.** This one is backwards: Kumo marks `asChild` as `@deprecated` and points to the `render` prop as the replacement. From `@cloudflare/kumo` `popover.d.ts`:

```ts
/**
 * @deprecated Use the `render` prop instead.
 * @example `<Popover.Trigger render={<Button />}>Open</Popover.Trigger>` instead of `<Popover.Trigger asChild><Button>Open</Button></Popover.Trigger>`
 */
asChild?: boolean;
```

So `Header.tsx` is on the deprecated form; this PR uses the current one. The forwarding concern is also unfounded — verified on the deploy preview that the chip button receives `aria-haspopup="dialog"` and `aria-expanded` toggling `false`→`true` on open, and the trigger's `onMouseDown` preventDefault, focus, and outside-click/Escape all work.

**2. Query-count snapshot changes — already resolved.** They aren't in this PR's net diff (`git diff main...HEAD` is just `CodeBlockNode.tsx` + the changeset). A CI job auto-committed a snapshot update, and the subsequent `main` merge folded those into the baseline since the SEO loader change already landed them on `main`. Nothing to revert.

**3. Focused unit test — fair.** Noting for the record: this regression is a ProseMirror focus/selection timing race that doesn't reproduce reliably under automation (Playwright with real key events, delays, and `pressSequentially` all passed against the broken build), which is why there's no e2e test. A React Testing Library test asserting the picker renders in a portal outside the editor container (plus Enter-commit / Escape-cancel) would be a durable structural guard against the fix being silently undone, and I'll consider adding one as a follow-up.

PR #1692, opened 2026-07-01
fix(db): restore idx_taxonomies_parent dropped by migration 036

### LINKED ISSUE #1665: Migration 036 drops idx_taxonomies_parent and never restores it
### Summary

The `idx_taxonomies_parent` index on `taxonomies(parent_id)` is silently lost during migration `036_i18n_menus_and_taxonomies` and never restored. Any install that has run migration 036 has no index backing parent-based (hierarchical) lookups on the `taxonomies` table.

### Details

Migration `015_indexes.ts` creates the index:

```ts
// Hierarchical queries filter on parent_id FK
await db.schema
  .createIndex("idx_taxonomies_parent")
  .on("taxonomies")
  .column("parent_id")
  .execute();
```

Migration `036_i18n_menus_and_taxonomies.ts` later **rebuilds** the `taxonomies` table to add `locale`/`translation_group` (SQLite path: create `taxonomies_new` → copy → `DROP TABLE taxonomies` → rename). SQLite drops every index attached to a table when that table is dropped. `rebuildTaxonomies` then recreates only three indexes:

```ts
await db.schema.createIndex("idx_taxonomies_name").on("taxonomies").column("name").execute();
await db.schema.createIndex("idx_taxonomies_locale").on("taxonomies").column("locale").execute();
await db.schema
  .createIndex("idx_taxonomies_translation_group")
  .on("taxonomies")
  .column("translation_group")
  .execute();
```

`idx_taxonomies_parent` is **not** among them, so it is gone after 036 runs.

This appears unintentional: the same migration deliberately recreates the parent index for the other rebuilt tables — `idx_menu_items_parent` in `rebuildMenuItems`, and `idx_content_taxonomies_term` in `rebuildContentTaxonomies` — so the omission for `taxonomies` looks like an oversight rather than a decision.

### Reproduction

On any DB migrated through 036:

```sql
SELECT name FROM pragma_index_list('taxonomies');
-- idx_taxonomies_translation_group
-- idx_taxonomies_locale
-- idx_taxonomies_name
-- sqlite_autoindex_taxonomies_2  (name+slug+locale unique)
-- sqlite_autoindex_taxonomies_1  (pk)
-- no idx_taxonomies_parent
```

(Confirmed against a live remote D1 database.)

### Impact

Parent/child queries against `taxonomies` (e.g. fetching a term's children, building hierarchical trees) no longer have a dedicated index on `parent_id`. SQLite can fall back to an automatic covering index for some query shapes, but the persistent index is missing and worth restoring for any plan that benefits from it.

### Suggested fix

Add `idx_taxonomies_parent` back. Since 036 has already shipped, recreate it in a new forward-only migration (`CREATE INDEX IF NOT EXISTS idx_taxonomies_parent ON taxonomies(parent_id)`) so existing installs get it, and also add the `createIndex` call to `rebuildTaxonomies` for parity with the other rebuilt tables.

MA2153:
## What does this PR do?

Restores the `idx_taxonomies_parent` index on `taxonomies(parent_id)`, which was silently dropped by the i18n table rebuild in migration `036`.

Migration `015` created the index. `036`'s SQLite/D1 rebuild of `taxonomies` (`taxonomies_new` → `DROP TABLE` → rename) drops every index attached to the old table, but `rebuildTaxonomies` only recreated `name`, `locale`, and `translation_group` — the parent index was left out. Its sibling rebuilds (`rebuildMenuItems` → `idx_menu_items_parent`, `rebuildContentTaxonomies` → `idx_content_taxonomies_term`) *do* restore their parent index, confirming this was an oversight. No later migration restored it, so any install migrated through `036` has been running hierarchical taxonomy lookups without it.

Impact verified with `EXPLAIN QUERY PLAN`: the repository's actual parent queries (`findChildren`, `getTaxonomyTerm`, `list({ parentId })`) are single-probe `WHERE parent_id = ?` lookups that plan as a full `SCAN taxonomies` without the index and `SEARCH … USING INDEX` with it. SQLite's automatic (transient) indexing does **not** cover these — it only kicks in for repeated probes (join inner side / correlated subqueries), which these queries are not.

Changes:
- Adds forward-only migration `047_restore_taxonomy_parent_index` (`createIndex().ifNotExists()`, dialect-independent; no-op on Postgres where the in-place `pgWiden` path kept the index).
- Recreates the index in `036`'s `rebuildTaxonomies` (fresh installs) and `rebuildTaxonomiesDown` (faithful rollback to the pre-036 / `015` state), for parity with the sibling rebuilds.

Closes #1665

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] ~User-visible strings in the admin UI are wrapped for translation~ — n/a, no admin UI or user-facing strings (DB migration + tests only)
- [ ] ~New features link to an approved Discussion~ — n/a, bug fix

## Screenshots / test output

Regression test (fails before the fix, passes after) plus the full migration suite:

```
Test Files  21 passed (21)
     Tests  202 passed (202)
```

`EXPLAIN QUERY PLAN` for `SELECT * FROM taxonomies WHERE parent_id = ? AND locale = ? ORDER BY label`:

```
== WITHOUT idx_taxonomies_parent ==
  SCAN taxonomies
  USE TEMP B-TREE FOR ORDER BY
== WITH idx_taxonomies_parent ==
  SEARCH taxonomies USING INDEX idx_taxonomies_parent (parent_id=?)
  USE TEMP B-TREE FOR ORDER BY
```

### DIFF
--- a/.changeset/restore-taxonomy-parent-index.md
+++ b/.changeset/restore-taxonomy-parent-index.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Restores the missing `idx_taxonomies_parent` index on `taxonomies(parent_id)`, which was silently dropped by the i18n table rebuild in an earlier version. Installs upgrade automatically; hierarchical (parent/child) taxonomy lookups are indexed again.
--- a/packages/core/src/database/migrations/036_i18n_menus_and_taxonomies.ts
+++ b/packages/core/src/database/migrations/036_i18n_menus_and_taxonomies.ts
@@ -199,6 +199,15 @@ async function rebuildTaxonomies(db: Kysely<unknown>, defaultLocale: string): Pr
 		.on("taxonomies")
 		.column("translation_group")
 		.execute();
+	// Dropping the old table dropped idx_taxonomies_parent (from 015); recreate
+	// it here for parity with rebuildMenuItems/rebuildContentTaxonomies. Existing
+	// installs recover it via migration 047.
+	await db.schema
+		.createIndex("idx_taxonomies_parent")
+		.ifNotExists()
+		.on("taxonomies")
+		.column("parent_id")
+		.execute();
 }
 
 async function rebuildTaxonomyDefs(db: Kysely<unknown>, defaultLocale: string): Promise<void> {
@@ -569,6 +578,13 @@ async function rebuildTaxonomiesDown(db: Kysely<unknown>): Promise<void> {
 	await db.schema.dropTable("taxonomies").execute();
 	await sql`ALTER TABLE taxonomies_old RENAME TO taxonomies`.execute(db);
 	await db.schema.createIndex("idx_taxonomies_name").on("taxonomies").column("name").execute();
+	// Restore the pre-036 (migration 015) parent index so the rollback is faithful.
+	await db.schema
+		.createIndex("idx_taxonomies_parent")
+		.ifNotExists()
+		.on("taxonomies")
+		.column("parent_id")
+		.execute();
 }
 
 async function rebuildTaxonomyDefsDown(db: Kysely<unknown>): Promise<void> {
--- a/packages/core/src/database/migrations/047_restore_taxonomy_parent_index.ts
+++ b/packages/core/src/database/migrations/047_restore_taxonomy_parent_index.ts
@@ -1,0 +1,29 @@
+import type { Kysely } from "kysely";
+
+/**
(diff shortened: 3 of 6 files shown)


ascorbic (approved):
Thanks. Good spot

PR #1702, opened 2026-07-01
fix(core): restore idx_content_taxonomies_term after partial-apply retry (#1701)

### LINKED ISSUE #1701: Migration 036 can permanently lose idx_content_taxonomies_term on a partial-apply retry (same class as #1665)
### Summary

Migration `036_i18n_menus_and_taxonomies` can permanently lose the
`idx_content_taxonomies_term` index on `content_taxonomies(taxonomy_id)` if its
SQLite/D1 table rebuild is retried after a partial apply. This is the **same bug
class as #1665** (which lost `idx_taxonomies_parent` and was fixed by migration
047) — but for `content_taxonomies`, and with an extra twist: the index recreate
is *latent* behind a guard rather than merely omitted, so a simple "recreate it"
migration fixes the existing damage but the trap in 036 remains for any future
partial apply.

Applies to the **SQLite/D1 path** only. The Postgres path alters in place and is
unaffected.

### The defect

In `rebuildContentTaxonomies` (packages/core .../036_i18n_menus_and_taxonomies.ts):

```ts
async function rebuildContentTaxonomies(db) {
  const fks = await sql`PRAGMA foreign_key_list(content_taxonomies)`.execute(db);
  if (fks.rows.length === 0) return;           // (A) guard: no FK -> skip whole body

  await sql`DROP TABLE IF EXISTS "content_taxonomies_new"`...
  // CREATE content_taxonomies_new (no FK) ; INSERT OR IGNORE ... SELECT
  await db.schema.dropTable("content_taxonomies").execute();          // (B) drops old table + its indexes, strips FK
  await sql`ALTER TABLE content_taxonomies_new RENAME TO content_taxonomies`...

  await sql`CREATE INDEX IF NOT EXISTS idx_content_taxonomies_term
            ON content_taxonomies(taxonomy_id)`.execute(db);         // (C) INSIDE guard (A)
}
```

`rebuildContentTaxonomies` is the **only** one of 036's five rebuild helpers that
guards on FK-presence (A). The other four guard on `hasColumn(<table>, "locale")`
and recreate their indexes independently. Because the rebuild at (B) both strips
the FK **and** drops the old table's indexes, the index recreate at (C) is the
**only** index creation in the entire migration whose execution is gated behind a
guard that a retry can trip.

### How the index is lost

D1 DDL is not transactional (each statement auto-commits), and Kysely's `Migrator`
records a migration as applied only after `up()` resolves. So:

1. **First run.** FK present -> guard (A) passes -> table is rebuilt at (B): FK
   stripped, old indexes dropped. The run then fails at/just before (C) — any
   transient D1 error, an isolate eviction mid-migration (Workers CPU/duration
   limits), or a deploy overlap. `up()` throws; 036 is **not** marked applied, but
   the DROP/CREATE/RENAME at (B) have already committed. `content_taxonomies` now
   has **no FK and no index**.
2. **Retry.** `up()` re-runs from the top. `rebuildContentTaxonomies` now sees
   **0 FKs** -> guard (A) early-returns -> **(C) is skipped**. The other four
   rebuilds still run (their `locale` column isn't present yet) and recreate their
   indexes. `up()` resolves; 036 is marked applied.

Net result: 036 is recorded as successfully applied, every other 036 index exists,
and **only `idx_content_taxonomies_term` is missing** — with no error surfaced and
no forward migration to restore it (unlike `idx_taxonomies_parent`, which 047
recovers).

### Observable signature

On an affected install, `content_taxonomies` carries only its PK autoindex; all
other 036-created indexes (menus/menu_items/taxonomies/taxonomy_defs
`_locale`/`_translation_group`, `idx_menu_items_menu/_parent`, `idx_taxonomies_*`)
are present. i.e. exactly one index missing, precisely the FK-guarded one. This
signature is diagnostic — it is not reproducible by data loss/restore (which would
drop many indexes or none), only by this guard-skip.

### Suggested fix (mirrors the #1665 / 047 approach, plus closing the trap)

1. **Recovery migration** (forward-only, like 047): idempotent
   `CREATE INDEX IF NOT EXISTS idx_content_taxonomies_term ON
   content_taxonomies(taxonomy_id)`. No-op where present; restores it where the
   guard-skip lost it.
2. **Close the trap in 036**: move the `CREATE INDEX` for
   `idx_content_taxonomies_term` **out** of the FK-presence early-return, so a
   retry after a partial rebuild restores the index even once the FK is already
   gone (run it unconditionally whenever the table exists). Otherwise the same loss
   can recur on any future partial apply.

Optionally, the runner could self-heal expected indexes, or the query could be made
to fail loudly rather than silently full-scan when its backing index is absent.

### Related

- #1665 / migration 047 — same bug class (036 table rebuild silently dropped
  `idx_taxonomies_parent`); this is the `content_taxonomies` sibling that 047 did
  not cover.
- #1692 — prior fix to this migration.

Environment: SQLite/D1. Observed on an install carrying `idx_content_taxonomies_term`
absent while all sibling 036 indexes are present.

MA2153:
## What does this PR do?

Fixes migration `036_i18n_menus_and_taxonomies` permanently losing the
`idx_content_taxonomies_term` index on `content_taxonomies(taxonomy_id)` on
SQLite/D1 after a partial-apply retry.

In `rebuildContentTaxonomies`, the index recreate lived **inside** an
FK-presence early-return. D1 auto-commits each DDL statement, but Kysely only
records a migration as applied after `up()` resolves. So a first run that
committed the table rebuild (stripping the FK, dropping the old index) then
failed before the `CREATE INDEX` leaves the table with **no FK and no index** —
and 036 is *not* marked applied. On retry, `up()` re-enters, sees zero FKs,
early-returns, and never recreates the index; the other rebuilds run and 036 is
then marked applied. Net result: 036 recorded as successfully applied with only
`idx_content_taxonomies_term` missing and no forward migration to recover it.

The fix:

1. Move the `CREATE INDEX IF NOT EXISTS idx_content_taxonomies_term` **out** of
   the FK guard so it runs unconditionally whenever the table exists — closing
   the trap for any future partial apply.
2. Add forward-only migration `048_restore_content_taxonomies_term_index` to
   restore the index on installs that already hit the trap (idempotent,
   `ifNotExists`; no-op on Postgres, which alters in place).

Same bug class as #1665 / migration 047 (which restored `idx_taxonomies_parent`),
but for `content_taxonomies`. The Postgres path is unaffected.

Closes #1701

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes 
- [x] `pnpm lint` passes 
- [ ] User-visible strings in the admin UI are wrapped for translation (if applicable) — n/a, no UI/string changes

## Screenshots / test output

```
Test Files  4 passed (4)
     Tests  56 passed (56)
```

New/updated tests:
- Unit (`036_..._test.ts`): reproduces the partial-apply retry (FK stripped,
  index dropped → re-run `up()`) and asserts the index is restored; plus a
  normal-run assertion that 036 recreates `idx_content_taxonomies_term`.
- Integration (`migrations.test.ts`): asserts the index survives the full
  migration chain (regression for #1701); registers 048 in the trailing list.

### DIFF
--- a/.changeset/restore-content-taxonomies-term-index.md
+++ b/.changeset/restore-content-taxonomies-term-index.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Restores the `idx_content_taxonomies_term` index on SQLite/D1 installs that lost it when migration 036 was retried after a partial apply, so taxonomy reverse-lookups are indexed again. The index is now recreated unconditionally during 036 to prevent the same loss on any future partial apply.
--- a/packages/core/src/database/migrations/036_i18n_menus_and_taxonomies.ts
+++ b/packages/core/src/database/migrations/036_i18n_menus_and_taxonomies.ts
@@ -259,27 +259,32 @@ async function rebuildContentTaxonomies(db: Kysely<unknown>): Promise<void> {
 	// is load-bearing — if the translation_group seed ever changes, this needs
 	// an explicit remap *after* `rebuildTaxonomies` runs.
 	const fks = await sql<{ id: number }>`PRAGMA foreign_key_list(content_taxonomies)`.execute(db);
-	if (fks.rows.length === 0) return;
+	if (fks.rows.length > 0) {
+		await sql.raw(`DROP TABLE IF EXISTS "content_taxonomies_new"`).execute(db);
+		await db.schema
+			.createTable("content_taxonomies_new")
+			.addColumn("collection", "text", (c) => c.notNull())
+			.addColumn("entry_id", "text", (c) => c.notNull())
+			.addColumn("taxonomy_id", "text", (c) => c.notNull())
+			.addPrimaryKeyConstraint("content_taxonomies_pk", ["collection", "entry_id", "taxonomy_id"])
+			.execute();
 
-	await sql.raw(`DROP TABLE IF EXISTS "content_taxonomies_new"`).execute(db);
-	await db.schema
-		.createTable("content_taxonomies_new")
-		.addColumn("collection", "text", (c) => c.notNull())
-		.addColumn("entry_id", "text", (c) => c.notNull())
-		.addColumn("taxonomy_id", "text", (c) => c.notNull())
-		.addPrimaryKeyConstraint("content_taxonomies_pk", ["collection", "entry_id", "taxonomy_id"])
-		.execute();
+		await sql`
+			INSERT OR IGNORE INTO content_taxonomies_new (collection, entry_id, taxonomy_id)
+			SELECT collection, entry_id, taxonomy_id FROM content_taxonomies
+		`.execute(db);
 
-	await sql`
-		INSERT OR IGNORE INTO content_taxonomies_new (collection, entry_id, taxonomy_id)
-		SELECT collection, entry_id, taxonomy_id FROM content_taxonomies
-	`.execute(db);
+		await db.schema.dropTable("content_taxonomies").execute();
+		await sql`ALTER TABLE content_taxonomies_new RENAME TO content_taxonomies`.execute(db);
+	}
 
-	await db.schema.dropTable("content_taxonomies").execute();
-	await sql`ALTER TABLE content_taxonomies_new RENAME TO content_taxonomies`.execute(db);
-
-	// SQLite drops indexes when the underlying table is dropped. Restore the
-	// taxonomy_id index from migration 015.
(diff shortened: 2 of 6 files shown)


ascorbic (approved):
Thanks

PR #1731, opened 2026-07-02
feat(templates): retheme-from-theme.css architecture for blog and marketing

ascorbic:
## What does this PR do?

Restructures theming in the blog and marketing templates (and their `-cloudflare` variants) so that a site can be rethemed entirely from `src/styles/theme.css` — which is the only styling surface the emdash-build agent (and the template docs) offer.

- **Real tokens file.** Design tokens move out of `Base.astro` into `src/styles/tokens.css` as actual `@layer base` defaults; `theme.css` becomes a pure unlayered override file. No more hand-maintained commented mirror to drift.
- **`light-dark()` colors.** Each color token is defined once carrying both modes, replacing the triplicated light / `@media` dark / `.dark` palettes. A plain-color override in `theme.css` now rethemes both modes; the footer switcher just pins `color-scheme`. A `@supports not (light-dark)` block gives older browsers (Safari < 17.5, Chrome < 123) the plain light palette.
- **Semantic tokens.** `--font-body` / `--font-heading` replace `--font-sans` (no more setting "sans" to a serif); `--color-brand` (+ `--color-on-brand`, `--color-brand-ring`) replaces blog `--color-accent*` and marketing `--color-primary*`; heading weights become `--font-weight-heading` / `--font-weight-display`.
- **Marketing token promotion.** The signature gradients become `--gradient-brand` / `--gradient-brand-strong` / `--gradient-brand-soft` / `--gradient-headline` (previously inlined in four variants across five files — a flat brand is now `--gradient-*: <solid color>`); hardcoded `color: white`, the frozen-indigo focus ring, and hardcoded error reds become `--color-on-brand`, `--color-brand-ring`, and `--color-danger`.
- **Marketing dedup.** Blocks share `.section-header` / `.section-headline` / `.section-subheadline` and `.icon-tile` base-layer classes; the nav CTA composes `.btn .btn-primary` instead of duplicating it.
- **Bug fixed along the way:** the contact form error color only tracked `prefers-color-scheme`, ignoring the cookie-based theme toggle; it now follows the active theme like everything else.

Template docs (`AGENTS-template.md`, regenerated `AGENTS.md`) are updated to match. Portfolio and starter templates still use the old architecture — follow-up. The emdash-build prompt update that teaches agents the new contract is committed in that repo and must deploy only after these templates release.

## Type of change

- [ ] Bug fix
- [x] Refactor (no behavior change)
- [ ] Translation
- [ ] Documentation
- [ ] Tests
- [ ] Performance improvement

Rendered output is intentionally unchanged (verified in browser, both modes); the theming *API* (variable names, file layout) changes deliberately. Maintainer-requested work, so no prior Discussion.

## Checklist

- [x] `pnpm typecheck` passes (`pnpm typecheck:templates` — all templates)
- [x] `pnpm lint` passes
- [ ] `pnpm test` passes (or targeted tests for my change) — n/a: templates have no test suite; changes are CSS/markup only and were verified in-browser (see below)
- [ ] I have added/updated tests for my changes (if applicable) — n/a, no template test infrastructure
- [ ] User-visible strings in the admin UI are wrapped for translation (if applicable) — n/a, no admin UI changes
- [ ] I have added a [changeset](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#changesets) — n/a: template packages are private and in the changesets ignore list, so they are never released
- [ ] New features link to an approved Discussion — n/a, maintainer-requested refactor

- [x] This PR includes AI-generated code — model/tool: Claude Code + Claude Fable 5

## Screenshots / test output

Verified by running both templates with `emdash dev` and screenshotting in agent-browser:

- Blog and marketing render pixel-equivalent to `main` in light and dark mode (home, post detail, contact).
- Dark mode works with no `.dark` class (pure `color-scheme`/OS), and the footer switcher pins it correctly under the opposite OS preference (`background` flips `rgb(255,255,255)` → `rgb(13,13,13)` when the class is applied).
- Retheme contract exercised end-to-end from `theme.css` alone: blog with teal `light-dark()` brand + serif headings + lowered weights (applies to both modes from one block); marketing with a flat teal brand via solid `--gradient-*` overrides (logo, CTA, headline all go flat, no component edits).
- An adversarial-review pass ran over the diff; its two findings (gradient text invisible on pre-`light-dark()` browsers, nav CTA hover transition regression) are fixed in the second commit.

### DIFF
--- a/.changeset/tidy-donkeys-shake.md
+++ b/.changeset/tidy-donkeys-shake.md
@@ -1,0 +1,8 @@
+---
+"@emdash-cms/template-blog": minor
+"@emdash-cms/template-blog-cloudflare": minor
+"@emdash-cms/template-marketing": minor
+"@emdash-cms/template-marketing-cloudflare": minor
+---
+
+Restructures theming so a site can be rethemed entirely from `src/styles/theme.css`. Design tokens now live in `src/styles/tokens.css` as real (not commented) defaults, colors are defined once with CSS `light-dark()` so a single override rethemes both light and dark mode, and font tokens are semantic (`--font-body`, `--font-heading`) so a serif theme no longer means setting `--font-sans` to a serif. The main brand color is `--color-brand` in both templates (was `--color-accent` in blog, `--color-primary` in marketing), with `--color-on-brand` and `--color-brand-ring` replacing hardcoded white text and focus rings. The marketing template's signature gradients are now `--gradient-*` tokens, its heading weights are `--font-weight-heading`/`--font-weight-display`, and form errors use a new `--color-danger` that follows the theme toggle (previously they only tracked the OS preference).
--- a/.changeset/tidy-donkeys-shake.md
+++ b/.changeset/tidy-donkeys-shake.md
@@ -1,8 +1,0 @@
----
-"@emdash-cms/template-blog": minor
-"@emdash-cms/template-blog-cloudflare": minor
-"@emdash-cms/template-marketing": minor
-"@emdash-cms/template-marketing-cloudflare": minor
----
-
-Restructures theming so a site can be rethemed entirely from `src/styles/theme.css`. Design tokens now live in `src/styles/tokens.css` as real (not commented) defaults, colors are defined once with CSS `light-dark()` so a single override rethemes both light and dark mode, and font tokens are semantic (`--font-body`, `--font-heading`) so a serif theme no longer means setting `--font-sans` to a serif. The main brand color is `--color-brand` in both templates (was `--color-accent` in blog, `--color-primary` in marketing), with `--color-on-brand` and `--color-brand-ring` replacing hardcoded white text and focus rings. The marketing template's signature gradients are now `--gradient-*` tokens, its heading weights are `--font-weight-heading`/`--font-weight-display`, and form errors use a new `--color-danger` that follows the theme toggle (previously they only tracked the OS preference).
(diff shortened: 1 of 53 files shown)


ascorbic:
Review findings addressed in 295eda10:

**Copilot (all four comments — `color-scheme` pins vs the light-only fallback):** applied as suggested. The `:root.light`/`:root.dark` pins in both templates are now wrapped in `@supports (color: light-dark(#000, #fff))`, so browsers on the plain-light fallback palette no longer get dark UA form controls when the toggle sets the dark cookie — the toggle is simply inert there.

**emdashbot #1/#3 (marketing link reset):** the invisible-inline-link behaviour predates this PR — `a { color: currentColor; text-decoration: none }` is unchanged from `main` — but it's a real gap, so fixed properly rather than argued: the reset now carries the same `:where(:not([class*="emdash"]):not([class*="ec-"]))` guard as the blog, and classless links inside `p`/`li` (Portable Text renders bare `<a>` tags) get `--color-brand` with an underline. Component links (nav, footer, contact methods, buttons) all carry classes and keep their explicit styles. The demo seed contains no classless content links, so rendered output is unchanged.

**emdashbot #2 (changeset omits shadow renames):** moot — the changeset was removed entirely (d6560e0e). Template packages are `private: true` and in the changesets `ignore` list, so they are never released and a changeset naming them can break `changeset version`. There is no released user whose `--shadow-dropdown`/`--shadow-btn-active` overrides could be silently dropped; templates only reach users at scaffold time.

PR #1746, opened 2026-07-03
Add media usage index write-side foundation

khoinguyenpham04:
## What does this PR do?

Adds the write-side foundation for the Media Usage index. This includes the database tables/status metadata, content media reference extraction, content source snapshots for `columns` and `draft_overlay`, repository APIs, runtime refresh wiring for normal content lifecycle writes, stale marking for known bypass paths, and generation-guarded source replacement/deletion so refresh races preserve current usage and mark coverage stale instead of failing content writes.

This PR intentionally stops before repair/rebuild, the usage read API, admin Used In UI, safe delete, replace-in-place behavior, or media delete behavior changes.

Related discussion: https://github.com/emdash-cms/emdash/discussions/1503

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [x] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable). Do not include `messages.po` changes except in translation PRs — a workflow extracts catalogs on merge to `main`. N/A: no admin UI strings in this PR.
- [x] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/1503

- [x] This PR includes AI-generated code — model/tool: OpenCode with GPT-5.5 (`openai/gpt-5.5`)

## Screenshots / test output

Non-visual backend/indexing change.

Verified:

- `pnpm lint`
- `pnpm typecheck`
- `pnpm format`
- `pnpm --filter emdash exec vitest run tests/integration/database/media-usage-status-migration.test.ts tests/integration/database/media-usage-content-fields.test.ts tests/integration/database/media-usage-repository.test.ts tests/integration/database/media-usage-migration.test.ts tests/integration/database/media-usage-content-snapshots.test.ts tests/integration/database/media-usage-content-refresh.test.ts tests/integration/database/media-usage-runtime-refresh.test.ts tests/integration/database/media-usage-stale-bypass.test.ts tests/unit/media/usage-extractor.test.ts tests/unit/media/usage-source-key.test.ts` (10 files, 98 tests)
- `git diff --check`

---

### Try this PR

**[Open a fresh playground →](https://feature-media-usage-index-core-emdash-playground.emdash-cms.workers.dev)**

<sub>Tracks `feature/media-usage-index-core`. Updated automatically when the playground redeploys.</sub>


.changeset/media-usage-index-hardening.md
```diff
@@ -0,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Adds internal media usage index hardening for future reference tracking.
```
ascorbic, line 5: It's borderline whether this needs a changeset at all because it doesn't affect anything user-facing, but it definitely doesn't need two

packages/core/src/media/usage/content-refresh.ts
```diff
@@ -0,0 +1,491 @@
...
+		}
+
+		let refreshedSourceCount = 0;
+		for (const snapshot of snapshotsResult.snapshots) {
+			const result = await repo.replaceSourceIfCurrent(
+				snapshot.source,
+				snapshot.occurrences,
+				observedGenerations.get(snapshot.source.sourceKey) ?? null,
+			);
+			if (!result.replaced) {
```
ascorbic: This should retry, because it means there's been a race condition so if it's changed then this will be conflicted and wrong until next revbuilt

packages/core/src/database/repositories/media-usage.ts
```diff
@@ -217,20 +456,287 @@ export class MediaUsageRepository {
...
+				join
+					.onRef("u.source_key", "=", "s.source_key")
+					.onRef("u.generation", "=", "s.current_generation"),
+			)
+			.select(currentUsageSelect);
+	}
+
+	private async deleteSourceKeys(
+		sourceKeys: readonly string[],
+		order: "source-first" | "usage-first",
```
ascorbic: I don't think this should have two modes: it should always be source-first. If it does usage-first, there's a risk of having invalid source if a deletion fails. This is down to D1 not having real transactions. With source-first there is a risk of orphan rows, but they can be cleaned up – they don't leave it in an incorrect state

ascorbic (commented):
Great stuff. Just a couple of notes. However this is a very large PR. For future ones, can you keep them smaller in scope so they're easier to review.

.changeset/media-usage-index-hardening.md
khoinguyenpham04: agreed. this was too large, I’ll keep follow-up PRs smaller and split them by surface area next time

packages/core/src/media/usage/content-refresh.ts
khoinguyenpham04: Fixed, refresh now retries generation conflicts.

packages/core/src/database/repositories/media-usage.ts
khoinguyenpham04: Fixed, deletes are now always source-first.

masonjames:
Tested this on a real production site (masonjames.com — Ghost→EmDash migration, Node + SQLite, downstream fork with local media patches). Results:

- Merge + tests: merged cleanly, 4364/4364 core tests passing including the new media-usage suites.
- Migration on live data: 050_media_usage_index_status applied cleanly at boot on the production SQLite DB — all three tables present, ledger correct, no boot noise.
- Write-side, end-to-end: inserting a media-library image into a post and publishing produced exactly the expected _emdash_media_usage row; lifecycle writes create content:{collection}:{id}:columns source rows as designed. Alt/caption edits and unreferenced uploads correctly created no rows. All collections initialized as stale / CONTENT_USAGE_STALE — the intended "incomplete until rebuild" state. No route errors, no interference with our local media patches.
- Field observation for later phases: pre-existing content yields zero usage rows here — migrated bodies embed media as absolute URLs on a custom media domain, so there are no managed references to extract. Expected given this PR's scope, but imported sites will depend on repair/rebuild to populate the index — and it may be worth attributing absolute URLs that match a configured media/public URL, since WordPress and Ghost migrations commonly end up in this shape.

Happy to re-test when the read API / rebuild phases land.

khoinguyenpham04:
@masonjames Thank you so much for testing this out! This is actually a really helpful validation :)

Also, I agree URL attribution for configured media/public domains is worth considering in the rebuild/import follow-up.

masonjames:
@khoinguyenpham04 awesome! I'm really excited about the media components in emdash. Feel free to tag me on PRs whenever you'd like additional testing. Always happy to do so. 🙌

ascorbic (approved):
Thanks! And thanks @masonjames for the review

PR #1818, opened 2026-07-04
fix(core): auto-select FTS5 snippet column instead of hardcoding title

### LINKED ISSUE #1817: Bug: search snippet always shows the title, never the field that actually matched
## What happened

Every search result's "snippet" is identical to its title, even when the query only
matched somewhere in the body. E.g. searching "houston" on a site with FTS search
enabled returns results where the highlighted `<mark>` term never appears in the
shown snippet, because the snippet is really just the title again.

## Root cause

`searchSingleCollection` in `packages/core/src/search/query.ts` calls:

    snippet("${ftsTable}", 2, '<mark>', '</mark>', '...', 32)

Column index `2` is hardcoded. In the FTS5 virtual table the column layout is
`[id(0), locale(1), ...searchableFields]`, and `searchableFields` comes from
`FTSManager.getSearchableFields()`, which returns fields in DB row order with no
`ORDER BY` — in every shipped seed, that's `title` first. So `snippet()` is always
told to extract from the title column, regardless of which column the row actually
matched through.

## Fix

FTS5 supports automatic column selection: passing a **negative** column index tells
SQLite to scan every indexed column itself and extract the snippet from whichever
one has the most matching tokens (https://sqlite.org/fts5.html — the `snippet()`
function, `-1` "let FTS5 choose"). Changing the hardcoded `2` to `-1` fixes this with
no schema change and no per-field logic.

## Follow-up (not in this issue's fix)

Body/richtext fields are indexed as raw Portable Text JSON (`content='...'` points
straight at the JSON column). If a Portable Text field ends up being the best-match
column, the auto-selected snippet will contain raw JSON tokens (`_type`, `block`,
`span`...) rather than prose. Core already ships `extractPlainText()` /
`extractSearchableFields()` (`search/text-extraction.ts`) to convert Portable Text to
clean text, but nothing wires them into FTS indexing yet. Flagging as separate,
larger follow-up work — not fixing in this PR.

I'll follow up with a PR shortly.

marcusbellamyshaw-cell:
## What does this PR do?

Fixes search snippets always showing the title verbatim, even when the search term
only matched in the body. `searchSingleCollection` called FTS5's `snippet()` with a
hardcoded column index of `2` (always the first searchable field / title). Switches
to FTS5's automatic column-selection mode (`-1`), so the snippet is drawn from
whichever indexed column actually contains the match.

Closes #1817

Note: if a query only matches a rich-text (Portable Text) body field, the
auto-selected snippet may show raw Portable Text JSON tokens rather than
clean prose — plain-text extraction isn't wired into indexing yet (tracked
as follow-up in #1817). Title/plain-text-field matches are unaffected.

## Type of change

- [x] Bug fix

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (targeted: `packages/core/tests/integration/search`, `packages/core/tests/unit/search` — 14/14)
- [ ] User-visible strings in the admin UI are wrapped for translation (n/a — no admin UI strings touched)
- [ ] New features link to an approved Discussion (n/a — bug fix, not a feature)

## Screenshots / test output

```
RUN  v4.1.5
 Test Files  4 passed (4)
      Tests  14 passed (14)
```

### DIFF
--- a/.changeset/search-snippet-auto-column.md
+++ b/.changeset/search-snippet-auto-column.md
@@ -1,0 +1,7 @@
+---
+"emdash": patch
+---
+
+Fixes search results always showing the title as the "snippet," even when the
+search term only appeared in the body. The snippet now reflects whichever field
+actually matched.
--- a/packages/core/src/search/query.ts
+++ b/packages/core/src/search/query.ts
@@ -290,7 +290,9 @@ async function searchSingleCollection(
 	// Using raw SQL because Kysely doesn't have FTS5 support
 	const bm25Expr = bm25Args ? `bm25("${ftsTable}", ${bm25Args})` : `bm25("${ftsTable}")`;
 
-	// Snippet column index is 2 (after id=0, locale=1, first searchable field=2)
+	// Negative column index tells FTS5 to scan every indexed column and
+	// extract the snippet from whichever one has the most matching
+	// tokens, instead of always the first searchable field (title).
 	let results;
 	try {
 		results = await sql<{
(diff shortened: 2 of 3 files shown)


marcusbellamyshaw-cell:
All checks green, no scope-gate applies (bug fix, closes #1817). Marking ready for review.

marcusbellamyshaw-cell:
Still active — no conflicts, checks green, just waiting on review.

marcusbellamyshaw-cell:
Still active — all checks green (closes #1817), just waiting on maintainer review. (Draft status reflects the change being small/low-risk, not incompleteness — happy to mark ready if that's blocking review.)

PR #1822, opened 2026-07-04
fix(core): reclaim object-cache epoch reads orphaned by cancelled requests

### LINKED ISSUE #1821: Object cache: epoch read started by a cancelled request never settles and wedges every later read of the namespace
### What happens

With the object cache enabled, a request that is **cancelled** (client disconnect) while its namespace **epoch read** is in flight permanently wedges that namespace's epoch reads for the whole isolate: every later request that needs the same namespace awaits a promise that will never settle and hangs until the edge kills it.

Production telemetry (EmDash 0.27.0, KV object cache, Cloudflare Workers, 100% invocation sampling):

- First request to a non-existent slug (e.g. `/clinics/<unknown>`) renders a 404 normally; the requester (typically a bot) disconnects early.
- The **next** request touching the same content namespace hangs: `outcome: canceled`, wall time pinned at the edge timeout (>30 s observed, up to ~100 s), single-digit-ms CPU — the render started, then parked on an await that never settles.
- Repeats indefinitely for the lifetime of the isolate; a redeploy (isolate recycle) heals it instantly.

This is the same failure class as the session-store hang fixed in #1538, on a different await.

### Why

`getEpoch` (`packages/core/src/object-cache/index.ts`) dedupes concurrent epoch reads by storing the in-flight promise in the isolate-global `epochCache` and returning it to later callers:

```ts
if (cached?.promise) return cached.promise;
```

The read itself is wrapped in `withTimeout`, and the comment next to the sharing site says the timeout "guarantees `promise` settles". That holds for a *stalled backend*, but not for a *cancelled request*: on workerd, when the request that created the promise is cancelled mid-await, its continuations never run — **including the timeout's own `setTimeout` callback**. So:

- the shared promise neither resolves nor rejects,
- the success/catch handler that would replace the cache entry never runs,
- and every subsequent request (a different request context, awaiting a promise created by the dead one) parks forever.

This is exactly the hazard `utils/init-lock.ts` and `utils/single-flight-cache.ts` document and defend against ("we deliberately never await a promise created by another request"); the epoch cache predates that pattern and still shares the bare promise.

Cancelled-mid-read requests are not exotic: bot traffic to dead slugs aborts early and 404 routes are typically uncached, so this path gets hit constantly on public sites.

### Fix direction

PR follows shortly, using the established pattern, scoped to `getEpoch`:

1. **Bounded waiters** — a caller that finds an in-flight read races it against its *own* timer (which fires even when the owner is dead) and degrades to the last-known epoch on timeout, matching the existing "backend errors are non-fatal" semantics.
2. **Deadline reclaim** — the entry records when the read started; once it is older than the read timeout plus grace, the owner must be dead and the next caller starts a fresh read instead of joining the dead one.
3. **Anchored read** — the owner's read is handed to `after()` (waitUntil), so where a lifetime extender exists, a cancelled originator's read still completes and publishes, preventing the wedge rather than merely recovering from it.

### Environment

- EmDash 0.27.0, `@emdash-cms/cloudflare` KV object cache, Astro 7, Cloudflare Workers (`compatibility_date: 2026-07-01`, `nodejs_compat`)
- Observed live on 2026-07-04; mitigated on our site by edge-caching 404 responses for 60 s so repeat requests stop reaching the wedged path — a workaround, not a fix.

swissky:
## What does this PR do?

Fixes #1821

`getEpoch` stores the in-flight namespace epoch read in the isolate-global `epochCache`, and later callers awaited that shared promise bare. On workerd, a request cancelled mid-await never runs its continuations — including the read's own timeout timer — so the shared promise never settles and every later read of that namespace hangs until the isolate is evicted. Observed in production as repeat requests to the same 404 slug hanging >30 s with `outcome: canceled` and ~0 CPU. This is the same hazard class as the session-store fix in #1538 and the documented init-lock/single-flight-cache contract ("never await a promise created by another request").

The fix keeps the dedup but makes it cancellation-safe:

1. **Bounded waiters** — a caller that finds a foreign in-flight read races it against its *own* timer (which belongs to the waiter's request context and therefore always fires) and degrades to the last-known epoch on timeout, matching the existing "backend errors are non-fatal" semantics.
2. **Deadline reclaim** — the cache entry records when the read started; a read older than the timeout plus grace means the owner is dead, and the next caller reclaims the entry with a fresh read instead of joining the dead one.
3. **Anchored read** — the owner's read is anchored via `after()`/waitUntil, so where the platform allows, a cancelled originator's read still settles and publishes instead of wedging.

Includes two regression tests and a changeset.

Verified failing-test-first: with the `index.ts` fix stashed, both new tests fail (time out awaiting the never-settling shared promise); with the fix restored, all 23 tests in `tests/unit/object-cache.test.ts` pass.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

## Screenshots / test output

Targeted file (`pnpm --filter emdash test run tests/unit/object-cache.test.ts`): 23/23 passed.

Full core suite (`pnpm --filter emdash test run`): 4271 passed, 8 failed — all 8 failures are pre-existing and unrelated (`tests/unit/astro/vite-config.test.ts` cannot resolve `@emdash-cms/admin/dist`, whose build crashes locally on a Node 24.1.0 internal ESM bug, `ERR_INTERNAL_ASSERTION: Unexpected module status 3`).

### DIFF
--- a/.changeset/object-cache-epoch-cancelled-owner.md
+++ b/.changeset/object-cache-epoch-cancelled-owner.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes object-cache reads hanging indefinitely after a request was cancelled mid-read. A namespace epoch read started by a request that disconnects (for example a bot aborting on a 404) could leave a never-settling shared promise behind, wedging every later read of that namespace until the isolate was recycled. Waiting requests now bound the shared read with their own timer, dead reads are reclaimed after a deadline, and in-flight reads survive the originating request's cancellation where the platform allows.
--- a/packages/core/src/object-cache/index.ts
+++ b/packages/core/src/object-cache/index.ts
@@ -77,6 +77,55 @@ interface EpochEntry {
 	at: number;
 	/** In-flight read, so concurrent callers share one backend round-trip. */
 	promise?: Promise<number>;
+	/**
+	 * `Date.now()` at which the in-flight read started. Callers use it to
+	 * decide whether the read's owner can still be alive (see
+	 * {@link epochReadDeadline}) — past the deadline the promise is presumed
+	 * dead and the next caller reclaims by starting a fresh read.
+	 */
+	promiseAt?: number;
+}
+
+/**
+ * Extra time past the configured read timeout before an in-flight epoch read
+ * is presumed dead. A live owner's read settles within `timeout` (enforced by
+ * {@link withTimeout}), so the grace only needs to absorb scheduling jitter.
+ */
+const EPOCH_READ_GRACE_MS = 1_000;
+
+/**
+ * How long an in-flight epoch read may be trusted. A read older than this can
+ * only exist because its owning request was cancelled mid-await — on workerd
+ * a cancelled request's continuations never run, *including the timeout's own
+ * `setTimeout` callback*, so the shared promise never settles and is never
+ * replaced. With the timeout disabled (`timeout: 0`) the default is used as
+ * the deadline; the worst case of guessing too short is one duplicate
+ * backend read.
+ */
+function epochReadDeadline(): number {
+	const timeout = holder.config.timeout > 0 ? holder.config.timeout : DEFAULT_TIMEOUT_MS;
+	return timeout + EPOCH_READ_GRACE_MS;
+}
+
+/**
+ * Await another request's in-flight epoch read, bounded by the waiter's own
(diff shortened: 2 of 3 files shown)


ascorbic (approved):
Thanks

PR #1845, opened 2026-07-06
fix(core): read OAuth env via virtual module, not locals.runtime.env

### LINKED ISSUE #1736: OAuth login (Google/GitHub) fails with "Failed to start OAuth flow" — route reads Astro.locals.runtime.env, removed in Astro 6+
### Description

On a Cloudflare deployment, clicking any OAuth login button ("Sign in with Google" / GitHub) never reaches the provider. `GET /_emdash/api/auth/oauth/<provider>` immediately 302-redirects back to `…/admin/login?error=oauth_error&message=Failed%20to%20start%20OAuth%20flow.`

Passkey login works fine, so the site is otherwise healthy, and the provider credentials are correctly configured (verified via `wrangler secret list`) — so this is **not** a config problem.

**Expected:** 302 redirect to `https://accounts.google.com/o/oauth2/v2/auth?...` to begin the OAuth flow.

**Actual:** 302 redirect to `/_emdash/admin/login?error=oauth_error&message=Failed%20to%20start%20OAuth%20flow.`

It reproduces identically for the GitHub provider, so it's not Google-specific.

## Root cause
`src/astro/routes/api/auth/oauth/[provider].ts` reads the OAuth config from `locals.runtime.env`:

```ts
const providers = getOAuthConfig(locals.runtime?.env ?? import.meta.env);
```

On Astro 6+, `locals.runtime.env` was removed and now **throws** on access — so the `?.` optional chain doesn't help (the getter throws rather than returning `undefined`), and the `?? import.meta.env` fallback is never reached. The `try/catch` swallows it into the generic "Failed to start OAuth flow." Because it throws before the `if (!providers[provider])` check, you get `oauth_error` instead of `provider_not_configured`, which masks the real cause even when credentials are set.

## Impact
All OAuth providers (Google + GitHub) are unusable on Astro 6+/Cloudflare; passkeys are unaffected. `emdash`'s `peerDependencies` declare `astro: ">=6.0.0-beta.0"`, so this config is within the supported range.

## Suggested fix
Read env via `import { env } from "cloudflare:workers"` (as elsewhere in the codebase) instead of `locals.runtime.env`:

```ts
const providers = getOAuthConfig(env ?? import.meta.env);
```

The identical line is still present in 0.24.0, 0.25.1, and 0.27.0. Happy to test a patch.

### Steps to reproduce

1. Enable a provider: `emdash({ authProviders: [google()] })` (GitHub reproduces identically).
2. Set the provider client id/secret as Worker secrets (`EMDASH_OAUTH_GOOGLE_CLIENT_ID` / `EMDASH_OAUTH_GOOGLE_CLIENT_SECRET`).
3. Deploy to Cloudflare (Astro 7 / `@astrojs/cloudflare` 14).
4. On the login page, click "Sign in with Google" (→ `GET /_emdash/api/auth/oauth/google`).

### Environment

- `emdash` 0.23.0 (identical buggy line also present in the current published source of 0.24.0, 0.25.1, and 0.27.0 — latest)
- `astro` 7.0.3
- `@astrojs/cloudflare` 14.0.1
- `output: "server"`, Cloudflare Workers (D1 + R2)
- Provider: `authProviders: [google()]`; client id/secret set as Worker secrets (confirmed present)

### Logs / error output

```shell
From `wrangler tail` when the OAuth start route is hit:


OAuth initiation error:
Error: Astro.locals.runtime.env has been removed in Astro v6. Use 'import { env } from "cloudflare:workers"' instead.
```

marcusbellamyshaw-cell:
## What does this PR do?

Astro 6 removed `Astro.locals.runtime.env` — accessing it now throws rather than returning `undefined`, so the existing `locals.runtime?.env ?? import.meta.env` fallback never reached `import.meta.env`: the throw propagated into the route's outer try/catch and surfaced as a generic `oauth_error`, masking the real `provider_not_configured` diagnosis. Both OAuth routes (start + callback) hit this; passkey auth was unaffected since it doesn't read this env.

Added a `virtual:emdash/env` module (mirrors the existing `virtual:emdash/wait-until` pattern exactly): re-exports Cloudflare's `env` from `cloudflare:workers` under that adapter, `undefined` otherwise. Core stays adapter-agnostic this way — a direct `cloudflare:workers` import would fail to resolve on a Node build. Both routes now read through it instead of `locals.runtime`.

Closes #1736

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] User-visible strings in the admin UI are wrapped for translation (n/a — server route, no UI strings)

## Screenshots / test output

`packages/core/tests/unit/astro/integration/virtual-modules.test.ts` + `tests/unit/auth/oauth-start-route.test.ts` — 20/20 pass; `pnpm lint:quick` clean; `pnpm typecheck` clean.

Opened as draft: fork already has several open PRs against this repo (non-collaborator open-PR limit).


afonsojramos:
Can confirm that this solves the issue

segmentationfaulter:
can it be merged? This is the only PR left which will fix oAtuh providers login.

ascorbic (approved):
Thanks

PR #1820, opened 2026-07-04
docs: add Edge Caching guide (Workers Cache + tag purge)

swissky:
## What does this PR do?

Adds a deployment guide (`/deployment/edge-cache/`) for caching rendered HTML in the Cloudflare Workers Cache via Astro route caching: the `cacheCloudflare` provider + `routeRules`, tag-based purge-on-publish through `content:afterSave`/`content:afterDelete` hooks, and the responses that must opt out of the edge cache (preview-token requests, error responses, middleware redirects).

It also registers the page in the docs sidebar (after Object Cache), adds an "Edge Caching" section to `deployment/cloudflare.mdx` pointing to the new page, and fixes the cross-link in `object-cache.mdx`, which sent readers to `/deployment/cloudflare/` for edge HTML caching although that page had no such section.

This grew out of the context around discussion #1742 (editors behind shared caches seeing stale content): the page documents the shared-cache behavior editors will encounter and links back to that discussion.

## Type of change

- [ ] Bug fix
- [ ] Feature (requires [maintainer-approved Discussion](https://github.com/emdash-cms/emdash/discussions/categories/ideas))
- [ ] Refactor (no behavior change)
- [ ] Translation
- [x] Documentation
- [ ] Performance improvement
- [ ] Tests
- [ ] Chore (dependencies, CI, tooling)

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md)
- [ ] `pnpm typecheck` passes
- [ ] `pnpm lint` passes
- [ ] `pnpm test` passes (or targeted tests for my change)
- [x] `pnpm format` has been run
- [x] I have added/updated tests for my changes (if applicable)
- [x] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable). Do not include `messages.po` changes except in translation PRs — a workflow extracts catalogs on merge to `main`.
- [x] I have added a [changeset](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#changesets) (if this PR changes a published package)
- [ ] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/...

[...]

Issue #1398, opened 2026-06-10
Editor toolbar injection doesn't set Cache-Control: private, no-store, poisoning shared CDN caches


virafb:
### Summary

The request-context middleware injects the editor toolbar into public HTML responses for a logged-in editor, but the plain-editor branch does **not** set `Cache-Control: private, no-store` — unlike the preview branch, which does. On any site fronted by a shared cache (Cloudflare, etc.) where HTML is cacheable, an editor simply *browsing* the public site primes the shared edge cache with toolbar-bearing HTML that is then served to **all anonymous visitors** until the entry expires.

### Where

`dist/astro/middleware/request-context.mjs`, `onRequest` (observed in `0.15.0`):

- The preview path sets the header correctly:

```js
if (preview) {
  response = new Response(response.body, response);
  response.headers.set("Cache-Control", "private, no-store");  // ✅
}
```

- The plain-editor path injects the toolbar with no such header:

```js
if (isEditor) return injectToolbar(await next(), renderToolbar({ editMode: false, isPreview: false }));  // ❌ no Cache-Control
```

### Impact

The toolbar markup (and the fact a session is active) leaks to anonymous visitors via the cache. It's session-specific response mutation, so it should never be shared-cacheable — exactly the reason the preview branch already sets `private, no-store`.

### Repro

1. Astro + EmDash site behind Cloudflare with a cache rule (or any CDN that caches HTML on a cacheable origin `Cache-Control`).
2. As a logged-in editor, request a cacheable public page (cache MISS) — response includes `#emdash-toolbar`.
3. As an anonymous visitor, request the same URL — cache HIT, the editor toolbar HTML is served.

### Suggested fix

Set `Cache-Control: private, no-store` on any toolbar-injected response, mirroring the preview branch — i.e. whenever `injectToolbar` runs.

Bonus (CSP, secondary): the toolbar's inline `<style>`/`<script>` aren't covered by a hash-based `security.csp`, since injection happens after render — they get CSP-blocked and render unstyled. A nonce/hash hook for the injected assets would round this out, but the cache header is the security-relevant half.

### Version

Observed in `0.15.0`. Happy to confirm against `0.17.x` if helpful.

MA2153:
This seems like it should be an Astro server island, or we invent a new client component to handle actions by authenticated users. So the original page stays cache-eligible.

PR #1855, opened 2026-07-07
fix(preview): opt preview and toolbar responses out of the route cache

swissky:
## What does this PR do?

With route caching enabled (Workers Cache on Cloudflare), preview responses and editor-toolbar HTML end up in the **shared edge cache** and get served to subsequent requests without any token verification or worker invocation.

Verified empirically on a stock Cloudflare deploy (0.27.0, the documented `routeRules` setup): a valid preview URL (`?_preview=<token>`) returned `cf-cache-status: MISS` with `Cache-Control: private, no-store` on request 1 — and shared-cache `HIT`s on requests 2 and 3. The draft content stays reachable until TTL/purge, **including after the token expires**. (First reported in the #1820 review thread.)

Root cause: the request-context middleware sets only the `Cache-Control` header on preview/toolbar responses. That governs browsers and downstream proxies — but the edge cache follows Astro's **route-cache options** (on Cloudflare the adapter emits `Cloudflare-CDN-Cache-Control` from them), which the middleware never touched.

The fix:

- Any request carrying a `_preview` param now calls `context.cache.set(false)` (opt out of route caching). This is deliberately keyed on the param's presence rather than token validity: preview URLs are per-token, so cached copies are useless at best and drafts at worst.
- Toolbar-injected editor responses opt out the same way (their `Cache-Control: private, no-store` guard from #1398 had the same blind spot). Non-HTML editor responses are unaffected and stay cacheable.
- The middleware uses Astro's typed `context.cache` API directly (EmDash requires Astro 6+, which always provides it — thanks @ascorbic). Dev mode and deployments without a cache provider are unaffected: `NoopAstroCache`/`DisabledAstroCache` accept `set(false)` as a no-op.

New unit tests cover: preview param present → opt-out; toolbar injected → opt-out; editor non-HTML response → cache untouched; anonymous fast path → cache untouched.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] User-visible strings in the admin UI are wrapped for translation *(n/a — no admin UI changes)*
- [ ] New features link to an approved Discussion: *(n/a — bug fix)*

## Screenshots / test output

Measurements from the stock demo deploy (before the fix):

```
request 1: cf-cache-status: MISS   cache-control: private, no-store
request 2: cf-cache-status: HIT
request 3: cf-cache-status: HIT
```

```
✓ tests/unit/astro/request-context-route-cache.test.ts (5 tests)
✓ tests/unit/astro/ (133 tests total)
```

### DIFF
--- a/packages/core/src/astro/middleware/request-context.ts
+++ b/packages/core/src/astro/middleware/request-context.ts
@@ -19,10 +19,36 @@ import { getRequestContext, runWithContext } from "../../request-context.js";
 import { renderToolbar } from "../../visual-editing/toolbar.js";
 
 /**
+ * The subset of Astro's route-cache API the middleware needs. `context.cache`
+ * ships with Astro's route caching; typed structurally (and called
+ * optionally) so the middleware stays compatible with Astro versions that
+ * predate it.
+ */
+interface RouteCacheLike {
+	set(input: false): void;
+}
+
+/**
+ * Opt the current request out of Astro's route cache (e.g. Workers Cache on
+ * Cloudflare). `Cache-Control` headers do NOT cover this: the adapter derives
+ * the shared-cache TTL from the route-cache options (on Cloudflare via
+ * `Cloudflare-CDN-Cache-Control`), so session-specific responses must
+ * explicitly disable it or they get stored in the shared cache and served to
+ * anonymous visitors without ever invoking the middleware again.
+ */
+function optOutOfRouteCache(cache: RouteCacheLike | undefined): void {
+	cache?.set(false);
+}
+
+/**
  * Inject toolbar HTML into a response if it's an HTML page.
  * Returns the original response if not HTML.
  */
-async function injectToolbar(response: Response, toolbarHtml: string): Promise<Response> {
+async function injectToolbar(
+	response: Response,
+	toolbarHtml: string,
+	routeCache: RouteCacheLike | undefined,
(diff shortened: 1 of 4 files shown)


packages/core/src/astro/middleware/request-context.ts
```diff
@@ -18,11 +18,37 @@ import { verifyPreviewToken, parseContentId } from "../../preview/tokens.js";
...
 
+/**
+ * The subset of Astro's route-cache API the middleware needs. `context.cache`
+ * ships with Astro's route caching; typed structurally (and called
+ * optionally) so the middleware stays compatible with Astro versions that
+ * predate it.
+ */
+interface RouteCacheLike {
+	set(input: false): void;
+}
```
ascorbic: EmDash requires Astro 6, which has the route cache API

swissky: Good point, thanks — dropped the structural interface and the defensive read in a5dbcb8. The middleware now uses `context.cache` directly (typed via `APIContext["cache"]`), which also let me remove the type-assertion lint suppression and the "absent cache" test.

ascorbic (approved):
Thanks!

PR #1857, opened 2026-07-07
perf(admin): serve the content-list search filter from the FTS5 index

### LINKED ISSUE #1517: Admin content-list search filter does an unindexable full-table scan
## Description

The admin content-list `q` filter (the search box on a collection list view) runs a substring `LIKE` that cannot use any index, so it scans every non-deleted row in the locale on each search — including just to compute the result count. On large collections and on D1 (where this routes to the primary and reads every row) the rows-read / rows-returned ratio is terrible.

The generated count query looks like:

```sql
select count("id") as "count"
from "ec_resources"
where "deleted_at" is null
  and "locale" = ?
  and (lower("slug") like lower(?) escape '\'
       or lower("title") like lower(?) escape '\')
```

Two things make it unindexable by construction:

1. **Leading `%` wildcard** — `pattern = `%${term}%``, a substring match that no B-tree index can serve.
2. **`lower(col)` wraps the column**, defeating the index on `slug`/`title` regardless of the wildcard.

Notably, EmDash already ships an FTS layer (fts5 + bm25, `packages/core/src/search/`) that the content-list filter does not use. For collections with `search_config.enabled`, the list filter could be backed by FTS and fall back to `LIKE` otherwise.

Source of the query: `ContentRepository.applySearchFilter` / `count`:
- https://github.com/emdash-cms/emdash/blob/main/packages/core/src/database/repositories/content.ts#L786-L805
- https://github.com/emdash-cms/emdash/blob/main/packages/core/src/database/repositories/content.ts#L835-L860

## Steps to reproduce

1. Create a collection with a non-trivial number of rows (e.g. `resources`).
2. Open the collection list view in the admin and type a search term.
3. Inspect the issued SQL / D1 analytics: the `count(...)` and `find(...)` queries scan the full non-deleted set for the locale; rows-read greatly exceeds rows-returned.

## Environment

- emdash version: 0.20.0
- Runtime: Cloudflare Workers (D1) — also applies to Node/SQLite

## Logs / error output

No error — this is a performance/scalability issue (full scan, poor rows-read/returned ratio), not a crash.

## Possible direction

- Back the content-list filter with the existing FTS5 index when the collection has search enabled; fall back to `LIKE` otherwise.
- Alternatively, drop or cap the redundant full-scan `count()` on search (e.g. `LIMIT n` + "n+").

swissky:
## What does this PR do?

Fixes #1517 (milestone 1.0): the admin content-list search box generated

```sql
lower("slug") like lower(?) escape '\' or lower("title") like lower(?) escape '\'
```

with a `%term%` pattern — doubly unindexable (leading wildcard + `lower()` wrapping the column) — so **every keystroke scanned all non-deleted rows in the locale, twice** (page fetch + the count for the pagination denominator). On D1 these are primary reads with a terrible rows-read/rows-returned ratio.

As the issue suggests, the filter now uses the FTS layer EmDash already ships. When the collection has search enabled and the index covers the display columns, the `q` filter becomes:

```sql
id IN (SELECT id FROM _emdash_fts_<slug> WHERE _emdash_fts_<slug> MATCH '"term"*')
OR slug GLOB 'term*'
```

Both sides are index-backed. `EXPLAIN QUERY PLAN` before/after on a seeded table:

```
OLD:  SCAN ec_posts
NEW:  MULTI-INDEX OR
        SCAN _emdash_fts_posts VIRTUAL TABLE INDEX 0:M3   (inverted index)
        SEARCH ec_posts USING INDEX idx_ec_posts_slug (slug>? AND slug<?)
```

Design notes:

- **Routing is conservative.** `useFts` is set by the handler only when: SQLite (FTS5 is SQLite-only), `search_config.enabled`, every non-slug display column (`title`/`name`) is in the searchable-field set (otherwise FTS would miss matches LIKE finds), and the FTS table exists. Everything else — Postgres, search disabled, uncovered columns — keeps the existing LIKE fallback, unchanged.
- **The MATCH expression cannot syntax-error.** A new `buildFtsPrefixMatch` helper quotes every term (interior quotes doubled) and treats FTS operators as plain words — unlike the public search API's `escapeQuery`, which intentionally passes operators through. A list filter box is not a search-syntax field.
- **Slug lookups keep working.** The slug is not in the FTS index, so the filter OR's an index-served `slug GLOB 'term*'` prefix (GLOB metacharacters escaped via character classes; GLOB's case-sensitive BINARY comparison is what lets SQLite use the ordinary slug index, and slugs are lowercase by construction).
- **Semantics change on FTS-enabled collections** from arbitrary substring to token-prefix matching (`need head` finds "Needle Headline"; mid-word `eedle` no longer matches). That's standard search behavior and what the FTS index can serve; the count and the page use the same filter, so the pagination denominator stays consistent.
- Index integrity is already handled elsewhere: boot-time `verifyAndRepairAll` plus migration 039.

Tests: unit tests for the two expression builders; integration tests for the FTS path (title token prefix, case-insensitivity, multi-word, slug prefix, filtered `total`, literal quotes, operators as plain words, soft-delete sync) and for the LIKE fallback when a display column isn't covered by the index. The existing `content-list-search.test.ts` (both dialects) passes unchanged.

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [x] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] User-visible strings in the admin UI are wrapped for translation *(n/a — no admin UI changes)*
- [ ] New features link to an approved Discussion: *(n/a — fixes the triaged milestone-1.0 issue #1517, which already outlines this approach)*

## Screenshots / test output

```
✓ tests/unit/search/match.test.ts (6 tests)
✓ tests/integration/content/content-list-search-fts.test.ts (10 tests)
✓ tests/integration/content + tests/integration/search + content-handlers (90 tests total)
```

### DIFF
--- a/.changeset/admin-search-fts.md
+++ b/.changeset/admin-search-fts.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Speeds up the admin content-list search on collections with full-text search enabled: the filter is now served from the FTS5 index (token-prefix matching plus slug-prefix lookup) instead of scanning every row. Collections without search enabled, and PostgreSQL databases, keep the previous substring matching.
--- a/packages/core/src/api/handlers/content.ts
+++ b/packages/core/src/api/handlers/content.ts
@@ -5,6 +5,7 @@
 import type { Kysely } from "kysely";
 import { sql } from "kysely";
 
+import { isSqlite } from "../../database/dialect-helpers.js";
 import { BylineRepository } from "../../database/repositories/byline.js";
 import type { ContentBylineInput } from "../../database/repositories/byline.js";
 import { CommentRepository } from "../../database/repositories/comment.js";
@@ -31,6 +32,7 @@ import type { Database } from "../../database/types.js";
 import { validateIdentifier } from "../../database/validate.js";
 import { getI18nConfig, isI18nEnabled } from "../../i18n/config.js";
 import { invalidateRedirectCache } from "../../redirects/cache.js";
+import { FTSManager } from "../../search/fts-manager.js";
 import { invalidateTermCache } from "../../taxonomies/index.js";
 import { isMissingTableError } from "../../utils/db-errors.js";
 import { encodeRev, validateRev } from "../rev.js";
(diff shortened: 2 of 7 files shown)


ascorbic (approved):
Nice. Thanks

PR #1898, opened 2026-07-09
fix(core): plugin storage where-filters fail on Postgres with boolean = integer (#920)

### LINKED ISSUE #920: Postgres: plugin-storage list/count fail with "operator does not exist: boolean = integer" when filtering
## Summary

When EmDash runs against Postgres, `PluginStorage.list()` and `PluginStorage.count()` throw `operator does not exist: boolean = integer` whenever a `where` clause is supplied. SQLite is unaffected because it silently coerces booleans to integers.

In practice this surfaces as 500s on plugin admin pages that filter their own storage. The most visible victim is `emdash-smtp` (the Providers and Logs admin pages both filter), but the bug applies to any plugin that calls `list`/`count` with a filter on Postgres.

## Reproduction

EmDash `0.9.0`, Postgres (Supabase pooler, but any Postgres reproduces it):

```ts
const storage = ctx.storage<MyDoc>('my-collection');
await storage.put({ id: 'a', name: 'foo' });
await storage.list({ where: { name: 'foo' } }); // throws on Postgres
await storage.count({ name: 'foo' });            // throws on Postgres
```

Stack tail:

```
error: operator does not exist: boolean = integer
  at … kysely … select \"id\", \"data\", \"created_at\" from \"_plugin_storage\" where … = $N
```

The `= $N` (where `$N` binds `'1'`) is the smoking gun.

## Root cause

[`packages/core/src/database/repositories/plugin-storage.ts`](https://github.com/emdash-cms/emdash/blob/main/packages/core/src/database/repositories/plugin-storage.ts) wraps the joined where SQL in an `eb(..., '=', sql.raw('1'))` call at two spots:

- line ~226 (in `list`)
- line ~305 (in `count`)

```ts
query = query.where(({ eb }) =>
  eb(sql.join(whereSqlParts, sql.raw('')), '=', sql.raw('1'))
);
```

`whereSqlParts` already builds a complete boolean SQL expression. Wrapping it in `… = 1` works on SQLite (`1 = 1` after boolean→int coercion) but Postgres rejects `boolean = integer` with no implicit cast.

## Suggested fix

Drop the wrapper in both spots; pass the joined fragment to `where()` directly:

```ts
query = query.where(sql.join(whereSqlParts, sql.raw('')));
```

`Kysely.where()` accepts a raw SQL fragment that evaluates to boolean, so this is the idiomatic form and works on both dialects.

Happy to open a PR with the change plus a regression test in the existing dialect-parameterised plugin-storage suite — let me know if that'd be welcome.

## Environment

- EmDash `0.9.0`
- Astro 6, `@astrojs/vercel`, `output: 'server'`
- Postgres via `emdash/db` `postgres()` driver, Supabase session pooler
- Node 22

## Workaround

Patch `node_modules/emdash/dist/search-*.mjs` via `patch-package`, dropping the same wrapper in both bundled occurrences. Rough but unblocks Postgres deploys until upstream lands.

swissky:
## What does this PR do?

Fixes #920: on PostgreSQL, every plugin storage `query()` / `count()` call with a `where` filter failed with `operator does not exist: boolean = integer` (reported via `emdash-smtp`, whose provider lookups filter storage).

Root cause: `PluginStorageRepository` fit the raw JSON-extraction predicate into Kysely's binary `eb(lhs, "=", rhs)` by wrapping it as `(json_extract(...) = ?) = 1`. SQLite coerces booleans to integers so the extra `= 1` was harmless there; Postgres has a strict boolean type and rejects the comparison outright — so filtered reads never worked on PG.

Fix: pass the predicate directly to `.where()` as a boolean raw expression (`rawWhereExpr` helper, shared by `query()` and `count()`), keeping the existing `?`-placeholder parameterization intact. No SQL semantics change on SQLite.

Tests: new `storage-dialect.test.ts` runs the filtered query/count paths end-to-end via `describeEachDialect`, so the regression is exercised against real Postgres in CI's PG job (equality filters, `where` + `orderBy`, range filters, filtered counts). Verified green on SQLite locally; I don't have a local Postgres, so relying on the CI PG matrix for the dialect that actually regressed — the failure mode is deterministic (the old SQL is a hard type error on PG).

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (or targeted tests for my change) — plugin storage unit + integration suites
- [ ] User-visible strings in the admin UI are wrapped for translation (if applicable) — n/a

- [x] This PR includes AI-generated code — model/tool: Cursor + Claude (Fable 5)

## Screenshots / test output

```
 Test Files  6 passed (6)
      Tests  116 passed (116)   (integration/plugins, SQLite)
 Test Files  1 passed (1)
      Tests  30 passed (30)     (unit plugin-storage)
```

### DIFF
--- a/.changeset/fix-plugin-storage-postgres-filters.md
+++ b/.changeset/fix-plugin-storage-postgres-filters.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes plugin storage `query()` and `count()` failing on PostgreSQL with "operator does not exist: boolean = integer" whenever a `where` filter was supplied.
--- a/packages/core/src/database/dialect-helpers.ts
+++ b/packages/core/src/database/dialect-helpers.ts
@@ -215,17 +215,23 @@ export function binaryType(db: Kysely<any>): ColumnDataType {
 }
 
 /**
- * SQL expression for extracting a field from a JSON/JSONB column.
+ * SQL expression for extracting a field from a JSON column stored as text.
  *
  * sqlite:   json_extract(column, '$.path')
- * postgres: column->>'path'
+ * postgres: (column)::jsonb->>'path'
+ *
+ * The Postgres cast is required because JSON columns (e.g.
+ * `_plugin_storage.data`) are `text`, and `text ->> unknown` is not an
+ * operator. The cast is immutable, so the same expression works in
+ * expression indexes — queries and indexes must build it through this
+ * helper so the planner can match them.
  */
 // eslint-disable-next-line @typescript-eslint/no-explicit-any -- accepts any Kysely instance
 export function jsonExtractExpr(db: Kysely<any>, column: string, path: string): string {
 	validateIdentifier(column, "JSON column name");
 	validateJsonFieldName(path, "JSON path");
 	if (isPostgres(db)) {
-		return `${column}->>'${path}'`;
+		return `(${column})::jsonb->>'${path}'`;
 	}
 	return `json_extract(${column}, '$.${path}')`;
 }
--- a/packages/core/src/database/repositories/plugin-storage.ts
+++ b/packages/core/src/database/repositories/plugin-storage.ts
@@ -302,7 +302,8 @@ export class PluginStorageRepository<T = unknown> implements StorageCollection<T
 		}
 
 		const result = await query.executeTakeFirst();
-		return result?.count ?? 0;
+		// Number() because the pg driver returns COUNT(*) (bigint) as a string.
+		return Number(result?.count ?? 0);
 	}
 }
 
--- a/packages/core/src/database/repositories/plugin-storage.ts
+++ b/packages/core/src/database/repositories/plugin-storage.ts
@@ -7,7 +7,7 @@
  * @see PLUGIN-SYSTEM.md § Plugin Storage > Full API Reference
  */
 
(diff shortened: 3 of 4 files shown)


swissky:
CI's PG job did its job and caught a second, deeper layer of the same bug — updated in 604d00cd:

The `= 1` wrapper wasn't the only reason filtered reads never worked on Postgres. `_plugin_storage.data` is a `text` column, and `jsonExtractExpr` emitted `data->>'field'` on PG — `text ->> unknown` isn't an operator, so the extraction itself was broken too (the CI failure: `operator does not exist: text ->> unknown`). The fix casts through `(data)::jsonb->>'field'` in `jsonExtractExpr`, which is shared by the query builder *and* the expression-index builder (`storage-indexes.ts`), so indexes and predicates keep emitting the identical expression and the planner can still match them. It also means plugin storage expression indexes were never creatable on PG before this — same operator error at CREATE INDEX time.

Also swapped two `.sort()` calls in the new test for `.toSorted()` to satisfy the type-aware lint (that was the Lint job failure).

The dialect-parity tests in this PR now exercise the fixed path end-to-end on the CI PG matrix.

PR #1909, opened 2026-07-10
Plugin registry labelling service

ascorbic:
## What does this PR do?

Establishes `feat/plugin-registry-labelling-service` as the integration branch for the plugin registry labelling service described in RFC #694.

This PR tracks the complete implementation without sub-issues. Implementation PRs should target `feat/plugin-registry-labelling-service`; this umbrella PR remains draft until the production launch gates in the implementation plan are complete.

The tracked planning documents are:

- `.opencode/plans/plugin-registry-labelling-service/spec.md`
- `.opencode/plans/plugin-registry-labelling-service/implementation-plan.md`

Related to #694.

### Design and feasibility

- [x] Freeze public NSIDs, assessment API shapes, and error/cursor contracts (#1910)
- [x] Freeze label vocabulary, subject/CID rules, override precedence, and executable policy fixtures (#1910)
- [x] Audit and plan the `security:yanked` to `security-yanked` cutover (#1912)
- [x] Prove DRISL label signing and verification interoperability in workerd (#1911)
- [x] Decide labeler identity, declaration, and localized-definition hosting
- [ ] Prove verified aggregator-mirror artifact consumption and fallback inputs
- [x] Select dependency/malware scanners, SBOM tooling, and initial Workers AI models (code AI model selected via calibration data — glm-5.2, #2058; image model kimi-k2.7-code; dependency/malware/SBOM scanner + tooling selection DISSOLVED — ratified 2026-07-17 that scanners are inapplicable (dependency-free plugins) and capability analysis redundant (runtime enforces declared access), so there is nothing to select. The production prompt + inputs still require a calibration re-validation sweep before enforcement launch)
- [x] Run the read-only `security:yanked` production preflight (zero legacy rows confirmed; Branch A, no compatibility path)

### Shared protocol and policy

- [x] Apply the label vocabulary cutover across aggregator, registry client, core, and admin (cutover landed incrementally with the labeler build; the complete blocking vocabulary is centralized in `packages/registry-moderation` and consumed by core/admin/registry-client/aggregator; a full-tree audit finds zero functional `security:yanked` colon-form values remaining — only a stale comment in the already-applied historical migration `aggregator/migrations/0001_init.sql`, left untouched for migration integrity; preflight confirmed zero persisted legacy rows so no dual-read compatibility path was needed — Branch A)
- [x] Add labeler assessment and policy lexicons (#1923)
- [x] Generate and export typed contracts and fixtures (#1923)
- [x] Add shared DRISL label signing/verification primitives (#1922, #1952)
- [x] Add the single shared moderation evaluator and subject-applicability rules (#1917)
- [x] Add accepted/content-labeler header parsing and serialization (#1965)
- [ ] Publish the versioned policy schema and localized definitions

### Service foundation and signed distribution

- [x] Scaffold `apps/labeler`, bindings, CI coverage, and real workerd/D1 tests (#1926)
- [x] Add D1 migrations, typed repositories, lifecycle state machine, and evidence boundary (#1976)
- [x] Publish DID/policy identity runtime and lazy Secrets Store-compatible signer (#1942)
- [ ] Provision staging custom domain, D1, signing key, and Secrets Store binding
- [x] Implement typed role/subject issuance proposals and atomic label sequence allocation (#1926)
- [x] Implement `com.atproto.label.queryLabels` (#1926)
- [x] Implement the Subscriber Durable Object and `com.atproto.label.subscribeLabels` (#1931)
- [x] Implement routine signing-key rotation (#1938)
- [x] Return explicit `NotSupported` for deferred standard report intake (#1938)

### Aggregator and client enforcement

- [x] Implement aggregator labeller configuration and DID resolution (#1949)
- [x] Implement signed subscription ingest (#1959)
- [x] Implement append-only label history and latest-state projection (#1959)
- [x] Implement accepted-labeler request policy and `atproto-content-labelers` (#1965)
- [x] Implement batched publisher/package/release hydration without N+1 queries (#1965)
- [x] Implement endpoint filtering, redaction, cascades, and label-aware latest release selection (#1965)
- [x] Integrate the shared evaluator into the registry client (#1972)
- [x] Enforce eligibility in core install/update handlers (#1972)
- [x] Surface eligibility in current CLI discovery paths (#1972)
- [x] Update admin registry UI, warning consent, cache keys, and installed-release alerts (#1977)

### Automated assessment

- [x] Implement independent Jetstream discovery and source-record verification (#1978)
- [x] Implement idempotent Workflow orchestration, supersession, and current assessment projection (#1978; production Workflow shell over the orchestrator with the run-key instance-id lock and finalization label-leak guard in #2072; the real analysis stages are wired and the deploy gate lifted across #2090 (code-AI + history) and #2091 (acquire consumer + SSRF egress + image stage + gate lift))
- [x] Implement initial, operator, and scanner-intelligence rerun triggers (initial + operator in #1978; the scanner-intelligence rerun trigger is DISSOLVED with the scanner cut — no scanner intelligence exists to rerun on)
- [x] Implement reconciliation for missed, stuck, stale, and deleted subjects (#1978)
- [x] Establish shared artifact-verification ownership with the delegated release service work (owned by the delegated release branch's `registry-verification`; W7 extends it from main)
- [x] Implement safe mirror/fallback artifact acquisition and canonical bundle validation (declared-URL-first acquisition + checksum/CID bundle validation in #2069 over the labeler→aggregator read binding #2070; mirror-first ordering deferred to the delegated release service per the ratified declared-URL-first decision)
- [x] Implement deterministic security, SBOM, dependency, and intelligence analysis (SCOPED DOWN and ratified 2026-07-17: plugins are dependency-free and the workerd sandbox enforces declared access at runtime, so SBOM/dependency/malware scanning is inapplicable and static capability analysis is redundant — the deterministic surface reduces to integrity/bundle validation, which lands in the acquire stage (#2069/#2091); the `deterministic`/`dependency` orchestrator stages are removed (#2090). Dissolves the scanner/SBOM-tooling selection gate — nothing to select)
- [x] Implement normalized finding contracts and evidence references (#1991)
- [x] Implement Workers AI code/metadata and image adapters through the Workers AI binding (#1994, #2003)
- [x] Implement bounded publisher-history context and versioned policy resolution (versioned policy resolver in #2004; publisher-history own-D1 context stage in #2033; aggregator-sourced verification-state input in #2074 over binding #2070; the history stage is now WIRED into the production orchestrator in #2090; the handle/profile-change input remains deferred to a follow-up ticket on aggregator identity-event ingestion; console surfacing pending)
- [x] Build the model calibration and false-positive review harness (#2058 — local harness over the real adapters + live Workers AI; found and fixed the OpenAI-envelope parse gap; drove the glm-5.2 default selection)

### Operator and publisher surfaces

- [x] Implement Access JWT verification, role mapping, CSRF, and idempotent mutations (#2005, #2008)
- [x] Build the Kumo operator dashboard, assessment, subject, audit, and system views (#2006, #2015)
- [x] Implement reviewer release/package actions and reruns (#2020, #2023)
- [x] Implement atomic false-positive overrides (#2023)
- [x] Implement admin-only emergency takedown, publisher-compromise, pause, and DLQ actions (#2026, #2027, #2028, #2029)
- [x] Publish machine-readable policy and public current/historical assessment APIs (#1989)
- [x] Implement public/private serializers and contact resolution (serializers landed; publisher contact resolution in #2073 (W10.4 slice A) over the aggregator `getPublisher` read, with confirm/unsubscribe/not-me token endpoints in #2071 (W10.4 slice C))
- [x] Implement notification outbox, email delivery, and manual reconsideration workflow (operator outbox/event scaffolding in #2024; publisher confirm/unsubscribe/not-me token endpoints in #2071; publisher-facing notification delivery table + CSPRNG confirm-token + gated double-opt-in send-core in #2076 (W10.5 slice 1) and the Cloudflare Email Sending adapter + five issuance triggers + retry sweep in #2077 (W10.5 slice 2); manual reconsideration case management + auditable open/note/resolve actions + outcome notice in #2083 (W10.6 server) with the operator console UI in #2087; the sixth event — the prolonged-error two-stage escalation (24h operator alert -> 72h publisher notice) built off the reconciliation cron — in #2088. Implementation complete; actual email send is gated on the operator-owned Cloudflare Email Sending domain onboarding (a Production-readiness item), and the handle/profile-change history input is deferred to a follow-up ticket)
- [ ] Complete browser, accessibility, localization, and Arabic RTL coverage (component axe/keyboard/RTL coverage landed in #2032; browser-e2e deferred; console not localized)

### Production readiness

- [ ] Harden staging/production bindings, CI, deploy validation, and secrets handling
- [ ] Add component metrics, dashboards, logs, and alerts
- [ ] Complete key-management, incident, outage, replay, and recovery runbooks
- [ ] Implement backup, restore, retention, and contact-data minimization
- [ ] Document self-hosting and third-party labeller consumption (operator guide, ATProto reference, moderation model, and deploy/keygen docs landed in #2031; self-hosting and third-party-consumer guides pending)
- [ ] Complete protocol, registry, adversarial, query-count, performance, and browser conformance suites
- [ ] Complete calibration review and external security review
- [ ] Pass key compromise, D1 restore, cursor loss, full replay, queue, AI, mirror, and notification drills
- [ ] Assess every first-party release and enable positive-assessment enforcement through the staged rollout (the assessment pipeline is now functionally complete and live — real acquire→code-AI→image-AI→history with the deploy gate lifted, #2090/#2091; enabling enforcement is gated on the calibration re-validation sweep (production prompt + capabilities-only + real-description inputs; finalize `model_id`/`prompt_hash` provenance) and the operator-owned Cloudflare Email Sending onboarding + staging/prod provisioning)

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable). Do not include `messages.po` changes except in translation PRs; a workflow extracts catalogs on merge to `main`.

RFC #694 is the approved design source. Individual implementation PRs carry their applicable changesets and verification evidence. No admin translation catalogs are included.

- [x] This PR includes AI-generated code; model/tool: OpenCode with GPT-5.6-sol

## Screenshots / test output

Integrated gate on `2d6ad448` after #1952 merged:

- `pnpm build` passes
- `pnpm typecheck` and aggregator typecheck pass
- Type-aware lint reports 0 warnings and 0 errors
- Aggregator: 216 tests pass
- Registry moderation: 66 tests pass
- Labeler workerd integration: 29 tests pass
- #1952 completed two independent review rounds with no remaining high or medium findings

---

### Try this PR

**[Open a fresh playground →](https://feat-plugin-registry-labelling-service-emdash-playground.emdash-cms.workers.dev)**

<sub>Tracks `feat/plugin-registry-labelling-service`. Updated automatically when the playground redeploys.</sub>


ascorbic:
## Execution ledger

| Sequence | Workstream | PR | Status | Gate |
| --- | --- | --- | --- | --- |
| 01 | W0.1 + W0.2 contracts and policy fixtures | #1910 | Draft, ready for maintainer ratification | Gate 0 |
| 02 | W0.3 vocabulary cutover audit | #1912 | Draft, ready for audit review; production preflight still required | Gate 0 |
| 03 | W0.4 crypto interoperability | #1911 | Draft, ready for maintainer ratification | Gate 0 |

Wave 0A coordinator verification:

- All three branches target `feat/plugin-registry-labelling-service` and are isolated to their assigned scope.
- `pnpm lint` passes with 0 warnings/errors on all three branches.
- Formatting and `git diff --check` pass.
- Contract fixtures: 30/30 expected moderation outcomes reproduced.
- Vocabulary audit: 19/19 structural-screen cases and 10/10 expiry/quarantine cases pass.
- Crypto: independent Node vector test 1/1, workerd interoperability 12/12, both targeted typechecks pass.
- Final independent reviews report no unresolved high/medium findings.

No umbrella checklist item is marked complete until the corresponding draft PR is ratified, merged, and reverified on the integration branch.

ascorbic:
## Wave 0A complete

| Sequence | Workstream | PR | Result |
| --- | --- | --- | --- |
| 01 | W0.1 + W0.2 contracts and policy fixtures | #1910 | Ratified and merged |
| 02 | W0.3 vocabulary cutover audit | #1912 | Conditionally ratified and merged; production zero-row preflight remains required before W1.1 rollout selection |
| 03 | W0.4 crypto interoperability | #1911 | P-256 contract ratified and merged |

Gate decisions recorded:

- Public NSIDs/API, ATProto state reduction, issuance rules, cross-labeller policy, overrides, and 30 moderation outcomes are frozen.
- Canonical `security-yanked` and collision-safe ingest identity are frozen. No compatibility path is added unless the production preflight finds legacy rows.
- P-256, canonical-CBOR atcute signing, strict DID key validation, and rotation/compromise ordering are frozen.
- Retained crypto fixtures live under `packages/atproto-test-utils/tests/fixtures`; no retained source/test depends on `.opencode/plans`.

The Cloudflare API connector and local Wrangler OAuth token both returned `10000 Authentication error` for EmDash CMS D1, so the production preflight remains an explicit operator action.

ascorbic:
W4.1 merged via #1949 at eeec5786 after the CodeQL prerequisite #1948 merged at 9b504bfb. Post-merge integration gate on eeec5786 passed: full workspace build, aggregator production build, full workspace typecheck, aggregator typecheck, type-aware lint with 0 diagnostics, 216 aggregator tests, 29 labeler tests, and 61 registry-moderation tests. No tracked worktree changes remain.

ascorbic:
Shared label-verification precursor #1952 merged at 2d6ad448. Post-merge integration gate passed on that SHA: full build, full and aggregator typechecks, type-aware lint with 0 findings, 216 aggregator tests, 66 registry-moderation tests, and 29 labeler workerd tests. The shared primitive tracker item now cites #1952; signed subscription ingest remains open.

apps/aggregator/migrations/0001_init.sql
```diff
@@ -229,8 +229,8 @@ CREATE TABLE label_state (
 CREATE INDEX idx_label_state_enforce ON label_state(uri, val, trusted)
 	WHERE neg = 0 AND trusted = 1;
 
--- Trusted/known labellers (operator config, edited via deployment).
-CREATE TABLE labellers (
+-- Trusted/known labelers (operator config, edited via deployment).
+CREATE TABLE labelers (
```
ascorbic: Existing databases have already recorded `0001` with a `labellers` table. On upgrade, this edited migration is skipped and `0003` never renames the table, so every new `labelers` query fails with `no such table`. Applied migrations must remain immutable; add a forward rename migration and an upgrade test that starts from main's schema.

~ Sol 🤖

packages/registry-moderation/src/index.ts
```diff
@@ -0,0 +1,813 @@
...
+	reasonCodes: string[];
+	blockingLabels: string[];
+	stateLabels: string[];
+	warningLabels: string[];
+	suppressedLabels: string[];
+	applicableLabels: ModerationLabel[];
+	redacted: boolean;
+}
+
+const AUTOMATED_BLOCKS = new Set<string>([
```
ascorbic: The policy declares `hateful-imagery`, `explicit-imagery`, and `graphic-violence` as release-scoped hard blocks, but the shared automated/release sets omit them. The labeler can issue these values while both the evaluator and latest-release SQL ignore them, leaving a policy-blocked release installable. Please derive or validate the enforcement vocabulary against the executable policy fixture.

~ Sol 🤖

apps/aggregator/src/routes/xrpc/views.ts
```diff
@@ -107,6 +134,57 @@ export function releaseColumns(prefix = ""): string {
...
+ * untruncated per-subject sets so redaction decisions see every label;
+ * this boundary cap trims only the final view array. */
+function capLabels(labels: LabelView[], uri: string): LabelView[] {
+	if (labels.length <= LABELS_MAX_LENGTH) return labels;
+	console.warn("[aggregator] view labels truncated to maxLength", {
+		uri,
+		count: labels.length,
+		cap: LABELS_MAX_LENGTH,
+	});
+	return labels.slice(0, LABELS_MAX_LENGTH);
```
ascorbic: With more than 64 applicable labels, this slice can remove the only non-redacting hard block. Explicit-version install, update, and artifact paths then evaluate the truncated response and allow the release. The wire cap must preserve all enforcement-relevant state or force downstream evaluation to fail closed.

~ Sol 🤖

apps/labeler/src/assessment-orchestrator.ts
```diff
@@ -0,0 +1,404 @@
...
+				action,
+				{
+					uri: assessment.uri,
+					cid: assessment.cid,
+					val,
+					...(neg ? { neg: true } : {}),
+					...proposal,
+				},
+				now,
+				false,
```
ascorbic: Automated pending, outcome, block, and negation labels commit without a `LabelPublisher`, and finalization marks them non-pending without calling the subscription DO. A connected aggregator therefore remains online indefinitely without receiving these sequences until reconnect or an unrelated notification. Every committed automated label needs a durable live-publication path.

~ Sol 🤖

apps/labeler/src/subscribe-labels.ts
```diff
@@ -0,0 +1,363 @@
...
+
+	private pendingSockets(): WebSocket[] {
+		return this.ctx.getWebSockets().filter((socket) => {
+			if (socket.readyState !== WebSocket.OPEN) return false;
+			const state = this.state(socket);
+			return state.lastSent < state.targetSequence;
+		});
+	}
+
+	private async labelsAfter(cursor: number, through: number): Promise<SubscriptionEvent[]> {
```
ascorbic, line 220: After routine rotation, the DID document exposes only the new key while retained rows still carry old signatures. `queryLabels` lazily re-signs its page, but WebSocket replay reads these rows unchanged, so a fresh aggregator rejects the first old-key row and cannot advance its cursor. Retained replay must remain verifiable after every supported rotation.

~ Sol 🤖

apps/labeler/src/assessment-orchestrator.ts
```diff
@@ -0,0 +1,404 @@
...
+		if (!stillCurrent) {
+			return transitionAssessmentState(this.db, {
+				id: assessment.id,
+				from: "running",
+				to: "stale",
+				now,
+			});
+		}
+
+		const results = await this.db.batch(statements);
```
ascorbic, line 450: If signing pauses after preparation but before this batch executes, guarded issuance inserts zero rows while the unguarded assessment CAS still commits the terminal state. Post-commit throws, but Workflow retry sees a terminal assessment and never repairs its pending negation, outcome, or block labels. Terminal state and the complete label set must share one successful guard.

~ Sol 🤖

apps/labeler/src/console-mutation-api.ts
```diff
@@ -0,0 +1,1781 @@
...
+/**
+ * Best-effort live broadcast of a freshly committed label, run off the response
+ * path. Keyed on the caller's own action id: a concurrent-race loser (whose
+ * batch rolled back) reads nothing and skips, so the winner alone publishes.
+ * Publication needs the trigger-assigned `sequence`, which only exists
+ * post-commit — this uses a plain read, not `postCommit()`, so a deliberately
+ * absent row is `null` (skip) rather than a signing-state diagnosis. A missed
+ * frame is recoverable: subscribers replay from their cursor on reconnect.
+ */
+export async function publishAfterCommit(env: Env, actionId: string): Promise<void> {
```
ascorbic, line 470: A transient notify failure is caught here after a label commits with `publication_pending=1`. Replay does not redrive `afterCommit`, there is no pending-publication sweep, and rotation refuses to activate while the row remains pending. This needs a durable retry path rather than a logged fire-and-forget failure.

~ Sol 🤖

apps/labeler/src/access-auth.ts
```diff
@@ -0,0 +1,152 @@
...
+		identity = { kind: "human", email, sub, roles: [] };
+	} else {
+		throw new AccessAuthError(
+			"invalid-token",
+			"Access assertion has neither email nor common_name",
+		);
+	}
+
+	const principal = identity.kind === "service" ? identity.commonName : identity.email;
+	const principals = new Set<string>([principal, ...groupPrincipals(payload.groups)]);
```
ascorbic: Cloudflare Access places custom identity claims under `custom`, but this reads top-level `groups`. With the shipped group-only reviewer/admin mapping, normal human identities receive no role. Read and validate the configured group claim from the verified `custom` object.

~ Sol 🤖

apps/labeler/src/console-mutation-api.ts
```diff
@@ -0,0 +1,1781 @@
...
+	};
+	const returned = await commitMutation(
+		deps.db,
+		ctx,
+		commitSpec,
+		[update, eventInsert],
+		descriptor,
+	);
+
+	if (status === "retried") deferDeadLetterReenqueue(deps, id, ctx.actionId);
```
ascorbic: This commits `status='retried'` before payload decoding and queue delivery. If deferred work fails, the row is no longer actionable, health no longer counts it, and later operator retries return 409. `retried` must imply durable queue acceptance, or a recovery sweep must retain and redrive the intermediate state.

~ Sol 🤖

apps/aggregator/src/routes/xrpc/label-enforcement.ts
```diff
@@ -0,0 +1,229 @@
...
+	const sql = `
+		AND NOT EXISTS (
+			SELECT 1 FROM label_state ls
+			WHERE ls.src IN (${inClause(srcs)})
+			  AND ls.neg = 0
+			  AND (ls.exp_epoch_ms IS NULL OR ls.exp_epoch_ms > ?)
+			  AND (
+					(
+						ls.uri = 'at://' || ${releaseAlias}did || '/${NSID.packageRelease}/' || ${releaseAlias}rkey
+						AND ls.val IN (${inClause(RELEASE_BLOCK_VALUES)})
```
ascorbic: An exact-CID `assessment-passed` plus `assessment-overridden` pair suppresses the same source's automated block in the shared evaluator. This SQL still excludes the release solely because that block row exists, so an eligible overridden release disappears from latest selection. SQL and hydrated evaluation need the same override semantics.

~ Sol 🤖

apps/labeler/src/notification-triggers.ts
```diff
@@ -0,0 +1,606 @@
...
+ */
+export async function isVerifiedPublisher(
+	aggregator: Pick<AggregatorClient, "getPublisherVerification">,
+	did: string,
+	now: Date,
+): Promise<boolean> {
+	try {
+		const state = await aggregator.getPublisherVerification(did);
+		if (!state || state.verifications.length === 0) return false;
+		return state.verifications.some((claim) => isInForce(claim.expiresAt, now));
```
ascorbic: This treats any unexpired verification claim as sufficient. A publisher can self-issue a claim, put a victim address in its profile, and bypass double opt-in; trusted issuer authority and current handle/display-name bindings are not checked. Only a currently valid claim from the configured verification trust graph should upgrade a contact.

~ Sol 🤖

apps/labeler/src/notification-triggers.ts
```diff
@@ -0,0 +1,606 @@
...
+ * subject URI. A record URI (`at://did/collection/rkey`) yields the DID and the
+ * rkey as the slug — for a package record the rkey IS the package slug (tier-1/2
+ * resolve); for a release the rkey misses the package tiers and resolution falls
+ * through to the DID-keyed publisher-profile contact (tier 3), which is the
+ * reliable channel. A bare DID subject (publisher-level action) resolves by DID
+ * alone. Anything unparseable returns null and the notice is skipped.
+ */
+export function contactTargetFromUri(uri: string): ContactTarget | null {
+	if (uri.startsWith("at://")) {
+		const [did, , rkey] = uri.slice("at://".length).split("/");
```
ascorbic: For a release URI such as `.../gallery:1.2.0`, this passes the entire release rkey as the package slug. `getPackage` therefore misses `gallery`, skipping its `security[]` and `authors[]` contacts and falling back to broader publisher metadata. Parse the parent package slug before contact resolution.

~ Sol 🤖

apps/labeler/src/assessment-dispatch.ts
```diff
@@ -0,0 +1,91 @@
...
+ *
+ *   - Redelivery of the SAME discovery event recomputes the SAME runKey → same
+ *     instance id → `create` collides → idempotent no-op, no duplicate run.
+ *   - A re-assessment (operator rerun, intel re-trigger) mints a run with a new
+ *     triggerId → different runKey → different instance id → it dispatches its
+ *     own instance. A subject-only id would instead collide with the RETAINED
+ *     id of the prior completed/failed run (Cloudflare keeps instance ids ~30
+ *     days) and strand the re-assessment `pending` forever.
+ *
+ * This is per-run serialization, not per-subject: two DIFFERENT triggers for one
```
ascorbic, line 21: Different run keys allow concurrent assessments for the same URI/CID, but all runs issue the same `assessment-pending` stream. If run A finishes while newer run B remains active, A's negation clears pending and can expose an old pass before B later blocks. Per-run execution needs per-subject serialization or a pending projection that remains active while any current run is active.

~ Sol 🤖

apps/labeler/src/assessment-workflow.ts
```diff
@@ -0,0 +1,217 @@
...
+	}
+}
+
+/**
+ * The shell body: load the pending run, drive it through the orchestrator, and
+ * return the finalized state. Separate from the entrypoint class so it is
+ * testable without Cloudflare's Workflow runtime (which has no local harness) —
+ * `run` above is the thin durable-step wrapper over it.
+ */
+export async function executeAssessmentInstance(
```
ascorbic, line 81: Automation pause is checked before discovery, but an already-dispatched Workflow never reads `automation_state` and finalization has no pause predicate. An admin can receive a successful pause response while in-flight runs continue issuing automated labels. The kill switch must be enforced again at Workflow/finalization boundaries.

~ Sol 🤖

packages/registry-client/src/discovery/index.ts
```diff
@@ -169,16 +182,23 @@ export class DiscoveryClient {
...
-		const handler: typeof baseHandler = acceptLabelers
-			? async (pathname, init) => {
-					const headers = new Headers(init.headers);
-					headers.set("atproto-accept-labelers", acceptLabelers);
-					return baseHandler(pathname, { ...init, headers });
-				}
-			: baseHandler;
+		const onResponseMeta = options.onResponseMeta;
+		const handler: typeof baseHandler =
+			acceptLabelers || onResponseMeta
```
ascorbic: The protocol distinguishes a missing header from an explicitly empty header, but this truthiness check turns `acceptLabelers: ""` into omission. Configurations intending to accept no sources instead activate the aggregator's trusted defaults. Preserve the empty value on the wire.

~ Sol 🤖

apps/labeler/src/pds-verify.ts
```diff
@@ -0,0 +1,243 @@
...
+
+function buildGetRecordUrl(pds: string, did: string, collection: string, rkey: string): string {
+	const url = new URL("/xrpc/com.atproto.sync.getRecord", pds);
+	url.searchParams.set("did", did);
+	url.searchParams.set("collection", collection);
+	url.searchParams.set("rkey", rkey);
+	return url.toString();
+}
+
+async function fetchCar(
```
ascorbic, line 266: This endpoint is publisher-controlled, but the fetch follows redirects without the scheme, DNS, reserved-address, and per-hop checks used for artifact acquisition. A hostile DID/PDS document can cause blind requests to forbidden destinations. Route this path through equivalent egress validation.

~ Sol 🤖

apps/aggregator/src/routes/xrpc/router.ts
```diff
@@ -97,6 +103,24 @@ export async function handleXrpc(env: Env, request: Request): Promise<Response |
...
 	}
 
+	// Resolve the accepted-labelers policy once, before dispatch, so every
+	// handler sees the same policy the response's `atproto-content-labelers`
+	// header reports below. A malformed header 400s in the same envelope
+	// shape the router itself produces for InvalidRequest; any other failure
+	// (e.g. a D1 error) propagates so the caller 500s instead of the request
+	// silently falling open to an empty policy.
+	let policy: Awaited<ReturnType<typeof resolveRequestLabelerPolicy>>;
+	try {
```
ascorbic, line 149: Only `XRPCError` is converted here. A D1 policy-resolution failure escapes before the normal CORS and `private, no-store` wrapper, producing an opaque browser failure and violating the endpoint cache invariant. Convert unexpected policy failures into the same wrapped response path.

~ Sol 🤖

apps/labeler/src/notification-endpoints.ts
```diff
@@ -0,0 +1,297 @@
...
+	if (!action) return htmlResponse(notFoundPage(), 404);
+
+	if (request.method === "GET") return htmlResponse(formPage(action, readParams(request, action)));
+	if (request.method !== "POST")
+		return new Response("method not allowed", {
+			status: 405,
+			headers: { allow: "GET, POST", "cache-control": "no-store" },
+		});
+
+	const params = await readPostParams(request, action);
```
ascorbic, line 85: RFC one-click providers POST `List-Unsubscribe=One-Click` to the header URL while `c` remains in its query string. This POST parser reads `c` only from the form body, so the request returns success without suppressing the recipient. Preserve the query capability for one-click POSTs.

~ Sol 🤖

apps/labeler/src/assessment-store.ts
```diff
@@ -0,0 +1,1200 @@
...
+				input.fromState,
+			),
+	];
+	let pointerUpdateIndex: number | null = null;
+	if (CURRENT_POINTER_STATES.has(input.toState)) {
+		pointerUpdateIndex = statements.length;
+		statements.push(
+			db
+				.prepare(
+					`INSERT INTO current_assessments (src, uri, cid, assessment_id, updated_at)
```
ascorbic, line 717: Other assessment ordering uses `(created_at_epoch_ms, id)`, but this pointer update accepts any equal epoch. If the higher-ID run finalizes first and the lower-ID run later, the lower-ID run replaces it. Apply the same ID tie-break used by list/current semantics.

~ Sol 🤖

ascorbic (commented):
Changes are needed before merge. The inline comments cover the directly anchored blockers; the remaining cross-file findings are:

- Label-state ordering differs between the labeler, aggregator, and the ratified equal-time collision contract. Use one full-precision timestamp/collision algorithm and a shared conformance corpus.
- Production assessment provenance is disconnected: findings are not persisted, model/prompt identity remains `unassigned`, artifact pins remain null, and supersession links are never supplied.
- Recovery stops at logging for failed operator Workflow dispatch and initial discovery cursor reads. Both can leave durable work stranded.
- Conditional DLQ and reconsideration mutations can lose their guarded update while still committing success audit/result records.
- Public query bounds can exceed D1 bind and round-trip budgets in hydration, `queryLabels`, and `listAssessments`.
- Stream lifecycle has no periodic DID refresh on a healthy socket and no application deadline for WebSocket upgrade.
- CLI `info` can evaluate a latest-release response with stale accepted-labeler policy and broadly hides non-404 failures as no releases.
- Notification retry needs durable pre-claim intent, an exclusive send lease, and repeated deletion/supersession checks.
- Aggregator build, typecheck, and its 313 workerd/D1 tests are absent from required CI; add an upgrade test starting from main's actual schema.

Verified at `7074e810`: root build/typecheck, explicit aggregator build/typecheck, lint with 0 diagnostics, 313 aggregator tests, and 1,008 labeler/console/calibration tests pass. Those suites do not exercise the failing upgrade, concurrency, replay, and publication sequences described here. Deployment testing remains a separate gate.

~ Sol 🤖

apps/labeler/src/notification-triggers.ts
```diff
@@ -0,0 +1,606 @@
...
+		{ type: "operator", id: input.actionId },
+		target,
+		overrideRetractNoticeContent(deps, input),
+	);
+}
+
+/** Emergency takedown or its retraction (console `emergency/takedown[-retract]`).
+ * The subject may be redacted; the aggregator read is unfiltered (blank
+ * accept-labelers), so contact resolution still works. */
+export async function notifyEmergencyTakedown(
```
ascorbic, line 341: When an emergency takedown has no resolvable contact, this records only a generic undeliverable notification row. The required operator signal that manual contact is needed is never emitted, so the most consequential delivery failure is easy to miss. Propagate the `no_contact` outcome into a dedicated operational event.

~ Sol 🤖

apps/aggregator/migrations/0001_init.sql
ascorbic: Confirmed and fixed in #2111. `0001` is restored byte-for-byte to main's version, the rename lands as forward migration `0004`, and a new upgrade test provisions a DB from a frozen copy of main's shipped migration set before applying the live set — it reproduces the exact `no such table: labelers` failure pre-fix. Fresh installs and main upgrades converge on the same schema.

~ 🤖 Claude Fable 5

packages/registry-moderation/src/index.ts
ascorbic: Confirmed and fixed in #2112. The three image values join `AUTOMATED_BLOCKS` (and `content-warning` joins `WARNINGS` — same un-synced batch), and a parity test now asserts the enforcement vocabulary against the executable policy fixture, including that every `officialEffect: "block"` label is enforced at release or package/publisher scope. Third dual-source miss; the test is the structural fix.

~ 🤖 Claude Fable 5

apps/aggregator/src/routes/xrpc/views.ts
ascorbic: Confirmed and fixed in #2112. `capLabels` now partitions by enforcement priority (hard blocks → assessment-state → informational) before slicing, so a truncated view can never drop a block while keeping informational labels; under-cap responses are byte-unchanged. Hydration only returns `neg = 0` rows at this boundary, so no negation-pairing hazard. The >64-hard-blocks pathological case keeps all blocks and logs.

~ 🤖 Claude Fable 5

apps/aggregator/src/routes/xrpc/label-enforcement.ts
ascorbic: Confirmed and fixed in #2112. The precise trigger is a post-override re-assessment: the console override negates existing blocks (those rows already pass the `neg = 0` filter), but a re-issued automated block was excluded by SQL while the evaluator suppresses it under the §10 pass+overridden pair. The SQL now mirrors `evaluateReleaseModerationCore` exactly — same-src, exact-CID pair suppresses automated blocks only; `security-yanked`/`!takedown` and the package/publisher cascade are never suppressed. Adversarially verified divergence-free (CID exactness, `label_state` latest-row semantics, bindings order).

~ 🤖 Claude Fable 5

packages/registry-client/src/discovery/index.ts
ascorbic: Confirmed and fixed in #2112. The wrapper now checks `!== undefined`, so `acceptLabelers: ""` sends an explicitly empty header (aggregator-side confirmed: absent → trusted defaults, empty → parsed as the documented opt-out) while an omitted option still sends no header. Changeset on `@emdash-cms/registry-client`.

~ 🤖 Claude Fable 5

apps/labeler/src/access-auth.ts
ascorbic: Confirmed and fixed in #2113. The Access application token has no top-level `groups` claim — IdP group membership arrives under the verified `custom` object (the token claim set is aud/email/exp/iat/nbf/iss/type/identity_nonce/sub/country/custom). Role mapping now reads a strictly-guarded `groups` array from `custom`; service-token path unchanged. One operator footgun to document: the Access group claim must be named exactly `groups`, else it fails closed to zero roles — adding that to the runbook.

~ 🤖 Claude Fable 5

apps/labeler/src/pds-verify.ts
ascorbic: Confirmed and fixed in #2113. The PDS fetch now enforces HTTPS-only and validates every hop (initial + each manual redirect, max 3) through `cloudflareDohResolver` + private/reserved-address rejection, fails closed if the resolver is absent, and bounds the body read with an absolute wall-clock deadline. A permanent `PDS_HOST_BLOCKED` dead-letter reason is threaded through; 404→RECORD_NOT_FOUND and 5xx→transient classification preserved (so a blocked host can never be mistaken for record-absence). Two sibling publisher-controlled fetches remain unguarded and are tracked for a follow-up: the `did:web` DID-document resolution, and the aggregator’s own `pds-verify.ts`.

~ 🤖 Claude Fable 5

apps/labeler/src/assessment-orchestrator.ts
ascorbic: Confirmed and fixed in #2115. The orchestrator now takes the same `createLabelPublisher(env)` the console path uses, commits finalization labels `publication_pending=1`, best-effort broadcasts each to the subscription DO post-commit, and a new `sweepPendingPublications` in the reconciliation cron (5-min threshold, per-row isolated) is the durable backstop. The rotation-drain contract is preserved — the sweep is exactly what unblocks a stuck rotation.

~ 🤖 Claude Fable 5

apps/labeler/src/subscribe-labels.ts
ascorbic: Confirmed and fixed in #2115. The WS replay reader now reuses `queryLabels`' `resignStaleLabels` — re-signing retired-key rows under the active key, persisting the result (no per-connection repeat), preserving `sequence`/ordering, and throwing on a signing pause like `queryLabels`' 503. The cursor is untouched on throw (lastSent only advances after a successful send), so reconnect is gap-free.

~ 🤖 Claude Fable 5

apps/labeler/src/assessment-orchestrator.ts
ascorbic: Confirmed and fixed in #2115. The finalization CAS `UPDATE` now carries the identical signing predicate the issuance statements use, so within the single atomic `db.batch` a mid-batch signing pause no-ops the CAS and the labels together — the run stays `running` and the Workflow retry re-finalizes after resume. (One theoretical bootstrap-only reverse-split residual is documented in the PR; it's self-healing and unreachable in practice.)

~ 🤖 Claude Fable 5

apps/labeler/src/console-mutation-api.ts
ascorbic: Confirmed and fixed in #2115. Both the `runLabelMutation` and `runEmergencyAction` replay branches now redrive `afterCommit`, and the #4 reconciliation sweep is the durable backstop — console-committed labels carry `publication_pending=1` and match the sweep filter, so a dropped notify no longer strands the row (and no longer permanently blocks rotation activation).

~ 🤖 Claude Fable 5

apps/labeler/src/notification-triggers.ts
ascorbic: Confirmed and fixed in #2116. The double-opt-in bypass is now gated on: (a) the claim's authenticated issuer (`issuer_did` = the PDS-verified repo owner, not a self-asserted field) being in a configured trusted-issuer set, (b) unexpired, (c) the claim's bound displayName still matching the publisher's current profile. The trusted set defaults EMPTY, so production falls back to double opt-in everywhere until an operator configures a trust source. Documented residual (confined to operator-trusted issuers): a trusted issuer vouches for identity, not address ownership — the address still comes from the publisher profile, which is the operator's explicit call when trusting an issuer.

~ 🤖 Claude Fable 5

apps/labeler/src/notification-triggers.ts
ascorbic: Confirmed and fixed in #2116. `contactTargetFromUri` now parses the parent package slug (before the first `:`) from a release rkey before calling `getPackage`, so a release notice resolves the package's security[]/authors[] contacts instead of falling through to publisher metadata. Package subjects and malformed rkeys degrade safely to the same-DID publisher tier.

~ 🤖 Claude Fable 5

apps/labeler/src/notification-endpoints.ts
ascorbic: Confirmed and fixed in #2116. The one-click POST handler now reads `c` (and the confirm token) body-first with a query fallback, so an RFC 8058 one-click request (token in the header URL query, `List-Unsubscribe=One-Click` in the body) actually suppresses the recipient. GET is unchanged and suppression is still authenticated by the peppered recipient hash.

~ 🤖 Claude Fable 5

apps/labeler/src/notification-triggers.ts
ascorbic: Confirmed and fixed in #2116. A takedown issuance (not retract) that resolves no contact now emits a dedicated `takedown-no-contact` operational event (severity high) plus its deployment-alert outbox row, deduped on replay — so the most consequential delivery failure surfaces to operators instead of only a generic undeliverable row.

~ 🤖 Claude Fable 5

apps/aggregator/src/routes/xrpc/router.ts
ascorbic: Confirmed and fixed in #2117. Unexpected (non-XRPC) errors — e.g. a D1 policy-resolution failure — now convert to a generic 500 via `internalErrorResponse()` through the same CORS + `private, no-store` wrapper as every other response; the internal detail is logged, the client gets the router's opaque envelope (no error.message/SQL/stack). XRPCError handling unchanged. (The same PR also hardens the aggregator's own pds-verify.ts against the SSRF twin of the labeler's.)

~ 🤖 Claude Fable 5

ascorbic:
Superseded

PR #1938, opened 2026-07-11
feat(registry): add labeler key rotation

ascorbic:
## What does this PR do?

Adds the persistence and service layer for routine signing-key rotation to the registry labeler. Issuance pauses across the DID update boundary, activation uses compare-and-swap state, historical labels are lazily re-signed without changing their creation timestamps, and rotation status and alerts are available to the later authenticated operator surface.

It also resolves the v1 identity strategy: discovery uses the base `#atproto_labeler` DID service plus the EmDash policy document, without an `app.bsky.labeler.service/self` declaration. Standard moderation reports return an explicit `NotSupported` response and are never persisted.

The production signing runtime and HTTP wiring remain part of W3.1, which provisions the DID, Secrets Store key binding, startup key validation, and policy endpoint. Subscription backfill may retain old-key signatures as explicitly permitted by the AT Protocol key-rotation guidance; labels issued after activation use the active key.

Part of #1909 and RFC #694.

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

Admin i18n is not applicable because this changes only the labeler Worker and design documents. A changeset is not applicable because `@emdash-cms/labeler` is private. This is an implementation slice of the approved registry labeler design tracked by RFC #694 and umbrella PR #1909.

- [x] This PR includes AI-generated code — model/tool: OpenCode with GPT-5.6-sol

## Screenshots / test output

- `pnpm build`
- `pnpm typecheck`
- `pnpm lint:json` reports 0 diagnostics
- `pnpm --filter @emdash-cms/labeler test` passes 23 workerd/D1 tests
- `git diff --check`


ascorbic:
Thanks. I checked these against the workstream boundaries and updated the PR description to make the scope explicit.

- Public query re-signing: confirmed intentionally deferred to W3.1. This slice provides and verifies the `VersionedLabelSigner` path, but W3.1 owns the production DID resolution, Secrets Store key binding, startup key validation, and Worker wiring. Calling the route without that runtime fails closed with 503 after rotation rather than serving a signature that cannot be validated.
- Subscription replay: no change needed. Labels issued after activation use the active key. Only retained backfill can carry old-key signatures, and the AT Protocol Labels specification explicitly says older labels in the event-stream backfill period may have invalid signatures during rotation and that this is acceptable.
- Status and alerts: the typed service functions are implemented here; exposing them over HTTP requires the authenticated operator surface tracked later in the umbrella workstream. Adding an unauthenticated endpoint here would be incorrect.

The approved design is tracked by RFC #694 and umbrella PR #1909. The labeler app is private, so the changeset warning is expected.

PR #1962, opened 2026-07-11
fix(loader): seek taxonomy-filtered listings via a denormalized pivot

### LINKED ISSUE #1834: Taxonomy-filtered collection queries do a full table scan (75k+ D1 row reads) on selective terms
# Taxonomy-filtered collection queries do a full table scan (75k+ D1 row reads) on selective terms

## Summary

The loader's SQL for a taxonomy-filtered collection listing (`collection?taxonomy=…` style query, ordered + paginated) is structured so that SQLite cannot use the selective taxonomy filter to drive the scan. Instead it walks the **entire** collection table in `ORDER BY` order, running the taxonomy filter as a correlated `EXISTS` per row. When the chosen term matches few entries, the `LIMIT` never fills, so the query scans the whole table.

On a collection of ~26k entries this produced **~75,000 D1 rows read** for a query returning a **single** row (≈295 ms). On D1, rows-read is the billed/throttled unit, so a routine category page can read the whole table.

Two independent causes, both fixable in the framework:

1. **Missing composite index** on the content↔taxonomy join table.
2. **Query shape** — the collection table is the scan driver instead of the taxonomy join table.

## Environment

- Single-locale site (e.g. `defaultLocale: "de"`), so the `locale` predicate is not selective — every published entry shares the one locale.
- A collection with a taxonomy that has many terms, each mapping to few entries (e.g. a fine-grained category/tag).
- Backend: Cloudflare D1 (SQLite).

## The generated query (shape)

For a listing filtered to a taxonomy term, ordered by `published_at DESC` with a `LIMIT`, the loader emits roughly:

```sql
SELECT *,
  (/* _emdash_terms correlated subquery */),
  (/* _emdash_bylines correlated subquery */)
FROM "ec_<collection>"
WHERE deleted_at IS NULL
  AND (status = 'published' OR (status = 'scheduled' AND scheduled_at <= <now>))
  AND locale = ?
  AND EXISTS (
    SELECT 1 FROM content_taxonomies ct
    WHERE ct.collection = ?
      AND ct.entry_id = "ec_<collection>".id
      AND ct.taxonomy_id IN (
        WITH RECURSIVE sub(grp) AS (
          SELECT COALESCE(translation_group, id) FROM taxonomies
          WHERE name = ? AND slug IN (?)
          UNION
          SELECT COALESCE(c.translation_group, c.id)
          FROM taxonomies c JOIN sub ON c.parent_id = sub.grp
        )
        SELECT grp FROM sub
      )
  )
ORDER BY published_at DESC, id DESC
LIMIT ?;
```

## Root cause

`EXPLAIN QUERY PLAN` for the above:

```
SEARCH ec_<collection> USING INDEX idx_ec_<collection>_deleted_published_id (deleted_at=?)
CORRELATED SCALAR SUBQUERY …   -- terms
LIST SUBQUERY … / MATERIALIZE sub   -- recursive taxonomy CTE (materialized once, fine)
CORRELATED SCALAR SUBQUERY …   -- bylines
```

The outer driver is `idx_ec_<collection>_deleted_published_id (deleted_at, published_at DESC, id DESC)`. The planner picks it to satisfy `ORDER BY published_at DESC, id DESC` without a sort — but it constrains **only** `deleted_at`. `locale` and `status` are not selective on a single-locale site, so this walks essentially the whole table in `published_at` order, evaluating the taxonomy `EXISTS` on every row.

Because the term is selective (few matching entries) and they are scattered through `published_at` order, the `LIMIT` is never satisfied early and the scan runs to the end of the table.

The `EXISTS` itself is indexed fine (covering seek on the `content_taxonomies` PK `(collection, entry_id, taxonomy_id)`), but it is invoked once per scanned row. Total reads ≈ (rows scanned) + (per-row EXISTS reads).

## Reproduction / measurements

Collection `ec_<collection>`: 26,224 rows (all one locale, 24,383 published). Join table `content_taxonomies`: 76,606 rows. Filtering by a term that maps to 1 entry:

| Variant | Plan | Rows read | Duration |
|---|---|---|---|
| Generated query (as above) | scan `ec_<collection>` by `(deleted_at, published_at)`, EXISTS per row | **75,454** | ~295 ms |
| Rewrite: `id IN (SELECT entry_id FROM content_taxonomies …)` | scans join table by `collection` prefix (76k) | 102,210 | ~40 ms |
| Rewrite: drive from join table, join to collection, **no hint** | still scans join table by `collection` prefix | 75,988 | ~23 ms |
| Same rewrite **forced onto `(taxonomy_id)` index** | seek by `taxonomy_id`, join by id, temp b-tree sort | **10** | ~1 ms |

The last row is the target behaviour: seek the selective side first.

## Fix 1 — add a composite index on the join table

`content_taxonomies` currently has:

- PK `(collection, entry_id, taxonomy_id)` — supports "terms **of** an entry", not "entries **with** a term".
- `idx_content_taxonomies_term (taxonomy_id)` — the right idea, but the planner **won't choose it** when `collection = ?` is also present: it's non-covering for `collection`/`entry_id`, so the cost model prefers the PK's `(collection=?)` covering scan (whole-collection scan of the join table). This is why the un-hinted rewrite above still read ~76k.

Proposed index:

```sql
CREATE INDEX idx_content_taxonomies_term_lookup
  ON content_taxonomies (taxonomy_id, collection, entry_id);
```

This makes the seek by `taxonomy_id` covering for `collection` and `entry_id`, so the planner picks it without a hint. (Belongs in the framework's migration that creates `content_taxonomies` / its indexes.)

## Fix 2 — restructure the taxonomy-filtered query

Even with the index present, the current shape still invites the planner to scan the collection table in `ORDER BY` order because the taxonomy filter is a correlated `EXISTS`. The query for a taxonomy-filtered listing should **drive from `content_taxonomies` seeking by `taxonomy_id`** and join to the collection table by `entry_id`, e.g.:

```sql
SELECT r.*, /* terms + bylines subqueries */
FROM content_taxonomies ct
JOIN ec_<collection> r ON r.id = ct.entry_id
WHERE ct.collection = ?
  AND ct.taxonomy_id IN (/* recursive CTE */)
  AND r.deleted_at IS NULL
  AND (r.status = 'published' OR (r.status = 'scheduled' AND r.scheduled_at <= <now>))
  AND r.locale = ?
ORDER BY r.published_at DESC, r.id DESC
LIMIT ?;
```

With the composite index this seeks the small set of matching entries, joins by primary key, and sorts a tiny result set — the 10-rows-read plan above. The tradeoff (a temp b-tree sort instead of a pre-ordered index walk) is negligible because the candidate set is small, and it is vastly cheaper than scanning the whole collection whenever the term is selective.

## Impact

- Any single-locale (or otherwise low-`locale`-selectivity) site paginating a collection filtered by a selective taxonomy term reads the entire collection table per request on D1, where rows-read drives cost and throttling.
- Worst case scales with collection size × requests, not with result size.

## Suggested resolution

1. Add `content_taxonomies (taxonomy_id, collection, entry_id)` to the schema/migrations.
2. Change the taxonomy-filtered collection query to drive from `content_taxonomies` (seek by `taxonomy_id`) and join to the collection table, rather than scanning the collection with a correlated `EXISTS`.

---

Relevant discussion: https://github.com/emdash-cms/emdash/discussions/1851

MA2153:
## What does this PR do?

Taxonomy-filtered collection listings (`getEmDashCollection("posts", { where: { category }, orderBy, limit })`) used to apply the taxonomy filter as a correlated `EXISTS` on `SELECT * FROM ec_<collection> … ORDER BY … LIMIT ?`. The pivot table `content_taxonomies` carried only `(collection, entry_id, taxonomy_id)`, so the filter (`deleted_at`/`status`/`locale`) and the ordering (`published_at`/`created_at`) could only be evaluated on the `ec_*` row. On stats-blind SQLite/D1 a **selective** term never fills the `LIMIT`, so the query walks the entire collection — a measured **~75,000 D1 rows read for a page returning one row** (rows-read is the billed/throttled unit on D1).

This denormalizes the filter + sort columns onto `content_taxonomies`, mirrors `ec_*`'s composite sort indexes onto the pivot keyed by `(taxonomy_id, collection, deleted_at, [locale,] <sort> DESC, entry_id DESC)`, and restructures the taxonomy-filtered listing to **drive from the pivot**: seek the term, walk a sort-ordered pivot index, let `LIMIT` short-circuit, and touch `ec_*` only by primary key to hydrate the page.

**The pivot is advisory; `ec_*` is authoritative.** D1 has no transactions, so the write-path re-stamp is a separate statement from the `ec_*` mutation and can be transiently stale. The outer `ec_*` join therefore **re-checks** `deleted_at`/`status`/`locale` on the joined row — a stale pivot can only cause a page to briefly under-fill (self-healing on the next request); it can never leak a deleted or wrong-status/locale row onto a public listing.

Implementation:
- **Schema**: six nullable columns added to `content_taxonomies` (`status`, `scheduled_at`, `deleted_at`, `locale`, `published_at`, `created_at`). `updated_at` is deliberately **not** denormalized — it moves on every edit, so denormalizing it would force a pivot re-stamp on the common edit path; `updated_at`-sorted taxonomy listings take a bounded temp-sort path instead.
- **Migration 051** (forward-only): adds the columns (idempotent), backfills per-collection, then builds four composite sort indexes.
- **Read path** (`loader.ts`): a caller-agnostic pivot-drive builder (parameterized on the `deleted_at` predicate and status condition, so it can also serve an admin trash/all-statuses shape). Both dialects run the identical query shape.
- **Write path**: synchronous re-stamp at the existing data-layer choke points (`ContentRepository`, `TaxonomyRepository`) — no stray SQL.

Supersedes the closed #1852 (cached-count runtime planner — wrong layer). This is also the substrate the future taxonomy-subtree filter needs to perform well.

**Follow-up (out of scope):** `idx_content_taxonomies_term (taxonomy_id)` from migration 048 is now redundant — all four new indexes lead with `taxonomy_id` and serve a bare `taxonomy_id` seek as a prefix. Dropping it is a separate additive-discipline call (048 was a deliberate restore for #1701), flagged here so the write cost isn't paid twice.

Closes #1834

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [x] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable). — n/a: no admin UI strings; this is a data-layer/loader change.
- [ ] New features link to an approved Discussion — n/a: this is a performance/bug fix with an approved design spec, not a feature.

- [x] This PR includes AI-generated code — model/tool: Claude Opus 4.8 (Claude Code). Code review pass also run.

## Screenshots / test output

- Full `emdash` core suite: **4675 passed, 3 skipped**. Lint: **0 diagnostics**. Typecheck clean. Formatted.
- Tests added: `tests/unit/loader-taxonomy-pivot.test.ts` (`describeEachDialect`: parity across selective/broad/OR-within-taxonomy/multi-term-AND/sort orders/locale/status; stale-pivot correctness under non-atomic writes; re-stamp sync; keyset pagination incl. the GROUP-BY/HAVING multi-group path; byline-in-pivot; admin trash/all-statuses builder shapes) and `tests/integration/loader-taxonomy-pivot-plan.test.ts` (SQLite `EXPLAIN QUERY PLAN`: asserts the correct pivot index per case, no `ec_*` full scan, and no temp B-tree for the indexed early-`LIMIT`).
- **CI note:** `EMDASH_TEST_PG` was not set locally, so only the SQLite dialect executed. The raw SQL (row-value `UPDATE … SET (cols) = (SELECT …)`, `MAX()`/GROUP BY/HAVING, CTE) was reasoned through as valid Postgres but not exercised — please confirm the `describeEachDialect` suite runs against real Postgres in CI.
- **Query-counts snapshot** — `+1` query on the taxonomy-filtered routes (`/category/*`, `/tag/*`), on both cold and warm. This is the pivot-drive's up-front term resolution (`resolveTermGroups`): the seek needs literal `translation_group` values, because on a stats-blind planner an inlined `taxonomy_id IN (SELECT …)` is treated as multi-valued and forces a temp-sort over the whole term — defeating the early-`LIMIT` seek (confirmed via `EXPLAIN QUERY PLAN`). So the values must be resolved in a separate statement. It's cheap in rows-read (rides `idx_taxonomies_name_locale`, touches one taxonomy's terms), but currently uncached, so it lands on cold *and* warm. **Caching that resolution to reclaim the warm cost is deferred** — the obvious reuse (`getTaxonomyTerms`) regresses localized filtering (#1480), so the right shape is an open call tracked in #1966. Net rows-read still collapses by ~4 orders of magnitude on a selective term page, which is the point of this PR.


ascorbic (approved):
Looks solid. Thanks

PR #1972, opened 2026-07-12
feat(registry): enforce moderation labels in clients

ascorbic:
## What does this PR do?

Implements W5 client eligibility enforcement (minus the admin UI, which is W5.5's own PR) — the layer that decides whether an EmDash site installs a plugin. Three commits:

- **Hydrated moderation evaluation** (`@emdash-cms/registry-moderation`): `evaluateHydratedReleaseModeration` evaluates aggregator-hydrated labels that carry no signature, sharing one evaluation body with the branded `verifyLabel` path so the two cannot drift. The branded entrance and its runtime check are untouched; nothing can mint the verified brand without cryptographic verification. Also corrects a latent invalid CID in the ratified gate-0 moderation corpus (present since ratification, inert until this path added structural CID validation — recorded in the gate-0 amendments log in this diff).
- **Registry-client adapter** (`@emdash-cms/registry-client`): `evaluateReleaseViews` (subject context from validated views, tolerant of structurally invalid aggregator labels), `resolveAcceptedPolicy` (response `atproto-content-labelers` header > configured `acceptLabelers` > empty; response-header parse failure warns and falls through, configured parse failure is a loud error), and `onResponseMeta` on `DiscoveryClient` for header access. Docs cut over to the canonical vocabulary.
- **Core + CLI enforcement**: install and update gate on the shared evaluation — package and publisher cascades, CID-bound labels, negation, expiry — immediately before artifact download (fetch-tripwire tested on both paths), replacing the three raw `security:yanked` string comparisons (zero occurrences remain in core, registry-client, or plugin-cli). Update now fetches the package view (it previously saw release labels only, missing every cascade) with the same aggregator-identity cross-check as install. The artifact proxy refuses media for blocked releases. Update-check reports additive per-plugin `moderation` state with no extra network calls. The CLI renders evaluated eligibility instead of raw label dumps. Error codes unified: `RELEASE_YANKED` when `security-yanked` is among the blocking labels, else `RELEASE_BLOCKED`; malformed `acceptLabelers` config is a loud `REGISTRY_POLICY_INVALID`.

Adversarial review (opus) finding fixed pre-merge: the gate originally required `eligibility === "blocked"`, but the evaluator ranks pending/error above automated blocks — so `malware` plus a co-present `assessment-pending` (a realistic state once a newer assessment run opens against a flagged release) re-ranked eligibility and **failed the gate open**. All three consumers now key solely on `blockingLabels`/`redacted`, with a regression test for the co-present case.

Decisions needing maintainer ratification (deviations from the plan, reasoned in-thread):

1. **Hydrated-trust evaluation instead of per-request signature verification.** A compromised aggregator attacks by *omitting* labels, which verifying the labels it chose to show cannot detect; the integrity roots (signed release record, artifact checksum) are already verified in core. Signature verification is reserved for the future positive-assessment gate, where fabricating a pass is the actual threat — the branded path exists and is required for it.
2. **No enforcement flag.** The plan's "disabled production flag" applies to the positive-assessment gate (not shipped here). Blocking-label enforcement is behavior-preserving-or-stronger versus the string checks it replaces, and `missing-assessment-pass` explicitly does not block — zero production change on merge (verified: a clean release with an accepted policy installs; no labels flow in production yet regardless).
3. **Gate-0 corpus amendment**: the ratified fixture's `oldPackageCid` was not a decodable CID (hard parse failure, traced to the original ratification commit); corrected to a verified round-tripping CID with no change to any case's expected outcome.

Implementation PRs target `feat/plugin-registry-labelling-service` per the umbrella PR. Related to #1909 and RFC #694.

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [x] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

Admin i18n is n/a — no admin changes in this PR (W5.5 handles the UI). Changesets included for all four published packages that changed (`@emdash-cms/registry-moderation`, `@emdash-cms/registry-client`, `emdash`, `@emdash-cms/plugin-cli`).

## Screenshots / test output

- Core: full unit suite 4618 passed / 3 skipped, including 16 new moderation-gate tests (install/update/update-check, cascades, CID-stale, warning/pending/error non-blocking, the co-present fail-open regression, gate-before-fetch tripwires) and 4 new artifact-proxy tests
- registry-moderation: 150 pass (ratified corpus now runs through BOTH the branded and hydrated entry points); registry-client: 85 pass; plugin-cli: 399 pass (6 new display tests)
- Typecheck clean everywhere; `pnpm lint:quick` 0 diagnostics; type-aware `lint:json` 0 findings repo-wide
- CLI rendering (verbatim): `Moderation: blocked (malware) by did:plc:labeler-a` on `info`; `[blocked]` marker + `blocked: !takedown` line on `search`; clean packages render identically to before

---

### Try this PR

**[Open a fresh playground →](https://feat-labeler-client-enforcement-emdash-playground.emdash-cms.workers.dev)**

<sub>Tracks `feat/labeler-client-enforcement`. Updated automatically when the playground redeploys.</sub>

### DIFF
--- a/.changeset/quiet-labels-moderate.md
+++ b/.changeset/quiet-labels-moderate.md
@@ -1,0 +1,5 @@
+---
+"@emdash-cms/registry-client": patch
+---
+
+Adds release moderation evaluation over hydrated aggregator labels and exposes the atproto-content-labelers response header.
--- a/.changeset/tame-labels-hydrated.md
+++ b/.changeset/tame-labels-hydrated.md
@@ -1,0 +1,5 @@
+---
+"@emdash-cms/registry-moderation": patch
+---
+
+Adds `evaluateHydratedReleaseModeration` for evaluating aggregator-hydrated labels that carry no signature.
--- a/.changeset/tame-labels-hydrated.md
+++ b/.changeset/tame-labels-hydrated.md
@@ -2,4 +2,4 @@
 "@emdash-cms/registry-moderation": patch
 ---
 
-Adds `evaluateHydratedReleaseModeration` for evaluating aggregator-hydrated labels that carry no signature.
+Adds `evaluateHydratedReleaseModeration` for evaluating aggregator-hydrated labels that carry no signature, and `isModerationBlocking`, the canonical blocking predicate for enforcement consumers.
--- a/.changeset/tidy-labels-enforce.md
+++ b/.changeset/tidy-labels-enforce.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Registry install and update now evaluate the full moderation label set (package and publisher cascades, CID-bound labels, negation and expiry) instead of a single yanked-label string match. The update-check and artifact proxy endpoints also gate on this evaluation.
--- a/.changeset/tidy-plugins-eligible.md
+++ b/.changeset/tidy-plugins-eligible.md
@@ -1,0 +1,5 @@
+---
+"@emdash-cms/plugin-cli": patch
+---
+
+Search and info now show moderation eligibility for registry packages instead of raw label values.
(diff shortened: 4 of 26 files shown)


ascorbic:
Review findings addressed in 103173a6:

**Collision bypass (needs fixing) — taken, with a different mechanism than the suggestion.** Surfacing colliding values into `blockingLabels` would break that field's ratified definition ("only active, applicable" labels — gate-0 result-fields contract). Instead the shared package now exports `isModerationBlocking` — the single canonical blocking predicate (applicable blocking label, redact-flagged takedown, or `label-state-collision`; never keyed on `eligibility`) — and all three consumers delegate to it. That also fixes the root cause of this whole bug class: the predicate was restated in three places, and every restatement was a fresh chance to get it wrong. Regression tests at both levels (predicate unit tests incl. the same-cts collision, and an end-to-end install test proving `RELEASE_BLOCKED` before any artifact fetch).

**Double parse (suggestion) — declined.** `evaluateReleaseViews` needs per-label tolerance (skip + warn) while the primitive stays strict; the re-parse preserves that split at a cost of microseconds on an operation that performs network fetches and archive extraction. Strictness-in-the-primitive is the deliberate design.

**Stub docstring (suggestion) — taken**: it now says only the package-scope cascade is excluded by the empty uri/cid; publisher-scope labels riding on the release response still evaluate through the real publisherDid.

PR #1976, opened 2026-07-12
feat(registry): labeler assessment persistence and automated issuance

ascorbic:
## What does this PR do?

Adds the labeler service's assessment persistence layer and automated label issuance — plan items W2.2–W2.5 plus the issuance-boundary extension the assessment orchestrator (a follow-up PR) needs. This is the storage and state-machine foundation; nothing in production issues an automated label yet (no code path drives a run to completion until discovery/orchestration lands).

- **Storage model** (migration 0003): `subjects`, `assessments` (state CHECK-constrained to the ten-state lifecycle vocabulary), `current_assessments`, `findings`, `evidence_objects`, plus the `dead_letters` and `ingest_state` tables the discovery PR will write to. Timestamp columns that get compared carry epoch-millisecond siblings so ordering never depends on RFC 3339 string comparison. `issuance_actions` gains a nullable `assessment_id` (rather than a parallel `actions` table) linking a label to the assessment that produced it.
- **Lifecycle + repositories**: CAS-guarded state transitions (the `signing-rotation` idiom — a lost race is a typed conflict, never a silent no-op), a deterministic `runKey` per spec §9.2 so a redelivered trigger converges on one run, and `buildFinalizationStatements` whose pointer update re-checks the state CAS *inside the same batch*, letting the orchestrator compose assessment completion and label issuance into one atomic transaction. The pointer update also carries a recency guard so a late-committing older run cannot regress a newer run's current pointer.
- **Automated issuance**: a new `automated-assessment` action whose proposals are validated from the ratified policy fixture (release-only subject, mandatory CID, blocking values only from critical automated-block finding categories, manual-only values unreachable). One shared `buildIssuanceStatements` backs both manual and automated paths — manual issuance behavior is byte-identical and its existing suite passes unmodified. Automation is additionally forbidden from negating a label whose active stream event was issued manually (§10), enforced at the issuance chokepoint.
- **Evidence boundary**: public/private finding types where the public serializer cannot carry private detail (compile-time + runtime tested).

Adversarial review (opus) findings all addressed pre-open: the §10 automated-negation guard (was reachable, now blocked + tested), the pointer-recency regression (guarded + tested with an older-commits-last ordering), unconditional CID on the automated path per §20.2, and a strengthened atomicity test that fails on a runtime constraint violation rather than a missing table.

Deferred by design (noted so reviewers don't flag them): the `notifications` table is W10 (nothing writes to it yet); `assessment_id` lives on `issuance_actions` and is recovered by join rather than duplicated onto `issued_labels` per the spec DDL (a deliberate normalization).

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [x] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

Admin i18n and changeset are n/a — `apps/labeler` is a private, unpublished Worker app (no UI, not on npm).

## Screenshots / test output

- `pnpm --filter @emdash-cms/labeler test`: 73 pass (29 pre-existing manual-issuance tests unmodified; 44 new across migration, lifecycle/CAS-race, finalization atomicity, supersession/recency, automated-issuance legality, evidence boundary, policy parse)
- Typecheck clean; `pnpm lint:quick` 0 diagnostics; type-aware lint clean on touched files

---

### Try this PR

**[Open a fresh playground →](https://feat-labeler-assessment-persistence-emdash-playground.emdash-cms.workers.dev)**

<sub>Tracks `feat/labeler-assessment-persistence`. Updated automatically when the playground redeploys.</sub>


ascorbic:
All three findings addressed in 8b03d379:

**Racy §10 guard (needs fixing) — fixed.** The negation-provenance check now runs as a condition inside the `issuance_actions` INSERT's WHERE, gated on an `isAutomatedNegation` flag, so the check and the write are one atomic statement — no manual label can commit in the window. I placed it on the *action* insert rather than the label insert: when the guard fires, no action row is written, so the label insert's `SELECT FROM issuance_actions` finds nothing and there's no orphan action (guarding only the label insert would leave one). The pre-batch SELECT stays as a fast, friendly-error fail. New test builds the negation statements while the stream is empty (pre-check passes), then commits a manual label before executing the batch — the exact TOCTOU window — and asserts the negation didn't land and no orphan action exists.

**Fragile release-rule lookup (needs fixing) — fixed.** Switched from `.find()` to `.some()` over all release rules, and the policy parser now rejects duplicate `subject` entries outright (test added) so the document is unambiguous at parse time too.

**Duplicated ASSESSMENT_ID (suggestion) — taken.** Exported from `assessment-lifecycle.ts`; both `service.ts` and `assessment-store.ts` import it.

75 tests pass; typecheck and lint clean.

ascorbic:
Fixed in 704b7332. `postCommit`'s no-row branch now re-runs the §10 guard first when the label was an automated negation, so a guard-suppressed insert surfaces the policy violation (`cannot negate the manually-issued label`) instead of misdiagnosing it as a signing-state failure and firing spurious alerts. If no manual label is present it's a no-op and falls through to the genuine signing diagnosis. The race test now also asserts `postCommit()` rejects with the policy error.

ascorbic:
Added the covering index in 9684daf7 — `idx_issued_labels_stream ON issued_labels(src, uri, val, sequence)` in the 0003 migration (unreleased, so it lands with the table's first deploy). Keeps the §10 guard's per-stream head lookup cheap as label streams grow.

ascorbic:
I don't think this one holds — the release-record and CID checks already run *before* the negation early-return. Current `validateAutomatedProposal` order (service.ts:571-590):

1. line 572-573: reject non-release URIs
2. line 574-575: reject missing CID (unconditional)
3. line 576-580: reject vals whose release rule lacks automated mode
4. line 581: `if (proposal.neg === true) return;` — skips **only** the finding-category/severity checks below it

So a negation already gets the release-record and mandatory-CID checks; only the finding requirements (which don't apply to negations) are skipped — which is the intended behavior your suggested fix describes. Added a test in 85531f16 that proves it: a negation with a publisher-DID URI rejects with "must target a release record", and a negation missing a CID rejects with "must include a release CID". If you were reading an earlier revision where the CID check sat after the early-return, this is already resolved.

PR #1977, opened 2026-07-12
feat(admin): registry moderation UI and blocked-install gating

ascorbic:
## What does this PR do?

Adds the admin registry moderation UI — plan items W5.5 and W5.6. This is the last consumer in the aggregator→client moderation arc; the server (core's install/update handlers, merged in #1972) remains the enforcement boundary, and this layer is defense-in-depth plus UX.

- **Detail page**: each release is evaluated client-side (`evaluateReleaseViews`) against the accepted-labeler policy that produced its data. A blocked release stays visible and selectable in the version picker but shows an explanation panel and a disabled Install button; a warning-labelled release surfaces its warnings in the install consent dialog; the version options are annotated with their moderation state. The legacy `security:yanked` (colon) string filter that silently hid yanked releases is removed — a `security-yanked` (hyphen) release is now visible-but-blocked.
- **Browse cards**: mark packages carrying a package- or publisher-scope block, evaluated with a package-only stub mirroring the CLI.
- **Plugin manager (W5.6)**: consumes the additive `moderation` field core's update-check now returns; a blocked update is disabled with a badge, a warned update keeps its button with a warning badge. Never keys off `eligibility` (core returns `blocked` for every clean plugin via missing-assessment-pass) — only `blockingLabels`/`warningLabels`.
- **Per-response policy**: each aggregator response's `atproto-content-labelers` header travels with its data via a per-call `DiscoveryClient` with a call-local closure, so a response's labels are only ever judged by the policy that produced them. Query keys that fetch labels now include the configured `acceptLabelers`. Package-scope and release-scope labels arrive on separate responses and are evaluated against the union of both responses' effective policies.
- **Localized labels**: a Lingui descriptor map over the 23 canonical label values, falling back to the raw value for anything unknown (never throws, never hides).
- **Structured errors**: `RELEASE_BLOCKED`/`RELEASE_YANKED` install/update responses are parsed into a typed error rendering a localized headline plus localized blocking-label names.

Adversarial review (opus) verdict was ship — no bug above LOW. The LOW (package labels evaluated against the releases response's policy) and both flagged test-coverage gaps (header-precedence path, package-scope block via the real pipeline) are fixed in the final commit: the two header-derived policies are now unioned, with tests for both.

### Verification note (please read)

RTL is verified statically: every layout class in the diff is logical (`ms`/`me`/`ps`/`pe`/`start`/`end`), zero physical-direction classes, confirmed by review. A **live Arabic browser pass of the blocked/warned states was not run** — those states require a real blocked release, which needs the labeler→aggregator label pipeline that isn't deployed yet. That live pass belongs to the staging-launch phase (the umbrella's consumer-shadow phase) when a blocked release can be produced end-to-end; flagging it as an open verification item rather than claiming a test I couldn't run. All new components reuse Kumo primitives and existing banner/badge patterns already RTL-tested in this admin.

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [x] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) — every new string via Lingui `t`/`Trans`/`msg`; no `messages.po` changes (catalog extraction runs on merge to `main`).

## Screenshots / test output

- `pnpm --filter @emdash-cms/admin test`: 89 files, 1088 pass (22 new moderation tests across detail, consent dialog, plugin manager, browse, and the registry-moderation lib; the security-yanked-now-visible regression and the eligibility-not-used property both asserted through the real evaluation pipeline)
- Typecheck clean; `pnpm lint:quick` 0 diagnostics; type-aware lint clean on touched files

---

### Try this PR

**[Open a fresh playground →](https://feat-labeler-admin-moderation-emdash-playground.emdash-cms.workers.dev)**

<sub>Tracks `feat/labeler-admin-moderation`. Updated automatically when the playground redeploys.</sub>

### DIFF
--- a/.changeset/lucky-labelers-admin-moderation.md
+++ b/.changeset/lucky-labelers-admin-moderation.md
@@ -1,0 +1,5 @@
+---
+"@emdash-cms/admin": patch
+---
+
+The registry browser now shows moderation state: blocked releases are visibly blocked with an explanation, warning-labelled releases ask for confirmation before install, and the plugin manager flags blocked or warned updates.
--- a/packages/admin/src/components/CapabilityConsentDialog.tsx
+++ b/packages/admin/src/components/CapabilityConsentDialog.tsx
@@ -15,6 +15,15 @@ import { describeCapability } from "../lib/api/marketplace.js";
 import { cn } from "../lib/utils.js";
 import { DialogError } from "./DialogError.js";
 
+/** A moderation label, pre-resolved to localized display text by the caller. */
+export interface ModerationLabelEntry {
+	value: string;
+	name: string;
+	description: string | null;
+	/** DID of the labeler that issued this label. */
+	issuerDid: string;
+}
+
 export interface CapabilityConsentDialogProps {
 	/** Dialog mode */
 	mode?: "install" | "update";
@@ -22,6 +31,13 @@ export interface CapabilityConsentDialogProps {
 	pluginName: string;
 	/** Capabilities the plugin requests */
 	capabilities: string[];
+	/**
+	 * Active moderation warning labels on the release being installed/updated.
+	 * Shown above the capabilities list whenever non-empty -- including when
+	 * `capabilities` is empty, in which case the dialog shows only this
+	 * section.
+	 */
+	moderationWarnings?: ModerationLabelEntry[];
 	/** Allowed network hosts (for network:fetch capability) */
 	allowedHosts?: string[];
 	/** New capabilities added in an update (highlighted differently) */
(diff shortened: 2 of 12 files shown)


ascorbic:
Fixed in 09104bb6. `formatHoldback` now routes its duration units through Lingui plurals (full words — minute/hour/day — since the abbreviated forms don't translate cleanly).

For the record: this function predates this PR (introduced in #1011) and isn't touched by the moderation diff, so it's strictly a pre-existing string. I fixed it anyway because it's in a file this PR already substantially edits and localization is a hard project rule — but flagging the provenance in case you'd rather it lived in its own change.

PR #1978, opened 2026-07-12
feat(registry): labeler discovery and assessment orchestration

ascorbic:
## What does this PR do?

Adds independent Jetstream discovery and assessment orchestration to the labeler — plan workstream W6, on top of the assessment persistence layer (#1976). It wires the production path up to `assessment-pending`; the full analysis pipeline ships with deterministic stub adapters, exercised only by tests, until W7/W8 provides real analyses.

- **Discovery**: a `LabelerDiscoveryDO` maintains an independent Jetstream subscription to release records (cursor persisted in `ingest_state`). The queue consumer verifies each event's exact URI+CID against the publisher's signed record (`fetchAndVerifyExactRecord`) before creating a subject and issuing a signed `assessment-pending` label. A forged or unverifiable event dead-letters and produces nothing; a CID mismatch (the PDS serving a newer version than the event named) is a permanent dead-letter; a transient PDS or DID-directory failure retries. Redelivery converges on one run and one pending label via the deterministic runKey and idempotency key.
- **Orchestration**: `AssessmentOrchestrator` drives `pending → running → stages → finalization`. Finalization composes the assessment CAS + current-pointer update with all label statements (outcome labels, `assessment-error` on transient exhaustion, negation of this run's own `assessment-pending`, and negation of prior automated labels this outcome no longer supports) into one atomic `db.batch`. It never negates a manually-issued label (`getNegatableAutomatedLabels` returns only automated-provenanced labels whose current stream head is automated).
- **Reconciliation**: a scheduled job flags runs stuck in a non-terminal state and subjects lacking any run (read-only; structured logs that W11.3 turns into alerts).

Adversarial review (opus) findings addressed: transient DID-resolution failures now retry instead of permanently dropping a release on a directory blip (#2); `getNegatableAutomatedLabels` computes the active stream head across all issuers so a manually-negated val is never a candidate (#4); the production-boundary test also catches bare side-effect imports (#6); and the finalization currency re-check was moved to immediately before the commit, shrinking the delete/cancel window from "spans every label's signing round-trip" to two adjacent D1 ops, with a mid-run-delete test.

**Two deferrals, called out for the maintainer** (both provably unreachable in production — the production-boundary test enforces that no production path constructs the orchestrator until W7/W8):

1. **Full closure of the finalization/delete race** requires the per-subject workflow lock the spec assigns to live wiring (§14.1), or an in-batch state guard on every issuance statement (which couples through the shared signer's `postCommit` diagnosis). The window is tightened and documented; the lock lands with the orchestrator's production wiring in W7/W8. The overstated "under the same lock" comment is corrected.
2. **A create-after-delete race** (a delete processed before a create's INSERT resurrects a tombstone) is a narrow eventual-consistency window, already partly mitigated by the create path's PDS verification, and needs a deletes-log to close cleanly. Documented as a W6.8/reconciliation follow-up.

If you'd rather either be fully closed before this merges, say so.

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [x] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] User-visible strings in the admin UI are wrapped for translation (if applicable)

Admin i18n and changeset are n/a — `apps/labeler` is a private, unpublished Worker app.

## Screenshots / test output

- `pnpm --filter @emdash-cms/labeler test`: 121 pass (75 from the persistence layer + 46 new: discovery consumer, record verification, orchestrator with stub stages, reconciliation, and the static production-boundary test)
- Typecheck clean; `pnpm lint:quick` 0 diagnostics
- The copied Jetstream/PDS primitives are byte-diff-clean against the aggregator originals

---

### Try this PR

**[Open a fresh playground →](https://feat-labeler-discovery-emdash-playground.emdash-cms.workers.dev)**

<sub>Tracks `feat/labeler-discovery`. Updated automatically when the playground redeploys.</sub>


ascorbic:
All four addressed in ee843c2a:

1. **Negation src scoping (needs fixing)** — `getNegatableAutomatedLabels` now filters `l.src = <this labeler DID>` (outer and inner MAX). A labeler only negates labels it issued. (In practice `issued_labels` only ever holds this labeler's own rows, so it was latent, but the scoping matches intent and is robust to any future change.)

2. **Mid-download error → retry (needs fixing)** — `fetchCar`'s catch re-wraps a post-headers stream error as a transient `PdsVerificationError(PDS_NETWORK_ERROR)`; our own `PdsVerificationError`s (e.g. RESPONSE_TOO_LARGE) pass through with their existing classification. The consumer now retries instead of dead-lettering `UNEXPECTED_ERROR`.

3. **Unverified deletes (needs fixing)** — the delete path now calls `confirmRecordAbsent` (verifies the record is genuinely gone at the PDS) before suppressing: a still-resolving record dead-letters as `DELETE_RECORD_PRESENT` and suppresses nothing, transient failures retry. Tests cover the confirmed-absent, forged-present (subject not tombstoned, run stays pending), and transient cases. Worth noting the practical exploitability was already low — a forged delete can't make a malicious release installable because the missing-positive-assessment gate stays fail-closed — but the create/delete verification asymmetry is now closed.

4. **Only first blocking label (needs fixing)** — `resolvePolicyOutcome` collects every distinct critical automated-block finding (matching the warning branch), with a two-blocker test.

124 tests pass; typecheck and lint clean.

ascorbic:
Fixed in 5afe8479. Extracted a single `classifyDiscoveryError` helper that both the delete and create paths route their catch through — transient failures retry, permanent verification failures dead-letter and ack, unexpected errors log + dead-letter + ack. The delete path no longer retries a permanent `INVALID_URI`/`DID_RESOLUTION_FAILED`, and the two paths can't diverge again since they share the classifier. Added a test asserting a permanent delete-verification failure dead-letters rather than retries.

ascorbic:
Both addressed in 817c85c2:

**Create-after-delete (needs fixing)** — took your suggested `ON CONFLICT DO UPDATE`. It's the right fix and it closes the deferral, not just papers over it: the create path only calls `createSubject` after PDS verification confirms the record exists, so clearing `deleted_at` on a verified re-observation is always correct — it reactivates a legitimately-republished subject and resolves the out-of-order delete-before-create race in one stroke. Test added (tombstone → verified re-observe → current again). This removes create-after-delete from the PR's deferral list.

**Same-instant currency tie (suggestion)** — `isSubjectCurrent` now breaks an equal `observed_at_epoch_ms` by `cid` (`observed_at_epoch_ms > ? OR (= ? AND cid > ?)`), so exactly one subject at a URI is current even when two are observed in the same millisecond. Test added with two same-instant CIDs.

That leaves only the finalization/delete-race full closure deferred to W7/W8 (the per-subject workflow lock per §14.1); the window is already tightened to two adjacent D1 ops.

127 tests pass; typecheck and lint clean.

ascorbic:
Both addressed in c7eec50f:

**Delete doesn't negate pending (needs fixing)** — the delete path now negates the `assessment-pending` label each cancelled run issued, so `queryLabels` stops returning a positive pending for a deleted release. Each negation uses the run's own assessment id and `automatedIdempotencyKey(runKey, "assessment-pending", true)`, so a redelivered delete converges and an already-finalized run's pending no-ops. `applyDiscoveryDelete` returns the runs that had reached pending; the consumer (which holds the signer) issues the negations. Test asserts the pending label's head is `neg=1` after delete.

**finalize keep-set for error (suggestion)** — `keep` now includes `assessment-error` when the outcome is error, so a second error run for the same subject can't negate its own freshly-issued error label.

127 tests pass; typecheck and lint clean.

ascorbic:
Both addressed in c68729c8:

**Issuance-during-rotation classification (needs fixing)** — the signing-pause/stale/refresh/state-changed conditions in `service.ts` now throw a typed `LabelIssuanceUnavailableError` (identical messages, so existing message-based assertions are unaffected), and the discovery consumer's classifier retries it. So a create or delete event arriving mid key-rotation retries until the rotation completes instead of dead-lettering and permanently losing the release's pending label. (The signer-integrity-mismatch throw stays a plain Error → unexpected → dead-letter, since that's a real integrity failure, not transient.) New test: a verified create with signing paused retries, no dead-letter.

**Missing error-preservation test (needs fixing)** — added a two-consecutive-error-runs test: the second error run for the same subject leaves `assessment-error` active at the stream head rather than self-negating. It fails against the pre-fix code (keep-set without assessment-error) and passes now.

129 tests pass; typecheck and lint clean.

ascorbic:
Both suggestions:

**Delete-path paused-issuance test** — added in 3ceeeb9c, mirroring the create-path one: a confirmed delete with signing paused retries (no ack, no dead-letter), since `negatePendingForDeletedRuns` throws the same `LabelIssuanceUnavailableError`.

**Mid-batch signing-state flip during finalization** — this is the same test-only atomicity family as the finalization/delete-race already documented in `finalize()`'s comment and the PR's deferral list: the label statements are signing-guarded and no-op on a flip, but the state CAS isn't, so a flip mid-batch could commit the terminal state without labels. Full closure lands with the per-subject workflow lock the spec assigns to production wiring (§14.1, W7/W8), which also closes the delete-race window; the orchestrator is unreachable in production until then (production-boundary test). I've noted the signing-flip case alongside the existing deferral rather than partially guarding it now, so both are closed together with the lock. Shout if you'd rather I guard the state CAS on signing_state here.

130 tests pass; typecheck and lint clean.

ascorbic:
Confirmed real, and thank you for the thorough trace — all three are genuine, and they're one family: `finalize()` isn't serialised against a concurrent state change (stale run's dangling pending, unchecked finalization CAS, uncaught transition conflict).

Decision (maintainer): defer all three to W7/W8, where the orchestrator is wired to production behind the per-subject workflow lock the spec assigns to finalization (§14.1). That lock serialises finalization per subject and closes the whole family — including the currency-window and signing-flip cases already noted. Fixing them piecemeal here (partial CAS checks, in-code negations) would be superseded by the lock and risks a false sense of closure. The production-boundary test proves nothing wires this method live until that work, so none is reachable today. Documented together in `finalize()`'s comment (70d5cc96) and the PR description so the lock work closes them as a set.

Everything production-reachable in this PR — discovery, PDS verification, delete verification, issuance retry-under-rotation, negation provenance — is fixed and tested. Merging on that basis.

PR #1994, opened 2026-07-12
feat(labeler): code/metadata AI adapter (W8.2)

ascorbic:
## What does this PR do?

Adds the labeler's **code/metadata AI adapter** — plan item W8.2. `analyzeCode(input, deps)` analyzes a plugin's source files and metadata with a code model (Kimi `@cf/moonshotai/kimi-k2.7-code`) via the **Workers AI binding** (`env.AI`, injected — not AI Gateway) and returns validated `NormalizedFinding[]` (source `"model"`), feeding the W8.5 policy resolver.

- **Pure module.** The `AI` binding is injected via `deps`; the module does no DB writes and imports nothing from `cloudflare:workers`, so it's fully testable against a fake binding and decoupled from the orchestrator's still-abstract `StageContext`. Wiring it into the `codeAi` stage — threading the acquired bundle and recording evidence objects — is deferred to the W7.3-consumer work once a real bundle-input producer exists.
- **Strict structured output.** `response_format: json_schema` constrains each finding's `category` to exactly the policy's allowed set (automated-block ∪ warning) and `severity`/`confidence` to their ranges. The enum is derived from the versioned policy, not hardcoded.
- **Untrusted-input hardening.** All plugin-controlled text — file **content**, file **paths**, and **metadata** — is wrapped in an untrusted fence with a unique per-request boundary token and sentinel-escaped, and the system prompt instructs the model to honor only an exact-token close. A malicious path or file can neither forge a boundary nor inject instructions.
- **Honest coverage.** Input is budgeted against the *assembled* prompt size (escaping expansion + fences + paths + metadata + system prompt), not raw content. An oversized bundle drops largest-rendered files first and reports `coverage: "partial"` with the dropped paths — it never truncates silently.
- **Failure semantics.** Model transport/parse failures, and model output that fails the W8.1 finding contract (e.g. an out-of-enum category), are `ModelTransientError` (retryable) — a flaky model is not the non-retryable stage-adapter bug a `FindingValidationError` signals. Model-supplied `source`/`evidenceRefs`/`sourceMetadata` are overridden so a model can't spoof provenance. A blank `promptVersion` is a caller bug and throws up front (not retried).

No migration, no lexicon change, no public-API change. Adds the `AI` binding to `apps/labeler/wrangler.jsonc` (regenerated `worker-configuration.d.ts`).

Reviewed with an opus adversarial pass that found and drove fixes for a fence-escape via unescaped paths, budget undercounting, a misclassified-retryable config bug, and two missing tests — all addressed here before opening.

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [x] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

Admin i18n and changeset are n/a — this touches only `apps/labeler`, a private, unpublished Worker app (no admin UI, not on npm).

- [x] This PR includes AI-generated code — model/tool: Claude Code with Claude Opus 4.8 (design, adversarial review, fixes) and a Sonnet subagent (implementation)

## Screenshots / test output

- `pnpm --filter @emdash-cms/labeler test`: 228 pass (16 files) — 17 new for the adapter (happy path; content-fence + malicious-path injection resistance; model-field override; schema enum = policy; transport/parse/out-of-enum → `ModelTransientError`; blank promptVersion → `TypeError`; oversized-bundle coverage/drops; no-cache option; promptHash stability + policy-sensitivity)
- Typecheck clean; `pnpm lint:json` diagnostics unchanged at 19 (0 in `apps/labeler`)
- Tests use a fake `AiBinding` only — never the live model (Workers AI bills even in local dev)

---

### Try this PR

**[Open a fresh playground →](https://feat-labeler-code-ai-adapter-emdash-playground.emdash-cms.workers.dev)**

<sub>Tracks `feat/labeler-code-ai-adapter`. Updated automatically when the playground redeploys.</sub>


ascorbic:
Fixed the budget undercount you flagged. `capFilesToBudget` was counting only rendered fence length; the assembled user content adds a blank-line separator per file plus a leading newline, so it undercounted by `N+1` and near the cap could report `complete` while the prompt was slightly over. Now it charges the exact overhead — +2 per file (fence + separator) and +1 fixed — and a new test asserts the assembled prompt (system + user messages) stays `<= MAX_MODEL_INPUT_CHARS` under a 55-file over-budget bundle. Re-triggering.

(Separately: once #2001 — the workers-types v5 bump — lands on the base and this rebases onto it, I'll pin the input type to the real `ChatCompletionsInput`/k2.6 ambient type, since CF confirms kimi-k2.7-code shares k2.6's exact interface.)

ascorbic:
Heads-up: pushed one more commit after your approval, so re-triggering. Now that the workers-types v5 bump (#2001) is on the base, I rebased onto it and pinned the request to the real Workers AI `ChatCompletionsInput` via `satisfies` (kimi-k2.7-code shares kimi-k2.6's chat-completions interface). That pin caught a real bug: the native binding's json_schema response format is `{ name, schema }`, not the bare schema object — our schema was nested wrong and would've been rejected at runtime. Fixed + the type now can't drift. 229 tests green, typecheck clean on v5.

ascorbic:
Both addressed:

- **Metadata budget gap**: `capFilesToBudget` now fails fast with a non-retryable `TypeError` when the fixed overhead (system prompt + metadata fence) alone exceeds `MAX_MODEL_INPUT_CHARS` — no amount of file-dropping fixes that, and retrying an oversized request can't succeed, so it must not surface as `ModelTransientError`. Test asserts the error class and that the model is never called.
- **Duplicate-path over-drop**: dropping is now by object identity, not path; a bundle with two files at the same path keeps the small one when only the large one is over budget. Regression test included.

231 tests green. Re-triggering.

~ 🤖 Fable

PR #2030, opened 2026-07-14
fix(core): honor siteUrl in /.well-known OAuth discovery on the anonymous fast path

### LINKED ISSUE #2016: siteUrl does not reach /.well-known/oauth-protected-resource — MCP discovery advertises http:// on Cloudflare
## Description

`siteUrl` cannot reach `/.well-known/oauth-protected-resource`, so on Cloudflare the MCP discovery document advertises `http://` and MCP clients refuse to attach.

The [configuration docs](https://docs.emdashcms.com/reference/configuration/#siteurl) say `siteUrl` fixes "passkeys, CSRF origin matching, OAuth redirects, login redirects, **MCP discovery**, snapshot exports, sitemap, robots.txt, and JSON-LD". It fixes all of those _except_ MCP discovery, which is the one that sends users looking for the option in the first place.

### Root cause

`getPublicOrigin()` resolves `config.siteUrl` → `EMDASH_SITE_URL` env → `url.origin` (`src/api/public-url.ts`), and the route reads the config off `locals`:

```ts
// src/astro/routes/api/well-known/oauth-protected-resource.ts
const origin = getPublicOrigin(url, locals.emdash?.config);
```

But the middleware only attaches `config` to `locals.emdash` on the full-runtime path. Anonymous requests to routes that are neither under `/_emdash` nor in `PUBLIC_RUNTIME_ROUTES` take a fast path whose `locals.emdash` carries handlers but **no `config`**:

```ts
// src/astro/middleware.ts:429
const PUBLIC_RUNTIME_ROUTES = new Set(["/sitemap.xml", "/robots.txt"]);

// src/astro/middleware.ts:515
    // ~:570 — locals.emdash = { handlePublicPluginApiRoute, collectPageMetadata,
    //                           collectPageFragments, getPublicMediaUrl, storage }
    //         no `config`
```

`/.well-known/oauth-protected-resource` is root-level and unauthenticated — which is exactly how an MCP client hits it — so `locals.emdash?.config` is `undefined`, `getPublicOrigin()` falls through to `url.origin`, and behind Cloudflare's proxy that is `http://`.

The env fallback does not save it either: `getEnvSiteUrl()` reads `process.env`, which is empty on Workers unless `nodejs_compat_populate_process_env` is set. The source comment in `public-url.ts` ("On Cloudflare Workers process.env is unavailable") points people away from the one mechanism that does work.

### Related: `security.allowedDomains` is set without a protocol

```ts
// src/astro/integration/index.ts (~410)
...(resolvedConfig.siteUrl
  ? { allowedDomains: [{ hostname: new URL(resolvedConfig.siteUrl).hostname }] }
  : {}),
```

The comment above it says this is done "so `Astro.url` reflects the public origin (helps user template code that reads `Astro.url` directly)". It does not, because only `hostname` is passed. Astro's `allowedDomains` accepts `protocol` too (`Partial<RemotePattern>[]` — "Each pattern can specify `protocol`, `hostname`, and `port`"), and without it `Astro.url.origin` stays `http://` on Cloudflare. Any template that builds canonical/OG/hreflang from `Astro.url.origin` — as the official starter layouts do — ships `http://` canonicals to crawlers.

Passing `protocol` through from `siteUrl` would likely fix both symptoms at the root, since a correct `url.origin` also makes the `getPublicOrigin()` fallback correct.

## Steps to reproduce

1. Deploy an EmDash site to Cloudflare Workers on a proxied custom domain.
2. Set `siteUrl` in `astro.config.mjs`:
   ```js
   emdash({ siteUrl: "https://cms.example.com" /* … */ });
   ```
3. Build and deploy, then compare a full-runtime route with an anonymous one:

```bash
# /sitemap.xml IS in PUBLIC_RUNTIME_ROUTES -> gets `config` -> correct
curl -s https://cms.example.com/sitemap.xml | grep -o '<loc>[^<]*</loc>' | head -1
# <loc>https://cms.example.com/sitemap-work.xml</loc>          ✅ https

# /.well-known/* is NOT -> anonymous fast path -> no `config` -> falls back to url.origin
curl -s https://cms.example.com/.well-known/oauth-protected-resource
# {"resource":"http://cms.example.com/_emdash/api/mcp", …}      ❌ http
```

Same Worker, same deployment, same request. The only difference is whether the route gets `config` on `locals`.

4. Add the site as a custom connector in Claude → it fails to attach, because the discovery document is not https.

### Workaround

Set the origin as an env var instead of (or in addition to) the config option, and turn on `process.env` population:

```jsonc
// wrangler.jsonc
"compatibility_flags": ["nodejs_compat", "nodejs_compat_populate_process_env"],
"vars": { "EMDASH_SITE_URL": "https://cms.example.com" }
```

`getEnvSiteUrl()` reads `process.env` at module scope, bypassing `locals`, so it reaches every route. Verified: discovery then returns `https://`. Both flags are required — with the var but without the compat flag, `process.env` is empty and the var is silently ignored.

For canonical/OG/hreflang, templates additionally need Astro's own `site` option, since `Astro.url.origin` remains `http://`.

## Environment

- `emdash` 0.29.0, `@emdash-cms/cloudflare` 0.29.0
- Astro 7.0.7, `@astrojs/cloudflare` 14.1.2
- Cloudflare Workers (D1 + R2 + KV), proxied custom domain
- `compatibility_date` 2026-02-24, `compatibility_flags: ["nodejs_compat"]`

## Logs / error output

Before the workaround, with `siteUrl` set in config and deployed:

```
$ curl -s https://cms.example.com/.well-known/oauth-protected-resource
{
  "resource": "http://cms.example.com/_emdash/api/mcp",
  "authorization_servers": ["http://cms.example.com/_emdash"],
  ...
}
```

Page head from the same deployment (layout builds these from `Astro.url.origin`):

```
<link rel="canonical" href="http://cms.example.com/">
<meta property="og:url" content="http://cms.example.com/">
<link rel="alternate" hreflang="en" href="http://cms.example.com/">
```

After adding `EMDASH_SITE_URL` + `nodejs_compat_populate_process_env` (and Astro `site` for the head):

```
$ curl -s https://cms.example.com/.well-known/oauth-protected-resource
{
  "resource": "https://cms.example.com/_emdash/api/mcp",
  "authorization_servers": ["https://cms.example.com/_emdash"],
  ...
}
```

## Suggested fixes

1. Attach `config` (or at least `siteUrl`) to `locals.emdash` on the anonymous fast path too — the `.well-known` routes are public by design, so they will always land there.
2. Pass `protocol` into `security.allowedDomains` from `siteUrl`, so `Astro.url` actually reflects the public origin as the comment intends.
3. Update the `public-url.ts` comment and the `siteUrl` docs: `process.env` _is_ available on Workers with `nodejs_compat_populate_process_env`, and it is currently the only mechanism that reaches the anonymous routes.

swissky:
## What does this PR do?

`siteUrl` never reached `/.well-known/oauth-protected-resource` (and `/.well-known/oauth-authorization-server`): these routes are root-level and unauthenticated — exactly how MCP clients fetch them — so they land on the middleware's anonymous fast path, where `locals.emdash` carries handlers but no `config`. `getPublicOrigin()` then fell through to `url.origin`, which is `http://` behind Cloudflare's proxy, and MCP clients refused to attach because the discovery document advertised non-https URLs.

The fix: both routes now fall back to the build-time `virtual:emdash/config` (which carries the origin-normalized `siteUrl`) when `locals.emdash?.config` is absent. This also covers the case where anonymous-path runtime init fails and `locals.emdash` is never set at all. Verified the symptom on a real Workers deployment before the fix (`curl .../.well-known/oauth-protected-resource` → `"resource":"http://…"`).

Also included, from the same report (#2016):

- Corrects the `public-url.ts` comment: `process.env` **is** readable on Workers when `nodejs_compat_populate_process_env` is enabled — the old comment pointed people away from the one mechanism that worked.
- Documents the Workers env-var setup (`nodejs_compat_populate_process_env` + `vars`) in the `siteUrl` reference, and clarifies that `siteUrl` doesn't change `Astro.url` in user templates (use Astro's `site` option for self-built canonical/OG URLs).

Deliberately **not** included: the issue's suggestion to pass `protocol` into `security.allowedDomains`. The Cloudflare adapter passes `request.url` straight through and never runs Astro's forwarded-header validation, so that change wouldn't affect `Astro.url` on Workers — and on the Node path it could regress proxy setups that send only `Host` (no `x-forwarded-host`), since a protocol pattern makes host validation fail for the internal http hop. Analysis posted on the issue.

Closes #2016

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/... (n/a — bug fix)

- [x] This PR includes AI-generated code — model/tool: Cursor + Fable

## Screenshots / test output

New regression tests in `packages/core/tests/unit/astro/well-known-public-origin.test.ts` (named `reproduces #2016`): siteUrl honored with config-less `locals.emdash`, with `locals.emdash` entirely absent, and locals-config precedence preserved. Full `emdash` package suite:

```
Test Files  356 passed | 1 skipped (357)
     Tests  4723 passed | 3 skipped (4726)
```


packages/core/src/astro/routes/api/well-known/oauth-authorization-server.ts
```diff
@@ -10,14 +10,19 @@
...
 import { getPublicOrigin } from "#api/public-url.js";
 import { VALID_SCOPES } from "#auth/api-tokens.js";
 
 export const prerender = false;
 
 export const GET: APIRoute = async ({ url, locals }) => {
-	const origin = getPublicOrigin(url, locals.emdash?.config);
+	// Same fallback as oauth-protected-resource.ts: this route is public and
+	// root-level, so it lands on the anonymous fast path where locals.emdash
+	// carries no config (#2016).
```
khoinguyenpham04: Could we shorten this while keeping the reason for the fallback? 

```suggestion
	// Anonymous discovery requests omit runtime config; use build-time config
	// so `siteUrl` can override the proxy's internal origin.
```

packages/core/src/astro/routes/api/well-known/oauth-protected-resource.ts
```diff
@@ -12,14 +12,21 @@
...
 
 export const prerender = false;
 
 export const GET: APIRoute = async ({ url, locals }) => {
-	const origin = getPublicOrigin(url, locals.emdash?.config);
+	// Anonymous requests outside /_emdash take the middleware fast path,
+	// which attaches locals.emdash WITHOUT config — and MCP clients hit this
+	// route unauthenticated by design. Fall back to the build-time config so
+	// `siteUrl` still reaches discovery (#2016); behind Cloudflare's proxy
+	// url.origin is http:// and clients refuse to attach.
```
khoinguyenpham04: Could we make this comment more concise? Maybe:

```suggestion
// The anonymous fast path omits runtime config, so fall back to build-time config.
```

khoinguyenpham04 (commented):
Everything looks good to me! I only left a couple of non-blocking suggestions.

Thanks for tackling this issue :)

packages/core/src/astro/routes/api/well-known/oauth-authorization-server.ts
swissky: Done in f9aaad83 — shortened to just the sibling reference, thanks.

packages/core/src/astro/routes/api/well-known/oauth-protected-resource.ts
swissky: Done in f9aaad83 — adopted your wording, kept the #2016 reference and the siteUrl rationale. Thanks!

khoinguyenpham04: thank you @swissky !! LGTM

khoinguyenpham04 (approved):

PR #2036, opened 2026-07-14
fix: actionable slug-conflict error when publishing a staged slug

### LINKED ISSUE #2034: Publish fails with opaque D1 UNIQUE error when staged slug collides with an existing published entry
## Summary

Publishing a `news` (or any collection) entry whose staged slug collides with
another entry's already-published `(slug, locale)` fails with a raw D1/SQLite
error and no indication of *what* failed or *how to fix it*.

## Repro

1. Create/edit a draft entry and set its slug to a value that another
   **already-published** entry in the same locale already uses.
2. Click Publish (or `POST /_emdash/api/content/{collection}/{id}/publish`).
3. The request 500s. The only feedback surfaced (both in the admin UI toast
   and in the Worker logs) is:

   ```
   Content publish error: Error: D1_ERROR: UNIQUE constraint failed:
   ec_news.slug, ec_news.locale: SQLITE_CONSTRAINT (extended: SQLITE_CONSTRAINT_UNIQUE)
   ```

   There is no mention of the slug value, the locale, or which existing entry
   already holds it. As an editor, all you see is "failed to publish" — you
   have to go spelunking in the database to find out that a totally unrelated
   article (created via an earlier CSV/legacy import with a numeric slug like
   `11`) happens to already own the slug you typed.

## Where this happens

`ContentRepository.publish()` writes the staged `_slug` from the draft
revision straight to the live `slug` column with a raw `UPDATE ... SET slug =
...` (see `packages/core/src/database/repositories/content.ts`, inside
`publish()`, the `if (typeof revision.data._slug === "string")` block). This
bypasses whatever slug-uniqueness pre-check exists for `handleContentUpdate`
(the comment right above `handleContentPublish` even calls this out: "a slug
edit is staged as `_slug` in the draft revision and only lands on the live
`slug` column here, inside `repo.publish()` — it never passes through
`handleContentUpdate`, where slug-change auto-redirects are normally
created"). So the uniqueness check that presumably guards normal slug edits
never runs on this path, and the DB constraint is the only thing that catches
it — as a generic 500.

## Suggested fix

Before writing `_slug` to the live column in `publish()`, check for an
existing published row with the same `(slug, locale)` (excluding the current
id) and return a proper `EmDashValidationError` / `VALIDATION_ERROR` result
naming the conflicting slug — the same shape already used elsewhere in this
function for other failure modes. That turns an opaque 500 + raw SQLite
message into an actionable "this slug is already used by entry X" error the
admin UI can show inline on the slug field.

## Environment

- `emdash` 0.28.1
- Cloudflare Workers + D1

swissky:
## What does this PR do?

Publishing an entry whose staged slug (stored as `_slug` in a draft revision) collides with another entry's `(slug, locale)` failed with an opaque 500 and a raw D1/SQLite message (`D1_ERROR: UNIQUE constraint failed: ec_news.slug, ec_news.locale`) — after draft data had already been synced into the live columns. The staged slug lands on the live column only inside `repo.publish()`, bypassing the uniqueness handling that guards direct slug edits.

This PR:

- Validates the staged slug inside `publish()` **before any write**, against all rows in the entry's locale — including drafts and trashed entries, since the `(slug, locale)` unique index covers them. On conflict it throws a validation error naming the slug and the conflicting entry id, which the handler maps to the same `SLUG_CONFLICT` (409) as direct slug edits in create/update.
- Reorders the publish writes so the slug `UPDATE` (the only statement that can hit the constraint) runs before `syncDataColumns` — if a concurrent write takes the slug in the race window, no partial publish is left behind on D1 (which has no transactions to roll back).
- Maps the unique-constraint fingerprint in `handleContentPublish`'s catch to `SLUG_CONFLICT` as a backstop for that race window, matching the existing create/update handling — a raw SQLite error never reaches the client.
- Skips the check for rows with a `NULL` locale (unique indexes treat NULLs as distinct, so those rows can't collide).

Closes #2034

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

## Screenshots / test output

Regression tests: staged slug conflicting with a published entry, with a trashed entry, republish with own slug, same slug across locales, and the handler-level `SLUG_CONFLICT` mapping.

```
Test Files  239 passed (239)
     Tests  3190 passed (3190)
```

### DIFF
--- a/.changeset/publish-staged-slug-conflict.md
+++ b/.changeset/publish-staged-slug-conflict.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes publishing an entry whose staged slug is already used by another entry in the same locale failing with an opaque database error. Publish now returns a validation error naming the conflicting slug and entry, which the admin can show inline, and leaves the live entry untouched.
--- a/packages/core/src/api/handlers/content.ts
+++ b/packages/core/src/api/handlers/content.ts
@@ -1487,10 +1487,18 @@ export async function handleContentPublish(
 			};
 		}
 		if (error instanceof EmDashValidationError) {
+			// The staged-slug pre-check tags its error so it maps to the same
+			// 409 SLUG_CONFLICT as direct slug edits in create/update.
+			const details: unknown = error.details;
+			const isSlugConflict =
+				typeof details === "object" &&
+				details !== null &&
+				"code" in details &&
+				details.code === "SLUG_CONFLICT";
 			return {
 				success: false,
 				error: {
-					code: "VALIDATION_ERROR",
+					code: isSlugConflict ? "SLUG_CONFLICT" : "VALIDATION_ERROR",
 					message: error.message,
 				},
 			};
--- a/packages/core/src/api/handlers/content.ts
+++ b/packages/core/src/api/handlers/content.ts
@@ -1495,6 +1495,23 @@ export async function handleContentPublish(
 				},
 			};
 		}
+		// Backstop for the pre-check inside repo.publish(): a concurrent write
+		// can still take the slug between the check and the UPDATE, in which
+		// case the `(slug, locale)` unique constraint fires. Same fingerprint
+		// mapping as create/update — never a raw SQLite error to the client.
+		const message = error instanceof Error ? error.message.toLowerCase() : "";
+		if (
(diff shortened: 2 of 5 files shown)


ascorbic (approved):
Thanks

PR #2050, opened 2026-07-15
fix(setup): allow admin recovery when setup_complete is set but no users exist

edrpls:
## What does this PR do?

Fixes an unrecoverable setup lockout. `GET /api/setup/status` deliberately resumes the wizard at the **admin** step when `emdash:setup_complete` is `true` but the `users` table is empty (`status.ts`: "If setup_complete but no users, jump to admin step") — a state that appears when a completed instance's users are wiped or a sanitised DB copy is shipped to a new environment. But `POST /api/setup/admin` and `POST /api/setup/admin/verify` rejected that state on the flag alone (`SETUP_COMPLETE`, 400), so the step status advertises was unreachable: the wizard UI shows the account form, the POST 400s, and the instance has **no UI path to an admin account** (recovery required a manual `DELETE FROM options` on the DB — hit in production while promoting a sanitised QA database to a fresh prod environment on 2026-07-15).

Both routes now reject `SETUP_COMPLETE` only when a user **also** exists. `ADMIN_EXISTS` (users present without the flag) is unchanged, and the empty-users fall-through lets the wizard re-create the admin — `admin-verify` re-sets `setup_complete` on success, exactly as before.

State matrix (only the third row changes):

| `setup_complete` | users | before | after |
|---|---|---|---|
| true | >0 | 400 `SETUP_COMPLETE` | 400 `SETUP_COMPLETE` |
| false | >0 | 400 `ADMIN_EXISTS` | 400 `ADMIN_EXISTS` |
| true | 0 | 400 `SETUP_COMPLETE` (dead end) | **allowed — recovery** |
| false | 0 | allowed | allowed |

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes (`packages/core` via tsgo after a workspace build)
- [x] `pnpm lint` passes (oxlint `--type-aware --deny-warnings` on the changed files)
- [x] `pnpm test` passes (or targeted tests for my change) — `tests/integration/astro/` suite; the one failure on my machine (`media-confirm-placeholder`) is environmental (sharp's native binding doesn't build locally) and fails identically without this diff
- [x] `pnpm format` has been run (oxfmt over the changed directories — no churn)
- [ ] User-visible strings in the admin UI are wrapped for translation (n/a — no UI strings)
- [x] I have added a [changeset](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#changesets) (`emdash`: patch)

## Screenshots / test output

New test file `packages/core/tests/integration/astro/setup-admin-recovery.test.ts` (mirrors the `setup-admin-nonce` harness — real in-memory SQLite, no mocking). **Reproduction proof:** with the route changes stashed (unpatched code), 3 of the 5 tests fail — the two recovery-allowed cases (`true` and `"true"` flag variants) and the verify-gate case; the two regression-pin cases (SETUP_COMPLETE with users, ADMIN_EXISTS without flag) pass, as expected since those behaviors are unchanged. With the fix: 5/5.

```
unpatched:  Tests  3 failed | 2 passed (5)
patched:    Tests  5 passed (5)
```

### DIFF
--- a/.changeset/setup-admin-recovery.md
+++ b/.changeset/setup-admin-recovery.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes admin-account recovery when `emdash:setup_complete` is set but the users table is empty (a wiped or sanitised database copied between environments). `/api/setup/status` already resumes the wizard at the admin step for this state, but the admin-step routes rejected it with `SETUP_COMPLETE`, leaving no UI path to an admin account. Both routes now reject on the flag only when a user also exists; the fully-set-up and admin-exists rejections are unchanged.
--- a/packages/core/src/astro/routes/api/setup/admin-verify.ts
+++ b/packages/core/src/astro/routes/api/setup/admin-verify.ts
@@ -33,18 +33,22 @@ export const POST: APIRoute = async ({ cookies, request, locals }) => {
 		// Check if setup is already complete
 		const options = new OptionsRepository(emdash.db);
 		const setupComplete = await options.get("emdash:setup_complete");
-
-		if (setupComplete === true || setupComplete === "true") {
-			return apiError("SETUP_COMPLETE", "Setup already complete", 400);
-		}
+		const isComplete = setupComplete === true || setupComplete === "true";
 
 		// Check if any users exist
 		const adapter = createKyselyAdapter(emdash.db);
 		const userCount = await adapter.countUsers();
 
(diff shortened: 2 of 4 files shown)


ascorbic (approved):
Thanks! Good catch

PR #2070, opened 2026-07-16
feat(labeler): aggregator service binding + read-only client (W8.4 slice-3 foundation)

ascorbic:
## What does this PR do?

Adds the read-only labeler→aggregator service binding and a typed client — the foundation the ratified W8.4 slice-3 read path (and W10.4 slice A contact resolution) build on. No consumer wiring yet; this lands the mechanism.

The aggregator is an HTTP/XRPC-only module Worker (no RPC entrypoint), so consumption is fetch-over-binding, matching the repo's one existing service-binding example (`infra/perf-monitor`). `AggregatorClient` wraps the `AGGREGATOR` `Fetcher` with three methods over the existing XRPC reads — `getPackage`, `getLatestRelease`, `listReleases` — typed against `@emdash-cms/registry-lexicons` view types. NSIDs are fixed per-method constants; only query-param values carry caller data and every value is `encodeURIComponent`-encoded, so no caller data can reach the host/path. One fetch per call, no caching (honors W10.4's fresh-read-at-notification rule). XRPC `NotFound` → `null`; any other non-2xx or a transport failure throws (callers treat a throw as transient).

**Moderation-semantics note (flagged for maintainer):** the client sends a **blank** `atproto-accept-labelers` header to obtain a genuinely unfiltered view. This is deliberate and counter-intuitive: *omitting* the header makes the aggregator apply its default trusted-labeler redaction, so a redacted/taken-down subject returns `NotFound`→`null` and silently blinds the labeler's own analysis to the subjects it most needs to see. The labeler is the moderation authority performing analysis, so it must read unfiltered. The re-exposure risk (an unfiltered read reaching a public surface) is bounded at the **serving** layer, not here — see the ratified sync.getRecord / artifact-mirror serving decisions. This was caught by an adversarial review pass (the original code omitted the header under a comment claiming the opposite).

Part of the plugin-registry labelling service (RFC #694, umbrella #1909). Targets the integration branch.

## Type of change

Covered by the approved RFC #694 umbrella.

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes (zero diagnostics in changed files; pre-existing baseline unchanged)
- [x] `pnpm test` passes (aggregator-client 11/11; full labeler suite green)
- [x] `pnpm format` has been run (touched paths; worktree path is gitignored so local oxfmt skips it — CI format-checks the branch checkout)
- [x] User-visible strings in the admin UI are wrapped for translation (if applicable) — n/a, no UI
- [x] I have added a changeset (if this PR changes a published package) — n/a, `@emdash-cms/labeler` is unpublished (`private: true`)
- [x] New features link to an approved Discussion — RFC #694

- [x] This PR includes AI-generated code — model/tool: Claude Opus 4.8 (implementation and adversarial review)

## Screenshots / test output

`vitest run test/aggregator-client.test.ts` → 11/11 pass (URL/param encoding per method, typed parse, NotFound→null, 5xx and transport-failure throw, and the blank-header assertion). A note for the consumer-wiring slices: verify via a real-binding integration test that the blank header survives the service-binding transport and yields the unfiltered view — if empty-valued headers are dropped in transport, escalate to an explicit aggregator internal-read affordance.

---

### Try this PR

**[Open a fresh playground →](https://feat-labeler-aggregator-binding-emdash-playground.emdash-cms.workers.dev)**

<sub>Tracks `feat/labeler-aggregator-binding`. Updated automatically when the playground redeploys.</sub>


ascorbic:
Addressed the review:

- **[needs fixing] binding-transport header survival** — added a workerd-level test (`AGGREGATOR binding transport`) that calls the real `env.AGGREGATOR` service binding with a blank `atproto-accept-labelers` header and asserts it arrives as `""` (the vitest stub echoes the inbound value as a marker, so a dropped empty header would read `absent`, not `empty`). This proves the unfiltered-read contract holds across the workerd binding transport, not just that the client sets the header. The stub still fails loud (501) for any accidental call.

Declining the two suggestions, with rationale:

- **Runtime response validation** — the client intentionally treats the aggregator as a first-party, in-process binding and re-verifies artifact bytes by checksum downstream; the module doc states this trust assumption. Adding `.parse()` at the boundary couples the client to lexicon-schema internals and adds per-call cost without defending against the only actor that could return a schema-valid-but-wrong view (a compromised first-party aggregator). Better revisited if a specific consumer needs strict shape guarantees than baked into the foundation.
- **Sharing `ACCEPT_LABELERS_HEADER` from `registry-moderation`** — it's a fixed ATProto protocol string with negligible drift risk; the shared-export refactor would touch a published package (+ changeset) and the aggregator's `request-policy.ts`, which is out of scope for this read-only foundation. Noted as a reasonable future tidy-up.

~ 🤖 Claude Fable 5

PR #2083, opened 2026-07-16
feat(labeler): reconsideration case management + outcome notice (W10.6 server)

ascorbic:
## What does this PR do?

Implements the **server side** of the W10.6 manual reconsideration workflow for the plugin-registry labelling service (RFC #694, umbrella #1909). Targets the `feat/plugin-registry-labelling-service` integration branch.

The workflow is deliberately operator-side only — **no authenticated appeals portal or SLA** (per the plan). A publisher who wants a decision reconsidered emails the monitored inbox (already published in `/.well-known/emdash-labeler-policy.json` via `contact.reconsiderationUrl`/`reconsiderationEmail`) quoting the opaque public assessment id (the existing `asmt_<ulid>`, already served by the public assessment API and every notice). An operator then manages the case in the console:

- **Open** a case for the subject of the quoted assessment, with a first private note.
- **Add** further private, operator-only notes over the life of the case.
- **Resolve** with an outcome (`granted` / `denied` / `withdrawn`), which — for `granted`/`denied` — fires a publisher **outcome notice** through the W10.5 double-opt-in pipeline. `withdrawn` notifies nothing.

Design decisions were ratified and recorded in the plan doc (`docs(labeler-plan): ratify W10.6 reconsideration design decisions`).

**What's here (slice 1 — server only):**

- **Migration `0009_reconsiderations.sql`:** `reconsiderations` (mutable case, keyed on subject `uri`+`cid` so it survives reruns; `state` open→resolved; a partial unique index bounds one open case per subject; a table CHECK binds `state` ↔ `outcome` ↔ resolved-provenance so no half-resolved row can exist) and `reconsideration_notes` (append-only, immutable via triggers). Private notes live here, deliberately **off** the semi-public `operator_actions.reason` so note text can never reach notice copy.
- **Three reviewer console mutations** (`reconsiderations/open`, `:id/note`, `:id/resolve`) via the existing `guardMutation`/`commitMutation` seam — Access JWT + reviewer role + CSRF + request-fingerprint idempotency + atomic audit-row-plus-effect batch, none of it hand-rolled. Duplicate-open and double-resolve are `409`s, backstopped by the partial unique index and a `state='open'`-guarded UPDATE respectively.
- **Outcome notice** (`reconsideration-outcome`) keyed on the resolve `operator_action` (`source_type: 'operator'`), so the W10.5 retry sweep re-renders byte-identical public copy. The notice is gated on the resolve **winning** its guarded UPDATE — on both the fresh and the replay paths — so it fires **at most once per case** even under a concurrent-resolve-plus-key-replay race.
- **Read API:** paginated case list + case-with-notes detail, reviewer-gated.

The operator console **UI** for this is **slice 2** (a follow-up PR). This PR is server + tests only.

Part of #1909. Related to #694.

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm lint` passes (0 diagnostics on touched files)
- [x] `pnpm test` passes (18 reconsideration integration tests; regression re-run of console-mutation-api + notification-triggers + notification-sweep green)
- [x] User-visible strings in the admin UI are wrapped for translation (if applicable) — **n/a**: slice 1 is server-only; no admin/console UI strings. The console UI is slice 2.
- [x] I have added a [changeset](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#changesets) (if this PR changes a published package) — **n/a**: `@emdash-cms/labeler` is `private: true`, not in the published set.
- [x] New features link to an approved Discussion — the umbrella #1909 tracks RFC #694 as the approved design source; this is one slice under it.

- [x] This PR includes AI-generated code — model/tool: **Claude Opus 4.8** (implementation + adversarial review), orchestrated by **Claude Fable 5**

## Screenshots / test output

```
Test Files  1 passed (1)
     Tests  18 passed (18)
```

The adversarial review pass confirmed one real defect and it is fixed here: the resolve **replay** branch fired the outcome notice without the win-check the fresh path uses, so a losing concurrent resolve whose idempotency key was later replayed could send a second — possibly contradictory — outcome email under its own source id. The replay branch now runs the identical `resolveWon` gate; a reproduction test (`does not re-notify when a losing concurrent resolve's key is replayed`) fails without the gate (two notices leak) and passes with it.

---

### Try this PR

**[Open a fresh playground →](https://feat-labeler-reconsideration-emdash-playground.emdash-cms.workers.dev)**

<sub>Tracks `feat/labeler-reconsideration`. Updated automatically when the playground redeploys.</sub>

### DIFF
--- a/apps/labeler/migrations/0009_reconsiderations.sql
+++ b/apps/labeler/migrations/0009_reconsiderations.sql
@@ -37,8 +37,11 @@ CREATE TABLE reconsiderations (
 	resolved_at_epoch_ms INTEGER,
 	outcome_action_id TEXT REFERENCES operator_actions(id),
 	CHECK (
-		(state = 'open' AND outcome IS NULL AND resolved_at IS NULL AND outcome_action_id IS NULL)
-		OR (state = 'resolved' AND outcome IS NOT NULL AND resolved_at IS NOT NULL)
+		(state = 'open' AND outcome IS NULL AND resolved_at IS NULL AND resolved_at_epoch_ms IS NULL
+			AND resolved_by_id IS NULL AND outcome_action_id IS NULL)
+		OR (state = 'resolved' AND outcome IS NOT NULL AND resolved_at IS NOT NULL
+			AND resolved_at_epoch_ms IS NOT NULL AND resolved_by_id IS NOT NULL
+			AND outcome_action_id IS NOT NULL)
 	)
 );
 
--- a/apps/labeler/migrations/0009_reconsiderations.sql
+++ b/apps/labeler/migrations/0009_reconsiderations.sql
@@ -1,0 +1,79 @@
+-- Publisher reconsideration case management (spec §18/§19, plan W10.6). Two
+-- tables mirroring the operator_actions(immutable)/notifications(mutable) split:
+--
+--   * `reconsiderations`      — MUTABLE case, keyed on the subject release
+--     (uri + cid) which is STABLE across reruns (a reconsideration is about
+--     "this release"; a rerun mints a fresh assessment id). State open →
+--     resolved; at most one open case per subject via a partial unique index.
+--   * `reconsideration_notes` — APPEND-ONLY private operator notes, immutable
+--     like the other audit logs. Kept OFF `operator_actions`, whose `reason` is
+--     semi-public (folded into notices/events): a private note must never reach
+--     notice copy.
+--
+-- `triggering_assessment_id` is context only (the assessment the publisher
+-- quoted when they wrote in); the case tracks the release, not that run.
+--
+-- Timestamp columns that queries order on gain an integer `*_epoch_ms` sibling,
+-- matching 0003/0004/0005 (RFC 3339 strings compare incorrectly across timezone
+-- offsets in SQL).
+
+CREATE TABLE reconsiderations (
+	id TEXT PRIMARY KEY,
+	subject_uri TEXT NOT NULL,
+	subject_cid TEXT NOT NULL,
+	triggering_assessment_id TEXT NOT NULL REFERENCES assessments(id),
+	state TEXT NOT NULL CHECK (state IN ('open', 'resolved')),
+	outcome TEXT CHECK (outcome IN ('granted', 'denied', 'withdrawn')),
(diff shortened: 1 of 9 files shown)


ascorbic:
Thanks — addressed the resolved-event tradeoff you flagged (commit 83232c90).

The `reconsideration-resolved` operational event is now gated on the resolve winning its `state = 'open'` UPDATE, via a new `gateOnResolvedReconsideration` in-batch gate that mirrors `gateOnIssuedLabelActionKey`: `INSERT ... SELECT ... WHERE EXISTS (SELECT 1 FROM reconsiderations WHERE id = ? AND outcome_action_id = ?)`. Batched after the guarded UPDATE, so a losing concurrent resolve emits no phantom event while its audit row still commits for idempotency replay.

I deliberately left the loser's HTTP response as a 200 with its own descriptor rather than a 409: `guardMutation` returns the stored descriptor on a same-key replay, so the original and replay responses must agree — a 409 would diverge. The outcome notice was already gated by `resolveWon`. So post-fix, a losing concurrent resolve commits only its audit row (no event, no notice) and echoes its requested outcome in its own 200; the authoritative case state, resolved-event, and notice all reflect only the race winner. A handler comment now spells this out, and the concurrent-loser test asserts exactly one resolved-event (the winner's).

~ 🤖 Claude Fable 5

ascorbic:
Took the schema-defensiveness suggestion (commit 2ffd8268). The `reconsiderations` CHECK now fully enforces the resolve-provenance it describes rather than leaving it to app code:

- **open** branch also requires `resolved_at_epoch_ms IS NULL AND resolved_by_id IS NULL` (an open row carries zero resolve provenance).
- **resolved** branch also requires `resolved_at_epoch_ms IS NOT NULL AND resolved_by_id IS NOT NULL AND outcome_action_id IS NOT NULL` (a resolved row carries all of it).

Two tests assert the DB rejects a partial-resolved row and an open row with stray provenance; the existing resolve tests confirm legitimate resolves (which always set the full tuple) still pass. 20/20 green.

~ 🤖 Claude Fable 5

PR #2126, opened 2026-07-18
fix(admin): echo _rev on editor save to prevent silent draft overwrite (#2121)

### LINKED ISSUE #2121: Admin editor can silently overwrite a newer draft: PUTs omit _rev and editor loads can miss the latest draft revision (0.29.0)
**Version:** emdash 0.29.0 (with @emdash-cms/admin 0.29.0), Cloudflare Workers runtime, D1 with `session: "auto"`, KV object cache.

**Summary:** the admin page editor can silently overwrite a newer draft revision. Two behaviours combine:

1. The editor's content PUTs never send `_rev`, so the server-side optimistic concurrency check (`validateRev`) never engages ("No `_rev` = blind write" per the comment in the update handler).
2. Shortly after a save there is a stale read window in which loading the editor returns the entry WITHOUT its newest draft revision. `ContentRepository.findById` / `findBySlug` / `mapRow` build `item.data` from the base content table columns only; the draft revision is hydrated separately, and in our reproductions the editor sometimes rendered base-only content while `draft_revision_id` pointed at a newer revision in the `revisions` table.

When 1 and 2 combine, a user who reloads the editor inside the stale window sees old content, edits it, and saves; the save repoints `draft_revision_id` at a new revision built from the stale state. The newer draft is orphaned with no warning. We reproduced this twice on a fresh 0.29.0 tenant; the revisions table showed the newer draft (585 bytes, custom blocks intact) and then a post-reload save created a smaller revision from base content and repointed the draft.

**Steps to reproduce (timing dependent):**
1. Edit a draft page, add content, save (draft revision written, correct).
2. Reload the editor shortly after; if the load falls in the stale window the editor shows content without the fresh draft.
3. Edit and save again: the newer draft is silently replaced.

**Suggested fixes:** have the admin editor echo `_rev` from the GET into its PUTs so `validateRev` refuses the stale save; and/or hydrate the draft revision atomically in the same read that loads the entry for the editor.

**Workaround we ship:** template middleware injects a small script into the admin shell that remembers `_rev` from content API responses and adds it to PUT bodies that lack one, letting the existing `validateRev` do its job.

swissky:
## What does this PR do?

The content API returns an `_rev` token on reads and honours it on writes (409 Conflict on a stale token), but the admin never echoed it back — every editor save and autosave was a blind write. When the editor loaded an entry inside the stale-read window and the user saved, the save silently repointed `draft_revision_id` at a new revision built from stale state, orphaning the newer draft with no warning.

This makes the editor round-trip `_rev` so the server-side `validateRev` check actually engages:
- `fetchContent`/`createContent`/`updateContent` lift the envelope-level `_rev` onto the returned item (the server returns it next to `item`, not inside it).
- `ContentItem`/`UpdateContentInput` carry the optional `_rev` field.
- `handleSave`/`handleAutosave` echo `rawItem._rev`, so a save based on a stale read is rejected with 409 and the editor shows a clear error instead of losing work.

Autosave stays consistent: it updates the existing draft revision in place (no `updated_at` bump on the content table), and the post-save cache invalidation refetches a fresh `_rev`.

Closes #2121

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

- [x] This PR includes AI-generated code — model/tool: Cursor + Kimi K3

## Screenshots / test output

`packages/admin/tests/lib/content-rev.test.ts` — 3 passing: fetchContent surfaces `_rev`, updateContent sends it, blind writes omit it.


ascorbic:
@swissky This is close. Could you give it one final pass so we can merge it:

- Add a router-level regression test proving the GET _rev reaches save/autosave PUTs.
- Pass `_rev` through the remaining updateContent paths: author, SEO, and publishedAt.
- Refresh the cached _rev from the PUT response to avoid a transient false 409 before refetch completes.

Small clarification on the existing review: metadata-only writes cannot overwrite draft content, but they should still use consistent concurrency protection.

Once those are addressed, this should be ready to land. Thanks!

ascorbic (changes requested):
See above

ascorbic (approved):

ascorbic:
I've implemented those changes so we can get this in. Thanks!

swissky:
@ascorbic thank you🙏🏻

PR #2129, opened 2026-07-19
fix(core): content API answers 500 NOT_CONFIGURED, not 401, when the runtime failed to init

### LINKED ISSUE #2094: Content API returns 401 UNAUTHORIZED when the runtime fails to initialize (should be 5xx)
## Summary

When the EmDash runtime fails to initialize (broken DB binding, downstream failure), the content API responds **`401 UNAUTHORIZED` / "Authentication required"** to requests carrying a perfectly valid token. The `NOT_CONFIGURED` 500 that exists for exactly this case is unreachable, because the permission check runs before it.

This misattributes a server fault to the caller's credentials. In our case it cost about half a day of debugging a token that was never broken, and it breaks retry logic: clients correctly treat 4xx as non-retryable.

**Affected:** verified in **0.1.0** and **0.29.0** (latest). Present in both; line numbers below for each.

## Reproduction

1. Deploy an EmDash site where runtime init fails (e.g. break the D1 binding).
2. `GET /_emdash/api/content/posts?status=published&limit=75` with a **valid** `Authorization: Bearer <token>`.
3. Observe: `401 {"error":{"code":"UNAUTHORIZED","message":"Authentication required"}}`, returned in ~0.15s.
4. `GET /` still returns 200 — the Worker is alive.
5. `GET /_emdash/api/schema` under the same fault correctly returns **5xx**.

Step 5 is the tell: the same codebase handles this correctly in one place and incorrectly in another.

## Root cause

Three steps:

**1.** The middleware swallows the init error, so `locals.emdash` is never set:

```js
// 0.29.0 src/astro/middleware.ts:796  (0.1.0: :323)
} catch (error) {
    console.error("EmDash middleware error:", error);
```

**2.** Auth middleware bails out before it ever looks at the token, leaving `locals.user` undefined:

```js
// 0.29.0 src/astro/middleware/auth.ts:350  (0.1.0: :273)
if (!emdash?.db) {
    return next();
}
```

**3.** The content route checks permission *before* checking initialization:

```js
// 0.29.0 src/astro/routes/api/content/[collection]/index.ts
const denied = requirePerm(user, "content:read");     // line 20
if (denied) return denied;                            // -> 401, user is undefined
// ...
if (!emdash?.handleContentList) {
    return apiError("NOT_CONFIGURED", "EmDash is not initialized", 500);   // line 27 — unreachable
}
```

`requirePerm` (`src/api/authorize.ts`) returns `apiError("UNAUTHORIZED", "Authentication required", 401)` when `!user`.

The 500 sits seven lines too late and never fires in the failure it was written for.

## Suggested fix

The schema routes already do it right — mirror them:

```js
// src/astro/routes/api/schema/index.ts:25-28 (0.29.0)
const dbErr = requireDb(emdash?.db);
if (dbErr) return dbErr;
const denied = requirePerm(user, "schema:read");
```

`requireDb` is already exported from `#api/error.js`. Applying the same order to the content routes (9 of 10 have the inverted order) would fix this without new concepts.

Optionally: the swallowed error in `middleware.ts` is the only trace of the real fault. Surfacing it in the 5xx body — or at least behind a debug flag — would make this diagnosable from the outside.

## Why it matters

A 401 tells the caller "your credentials are wrong". Clients act on that: ours excludes 4xx from retries by design, so a transient server fault aborts the build immediately and points the on-call person at the wrong thing.

dchaudhari7177:
## What does this PR do?

When the EmDash runtime fails to initialize (broken DB binding, downstream failure), `locals.emdash` is never set and the auth middleware never resolves a user — even for a valid token. The content routes ran `requirePerm(user, …)` first, so they answered **401 UNAUTHORIZED** and the `NOT_CONFIGURED` 500 a few lines below was unreachable. A server fault got blamed on the caller's credentials, and clients correctly treating 4xx as non-retryable aborted instead of retrying.

This PR moves each route's existing initialization check (handler-presence check or `requireDb`) **ahead of** the permission check in the 13 content route files that had the order inverted, mirroring the schema routes which already do this correctly (`requireDb` before `requirePerm` in `src/astro/routes/api/schema/index.ts`). Routes that already checked init first (`publish`, `unpublish`, `restore`, `schedule`, `discard-draft`) are untouched. No new concepts — each route keeps its own check, just earlier.

Tests: the authz mock runtime now carries a `db` (a configured runtime) so all existing 403 assertions still exercise the authz path, and a new describe block pins the regression — uninitialized runtime → **500 `NOT_CONFIGURED`** on list / get / trash / a `requireDb`-gated route (references/parents) / a handler-presence write route (permanent DELETE), including for an authenticated admin. A patch changeset is included.

Closes #2094

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

## Screenshots / test output

`vitest run` on `content-route-permissions`, `content-routes-authz` (26 tests incl. 6 new), and `content-route-locale`: all pass. Prettier clean.

### DIFF
--- a/.changeset/content-api-init-500.md
+++ b/.changeset/content-api-init-500.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+The content API now answers 500 NOT_CONFIGURED instead of 401 UNAUTHORIZED when the EmDash runtime failed to initialize. Previously the permission check ran first, so a server fault (broken DB binding, init failure) was reported as an authentication error even for requests carrying a valid token — and retry logic correctly treating 4xx as non-retryable gave up instead of retrying. The 13 affected content routes now check initialization before permissions, matching the schema routes.
--- a/packages/core/src/astro/routes/api/content/[collection]/[id].ts
+++ b/packages/core/src/astro/routes/api/content/[collection]/[id].ts
@@ -18,15 +18,14 @@ export const prerender = false;
 
 export const GET: APIRoute = async ({ params, url, locals }) => {
 	const { emdash, user } = locals;
+	if (!emdash?.handleContentGet) {
+		return apiError("NOT_CONFIGURED", "EmDash is not initialized", 500);
+	}
 	const denied = requirePerm(user, "content:read");
 	if (denied) return denied;
 	const collection = params.collection!;
 	const id = params.id!;
 	const locale = url.searchParams.get("locale") || undefined;
-
-	if (!emdash?.handleContentGet) {
-		return apiError("NOT_CONFIGURED", "EmDash is not initialized", 500);
-	}
 
 	const result = await emdash.handleContentGet(collection, id, locale);
 
--- a/packages/core/src/astro/routes/api/content/[collection]/[id]/compare.ts
+++ b/packages/core/src/astro/routes/api/content/[collection]/[id]/compare.ts
@@ -13,14 +13,13 @@ export const prerender = false;
 
 export const GET: APIRoute = async ({ params, locals }) => {
 	const { emdash, user } = locals;
+	if (!emdash?.handleContentCompare) {
+		return apiError("NOT_CONFIGURED", "EmDash is not initialized", 500);
+	}
 	const denied = requirePerm(user, "content:read_drafts");
 	if (denied) return denied;
 	const collection = params.collection!;
 	const id = params.id!;
-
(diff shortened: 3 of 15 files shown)


ascorbic (approved):
Thanks!

Issue #1158, opened 2026-05-24
Bug Report: Plugin block edits do not persist after save


jimiryquai:
### Description

Summary
Editing plugin block settings (e.g. theme, language) in the admin editor does not persist. The first insert works, but subsequent edits revert to previously saved values on save/reload.

Environment
@emdash-cms/admin (bundled in dist/index.js)
Reproduced with a native-format plugin contributing a Portable Text block with Block Kit fields
Root Cause
Three bugs in ContentEditor combine to cause this:

Bug 1: handleFieldChange doesn't sync formDataRef or cancel autosave
handleFieldChange calls setFormData((prev) => ...) which is async. It never updates formDataRef.current, so a pending autosave timer reads stale data from the ref.

Bug 2: handleSubmit reads from stale closure
handleSubmit reads formData from the React closure instead of formDataRef.current. When the user saves quickly after editing a block, the closure still holds the old value.

Bug 3: handlePluginBlockInsert bypasses TipTap for edits
When editing an existing plugin block (as opposed to inserting a new one), handlePluginBlockInsert uses editor.view.dispatch(tr). This bypasses TipTap's dispatch pipeline, so the onUpdate callback never fires. The block data is updated in the ProseMirror document but ContentEditor never learns about it.

Proposed Fixes
Fix 1: handleFieldChange — sync ref + cancel pending autosave
diff
```javascript
const handleFieldChange = React$1.useCallback((name, value) => {
+    if (autosaveTimeoutRef.current) {
+        clearTimeout(autosaveTimeoutRef.current);
+        autosaveTimeoutRef.current = null;
+    }
-    setFormData((prev) => ({ ...prev, [name]: value }));
+    const next = { ...formDataRef.current, [name]: value };
+    formDataRef.current = next;
+    setFormData(next);
 }, [slugTouched]);
```

Rationale: Ensures formDataRef.current is always up to date before any autosave fires, and cancels any in-flight autosave that would snapshot stale state.

Fix 2: handleSubmit — read from ref, not closure
diff

```javascript
const handleSubmit = (e) => {
     e.preventDefault();
-    if (hasInvalidUrls(formData)) return;
+    if (hasInvalidUrls(formDataRef.current)) return;
     onSave?.({
-        data: formData,
-        slug: slug || void 0,
+        data: formDataRef.current,
+        slug: slugRef.current || void 0,
         bylines: activeBylines
     });
 };
```

[...]

PR #1119, opened 2026-05-20
fix: auto-save not detecting plugin block field changes

adentdk:
## What does this PR do?

Fixes auto-save not detecting changes when editing plugin block fields via the Block Kit modal. When a user edits an existing block's attributes through the modal, the change was silently lost unless they also made a separate text edit.

Closes #1118

## Type of change

- [x] Bug fix
- [ ] Feature (requires [maintainer-approved Discussion](https://github.com/emdash-cms/emdash/discussions/categories/ideas))
- [ ] Refactor (no behavior change)
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests
- [ ] Chore (dependencies, CI, tooling)

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md)
- [ ] `pnpm typecheck` passes
- [ ] `pnpm lint` passes
- [ ] `pnpm test` passes (or targeted tests for my change)
- [ ] `pnpm format` has been run
- [ ] I have added/updated tests for my changes (if applicable)
- [ ] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable). Do not include `messages.po` changes except in translation PRs — a workflow extracts catalogs on merge to `main`.
- [x] I have added a [changeset](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#changesets) (if this PR changes a published package)
- [ ] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/...

## AI-generated code disclosure

- [x] This PR includes AI-generated code — model/tool: Claude Opus 4.6 via OpenCode

## Root Cause

When editing an existing plugin block's attributes through the Block Kit modal, `handlePluginBlockInsert` in `PortableTextEditor.tsx` used `editor.view.dispatch(tr)` with `setNodeMarkup` to update the node. For atom nodes (`atom: true`) with attribute-only changes, this bypasses TipTap's command pipeline and does not reliably trigger the `onUpdate` callback. As a result:

1. `PortableTextEditor`'s `onChange` never fires
2. `ContentEditor`'s `formData` never updates
3. `serializeEditorState()` produces the same string as `lastSavedData`
4. Auto-save's `isDirty` stays `false`
5. The change is silently lost

## Fix

[...]

PR #2130, opened 2026-07-19
fix(core): CAS-guard the draft-revision pointer flip on concurrent saves

marcusbellamyshaw-cell:
## What does this PR do?

Fixes a save race reported as a follow-up comment on closed issue #1158. `EmDashRuntime.handleContentUpdate`'s draft-revision path does read-existing → merge → create-revision → flip-pointer with no wrapping transaction (D1 has no multi-statement transactions) and an unconditional pointer-flip `UPDATE`. Two concurrent saves (autosave racing an explicit save, or two explicit saves) that both read `draft_revision_id: null` before either writes both create a new revision and race to flip the pointer — whichever `UPDATE` lands last wins regardless of which read the fresher data, so the other save's edit is silently discarded even though the API reports success. This matches the DCathal repro in the #1158 thread (two revisions created in the same second, the stale one becomes `draft_revision_id`).

Note: the *original* #1158 report (TipTap `editor.view.dispatch` bypassing `onUpdate`) was already fixed by #1119 and is unaffected here — this addresses the separate, still-reproducing save race described in the newest comment on that thread (0.28.1/0.29.0).

Fix: the pointer-flip `UPDATE` is now a compare-and-set keyed on the `draft_revision_id` read at the start of the attempt (`COALESCE(draft_revision_id, '') = <value read>`, NULL-safe across SQLite/D1 and Postgres). If the CAS fails, the losing save retries — re-reading the now-current draft, re-merging its own edit on top, and retrying the create+CAS — instead of silently overwriting. Bounded to 3 attempts. The orphaned revision row from a lost race is harmless; `pruneOldRevisions` sweeps it up on a later successful save since it's keyed by `entry_id`, not by pointer.

Known residual, out of scope for this PR: the autosave-only "update existing draft revision in place" path (`skipRevision` when a draft already exists) still isn't CAS-guarded — two concurrent autosaves (no explicit save involved) can still last-write-wins on that single row's `data` column. Closing that fully would need a version column on `revisions`, which felt like more surface than this fix needs; flagging in case a maintainer disagrees.

Closes (comment thread on) #1158

## Type of change

- [x] Bug fix

## Checklist

- [x] `pnpm typecheck` passes (pre-existing unrelated errors in `oauth-state-store.ts` / `plugins/context.ts` confirmed present without this change)
- [x] `pnpm lint` passes (oxlint, scoped to changed files)
- [x] `pnpm test` passes (targeted: new race test + adjacent content/revision/media-usage integration suites, both sqlite and postgres dialects)
- [ ] `pnpm format` has been run (skipped locally — CRLF checkout makes `format:check` unreliable on Windows; will check CI)
- [x] I have added a changeset (patch, `emdash`)

## Test plan

Added `packages/core/tests/integration/database/content-update-save-race.test.ts`: creates a published entry, then fires two concurrent `handleContentUpdate` calls editing different fields, with a `RevisionRepository.create` spy that gates the first call until the second also arrives — forcing both to read `draft_revision_id: null` before either writes, regardless of driver timing. Verified the test fails against the pre-fix code (save B's edit silently dropped, matching the report) and passes with the fix (both edits survive).

### DIFF
--- a/.changeset/draft-revision-save-race.md
+++ b/.changeset/draft-revision-save-race.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes a race where two saves to the same draft (e.g. an autosave firing just before or after a manual save) could silently discard one of the edits. The losing save now retries against the latest draft instead of overwriting it.
--- a/.changeset/draft-revision-save-race.md
+++ b/.changeset/draft-revision-save-race.md
@@ -2,4 +2,4 @@
 "emdash": patch
 ---
 
-Fixes a race where two saves to the same draft (e.g. an autosave firing just before or after a manual save) could silently discard one of the edits. The losing save now retries against the latest draft instead of overwriting it.
+Fixes a race where two saves to the same draft (e.g. an autosave firing just before or after a manual save) could silently discard one of the edits. The losing save now retries against the latest draft instead of overwriting it; if the draft is still contested after retrying, the save now fails with a conflict error instead of silently dropping the edit.
--- a/packages/core/src/emdash-runtime.ts
+++ b/packages/core/src/emdash-runtime.ts
@@ -2906,7 +2906,7 @@ export class EmDashRuntime {
 						// comparison NULL-safe across both SQLite/D1 and Postgres without
 						// relying on dialect-specific NULL operators. No updated_at stamp:
 						// draft staging leaves live content untouched, so public "last
-						// modified" consumers must not see a change (#2143).
+						// modified" consumers must not see a change.
 						const pointerUpdate = await sql`
 							UPDATE ${sql.ref(tableName)}
 							SET draft_revision_id = ${revision.id}
@@ -2927,7 +2927,12 @@ export class EmDashRuntime {
 						// instead of silently discarding this save.
 						lostEveryCasAttempt = true;
 					}
-					if (lostEveryCasAttempt && !draftStorageChanged) {
+					if (
+						lostEveryCasAttempt &&
+						!draftStorageChanged &&
+						processedData &&
+						Object.keys(processedData).length > 0
+					) {
 						return {
 							success: false as const,
 							error: {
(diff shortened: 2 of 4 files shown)


marcusbellamyshaw-cell:
All checks green, no scope-gate applies (bug fix, save-race repro from #1158). Marking ready for review.

marcusbellamyshaw-cell:
Addressed all "needs fixing" items:

- CAS exhaustion no longer falls through to a metadata-only update: after 3 lost retries, the save now returns `DRAFT_SAVE_CONFLICT` instead of silently discarding the edit. Added a test that forces exhaustion (another save wins the pointer flip on every attempt) and asserts the conflict, not a false success.
- Removed the `#1158`/`#1119` references from the test file's header comment and describe title.
- Removed the accidental `pnpm-lock.yaml` entry for `demos/emdash-dev-main` (a local-only workspace member, git-excluded but still picked up by `pnpm install`).

Left the `handleContentRestore` follow-up (suggestion, not blocking) out of scope for this PR.

marcusbellamyshaw-cell:
Pushed a follow-up: my previous commit broke \`tests/integration/mcp/concurrency.test.ts\` in CI — 10 concurrent updates to the same item now legitimately exhaust 3 CAS attempts with zero backoff (all racers retry in lockstep and keep re-colliding on the same tick). That test only passed before because of the exact silent-drop bug this PR fixes — unlucky racers were succeeding with a metadata-only update instead of erroring.

Bumped \`MAX_ATTEMPTS\` to 8 and added jittered backoff between retries. Verified the concurrency test 5x in a row (it's timing-sensitive) plus a full core \`vitest run\` — the only remaining failures are pre-existing Windows-environment flakes (tar/vite-config/sqlite-file-persist), confirmed identical against a stash of this branch's changes.

marcusbellamyshaw-cell:
Superseded by #2349, merged to main. That PR added `repo.replaceDraftRevision()` — a full-row optimistic-concurrency CAS keyed on `version`/`live_revision_id`/`draft_revision_id` with the same read-merge-retry shape this PR used, but broader: it also CAS-guards the autosave in-place-update path this PR's description flagged as an out-of-scope residual. Closing without merging.

PR #2131, opened 2026-07-19
fix(core): invalidate edge cache on settings/menu writes (#2042)

### LINKED ISSUE #2042: Publishing content does not invalidate the Astro cache provider — edge-cached HTML stays stale for the full TTL
## Summary

On a Cloudflare deployment using Astro's cache provider (`cacheCloudflare()` from `@astrojs/cloudflare/cache` + `routeRules`), **publishing in the admin does not invalidate the cached HTML**. The rendered page keeps being served from the Workers Cache until its TTL lapses — no matter how long that TTL is.

EmDash's publish path invalidates its **own object cache** and runs the `content:afterPublish` plugin hooks, but never calls `cache.purge()` / `Astro.cache.invalidate()`. As far as I can tell there is no such call anywhere in the package, so the provider's `invalidate()` (which exists and works) is only ever reachable if the site wires a purge endpoint itself.

This is easy to get wrong precisely because everything *looks* wired: the provider tags every response (`Cache-Tag`), `getEmDashEntry`/`getEmDashCollection` hand back a `cacheHint`, and the natural reading is that publishing closes the loop. We set `maxAge: 604800` on our content pages on exactly that assumption — an editor changing the address in Site Settings would not have seen it on the contact page **for a week**.

## Versions

- `emdash` 0.29.0
- `astro` 7.0.7
- `@astrojs/cloudflare` 14.1.2
- `wrangler.jsonc`: `"cache": { "enabled": true, "cross_version_cache": false }`

## Reproduction (measured on a deployed worker)

`astro.config.mjs`:

```js
cache: { provider: cacheCloudflare() },
routeRules: {
  "/[slug]": { maxAge: 604800, swr: 2592000, tags: ["pages"] },
},
```

The page renders a singleton fetched with `getEmDashEntry(...)` and sets its hint via `Astro.cache?.set?.(cacheHint)`.

**1. Page is cached, entry is aging:**

```
cf-cache-status: HIT  age: 250
cf-cache-status: HIT  age: 255
```

**2. Publish that singleton in the admin** (content unchanged; `updated_at` → `2026-07-14T11:03:54Z`).

**3. The edge does not notice:**

```
cf-cache-status: HIT  age: 405  last-modified: Sun, 12 Jul 2026 20:24:59 GMT
cf-cache-status: HIT  age: 409  last-modified: Sun, 12 Jul 2026 20:24:59 GMT
cf-cache-status: HIT  age: 421  last-modified: Sun, 12 Jul 2026 20:24:59 GMT
```

`age` climbs straight through the publish, and the served document still carries the **pre-publish** `Last-Modified`.

**4. A manual tag purge does evict it** — so the tagging and the purge API are both fine, nothing calls them:

```
cache.purge({ tags: ["pages"] })   →  purged

cf-cache-status: MISS        last-modified: Tue, 14 Jul 2026 11:03:54 GMT
cf-cache-status: HIT  age: 2 last-modified: Tue, 14 Jul 2026 11:03:54 GMT
```

## What I'd expect

A publish (and unpublish / delete / scheduled publish) invalidates the edge-cache tags for the affected content, the same way it already invalidates the object cache.

The tags are all there to do it with — though they'd need to line up, which is worth a look on its own: `getEmDashEntry` hints tag with the **row's ULID**, while `getEmDashCollection` hints tag with the **collection slug**. So a purge keyed on the collection slug would miss every page that only ever queried the single entry (singletons, detail pages) — which is most of them.

## Two gaps a fix should probably cover

1. **Menus and site settings have no publish hook at all.** `getMenu()` / `getSiteSettings()` return no `cacheHint`, and the menu/settings write paths have nothing to hang an invalidation on — yet on a typical site the nav and the footer render on **every** page. Editing the navigation cannot invalidate anything today, by any mechanism.

2. **The docs point the other way.** [Deploy to Cloudflare → Workers Cache](https://docs.emdashcms.com/deployment/cloudflare/) explains that the cache sits in front of the Worker and that pages control their own caching, but never says that a publish will not reach it. Given `routeRules` invites long TTLs, that omission is what turns a cache into stale content. Even without a code change, a line saying "nothing invalidates rendered HTML on publish — the TTL is your only freshness control" would have saved us the week-long staleness.

Happy to send a PR if you can point me at where you'd want the invalidation to sit (runtime `handleContentPublish`, or a provider-agnostic hook near the object-cache invalidation).

swissky:
## What does this PR do?

Editing site settings or a menu never evicted rendered HTML from the Cloudflare Workers route cache. Pages went stale until their `maxAge` lapsed, even though the in-memory object cache was invalidated.

This PR closes the gap:

- Adds reserved route-cache tags `emdash:settings` and `emdash:menus` plus an `invalidateEdgeTag()` helper (`packages/core/src/astro/edge-cache-tags.ts`).
- The settings `POST` route and every state-changing menu route (create / update / delete / reorder / translate / item update / item delete) now purge their reserved tag on success — no-op when no cache provider is configured, same contract as the content write paths.
- Pages that render settings or menus are tagged so the purges reach them: `<EmDashHead>` adds `emdash:settings`, and menu widgets (`WidgetRenderer`) add `emdash:menus`.
- Sites calling `getMenu()` directly can opt a page into the same invalidation with `Astro.cache.set({ tags: [EDGE_TAG_MENUS] })` (both tags are now exported from `emdash`).
- The Cloudflare deployment docs now explain that a publish only evicts cached pages whose tags line up, and show how to tag content, menus, and settings.

Closes #2042

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/...

- [x] This PR includes AI-generated code — model/tool: Cursor + Kimi K3

### DIFF
--- a/packages/core/src/components/EmDashHead.astro
+++ b/packages/core/src/components/EmDashHead.astro
@@ -34,6 +34,7 @@ import {
 import { renderSiteIdentity } from "../page/site-identity.js";
 import { getPageRuntime } from "../page/index.js";
 import { getSiteSettings } from "../settings/index.js";
+import { EDGE_TAG_SETTINGS } from "../astro/edge-cache-tags.js";
 import { absolutizeMediaUrl } from "../page/absolute-url.js";
 import { getHreflangAlternates } from "../seo/hreflang.js";
 import { isI18nEnabled } from "../i18n/config.js";
@@ -67,6 +68,11 @@ if (runtime) {
 		runtime.collectPageMetadata(page),
 		runtime.collectPageFragments(page),
 	]);
+
+	// This page renders site settings, so it must be evicted when they change.
+	// Tag it with the reserved settings tag so the settings write path's purge
+	// reaches it (#2042). Merges with any tags the route already set.
+	Astro.cache.set({ tags: [EDGE_TAG_SETTINGS] });
 
 	// Site-level default OG image: applied per-page in base contributions
 	// rather than emitted unconditionally, so per-content images still win.
--- a/.changeset/edge-cache-invalidate-2042.md
+++ b/.changeset/edge-cache-invalidate-2042.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Closes the edge-cache gap where editing site settings or a menu never invalidated rendered HTML. Settings and menu write paths now purge reserved route-cache tags (`emdash:settings`, `emdash:menus`) on Cloudflare, and `<EmDashHead>` / menu widgets tag the pages they render so those purges reach them. Sites that call `getMenu()` directly can opt a page into the same invalidation with `Astro.cache.set({ tags: [EDGE_TAG_MENUS] })`. The Cloudflare deployment docs now explain that a publish only evicts cached pages whose tags line up, and show how to tag content, menus, and settings.
--- a/demos/cloudflare/src/layouts/Base.astro
+++ b/demos/cloudflare/src/layouts/Base.astro
@@ -1,5 +1,5 @@
 ---
-import { getMenu, getEmDashCollection, getSiteSettings } from "emdash";
+import { getMenu, getEmDashCollection, getSiteSettings, EDGE_TAG_MENUS } from "emdash";
 import {
 	WidgetArea,
 	EmDashHead,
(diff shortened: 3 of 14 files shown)


packages/core/src/components/EmDashHead.astro
```diff
@@ -68,6 +69,11 @@ if (runtime) {
 		runtime.collectPageFragments(page),
 	]);
 
+	// This page renders site settings, so it must be evicted when they change.
+	// Tag it with the reserved settings tag so the settings write path's purge
+	// reaches it (#2042). Merges with any tags the route already set.
+	Astro.cache.set({ tags: [EDGE_TAG_SETTINGS] });
```
ascorbic, line 75: This is risky to put inside a component, because depending on how it's used, there's a chance the headers will have already been sent.

ascorbic (commented):
I think middleware is probably the best place to set these

swissky:
Superseded by #2945, which covers the settings and menu write paths (plus taxonomies and widget areas) with per-name tags and cache hints from the read helpers — addressing the review feedback here about component-side tagging. Closing.

Issue #2040, opened 2026-07-14
Default D1 config: one canceled request permanently deadlocks all queries on the isolate (site-wide hang)


blazeITHugo:
### Summary

With the **default** D1 config (`d1({ binding: "DB" })`, sessions disabled), a single request canceled mid-query permanently deadlocks **every subsequent D1 query on that Worker isolate** — including EmDash's own per-request middleware — so the whole site hangs until the isolate is recycled or redeployed. Verified in `@emdash-cms/cloudflare` 0.27.0 and in the published 0.29.0 dist.

This caused two full production outages for us (2026-07-07 and 2026-07-14) before we found the mechanism.

### Root cause

1. In default mode all runtime queries go through one **module-scope Kysely singleton** (`getDb()` in the loader).
2. The default dialect (`createDialect` → `EmDashD1Dialect` → kysely-d1) uses kysely's `SqliteAdapter`, which reports `supportsMultipleConnections: false`.
3. Kysely's `RuntimeDriver` therefore serializes every query through a `ConnectionMutex`: `obtainLock()` → execute → `releaseConnection()` → `releaseLock()`.
4. On Workers, when a request is canceled while its query is in flight (visitor clicks away, client timeout), the pending I/O promise is **canceled without ever settling**. The `await` never returns, `releaseLock()` never runs.
5. Every later `obtainLock()` waiter awaits the dead promise forever. Since the CMS middleware itself queries D1 on every request, **all SSR requests on that isolate hang** from that moment on.

### Symptom signature (for anyone else searching)

- All SSR routes hang / time out; static assets stay fast.
- `wrangler tail --format json`: `outcome: "canceled"`, `cpuTime` ~1–2 ms, `exceptions: []` — nothing in the logs.
- Only a redeploy (or isolate recycling) recovers.

This is the same signature your docs describe for the `global_fetch_strictly_public` + sessions incompatibility (#1273), but it happens **without that flag and without sessions** — plain default config.

### Reproduction

1. Deploy any EmDash site on Workers + D1 with `database: d1({ binding: "DB" })` (no `session` option).
2. Have a page that does multi-second D1 work (e.g. a large collection drain on a cold isolate).
3. Request it and abort mid-flight: `curl --max-time 2 https://your-site/slow-page`.
4. Request any other page served by the same isolate → it hangs forever.

We reproduced this on demand in production; one `curl --max-time 2` was enough.

### Suggested fix

[...]

PR #2135, opened 2026-07-19
fix(cloudflare): reclaim stranded coalescing flush after request cancel (#1927)

### LINKED ISSUE #1927: CoalescingD1Connection: a cancelled request strands the shared flush, wedging all subsequent queries isolate-wide
### Description

The Cloudflare coalescing D1 dialect (CoalescingD1Dialect → CoalescingD1Driver → single shared CoalescingD1Connection per isolate) is not cancellation-safe on workerd. One cancelled request can permanently strand the shared flush machinery; every later query on that connection then returns a promise that never settles. The isolate stays wedged until eviction. Nothing throws, so nothing appears in worker logs.

This is the same hazard class src/utils/single-flight-cache.ts documents and defends against ("a cancelled request settles neither way… every later request that awaits that shared promise then hangs until the isolate is evicted") — the coalescing connection escaped that hardening.

1. #scheduleFlush() sets #flushScheduled = true and schedules #flush() via setTimeout(…, 0) — a timer created inside whichever request queued first.
2. If that request is cancelled before the timer fires (client disconnect / stop / reload), workerd drops its timer and continuations. The flush never runs and #flushScheduled stays true forever.
3. Every subsequent #scheduleFlush() returns early, so new queries buffer into #buffer and their promises neither resolve nor reject. A cancellation mid-batch() similarly strands all coalesced waiters.

**Potential fix**
Likely the same treatment the single-flight cache got: anchor the flush so a cancelled originator can't strand it (after()/waitUntil), and make #flushScheduled reclaimable (dead-timer/deadline detection) so waiters can never queue behind a flush that will never run. Happy to PR.

### Steps to reproduce

1. Route A: awaits any plugin kv/storage read (shared coalescing connection) per request.
2. Issue a request to route A and abort it immediately (curl --max-time 0.05), repeat under light concurrency until an abort lands between enqueue and flush.
3. All subsequent requests to any readDb-touching route on that isolate hang indefinitely; non-readDb routes keep responding.

### Environment

- **emdash**: 0.27.0 (`@emdash-cms/cloudflare` ^0.27.0)
- **Astro**: 6.4.8, `@astrojs/cloudflare` 13.7.0 (`output: 'server'`)
- **Platform**: Cloudflare Workers (workerd 1.20260708.1, wrangler 4.109.0), custom domain
- **compatibility_date**: 2026-04-23, flags: `["nodejs_compat"]`

### Logs / error output

```shell

```

swissky:
## What does this PR do?

Fixes a hang where a request cancelled mid-query on Cloudflare D1 could wedge the query-coalescing buffer. On workerd, a request that is cancelled after queueing a coalesced query drops its pending `setTimeout(0)`, so the flush it scheduled never fires and `#flushScheduled` stays `true` — every later query on the connection then awaits a flush that never runs. This is the same hazard class that `single-flight-cache.ts` already documents and defends against; the two coalescing connections had escaped that hardening.

The flush flag is now **reclaimable**: each schedule stamps a short deadline (1s — a live `setTimeout(0)` fires within a turn, so this only ever trips on a genuinely dropped timer), and if that deadline lapses with the flag still set, the next caller clears the stale flag and reschedules instead of queueing behind a dead timer. Applied to both the coalescing D1 connection and the Durable Object SQL connection (the latter is long-lived and its buffer can be shared across requests, so it is the more exposed of the two).

Context: this came up alongside #2040 (the default-config `ConnectionMutex` deadlock, fixed in #2125). As analysed in #1927, the coalescing connections are not the default path and are mostly short-lived/per-request, but the hazard is real and the fix is cheap defence-in-depth.

Closes #1927

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

- [x] This PR includes AI-generated code — model/tool: Cursor + Claude (Kimi K3)

## Screenshots / test output

New regression tests in `packages/cloudflare/tests/db/coalescing-d1.test.ts` (`cancellation-safe flush (#1927)`): one reclaims a stranded flush so a later query is not left hanging, one confirms a live flush within its deadline is not prematurely reclaimed. Full `tests/db` suite passes (165 tests).

### DIFF
--- a/.changeset/coalescing-flush-cancellation.md
+++ b/.changeset/coalescing-flush-cancellation.md
@@ -1,0 +1,5 @@
+---
+"@emdash-cms/cloudflare": patch
+---
+
+Fixes a hang where a request cancelled mid-query on Cloudflare D1 could wedge the query-coalescing buffer: the flush timer scheduled by the cancelled request was dropped with it, leaving the coalescing flag stuck so every later query on the connection waited on a flush that never ran. The coalescing D1 and Durable Object SQL connections now treat a flush left pending past a short deadline as stranded and reschedule it, so a cancelled request can no longer stall subsequent queries.
--- a/packages/cloudflare/src/db/coalescing-d1.ts
+++ b/packages/cloudflare/src/db/coalescing-d1.ts
@@ -69,6 +69,18 @@ export class CoalescingD1Connection implements DatabaseConnection {
 	#buffer: PendingQuery[] = [];
 	#flushScheduled = false;
 	/**
+	 * Time by which the currently-scheduled flush must have run. If it hasn't
+	 * (its timer was dropped by a cancelled owner), the next caller reclaims
+	 * the flag and reschedules. See #scheduleFlush.
+	 */
+	#flushDeadline = 0;
+	/**
+	 * How long a scheduled flush may go unfired before a later caller treats
+	 * it as stranded. Generous — a live setTimeout(0) fires within a turn —
+	 * so reclaim only ever trips on a genuinely dropped timer, never a slow one.
+	 */
+	static readonly #FLUSH_RECLAIM_MS = 1_000;
+	/**
 	 * Tail of a promise chain that serializes every physical call against the
 	 * shared D1DatabaseSession (direct-path statements and batch flushes
 	 * alike). See #enqueue.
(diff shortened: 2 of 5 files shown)


swissky:
@emdashbot both suggestions addressed in 6538d33:

**Structural parity (coalescing-do-sql.ts:145-147):** the DO SQL connection now extracts the stale-flush check into a `#reclaimStrandedFlush()` method mirroring `CoalescingD1Connection`, so a future edit updates one shape, not two divergent inlines.

**DO SQL regression coverage (coalescing-do-sql.test.ts):** added the same `cancellation-safe flush (#1927)` describe block to the DO SQL suite — one test reclaims a stranded flush so a later query is not left hanging (the stranded query stays buffered and coalesces with the reclaiming query into a single two-statement batch), one confirms a live flush within its deadline is not prematurely reclaimed.

Full `tests/db` suite: 167 passed. Typecheck + lint clean.

ascorbic (approved):
Nice

PR #2136, opened 2026-07-19
fix(core): trusted standard-format route handlers get the public single-arg RouteContext

### LINKED ISSUE #2079: Trusted-mode (plugins: []) route adapter calls standard-format handlers with the old two-arg signature, breaking the public single-arg RouteContext API
### Description

`adaptSandboxEntry` (`src/plugins/adapt-sandbox-entry.ts`) is used to run a standard-format plugin (one exporting `definePlugin({ hooks, routes })` as its default export) trusted/in-process via `plugins: [somePlugin()]`. When wrapping a plugin's route handler, it always invokes it with the **old two-argument** `SandboxedPlugin` calling convention:

```ts
// adapt-sandbox-entry.ts (~L219-225)
const routeCtx = { input: ctx.input, request: requestShape, requestMeta: ctx.requestMeta };
const { input: _, request: __, requestMeta: ___, ...pluginCtx } = ctx;
return handler(routeCtx, pluginCtx);
```

But the **public, exported, type-checked** route handler contract from `emdash`'s own types is single-argument:

```ts
// from `emdash`'s exported types (index-BXOCyLj0.d.mts)
interface PluginRoute<TInput = unknown> {
  handler: (ctx: RouteContext<TInput>) => Promise<unknown>;
}
```

A plugin author who writes `routes: { myRoute: { handler: async (ctx: RouteContext) => { ... ctx.storage.foo ... } } }` — which is what the exported `RouteContext`/`definePlugin` types require and what `tsc` accepts — gets `ctx = routeCtx` (the adapter's first positional argument) at runtime when run trusted, since JS silently drops the second argument for a function declared with one parameter. `routeCtx` only has `{ input, request, requestMeta }` — no `storage`, `email`, `kv`, `cron`, `users`, etc. — so any handler that touches those throws `Cannot read properties of undefined (reading '<whatever>')`.

This means **every standard-format plugin using the public single-arg `RouteContext` API is broken the moment it's run trusted (`plugins: []`)** if any route handler touches anything beyond `ctx.input`/`ctx.request`/`ctx.requestMeta`.

Related to (but distinct from) #2078 — that one is about sandboxed (`sandboxed: []`) route auth metadata never being populated for config-declared plugins. Between the two, there is currently no way to run a standard-format plugin with public routes that touch `ctx.storage`/`ctx.email`/etc. correctly: `sandboxed: []` 401s every non-admin route, and `plugins: []` throws on any route handler using more than the request-shaped fields.

### Steps to reproduce

1. Define a standard-format plugin with a route that reads `ctx.storage`:
   ```ts
   export default definePlugin({
     id: "demo",
     version: "1.0.0",
     storage: { things: { indexes: [] } },
     routes: {
       ping: {
         public: true,
         handler: async (ctx: RouteContext) => {
           const items = await ctx.storage.things!.query({ limit: 1 });
           return { ok: true, count: items.items.length };
         },
       },
     },
   });
   ```
2. Register it trusted: `emdash({ plugins: [demoPlugin()] })`.
3. `curl http://localhost:4321/_emdash/api/plugins/demo/ping`
4. Server log shows: `TypeError: Cannot read properties of undefined (reading 'query')` (or whichever field is accessed), thrown from inside the plugin's own handler at the line touching `ctx.storage`, because `ctx` at runtime is the adapter's `routeCtx` (missing everything but `input`/`request`/`requestMeta`), not the full `PluginContext`.

Found while trying to work around #2078 by running [emdash-plugin-engagement](https://github.com/marcusbellamyshaw-cell/emdash-plugin-engagement) trusted instead of sandboxed — its `leaderboard` route (`ctx.storage.points!.query(...)`) throws exactly this way.

### Environment

- emdash version: 0.29.0
- Node.js version: 24.15.0 (local dev)
- Runtime: Node (Astro dev server, `@astrojs/cloudflare` adapter)
- OS: Windows 11

### Logs / error output

```
[ERROR] [vite] [plugin:engagement] Route handler failed: TypeError: Cannot read properties of undefined (reading 'points')
    at handler (.../emdash-plugin-engagement/dist/sandbox-entry.js:301:44)
    at Object.handler (.../emdash/src/plugins/adapt-sandbox-entry.ts:225:13)
    at PluginRouteHandler.invoke (.../emdash/src/plugins/routes.ts:148:31)
```

dchaudhari7177:
## What does this PR do?

`adaptSandboxEntry` (the in-process adapter behind `plugins: []`) always invoked route handlers with the **sandboxed two-arg** `(routeCtx, pluginCtx)` convention. A standard-format plugin authored against the **public, exported, type-checked single-arg `RouteContext`** API silently received `routeCtx` as its only argument (JS drops the extra parameter for a one-param function), so `ctx.storage`, `ctx.email`, `ctx.kv` and every other capability surface were `undefined` — any route touching more than `input`/`request`/`requestMeta` threw `TypeError: Cannot read properties of undefined`.

The fix keys off the format signal the codebase already documents (`define-plugin.ts`): `definePlugin` **requires** `id`, and sandbox-format default exports **never** carry one (identity comes from the manifest's slug + publisher). When the definition has an `id`, the wrapper passes the incoming `RouteContext` through unchanged (full `PluginContext` + `input` + real WHATWG `Request` + `requestMeta`, exactly what the public contract promises). Sandbox-format entries keep the existing two-arg flattened convention, byte for byte.

This covers both trusted paths: config `plugins: [somePlugin()]` and marketplace bundles built by plugin-cli (which bundle the plugin's own `definePlugin` entry, so the `id` survives).

Closes #2079

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

## Screenshots / test output

New tests: a `definePlugin`-shaped definition's handler receives the full `RouteContext` as its **only** argument (same object identity, real `Request`, `storage` reachable); a sandbox-format definition (no `id`) still gets the flattened two-arg convention. Full plugin unit suite: **622/622 pass** (`tests/unit/plugins`, 30 files). Prettier clean.


ascorbic (approved):
Thanks

PR #2140, opened 2026-07-20
fix(secrets): read site secrets from process.env only, never import.meta.env

### LINKED ISSUE #2139: Secrets read via bare `import.meta.env` are inlined into production server bundles and shadow runtime Worker secrets (EMDASH_ENCRYPTION_KEY et al.)
### Description

`readDefaultEnv()` in the secrets module (`src/config/secrets.ts`, shipped as `dist/secrets-*.mjs`) reads `EMDASH_ENCRYPTION_KEY` — and five other secrets (`EMDASH_PREVIEW_SECRET`, `PREVIEW_SECRET`, `EMDASH_IP_SALT`, `EMDASH_AUTH_SECRET`, `AUTH_SECRET`) — through the bare `import.meta.env` object, with a `process.env` fallback:

```js
function readDefaultEnv() {
	const meta = import.meta.env ?? {};
	const proc = typeof process !== "undefined" && process.env ? process.env : {};
	return {
		EMDASH_ENCRYPTION_KEY: meta.EMDASH_ENCRYPTION_KEY ?? proc.EMDASH_ENCRYPTION_KEY,
		// ... same pattern for the other five
	};
}
```

In an Astro production build (`DEPLOY_TARGET=cloudflare` with `@astrojs/cloudflare`), Vite statically replaces the bare `import.meta.env` expression with an object literal built from the env loaded at build time. If `EMDASH_ENCRYPTION_KEY` is present in `.env` when the operator builds — which is exactly the state a machine that runs `emdash dev` is in — two things happen:

1. **The key is embedded in plaintext** in `dist/server/chunks/secrets-*.mjs` (and `@astrojs/cloudflare` additionally writes it to `dist/server/.dev.vars`). It ships to Cloudflare inside the Worker bundle and lingers in any build artifact.
2. **The inlined value takes precedence over the real runtime secret.** Because the chain is `meta.X ?? proc.X`, a key set properly via `wrangler secret put EMDASH_ENCRYPTION_KEY` is silently ignored — the deployed Worker encrypts plugin secrets under the developer's local dev key while the operator believes the production secret is in use.

The failure is completely silent: no warning, nothing in logs. The operator only finds out when they rotate "the" key, inspect key fingerprints, or grep their own bundle. It also quietly undermines the multi-key rotation story being reviewed in #1688 — the key actually in use isn't any of the ones the operator manages.

The docstring on `readDefaultEnv` says `import.meta.env` deliberately wins "so build-time substitutions are honored when present." That's the right call for public config, but for secrets it's inverted: a build-time substitution of a secret is precisely the case that should never win (and arguably should never exist).

**Expected:** runtime secrets configured on the deployment platform are the values actually used, and no secret value appears in build output.

### Steps to reproduce

1. Scaffold an EmDash project configured for Cloudflare (`@astrojs/cloudflare`, `DEPLOY_TARGET=cloudflare`).
2. Put a key in `.env` (throwaway generated for this report, never used anywhere):
   ```
   EMDASH_ENCRYPTION_KEY=emdash_enc_v1_3U363i_E4JWm5WsRrBPGHomuSYn6JZCEPIiwSHqKSKc
   ```
3. `DEPLOY_TARGET=cloudflare pnpm build`
4. `grep -rn "emdash_enc_v1_" dist/` → the full key value is present in `dist/server/chunks/secrets-*.mjs` and `dist/server/.dev.vars`.
5. Deploy that bundle with a *different* key set as a Worker secret → the Worker uses the inlined `.env` key, not the Worker secret (observable via the key fingerprint, or by the fact that values encrypted before/after deploy don't round-trip under the Worker secret).

### Environment

- emdash version: 0.28.1
- astro: 7.1.1, @astrojs/cloudflare: 14.1.3
- Node.js version: 22.22.3 (pnpm 10.6.5)
- Runtime: Cloudflare Workers (build host: macOS)

### Logs / error output

Transformed `readDefaultEnv` in the built server chunk (`dist/server/chunks/secrets-kp760LlD_*.mjs`) — note the inlined literal and that it wins over `process.env`:

```js
	}, { EMDASH_ENCRYPTION_KEY: "emdash_enc_v1_3U363i_E4JWm5WsRrBPGHomuSYn6JZCEPIiwSHqKSKc" }) ?? {};
	const proc = typeof process !== "undefined" && process.env ? process.env : {};
	return {
		EMDASH_ENCRYPTION_KEY: meta.EMDASH_ENCRYPTION_KEY ?? proc.EMDASH_ENCRYPTION_KEY,
		...
```

```
$ grep -rl "emdash_enc_v1_3U363i" dist/
dist/server/.dev.vars
dist/server/chunks/secrets-kp760LlD_WVSrxe7u.mjs
```

### Suggested fixes

In rough order of preference:

1. **Read secrets exclusively from `process.env` at runtime.** Never touch `import.meta.env` for the six secret fields — then nothing can be inlined. This also covers the CLI surface the docstring mentions: outside the bundle `import.meta.env` is empty anyway, so `process.env` is already the effective source there. On Workers, `process.env` carries Worker secrets under `nodejs_compat`.
2. **Or use `astro:env` server secrets** (`envField.string({ context: "server", access: "secret" })`), which Astro guarantees are runtime-only and never statically replaced.
3. **At minimum, invert the precedence for secrets** (`proc.X ?? meta.X`) so a real runtime secret always beats a baked-in build-time value. This fixes the silent-wrong-key problem but still leaves the plaintext copy in the bundle, so (1) or (2) is preferable.

A related nicety: since keys are prefixed (`emdash_enc_v1_`), a post-build check that greps the output for the prefix pattern would make this class of leak fail loudly — we run exactly that as a local guard and it's cheap.

Happy to PR whichever direction you prefer.

dchaudhari7177:
## What does this PR do?

In an Astro production build, Vite statically replaces the bare `import.meta.env` expression in `readDefaultEnv()` (`packages/core/src/config/secrets.ts`) with an object literal built from the env loaded at build time. If `EMDASH_ENCRYPTION_KEY` (or any of the other five secrets the module reads) is present in `.env` when the operator builds:

1. the secret is **embedded in plaintext** in `dist/server/chunks/secrets-*.mjs` and ships inside the Worker bundle, and
2. because the chain was `meta.X ?? proc.X`, the inlined build-time value **silently shadowed** the real runtime secret set via `wrangler secret put` — the deployed Worker encrypted plugin secrets under the developer's local dev key.

This PR makes the secrets module read **only** `process.env` (the same runtime source `resolveS3Config` already uses, available on Node/container deployments and on Cloudflare Workers with Node compatibility). With no `import.meta.env` reference left in the module, Vite has nothing to substitute, so no secret can be embedded in build output, and runtime-configured secrets are always the values actually used.

Also included:

- a regression test that `resolveSecrets({ db })` picks up `process.env` when no explicit env map is passed
- a source-hygiene test asserting the module never references `import.meta.env` in code (comments stripped), so the inlining vector can't be silently reintroduced
- updated docstrings (`readDefaultEnv`, `ResolveSecretsOptions.env`) documenting why secrets are excluded from the `import.meta.env` convention
- a changeset (patch, `emdash`) with operator guidance to rotate any key that shipped inside a bundle

Closes #2139

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] User-visible strings in the admin UI are wrapped for translation (not applicable)
- [ ] New features link to an approved Discussion (not applicable)

- [x] This PR includes AI-generated code — model/tool: Claude Code

## Screenshots / test output

```
Test Files  1 passed (1)
     Tests  33 passed (33)
```

(`packages/core/tests/unit/config/secrets.test.ts` — 31 existing + 2 new; `tsc --noEmit` and `oxlint` clean on the touched files.)

### DIFF
--- a/.changeset/secrets-no-buildtime-inlining.md
+++ b/.changeset/secrets-no-buildtime-inlining.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Security: site secrets (`EMDASH_ENCRYPTION_KEY`, `EMDASH_PREVIEW_SECRET`, `EMDASH_IP_SALT`, and their legacy aliases) are no longer read through `import.meta.env`. In production builds Vite statically replaced that expression with the env loaded at build time, which embedded any secret present in `.env` on the build machine into the server bundle and made the stale build-time value silently shadow the real runtime secret (e.g. one set with `wrangler secret put`). The secrets module now reads only the runtime `process.env`. If you previously relied on a secret being baked in at build time, set it as a runtime secret/env var on your deployment platform instead — and rotate any key that shipped inside a bundle.
--- a/packages/core/src/config/secrets.ts
+++ b/packages/core/src/config/secrets.ts
@@ -108,8 +108,9 @@ export interface ResolveSecretsOptions {
 	db: Kysely<Database>;
 	/**
 	 * Optional explicit env override map. When omitted, falls back to
-	 * `import.meta.env` via the global accessor below. Tests pass an
-	 * explicit map to avoid leaking process state.
+	 * `process.env` via `readDefaultEnv` below (never `import.meta.env` —
+	 * see that function's docstring). Tests pass an explicit map to avoid
+	 * leaking process state.
 	 */
 	env?: SecretsEnv;
 	/**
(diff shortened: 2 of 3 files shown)


ascorbic (approved):
Good call. Thanks

PR #2156, opened 2026-07-20
feat(core): built-in SMTP email transport for generic SMTP credentials

### LINKED ISSUE #1541: Feature request: built-in SMTP email transport in core
# Feature request: built-in SMTP email transport in Emdash core

## Summary

Provide a first-class **SMTP email transport in Emdash core** so operators can
send mail through any standard SMTP server (host / port / username / password)
instead of only through HTTP transactional APIs. A plugin cannot do this, by
design — only the host runtime can open the TCP socket SMTP requires.

## Motivation

Today, sending email from Emdash on Cloudflare Workers means using a provider
with an **HTTP API** (e.g. the `emdash-plugin-brevo` plugin, which calls
`POST https://api.brevo.com/v3/smtp/email`). That works well, but a large class
of operators only have **generic SMTP credentials**:

- Brevo's own **SMTP relay** (`smtp-relay.brevo.com:587`, authenticated with an
  `xsmtpsib-` SMTP key) — distinct from its HTTP API key (`xkeysib-`).
- Office365 / Google Workspace / Fastmail / Amazon SES SMTP endpoints.
- Self-hosted Postfix and corporate mail relays.

None of these can be reached from a plugin.

## Why this can't be a plugin

Emdash sandboxed plugins are given exactly **one** network primitive —
`ctx.http.fetch()` — which speaks **HTTP/HTTPS only** and is gated by the
plugin's `allowedHosts` allowlist. There is no TCP/socket capability in the
plugin capability vocabulary. SMTP is a raw TCP protocol (EHLO / STARTTLS /
AUTH / MAIL FROM / RCPT TO / DATA), so a plugin physically cannot open an SMTP
connection.

Cloudflare Workers *can* open TCP sockets via `connect()` from
`cloudflare:sockets`, **but only host code can** — that module is not exposed to
the sandbox. Therefore an SMTP transport has to live in core (or in a
host-trusted/native extension point), not in a sandboxed plugin.

## Proposal

Add an optional core email transport, `smtp`, that registers as the email
delivery provider so existing `ctx.email.send()` callers and the
`email:deliver` hook contract keep working unchanged.

### Configuration

Via env/wrangler vars (secrets) and/or the admin Email settings page:

| Key | Example | Notes |
|---|---|---|
| `EMAIL_SMTP_HOST` | `smtp-relay.brevo.com` | |
| `EMAIL_SMTP_PORT` | `587` | 587 (STARTTLS) or 465 (implicit TLS) |
| `EMAIL_SMTP_SECURE` | `starttls` \| `tls` | |
| `EMAIL_SMTP_USER` | `you@example.com` | |
| `EMAIL_SMTP_PASS` | `xsmtpsib-…` | stored as a secret |
| `EMAIL_SMTP_FROM` | `Site <noreply@example.com>` | default sender |

### Implementation sketch

- Use `connect(\`${host}:${port}\`, { secureTransport: port === 465 ? "on" : "starttls", allowHalfOpen: false })` from `cloudflare:sockets`.
- Minimal SMTP client: read greeting → `EHLO` → (`STARTTLS` + `socket.startTls()` when `secure=starttls`) → `AUTH LOGIN`/`PLAIN` → `MAIL FROM` → `RCPT TO` → `DATA` → dot-stuffed MIME body → `QUIT`.
- Build a compliant MIME message (headers, `text/plain` + `text/html` multipart/alternative, quoted-printable/base64, message-id, date).
- Enforce a connect/overall timeout; surface SMTP reply codes as structured errors.

### Caveats to call out in the implementation

- **Cloudflare blocks outbound port 25**; this transport targets submission
  ports **587/465** only, and should validate/refuse 25 with a clear message.
- TLS must be required (STARTTLS upgrade verified, or implicit TLS on 465);
  never fall back to plaintext auth on an unencrypted channel.
- Some hosts (corporate relays) may not be reachable from Cloudflare egress IPs;
  document this alongside the existing dynamic-egress-IP guidance.

## Alternatives considered

1. **HTTP→SMTP gateway service** — works from a plugin (it's just HTTP), but adds
   a third party in the mail path and means handing SMTP credentials to that
   service. Rejected as a default.
2. **Per-plugin SMTP** — impossible (no TCP capability in the sandbox), which is
   the reason for this request.
3. **Status quo (HTTP transactional APIs only)** — fine for Brevo/SendGrid/etc.
   API users, but excludes everyone who only has SMTP credentials.

## Backward compatibility

Additive. Existing HTTP-API providers (including the Brevo plugin) continue to
work; `smtp` is just another selectable transport. If both a core SMTP transport
and an exclusive `email:deliver` plugin are configured, document the precedence
(suggest: an explicitly-configured plugin transport wins, else core SMTP).

## Acceptance criteria

- [ ] `ctx.email.send()` delivers via SMTP when the `smtp` transport is configured.
- [ ] STARTTLS (587) and implicit TLS (465) both supported; port 25 refused.
- [ ] AUTH LOGIN and PLAIN supported; credentials stored as secrets, never logged.
- [ ] Connect + overall timeouts; SMTP error codes surfaced to callers.
- [ ] Docs: setup guide + the Cloudflare port-25 / egress-IP caveats.

swissky:
## What does this PR do?

Adds a built-in `email:deliver` transport that works with any standard SMTP server (Brevo relay, Office365, Google Workspace, Fastmail, Amazon SES, self-hosted Postfix) via raw TCP — the one network primitive sandboxed plugins cannot use. This unblocks operators who only have generic SMTP credentials and cannot use HTTP transactional APIs (Brevo HTTP API, SendGrid, etc.).

**How it works:**
- Registered as a built-in plugin (same pattern as the dev console email provider), participating in exclusive hook resolution. Explicitly selected plugin transports still take precedence.
- Configuration via env vars (`EMAIL_SMTP_HOST`, `EMAIL_SMTP_PORT`, `EMAIL_SMTP_USER`, `EMAIL_SMTP_PASS`, optional `EMAIL_SMTP_FROM`) or via the new Settings → Email admin UI (password stored encrypted).
- Supports STARTTLS (port 587) and implicit TLS (port 465). Port 25 is refused with a clear error because Cloudflare blocks it. TLS is always required; plaintext auth is never attempted.
- Runs on Cloudflare Workers via `cloudflare:sockets` and on Node via `node:net`/`node:tls`, with automatic fallback.

**Cloudflare Workers specifics fixed:**
- `WritableStream` lock: the plaintext socket's writer must be closed before `startTls()` upgrades the socket, otherwise the first write to the TLS socket throws "WritableStream is currently locked".
- `sock.opened` must be awaited before writing — unlike Node, where the connect callback signals readiness.
- `secureTransport: "starttls"` is required during the initial connection to later allow `sock.startTls()`.
- SMTP timeout (25s) must be shorter than the hook timeout (30s) and raced as a rejecting promise — a `throw` inside `setTimeout` never reaches the awaiting promise and becomes an unhandled exception.

**Admin UI:**
- Provider selection (None / SMTP / Cloudflare Email) with an explicit `__none__` sentinel for exclusive hooks.
- SMTP form with host, port, security, username, password (encrypted), from name/email, reply-to.
- Test email button that surfaces the underlying SMTP error (e.g. "535 Authentication failed") instead of a bare 500.
- Cloudflare Email "Test Binding" button that checks the `send_email` binding.
- Port change auto-suggests the matching security mode (WP Mail SMTP's autotls pattern).

Closes #1541

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/2157

- [x] This PR includes AI-generated code — model/tool: Kimi K3 (Moonshot AI)

## Screenshots / test output

Unit tests: `packages/core/tests/unit/plugins/email-smtp.test.ts` — 21 tests covering env config parsing, port-25 refusal, full STARTTLS session flow, AUTH failure, dot-stuffing, the plugin handler contract, and Cloudflare-specific regressions (timeout race, STARTTLS upgrade flow).

```
Test Files  1 passed (1)
     Tests  21 passed (21)
```

Live verification on Cloudflare Workers (Brevo SMTP, port 465 implicit TLS): test email delivered successfully.


swissky:
Rebased onto current main (including the restyled settings layout) and reworked based on a full self-review:

**Security hardening**
- AUTH credentials are redacted from the SMTP transcript; the transcript goes to server logs only, and thrown errors (`SmtpDeliveryError`) carry only humanized, credential-free messages. Unknown errors surface a generic message plus the OS error code.
- Envelope addresses are validated against CRLF/SMTP command injection before `MAIL FROM`/`RCPT TO`; the `To:` header is sanitized.
- Bodies are base64-encoded and wrapped at 76 columns (dot-stuffing removed — base64 never emits a leading dot).
- Port 25 is refused at delivery time regardless of config source (env, DB, or future callers).

**Behavior fixes**
- Provider auto-selection: hooks gained an `autoSelect` flag; unconfigured built-ins register with `autoSelect: false`, so a site's single real email plugin keeps being auto-selected. The dev console provider is treated as ephemeral — its selection is never persisted and a stale persisted value (e.g. from `wrangler dev --remote` against a shared D1) never wedges email off in production.
- The provider dropdown now lists installed email plugins and re-saves a plugin selection verbatim (new `plugin` variant in the settings PUT) instead of misreporting it as "None".
- Cloudflare Email config is loaded lazily per send (settings apply without a restart; no cold-start query), batched into one query.
- `EMAIL_SMTP_*` env vars work as a fallback even when `EMDASH_ENCRYPTION_KEY` is set, and stored passwords support key rotation (comma-separated key list).
- Sockets complete pending reads on mid-session errors instead of hanging until the timeout.

**Docs & tests**
- Email docs intro rewritten for the three provider options; built-in Cloudflare env vars and the full SMTP env var table documented; Workers example uses port 465.
- New route tests (keep-password path, none sentinel, plugin selection, authz), admin form tests, and regression tests for redaction, injection, port 25, ephemeral selection, and no-restart config freshness.

One known divergence left as-is for maintainer input: the generic `PUT /api/admin/hooks/exclusive/:hookName` route still clears a selection by deleting the option, while the email settings route stores the `__none__` sentinel so "off" survives restarts. Happy to align the generic route on the sentinel in a follow-up if wanted.

swissky:
Pushed 8543c82:

- **CI root cause** (Smoke, Measure, both E2E shards): the literal `await import("cloudflare:sockets")` in `email-smtp.ts` made every Node-adapter `astro build` fail ("Rolldown failed to resolve import"). The transport now imports `connect` from a new `virtual:emdash/sockets` module that the integration generates per adapter, same pattern as `virtual:emdash/env`: re-export from `cloudflare:sockets` under `@astrojs/cloudflare`, `undefined` elsewhere (falls through to `node:net`/`node:tls`). Smoke suite passes locally on all templates, Node and Cloudflare.
- `encodeHeader()` splits non-ASCII headers into RFC 2047 encoded-words of ≤75 octets. Chunked by character rather than by base64 offset so a multi-byte character is never split across words (RFC 2047 §5). Test decodes each word independently with a fatal UTF-8 decoder.
- Remaining SMTP/Cloudflare form placeholders wrapped in `t`.
- Stale `{@link createSmtpPlugin}` JSDoc replaced; duplicated "Configure via environment variables:" sentence removed from the Cloudflare guide.

PR #2161, opened 2026-07-21
fix(admin): echo _rev from content reads into editor saves so stale saves 409

### LINKED ISSUE #2121: Admin editor can silently overwrite a newer draft: PUTs omit _rev and editor loads can miss the latest draft revision (0.29.0)

dchaudhari7177:
## What does this PR do?

The admin editor can silently overwrite a newer draft revision (#2121): its content PUTs never send `_rev`, so the server-side optimistic concurrency check (`validateRev`) never engages — "No `_rev` = blind write". Combined with the stale-read window described in the issue, a reload inside that window shows old content, and saving repoints `draft_revision_id` at a revision built from stale state, orphaning the newer draft.

This implements the issue's first suggested fix at the API-client level, so no caller has to thread tokens around:

- `fetchContent` captures the `_rev` the server already returns alongside `item`
- `updateContent` echoes the latest known token for that id into the PUT body — including `skipRevision` autosaves, so autosave can't stomp a newer draft either — and refreshes the token from the PUT response, keeping sequential saves valid
- on a **409** the stale token is dropped, so the editor's refetch re-arms the check instead of failing every later save
- a save with no known token (no prior read in this session) keeps today's blind-write behavior, per the server's backwards-compatibility rule

The issue's second suggestion (atomic draft hydration in `ContentRepository`) is intentionally left for a follow-up — this change makes the stale window harmless (the save is refused) rather than closing the window itself.

Closes #2121

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes (for the changed files — `pnpm lint` currently fails on my machine in `packages/registry-verification/src/records.ts`, which fails identically on a clean `main` checkout)
- [x] `pnpm test` passes (targeted: `packages/admin tests/lib/content-rev-echo.test.ts`, 4/4 in the browser-mode harness)
- [ ] User-visible strings in the admin UI are wrapped for translation (n/a — no UI strings)
- [x] I have added a [changeset](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#changesets) (`@emdash-cms/admin` patch)

## Tests

`packages/admin/tests/lib/content-rev-echo.test.ts`:
- GET captures `_rev`; the next PUT body carries it; the PUT response's fresh token replaces it
- no known token → PUT body unchanged (backwards-compatible)
- 409 drops the stale token; a refetch re-arms
- autosave-style (`skipRevision`) updates carry the token too

### DIFF
--- a/.changeset/admin-editor-rev-echo.md
+++ b/.changeset/admin-editor-rev-echo.md
@@ -1,0 +1,5 @@
+---
+"@emdash-cms/admin": patch
+---
+
+The admin content editor now echoes the `_rev` token from content API reads into its update requests, so the server's optimistic concurrency check can refuse a stale save with a 409 instead of silently overwriting a newer draft revision. On a conflict the stale token is dropped so the editor's refetch re-arms the check. Saves made before any read (no known token) still behave as before.
--- a/packages/admin/src/lib/api/content.ts
+++ b/packages/admin/src/lib/api/content.ts
@@ -203,6 +203,23 @@ export async function fetchContentAuthors(collection: string): Promise<ContentAu
 }
 
 /**
+ * Latest known `_rev` token per content id, captured from GET/PUT responses.
+ * `updateContent` echoes it into PUT bodies so the server's optimistic
+ * concurrency check (`validateRev`) can refuse stale saves instead of
+ * blind-writing over a newer draft revision (#2121). Centralized here so
+ * callers don't have to thread the token through.
+ */
+const knownRevs = new Map<string, string>();
+
+/** Test-only helpers for the `_rev` cache. */
+export function _getKnownRev(id: string): string | undefined {
+	return knownRevs.get(id);
+}
+export function _clearKnownRevs(): void {
+	knownRevs.clear();
+}
+
+/**
  * Fetch single content item
  */
 export async function fetchContent(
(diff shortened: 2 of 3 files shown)


ascorbic:
Hi. I'm closing it because there's an earlier PR (#2126) that covers some extra cases. Thanks for the contribution.

PR #1985, opened 2026-07-12
feat(plugins): opt-in Cache-Control for public plugin routes

swissky:
## What does this PR do?

Adds an opt-in `cacheControl` option for plugin routes. Public routes that set it have their successful GET/HEAD responses served with that `Cache-Control` value instead of the API default `private, no-store` — enabling CDN and browser caching for public endpoints that serve identical data to every visitor (catalog listings, public search, widget feeds).

```ts
routes: {
	catalog: {
		public: true,
		cacheControl: "public, max-age=60, stale-while-revalidate=300",
		handler: async (ctx) => listProducts(ctx),
	},
},
```

Guardrails so this can't leak private data:

- `getRouteMeta` only ever exposes `cacheControl` for routes that are also `public: true` — setting it on a private route has no effect, authenticated responses always keep `private, no-store`.
- The header is applied only to **successful** GET/HEAD responses; errors and other methods keep the default.
- Additive and default-off: existing routes are untouched.

Discussion: https://github.com/emdash-cms/emdash/discussions/1984

## Type of change

- [ ] Bug fix
- [x] Feature (requires [maintainer-approved Discussion](https://github.com/emdash-cms/emdash/discussions/categories/ideas))
- [ ] Refactor (no behavior change)
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests
- [ ] Chore (dependencies, CI, tooling)

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md)
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (or targeted tests for my change)
- [x] `pnpm format` has been run
- [x] I have added/updated tests for my changes (if applicable)
- [x] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable). Do not include `messages.po` changes except in translation PRs — a workflow extracts catalogs on merge to `main`. (n/a — no admin UI changes)
- [x] I have added a [changeset](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#changesets) (if this PR changes a published package)
- [x] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/1984

## AI-generated code disclosure

[...]

PR #2162, opened 2026-07-21
feat(plugins): return a Response from a route handler to serve non-JSON content

### LINKED ISSUE #2110: Plugin API routes can't return non-JSON responses (e.g. images)
### Description

Plugin API route handlers can only return JSON. Whatever a handler returns is wrapped in the standard `{ success, data }` response, so there's no way to serve an image, a file, or anything with a custom content type from a plugin route.

I ran into this building a plugin whose route returns an image: there's no way to actually send the image back to the caller. Returning a `Response` from the handler doesn't help either.

**Expected:** a plugin route should be able to return non-JSON content such as an image.

### Steps to reproduce

1. Add a plugin route whose handler returns an image:

   ```ts
   routes: {
     image: {
       handler: async () =>
         new Response(pngBytes, { headers: { "Content-Type": "image/png" } }),
     },
   }
   ```

2. Call the route: `GET /_emdash/api/plugins/<plugin>/image`.
3. The image isn't returned

### Environment

- emdash version: v0.29.0
- Node.js version: 22.16.0
- Runtime: Cloudflare Workers

### Logs / error output

```shell

```

dchaudhari7177:
## What does this PR do?

Plugin API route handlers could only return JSON — whatever a handler returned was wrapped in the standard `{ success, data }` envelope, so there was no way to serve an image, a file, or anything with a custom content type from a plugin route. Returning a `Response` didn't help either; it was JSON-serialized.

A handler that returns a `Response` now has it passed through verbatim:

- `PluginRouteHandler.invoke` detects `result instanceof Response` and carries it on `RouteResult.response` with the Response's own status
- the plugin API endpoint (`/_emdash/api/plugins/{pluginId}/{...path}`) returns that Response directly instead of `apiSuccess()`-wrapping it
- ordinary return values are unchanged — they still get the `{ success, data }` envelope
- passthrough is **trusted (configured) plugins only**; sandboxed plugin results cross a serialization boundary in `handleSandboxedRoute`, so they stay JSON-only (documented on the type and in the changeset)

The exact repro from the issue now works:

```ts
routes: {
  image: {
    handler: async () =>
      new Response(pngBytes, { headers: { "Content-Type": "image/png" } }),
  },
}
```

Closes #2110

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

(Filed as a bug — an image route currently has no way to return its image. The change is additive and backwards-compatible; happy to re-file under a Discussion if you'd rather treat it as a feature.)

## Checklist

- [x] `pnpm typecheck` passes (`packages/core`)
- [x] `pnpm lint` passes for the changed files (`pnpm lint` overall currently fails on a pre-existing error in `packages/registry-verification/src/records.ts` that reproduces on a clean `main`)
- [x] `pnpm test` passes (targeted: `packages/core tests/unit/plugins/routes.test.ts`, 39/39)
- [ ] User-visible strings wrapped for translation (n/a — no UI strings)
- [x] I have added a [changeset](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#changesets) (`emdash` patch)
- [ ] New features link to an approved Discussion (see note above)

## Tests

Three new cases in `packages/core/tests/unit/plugins/routes.test.ts`:
- an `image/png` Response is returned verbatim (`result.response` is the Response, `result.data` undefined, body bytes and content-type intact)
- a non-200 Response (302 + `Location`) carries its status through to the result
- ordinary return values still JSON-wrap (`result.data` set, `result.response` undefined)


dchaudhari7177:
Rebased onto current `main`, which was 321 commits ahead. The earlier red Typecheck/Tests runs were from 2026-08-14 and failed in files this PR doesn't touch (`content.ts`, the MCP drafts/search tests).

The rebase needed one real decision. #1985 has since added opt-in `cacheControl` for public plugin routes, and a Response returned verbatim would have bypassed it: a **private** route's handler could send `Cache-Control: public, max-age=600` and get authenticated content cached. The passed-through Response now follows the envelope's caching rule:

| route | handler's Response | `Cache-Control` sent |
|---|---|---|
| private | anything, including `public, max-age=600` | `private, no-store` |
| public, no `cacheControl` | none | `private, no-store` (same default as the envelope) |
| public, `cacheControl` set, GET/HEAD | none | the route's `cacheControl` |
| public, `cacheControl` set, GET/HEAD | its own header | the handler's header |
| public, `cacheControl` set, POST | none | `private, no-store` |

The Response is copied (`new Response(body, response)`) so its headers are mutable. Body, status and other headers are unchanged. The `handler` JSDoc and the changeset describe the rule.

**Tests.** `tests/unit/astro/plugin-api-route-passthrough.test.ts` (new) drives the catch-all endpoint for each row above, plus body, content type and a 302 with `Location`. Against `main`'s route file, 4 of its 7 tests fail; the 3 that pass assert defaults `main` already produces. Against a plain `return passthrough`, 4 fail. The passthrough, cache, auth and public-route unit files plus `routes.test.ts`: 77 passed. `tsc --noEmit` on `packages/core` is clean, and `oxfmt` is applied to the touched files. The commit author is also corrected to my GitHub account.

ascorbic:
Thanks. I'm closing this as it's been superseded by [#3190](https://github.com/emdash-cms/emdash/pull/3190).

Issue #19, opened 2026-04-01
April fools or real?


SourceCodeplz:
Is this real or April fools?

dandean:
Or a one-shotted Claude Code abandonware project? It's impossible to tell, these days!

mikedotmx:
> Is this real or April fools?

Yes.

alexasomba:
Real, we all knew this had to be done and saw it coming. Just didn't expect it on April Fool's day.

PR #2166, opened 2026-07-21
fix(labeler): kill-switch on in-flight runs, durable dead-letter retry, pointer tie-break (held review findings)

ascorbic:
## What does this PR do?

Closes three held findings from the adversarial review of the labeler (RFC #694, umbrella #1909) — parked during the earlier fix PRs because they overlapped files those PRs were rewriting, now addressed off the merged tip. Each is TDD'd and the set passed an independent adversarial review. Targets the `feat/plugin-registry-labelling-service` integration branch.

**1. `fix(labeler)`: enforce the automation kill-switch on in-flight runs (#14).** Automation pause was only checked before discovery; an already-dispatched Workflow never re-read `automation_state` and finalization had no pause predicate, so an admin could pause automation, get a success response, and in-flight runs kept issuing automated labels. Now enforced at two points: the Workflow entry (`executeAssessmentInstance` re-reads `isAutomationPaused` after the terminal short-circuit and throws `AutomationPausedError` before spending AI/network work) and, fail-closed, in the finalization CAS — a `guardAutomationNotPaused` predicate (`EXISTS(automation_state WHERE paused=0)`) AND'd onto the same batch as the signing and subject-not-deleted guards, so a mid-flight pause no-ops the whole batch and leaves the run `running` for the Workflow retry (halt-and-leave-resumable, matching the signing-pause precedent). Unknown/unreadable pause state fails closed (no issuance).

**2. `fix(labeler)`: make dead-letter retry durable before marking `retried` (#9).** The dead-letter retry handler committed `status='retried'` in the mutation batch, then decoded the payload and enqueued in a deferred tail — so a failed or undecodable delivery stranded the row `retried`: health counts only `new`, no sweep re-drives `retried`, and a later operator retry hit 409. The decode + enqueue now happen **before** `commitMutation`: an undecodable payload returns 422 (permanent) and an enqueue failure returns 503 (transient), both leaving the row `new`/actionable; a `retried` row now implies the job was durably accepted. Verified replay-safe (idempotency/audit rows are written only at commit, the guard's replay branch short-circuits before re-enqueue) and double-retry-safe (the consumer's `runKey` dedup absorbs a duplicate).

**3. `fix(labeler)`: tie-break the current-pointer update on assessment id (#19).** The current-assessment pointer upsert allowed overwrite at any equal `created_at_epoch_ms`, but list/current ordering is `(created_at_epoch_ms DESC, id DESC)` — so a later lower-id run sharing a millisecond could displace a higher-id current. The upsert guard now requires `excluded.epoch > current.epoch OR (equal epoch AND excluded.id >= current.id)`, matching the canonical ordering (`>=` on the unique ULID id keeps idempotent same-run replay).

Known follow-up (not introduced here): a pause longer than the Workflow retry window strands the in-flight run `running` until an operator re-drives it — the fix is correctly fail-closed (no labels issued), and automatic resumption of a long-paused run folds into the existing "reconciliation should re-dispatch stuck runs" hardening item.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm test` passes (labeler: 907 + 136 + 39) — TDD: each fix has a pre-fix failing repro (#14 finalized `passed` with labels under a pause; #9 returned 200 and flipped the row to `retried` on a failed delivery; #19 a lower-id equal-epoch run displaced the higher-id current)
- [x] New features link to an approved Discussion — **n/a**: bug fixes; umbrella #1909 tracks RFC #694.

- [x] This PR includes AI-generated code — model/tool: **Claude Opus 4.8** (implementation + independent adversarial review), orchestrated by **Claude Opus 4.8**

## Screenshots / test output

```
labeler: 907 (workerd) + 136 (console) + 39 (calibration) passed
adversary: PASS on all three (kill-switch composes fail-closed with the finalization CAS; dead-letter reorder replay/dedup-safe; tie-break matches canonical ordering)
```

### DIFF
--- a/apps/labeler/src/assessment-orchestrator.ts
+++ b/apps/labeler/src/assessment-orchestrator.ts
@@ -294,6 +294,10 @@ export class AssessmentOrchestrator {
 			// subject after the currency re-check below no-ops this CAS at commit
 			// time, so no outcome/block label commits for a deleted subject.
 			guardSubjectNotDeleted: true,
+			// Halt an in-flight run whose automation kill-switch was paused after
+			// dispatch: the CAS (and every label gated on `toState`) no-ops, leaving
+			// the run `running` for the Workflow retry to re-run after resume.
+			guardAutomationNotPaused: true,
 		});
 
 		const statements = [...finalization.statements];
--- a/apps/labeler/src/assessment-store.ts
+++ b/apps/labeler/src/assessment-store.ts
@@ -706,8 +706,15 @@ export function buildFinalizationStatements(
 					 ON CONFLICT(src, uri, cid) DO UPDATE SET
 					   assessment_id = excluded.assessment_id, updated_at = excluded.updated_at
 					 WHERE EXISTS (SELECT 1 FROM assessments WHERE id = ? AND state = ?)
-					   AND (SELECT created_at_epoch_ms FROM assessments WHERE id = excluded.assessment_id)
-					       >= (SELECT created_at_epoch_ms FROM assessments WHERE id = current_assessments.assessment_id)`,
+					   AND (
+					     (SELECT created_at_epoch_ms FROM assessments WHERE id = excluded.assessment_id)
+					         > (SELECT created_at_epoch_ms FROM assessments WHERE id = current_assessments.assessment_id)
+					     OR (
+					       (SELECT created_at_epoch_ms FROM assessments WHERE id = excluded.assessment_id)
+					           = (SELECT created_at_epoch_ms FROM assessments WHERE id = current_assessments.assessment_id)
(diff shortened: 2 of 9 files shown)


ascorbic:
Fixed the stale `sendDiscoveryJob` docstring in `69596879`: it now states the queue op is awaited before `retried` is committed (a durability gate on the request path, not deferred), and that `retried` is a durable claim only once the job is accepted. Dropped the design-section reference per the repo comment policy (no doc/issue refs in comments).

~ 🤖 Claude Fable 5

PR #2168, opened 2026-07-21
fix(labeler): keep assessment-pending live until the last concurrent run resolves (held review finding #13)

ascorbic:
## What does this PR do?

Closes held finding #13 from the adversarial review of the labeler (RFC #694, umbrella #1909) — parked during the earlier fix PRs, now addressed off the merged tip. TDD'd and independently adversary-reviewed (PASS, no #2115 regression). Targets the `feat/plugin-registry-labelling-service` integration branch.

**The race.** `assessment-pending` is a real atproto label whose subject state is the highest-`sequence` op per `(src,uri,val)`. Concurrent assessment runs for the same `(uri,cid)` (different run keys → separate Workflow instances, by design so an operator rerun mints a fresh run) each issue their own positive/negation, but all target the same `val` — so **any** run's negation becomes the stream head and clears the gate for the whole subject. If run A finalizes `passed` while run B is still active, A's negation clears the pending gate, briefly flipping the release from `pending` to eligible (exposing A's stale pass) until B finalizes and re-blocks.

**The fix — a pending projection via last-run-clears.** The `assessment-pending` negation is now gated, in the same `db.batch`, on *no other non-terminal run existing for the same `(uri,cid)`* (a new `requireNoOtherActiveRun` guard on `buildIssuanceStatements`, alongside the existing signing / `requireAssessmentState` / subject-not-deleted guards). D1 serializes batch transactions and each run's `running→terminal` CAS commits in the same batch, so the *last* run to resolve is the only one that sees no non-terminal sibling and negates — the gate clears exactly once, when the last run finishes. Self-exclusion is by run id, so it holds regardless of batch ordering. Keyed on `(uri,cid)`, so distinct CID versions never entangle. Serialization was rejected in favor of the projection because it would regress the deliberate rerun-mints-a-fresh-run semantics.

Because a run can now finalize `passed`/`blocked` while still holding a live positive (its own negation suppressed), the delete cleanup's pending scan (`listPendingBearingAssessmentsForUri`) is made **label-driven** — it now negates any run of any state that holds a committed, un-negated positive, plus all non-terminal runs — so the #2115 delete-vs-issuance closure is preserved (the adversary confirmed it's a strict superset of the old state-based scan). The cleanup's cancel-skip widens from `state==="stale"` to all terminal states (a terminal run is negated but not fed an illegal `→cancelled` transition).

**Known follow-up (deferred, not a regression):** the pure CID-supersession case — a run self-`stale`d because its CID was superseded (no delete event) — leaves an old-CID positive that this fix doesn't negate. Negating it would clobber the *new* CID's legitimate pending (the negation would be the CID-mismatched stream head). Both real consumers (enforcement and `getActiveLabelState`) are CID-scoped, and the new-CID run's own pending is the stream winner, so the lingering old-CID positive is a stale audit signal on a superseded CID, not a false-eligibility signal. Tracked as a CID-aware pending-reduction follow-up if the audit-view wart matters.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm test` passes (labeler: 904 + 136 + 39) — TDD: the pending-gate test asserts A's negation is suppressed while a sibling is active (fails pre-fix, where finalize unconditionally negated); the delete test asserts a terminal decision run's live positive is negated on delete (fails pre-fix, where the state-based scan excluded `passed`)
- [x] New features link to an approved Discussion — **n/a**: bug fix; umbrella #1909 tracks RFC #694.

- [x] This PR includes AI-generated code — model/tool: **Claude Opus 4.8** (implementation + independent adversarial review), orchestrated by **Claude Opus 4.8**

## Screenshots / test output

```
labeler: 904 (workerd) + 136 (console) + 39 (calibration) passed
adversary: PASS — no stranding/both-skip; label-driven scan is a strict superset (no #2115 regression); read-back can't drop a signing-paused negation; supersession deferral verified CID-safe
```

---

### Try this PR

**[Open a fresh playground →](https://fix-labeler-sol-13-concurrency-emdash-playground.emdash-cms.workers.dev)**

<sub>Tracks `fix/labeler-sol-13-concurrency`. Updated automatically when the playground redeploys.</sub>

### DIFF
--- a/apps/labeler/src/assessment-orchestrator.ts
+++ b/apps/labeler/src/assessment-orchestrator.ts
@@ -32,6 +32,7 @@ import type { ModerationPolicy } from "./policy.js";
 import {
 	buildIssuanceStatements,
 	markPublicationAccepted,
+	readIssuedLabelByActionKey,
 	type AutomatedIssuanceAction,
 	type AutomatedLabelProposal,
 	type IssuedLabel,
@@ -347,8 +348,40 @@ export class AssessmentOrchestrator {
 		if (toState === "error") {
 			await issue("assessment-error", false);
 		}
-		// Spec §9.9 point 9: always negate this run's own assessment-pending.
-		await issue("assessment-pending", true);
+		// Spec §9.9 point 9: negate this run's own assessment-pending — but only when
+		// this is the last run in flight for the subject. `requireNoOtherActiveRun`
+		// no-ops the negation while a sibling run for the same (uri, cid) is still
+		// non-terminal, so the positive assessment-pending stays the stream head and
+		// the release stays gated; the last run to finalize clears it.
+		const pendingNegationKey = automatedIdempotencyKey(
+			assessment.runKey,
+			"assessment-pending",
+			true,
+		);
+		const pendingNegation = await buildIssuanceStatements(
+			this.db,
+			this.config,
+			this.signer,
+			{
+				actor: this.config.labelerDid,
+				type: "automated-assessment",
+				assessmentId: assessment.id,
+				reason: `assessment ${toState}`,
+				idempotencyKey: pendingNegationKey,
+			},
+			{ uri: assessment.uri, cid: assessment.cid, val: "assessment-pending", neg: true },
+			now,
+			this.publisher !== undefined,
+			{
+				requireAssessmentState: toState,
+				requireNoOtherActiveRun: {
+					uri: assessment.uri,
+					cid: assessment.cid,
+					assessmentId: assessment.id,
+				},
+			},
+		);
+		statements.push(...pendingNegation.statements);
 
 		// Spec §9.9 point 8 / §10: negate prior active automated labels this
 		// outcome no longer supports.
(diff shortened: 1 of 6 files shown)


ascorbic:
Fixed the stale comment in `83223581`: the negate loop comment now reads "any run — non-terminal or terminal —" to match the widened `listPendingBearingAssessmentsForUri` scan (any run holding a committed un-negated positive) and the `TERMINAL_STATES.has(state)` cancel-skip.

That commit is also the reconciliation merge with the integration branch after #2166 (the 9/14/19 batch) merged: the finalization CAS now carries the automation-not-paused guard (14), the subject-not-deleted guard (#2115), and the signing guard together with this PR's `requireNoOtherActiveRun` on the pending negation; the current-pointer id tie-break (19) is present; and both PRs' tests coexist. Full labeler suite 909 + 136 + 39 green.

~ 🤖 Claude Fable 5

PR #2169, opened 2026-07-21
feat(plugins): predicate-guarded atomic updateIf for plugin storage (no-oversell)

vedanshujain:
## What does this PR do?

Adds `ctx.storage.<collection>.updateIf(id, { where, set?, delta? })` — a **predicate-guarded atomic update** for plugin storage, and the **no-oversell** primitive. Today plugin storage offers only unconditional `put` (whole-doc upsert), so "sell one iff in stock" is impossible in-process without a racy read-modify-write.

`updateIf` runs the guard **and** the arithmetic in one statement:

```sql
UPDATE _plugin_storage
SET data = <json_set/jsonb_set expr>, updated_at = ?
WHERE plugin_id = ? AND collection = ? AND id = ? AND <guard>
RETURNING data
```

No read-then-write, no interactive transaction — so N concurrent guarded decrements serialize correctly (exactly M of N apply, final stock 0, never oversell).

- **`where`** reuses the numeric-correct `WhereClause` translation from `query()` (landed upstream in #1898 + the numeric fix), so a multi-digit guard like `stock >= 10` compares **numerically** on Postgres.
- **`set`** writes wholesale field values; **`delta`** applies integer `inc`/`dec` in-SQL over `COALESCE(base, 0)` (a delta on a missing/null field starts from 0). Integer-only deltas enforced at runtime; a field may not appear in both `set` and `delta`; at least one is required. `set`/`delta` are separate args so `{inc:n}` is never mistaken for a value.
- Returns `{ applied: true, data }` or `{ applied: false }` (row absent **or** guard failed — intentionally indistinguishable). **Never inserts.**

Backed by `pluginDataWriteExpr` (dialect-correct `json_set`/`jsonb_set`, values bound as params). **No new column, no migration.**

**Isolation:** the `{ applied: false }` contract assumes **READ COMMITTED** (Postgres' default), where the losing concurrent writers cleanly resolve to `{ applied: false }` via EvalPlanQual. Under REPEATABLE READ / SERIALIZABLE a loser aborts instead, so `updateIf` traps SQLSTATE `40001`/`40P01` and throws a typed `StorageSerializationError` (with `cause` + `sqlState`) telling the caller to retry — rather than surfacing a bare driver error. The no-oversell **safety** invariant holds under every isolation level (a losing writer never applies); only the *result shape* degrades to a throw. SQLite/D1 serialize writes and never hit this path.

### Tests
- `storage-updateif.test.ts` — guard pass/fail/missing, integer inc/dec + integer round-trip, float-delta rejection, `COALESCE`-from-0 on missing/null, wholesale `set`, `set`+`delta`, both-fields / neither-provided rejection, guard operator coverage (equality, multi-digit `gte`, `in`, `startsWith`, non-indexed field), empty-`in` no-op.
- `storage-no-oversell.test.ts` — M-of-N concurrent guarded decrements (the acceptance bar; Postgres is the real race, SQLite proves SQL correctness).

Both dialects; Postgres via `EMDASH_TEST_PG`. Full core plugin suites green (795 tests), typecheck / lint / format clean.

## Type of change

- [ ] Bug fix
- [x] Feature (requires [maintainer-approved Discussion](https://github.com/emdash-cms/emdash/discussions/categories/ideas)) — see https://github.com/emdash-cms/emdash/discussions/632
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (or targeted tests for my change) — full core plugin suites, both dialects
- [ ] User-visible strings in the admin UI are wrapped for translation (if applicable) — N/A
- [x] New features link to an approved Discussion — approved by @swissky in https://github.com/emdash-cms/emdash/discussions/632


logelog:
@vedanshujain, would you be open to me contributing to this PR? I've prepared and tested additional changes from my earlier combined implementation against your `b25bf0e5` commit:

- Argument validation so malformed guards cannot become unconditional writes, plus safe integer counter and overflow checks.
- Public exports for the update types and serialization error.
- Cloudflare and Workerd sandbox support, including preserving guards during JSON transport and carrying safe retry metadata.
- Documentation for the arguments, counter behavior and transaction retry requirements.

These preserve your existing return shape and add no revision migration. The changes have been tested on SQLite, PostgreSQL and D1, including the sandbox paths.

If this approach works for you, I'd like to contribute these fixes here and help get your PR ready to merge first. After that, I'd welcome your contribution to [my revision based writes PR](https://github.com/emdash-cms/emdash/pull/2980) so we can finish that part together. Let me know how you'd prefer to receive the commits.

vedanshujain:
@logelog sounds good to me, please do. I added the access to the fork so you can push commit. I was already working on https://github.com/vedanshujain/emdash/pull/9 before I saw your message, it addresses the argument validation but rest three are still open.

vedanshujain:
I am marking the PR as draft again so we can get it polished before asking maintainer's time for review. Once @logelog adds their commit, we will mark this ready for review and merge again

logelog:
Thanks @vedanshujain. I've pushed three commits directly to this branch and merged the latest main. Your fix for range filters with no defined bound is retained. The additions cover:

- Safe integer counter validation and overflow checks, with the whole update left unapplied when a stored counter or result is invalid.
- Public exports for the update types and serialization error.
- Cloudflare and Workerd sandbox support, including preserving guards during transport and carrying safe retry metadata.
- Documentation for the arguments, counter behavior and retrying an entire transaction after a serialization failure or deadlock.

The shared filter precedence and return shape are preserved, with no migration added. Local checks passed on SQLite, PostgreSQL and D1, including the sandbox paths; build, typecheck and lint also passed.

Please take a look at the additions. Once you're happy with them and CI passes, we can mark this ready for review and move on to the revision work together.

vedanshujain:
Looks great, thanks so much! I am marking this for review.

ascorbic (approved):
Thanks! This is a great addition

PR #2238, opened 2026-07-25
fix(core): drive taxonomy term counts from the pivot

### LINKED ISSUE #2237: Taxonomy term-count query (#1892) reads orders of magnitude more D1 rows than expected despite covering indexes
## Summary

The consolidated taxonomy term-count query introduced in #1892 (`packages/core/src/taxonomies/term-counts.ts`) reads an order of magnitude more D1 rows than expected on databases with a non-trivial number of content rows, despite `content_taxonomies` having covering indexes on `(taxonomy_id)` and composite indexes on `(taxonomy_id, collection, deleted_at, ...)`.

## Observed impact

On a production site with covering indexes already present, a single call to the term-count query (one taxonomy, `UNION ALL` across 3 collections) was observed to:

- Read **~35 million rows** in one execution
- Take **~29 seconds** of D1 `sql_duration_ms`
- Recur on essentially every admin content-list/editor page load (the term-count endpoint is called once per taxonomy to populate filter UI), multiplying the cost across normal editorial usage — not just an edge case or bulk operation

CPU time for the invoking Worker stayed low (single/double-digit ms) while wall time matched the D1 query duration, confirming the cost is entirely in D1 query execution, not Worker-side processing.

## Suspected cause

The query combines, per branch of the `UNION ALL`:

```sql
SELECT ct.taxonomy_id, COUNT(*)
FROM content_taxonomies AS ct
INNER JOIN "<collection_table>" AS e ON e.id = ct.entry_id
WHERE ct.collection = ?
  AND ct.taxonomy_id IN (SELECT translation_group FROM taxonomies WHERE name = ?)
  AND (e.status = 'published' OR (e.status = 'scheduled' AND e.scheduled_at <= strftime('%Y-%m-%dT%H:%M:%fZ', 'now')))
  AND e.deleted_at IS NULL
GROUP BY ct.taxonomy_id
```

The `taxonomy_id IN (SELECT translation_group FROM taxonomies WHERE name = ?)` predicate, combined with the `GROUP BY` and the cross-table `UNION ALL`, appears to prevent D1/SQLite's query planner from using the existing composite indexes on `content_taxonomies` — despite indexes covering `taxonomy_id`, `collection`, and `deleted_at` in the right order for this access pattern. The result looks like a near-full-table scan per `UNION ALL` branch rather than an indexed lookup.

## Repro sketch (generic, not tied to any specific dataset)

1. Seed a `content_taxonomies` table with a large number of rows relative to a small number of distinct `taxonomy_id` values (i.e., many entries per term) across at least two collections.
2. Confirm indexes exist: `idx_content_taxonomies_term (taxonomy_id)` plus the composite `(taxonomy_id, collection, deleted_at, ...)` indexes shipped alongside #1892.
3. Call `GET /_emdash/api/taxonomies/{name}/terms` and inspect D1 query stats (`rows_read`, `sql_duration_ms`) for the underlying `term-counts` query.
4. Compare `rows_read` against the actual number of rows that satisfy the `WHERE` clause — expect a large gap if the planner isn't using the covering indexes.

## Suggested areas to investigate

- Whether the `IN (subquery)` predicate on `taxonomy_id` should instead be resolved to concrete IDs by the application layer before the main query runs (avoiding a correlated/uncorrelated subquery the planner may not optimize well against a covering index).
- Whether `EXPLAIN QUERY PLAN` on D1/SQLite for this exact shape shows an index scan or a table scan, and whether reordering the `WHERE` clause or restructuring the `UNION ALL` changes the chosen plan.
- Whether the query-count CI snapshots (`scripts/query-counts.snapshot.d1.json`) captured row-count/cost at a data scale small enough to miss this — i.e., the regression may only manifest once collection tables cross some row-count threshold that the CI fixtures don't reach.

## Environment

- `emdash` 0.31.1 (npm registry release)
- Cloudflare Workers + D1
- Astro, `output: "server"`

MA2153:
## What does this PR do?

Fixes the taxonomy term-count query reading `entries × terms` rows instead of seeking the pivot.

`collectionBranch` in `packages/core/src/taxonomies/term-counts.ts` joined `content_taxonomies` to the content table with an `INNER JOIN`. On stats-blind SQLite/D1 (no `ANALYZE`, no `sqlite_stat1`) the planner picked `ec_*` as the **outer** table and re-ran the whole `taxonomy_id IN (SELECT translation_group FROM taxonomies WHERE name = ?)` term list as a pivot-primary-key probe for every visible entry in the collection:

```
SEARCH e USING INDEX idx_ec_<collection>_deleted_status (deleted_at=?)
SEARCH ct USING COVERING INDEX sqlite_autoindex_content_taxonomies_1 (collection=? AND entry_id=? AND taxonomy_id=?)
LIST SUBQUERY 1          ← re-probed once per outer row
SEARCH taxonomies USING INDEX idx_taxonomies_name_locale (name=?)
```

That is why the composite indexes added alongside #1892 / migration 051 never helped: they are all `taxonomy_id`-leading, and the pivot was never the driving table, so nothing could seek them. The cost is a nested loop, not a scan — it scales as `visible entries × terms in the taxonomy`.

The fix is `CROSS JOIN` with the join predicate moved into `WHERE`. In SQLite that is a **join-order hint, not a cartesian product**: it pins the pivot as the outer table. The terms are then seeked on a `(taxonomy_id, collection)` index and the content row is touched exactly once per assignment, by primary key:

```
SEARCH ct USING COVERING INDEX idx_content_taxonomies_loc_crt (taxonomy_id=? AND collection=?)
SEARCH e USING INDEX sqlite_autoindex_ec_<collection>_1 (id=?)
```

Postgres has statistics and reorders freely, so `CROSS JOIN` there is a plain inner join.

**No semantic change.** The predicates are untouched — this only changes join order. Counts are byte-identical (verified independently against the same dataset).

### Measured on a production D1

One call, one taxonomy, `UNION ALL` across 3 collections, ~26k entries in the dominant collection and ~1.4k terms:

| | rows_read | `sql_duration_ms` |
| --- | --- | --- |
| before | 35,627,677 | 28,892 |
| after | 63,854 | 120 |

**558× fewer rows read, 241× faster.**

### Rejected alternative

Counting the denormalized `content_taxonomies.status` / `deleted_at` columns (migration 051) and dropping the content-table join entirely. Those columns are advisory — 051's contract is that the read path re-checks the authoritative `ec_*` row — and measuring both on the same data showed the trusting version actually reads **more** (54,610 vs 52,282 rows), because no pivot index covers `status`/`scheduled_at`, so every candidate costs a table lookup anyway. Staying authoritative was both more correct and cheaper.

### Unrelated observation, not addressed here

On the same request, `TaxonomyRepository.findByName` (`select * from taxonomies where name = ? and locale = ?`) read 2,899 rows to return 1,449 terms — one index entry plus one table row each, since `idx_taxonomies_name_locale` does not cover `select *`. That is correct behaviour and simply the cost of a taxonomy that large; flagging it because it shows up in the same trace and could be mistaken for part of this bug. Out of scope for this PR.

Closes #2237

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [x] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes (`pnpm lint:json` → 0 diagnostics)
- [x] `pnpm test` passes (full `packages/core` suite: 5020 passed on SQLite; 5394 passed with `EMDASH_TEST_PG` against Postgres 17, so the new join shape is exercised on both dialects)
- [ ] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable). Do not include `messages.po` changes except in translation PRs — a workflow extracts catalogs on merge to `main`. — n/a, no admin UI strings

## Screenshots / test output

New regression test `packages/core/tests/integration/taxonomy-term-counts-plan.test.ts`, modelled on the existing `loader-taxonomy-pivot-plan` test. Written **before** the fix and confirmed failing against the unfixed source — the local stats-blind plan reproduced the production plan exactly:

```
 FAIL  tests/integration/taxonomy-term-counts-plan.test.ts > touches the content table only by primary key

- Expected:
/SEARCH e USING (COVERING )?INDEX sqlite_autoindex_ec_post_1 \(id=\?\)/

+ Received:
"CO-ROUTINE per_collection
SEARCH e USING INDEX idx_ec_post_loc_crt (deleted_at=?)
SEARCH ct USING COVERING INDEX sqlite_autoindex_content_taxonomies_1 (collection=? AND entry_id=? AND taxonomy_id=?)
LIST SUBQUERY 1
SEARCH taxonomies USING INDEX idx_taxonomies_name_locale (name=?)
..."

 Test Files  1 failed (1)
      Tests  2 failed (2)
```

With the fix:

```
 Test Files  1 passed (1)
      Tests  2 passed (2)
```

Full suite, both dialects:

```
 Test Files  388 passed | 1 skipped (389)
      Tests  5020 passed | 3 skipped (5023)      # SQLite

 Test Files  389 passed (389)
      Tests  5394 passed (5394)                  # + Postgres 17 parity
```


khoinguyenpham04 (approved):
@MA2153 LGTM! Thanks for tackling this issue.

PR #2243, opened 2026-07-26
fix(core): make term-list counts opt-out for the editor picker

MA2153:
## What does this PR do?

The content editor's taxonomy sidebar (`TaxonomySidebar`, rendered by `ContentSettingsPanel` on every editor open with no `staleTime`) lists terms to pick from and never renders a usage count. It shares `GET /_emdash/api/taxonomies/:name/terms` with the Taxonomies settings page, which does render counts — and `handleTermList` aggregated counts unconditionally. Opening any entry therefore ran the `content_taxonomies × ec_*` UNION ALL aggregate once per applicable taxonomy, with one branch per collection the taxonomy declares:

```sql
SELECT taxonomy_id, SUM(count) FROM (
  SELECT ct.taxonomy_id, COUNT(*) FROM content_taxonomies AS ct
  JOIN "ec_resources" AS e ON e.id = ct.entry_id WHERE ... GROUP BY ct.taxonomy_id
  UNION ALL ... -- one branch per declared collection
) AS per_collection GROUP BY taxonomy_id
```

The endpoint now takes an `includeCounts` query param, defaulting to `true`. When `false`, the aggregate is skipped and `count` is omitted from each term rather than reported as a misleading `0`. The editor sidebar passes `includeCounts=false`; the Taxonomies settings page keeps the default and still shows real counts.

#2219 made the public render path demand-driven (`getTaxonomyTerms({ includeCounts })`); this is the admin-side counterpart, and the param name matches.

**Compatibility:** the default is unchanged, so existing API consumers see identical responses. The one ripple is type-level — `TermWithCount.count` becomes `number | undefined`, so TS consumers reading it must handle `undefined`. The only consumer in this repo (`TaxonomyManager.tsx`) already did (`{term.count || 0}`).

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [x] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes (`pnpm lint:json` → 0 diagnostics, matching the pre-change baseline)
- [x] `pnpm test` passes (full `packages/core` suite: 5022 passed, 3 skipped, 0 failed)
- [x] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable) — n/a, no new strings; the one touched call site keeps its existing `msg` -wrapped error. No `messages.po` changes included.

## Screenshots / test output

New test asserts on the SQL actually executed (logging Kysely, filtering for the aggregate's `per_collection` subquery alias) — a response-shape assertion alone can't prove the work was skipped. Written first, failing on the two opt-out cases before the fix.

```
 ✓ tests/unit/taxonomies/term-list-counts.test.ts (4 tests)
   ✓ counts by default, so existing callers are unaffected
   ✓ omits the aggregate and the count field when the caller opts out
   ✓ honours ?includeCounts=false on the route
   ✓ still counts on the route when the param is absent

 Test Files  1 passed (1)
      Tests  4 passed (4)
```

Full core suite:

```
 Test Files  388 passed | 1 skipped (389)
      Tests  5022 passed | 3 skipped (5025)
```

### DIFF
--- a/packages/admin/src/components/TaxonomySidebar.tsx
+++ b/packages/admin/src/components/TaxonomySidebar.tsx
@@ -337,8 +337,10 @@ function TaxonomySection({
 	const [newCategoryLabel, setNewCategoryLabel] = React.useState("");
 	const [showCategoryInput, setShowCategoryInput] = React.useState(false);
 
+	// The count mode belongs in the key: the Taxonomies settings page reads the
+	// same endpoint with counts and must not be served this count-free list.
 	const { data: terms = EMPTY_TERMS } = useQuery({
-		queryKey: ["taxonomy-terms", taxonomy.name, entryLocale],
+		queryKey: ["taxonomy-terms", taxonomy.name, entryLocale, { includeCounts: false }],
 		queryFn: () => fetchTerms(taxonomy.name, entryLocale),
 	});
 
--- a/packages/admin/src/components/TaxonomySidebar.tsx
+++ b/packages/admin/src/components/TaxonomySidebar.tsx
@@ -78,10 +78,13 @@ export function useHasApplicableTaxonomies(collection: string): boolean {
 
 /**
  * Fetch terms for a taxonomy, scoped to the entry's locale so only the matching
- * translation variants are offered.
+ * translation variants are offered. The picker shows no usage counts, so it
+ * opts out of the per-collection count aggregate the endpoint runs by default.
  */
 async function fetchTerms(taxonomyName: string, locale?: string): Promise<TaxonomyTerm[]> {
-	const res = await apiFetch(withLocale(`/_emdash/api/taxonomies/${taxonomyName}/terms`, locale));
+	const res = await apiFetch(
+		withLocale(`/_emdash/api/taxonomies/${taxonomyName}/terms?includeCounts=false`, locale),
+	);
 	const data = await parseApiResponse<{ terms: TaxonomyTerm[] }>(
 		res,
 		i18n._(msg`Failed to fetch terms`),
--- a/.changeset/tidy-pandas-repeat.md
+++ b/.changeset/tidy-pandas-repeat.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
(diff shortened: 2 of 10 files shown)


packages/admin/src/components/TaxonomySidebar.tsx
```diff
@@ -78,10 +78,13 @@ export function useHasApplicableTaxonomies(collection: string): boolean {
...
 /**
  * Fetch terms for a taxonomy, scoped to the entry's locale so only the matching
- * translation variants are offered.
+ * translation variants are offered. The picker shows no usage counts, so it
+ * opts out of the per-collection count aggregate the endpoint runs by default.
  */
 async function fetchTerms(taxonomyName: string, locale?: string): Promise<TaxonomyTerm[]> {
-	const res = await apiFetch(withLocale(`/_emdash/api/taxonomies/${taxonomyName}/terms`, locale));
+	const res = await apiFetch(
+		withLocale(`/_emdash/api/taxonomies/${taxonomyName}/terms?includeCounts=false`, locale),
```
khoinguyenpham04, line 86: can we pls add the count mode to the query key? the editor and `TaxonomyManager` currently share the same cache key, but this result has no counts. on multi-language sites, the manager can reuse it and show every count as `0` until it refetches or the query is invalidated.

khoinguyenpham04 (commented):
just one thing before merge. the editor and TaxonomyManager need separate cache keys since one fetches counts and the other doesn’t.

packages/admin/src/components/TaxonomySidebar.tsx
MA2153: Great catch — thank you. Both keys now carry the count mode: the sidebar uses `["taxonomy-terms", name, locale, { includeCounts: false }]` and `TaxonomyManager` the `{ includeCounts: true }` variant.

Worth noting it bites single-locale sites too, and not just transiently: `App.tsx` sets a global `staleTime` of 1 minute, so the settings page mounting after an editor open is served the cached count-free list *without* refetching, and shows `0` for that whole window.

The existing invalidations are unaffected — they're prefixes of the new keys, so create/edit/delete/translate still refresh both consumers.

Added `packages/admin/tests/components/taxonomy-term-cache.test.tsx`: it mounts the sidebar and then the settings page under one `QueryClient` at the app's real `staleTime` and asserts the manager renders the true count. Fails on the previous commit (renders `0`), passes now. Full admin suite: 1261 passed.

khoinguyenpham04: thanks for fixing it

khoinguyenpham04 (approved):
LGTM! Thanks @MA2153

Issue #1604, opened 2026-06-24
Integration tests flaky: concurrent Astro dev servers collide on shared fixture lock ("Another astro dev server is already running")


MA2153:
## Summary

The **Integration Tests** CI job (`vitest run --config vitest.integration.config.ts`) fails intermittently with:

```
Error: Server at http://localhost:4397/_emdash/api/setup/dev-bypass did not start within 90000ms

Server output (last 5000 chars):
Another astro dev server is already running.

  URL:  http://localhost:4398
  PID:  3512

Run `astro dev stop` to stop it, or use `astro dev --force` to replace it.
```

This is an infra/harness flake, **not** a product or test-logic bug. In the failing run, 61 tests passed and only 2 suites (`field-widgets.test.ts`, `oauth-discovery.test.ts`) failed — both for the identical "already running" reason.

Seen on PR #1535 — [job log](https://github.com/emdash-cms/emdash/actions/runs/28100064484/job/83199420620).

## Root cause

Each integration suite boots a real Astro dev server via `createTestServer()` in `packages/core/tests/integration/server.ts`. Two facts combine to cause the race:

1. **All suites share one fixture directory.** The server runs the fixture *in-place* — `const workDir = FIXTURE_DIR; ... spawn(astroBin, ["dev", ...], { cwd: workDir })` (server.ts:160, 181). The comment explains the in-place choice avoids Astro virtual-module resolution issues with symlinked temp dirs. Only the DB/uploads live in a temp dir.

2. **Vitest runs test files in parallel by default.** `vitest.integration.config.ts` sets no `pool`/`fileParallelism` options, so each `*.test.ts` runs in its own worker concurrently.

Astro's "another dev server is already running" guard is scoped to the **project directory** (a lockfile under the shared `node_modules/.astro`), **not** to the port. So even though each suite picks a distinct hardcoded port:

| Suite | Port |
| --- | --- |
| `comments.test.ts` | 4396 |
| `field-widgets.test.ts` | 4397 |
| `cli.test.ts` | 4398 |
| `client.test.ts` | 4399 |
| `oauth-discovery.test.ts` | 4401 |

…the *second* server to start from the same fixture dir aborts immediately because another is already holding the lock. The tell: the suite requested port 4397 but the error reports a conflicting server at **4398** (a different suite's port). `waitForServer` then spins for the full 90s and the suite fails.

It's timing-dependent (whoever wins the lock first survives), which is why it only flakes sometimes.

[...]

PR #2246, opened 2026-07-26
fix(admin): keep content list view state in the URL

MA2153:
## What does this PR do?

Reported by an editor and reproduced: on a collection's entry list, go to page 2 or 3, open an entry, then press the browser's back button — you land back on page 1.

`page` lived in `ContentList`'s `useState`, and sort/search/status/author/date lived in `ContentListPage`'s. Opening an entry unmounts the list, so back threw all of it away.

This moves the whole list view onto the route's search params, so back-navigation restores the view the editor left. List URLs become shareable and bookmarkable as a side effect.

- **Defaults are never serialized.** A pristine list is `/content/posts` with no query string; clearing a control drops its param again. Page is 1-based in the URL and omitted on page 1.
- **Every filter/sort change resets the page** — page 4 of one filter set indexes nothing in another, and the cursor chain restarts from a fresh first page anyway.
- **History stays usable.** Discrete clicks (page, sort, status, author) push an entry so back steps through them; the debounced search and the date inputs replace, since both fire on partial input rather than on commit.
- **Params are validated, not trusted.** `?sort=authorId`, `?page=2.5`, `?dateFrom=yesterday` are dropped rather than forwarded to the API.

`ContentList` gains optional `page`/`onPageChange` and `initialSearchQuery` props, following the existing opt-in pattern used by `sort`/`onSortChange` and the filter callbacks. The component is exported from `@emdash-cms/admin`, so all three are additive — callers that don't pass them keep the current internal-state behavior and a working pager.

### Not included: the All/Trash tab

`activeTab` is still `useState` in `ContentList`, so back always lands on **All**. It was out of scope for the reported bug and the trash tab has no pagination of its own.

The infrastructure does generalize to it — adding it later is the same three moves this PR already makes for `page`: a `tab` entry in `parseContentListSearch`, an `activeTab`/`onTabChange` controlled prop pair on `ContentList` mirroring `page`/`onPageChange`, and a page reset on tab switch. Happy to fold it in here if a reviewer would rather have it in one go.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes — `pnpm lint:json` reports 0 diagnostics
- [x] `pnpm test` passes (or targeted tests for my change) — full `packages/admin` suite, 1251 passed
- [x] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable) — n/a, no new strings or markup; no `messages.po` changes included

## Screenshots / test output

Failing tests written first, then the fix.

Four browser-level regression tests in `tests/router.test.tsx` — page survives back-navigation, filters and sort survive back-navigation, pristine URLs stay free of query params, and page + search term restore together. Nine parser unit tests in `tests/router-search.test.ts`, alongside the existing `parseBylinesLocaleSearch` ones.

Two notes for reviewers:

- My first draft of the pagination test **passed against the broken code**. The mocked `ContentEditor` in that file renders the entry title, so `getByText("Post 21")` matched the editor rather than the list. The assertions now target the page indicator and the row links. Worth knowing when adding tests to that file.
- The subtlest guard in the change is in `handleSearchChange`: `ContentList` reports its seeded search query once when the debounce settles on mount, and treating that as a user edit wipes the page just restored from the URL. I removed the guard, confirmed `restores a page and a search term together` fails, and put it back.

```
 Test Files  104 passed (104)
      Tests  1251 passed (1251)
```

Not verified by hand against a seeded site — the suite runs in Chromium and covers the round trip, but I did not drive the real admin through `dev-bypass`. No new UI strings or layout, so there is no RTL/Arabic surface to re-check.

### DIFF
--- a/packages/admin/src/components/ContentList.tsx
+++ b/packages/admin/src/components/ContentList.tsx
@@ -211,7 +211,15 @@ export function ContentList({
 	const [searchQuery, setSearchQuery] = React.useState(controlledSearchQuery);
 	const [internalPage, setInternalPage] = React.useState(0);
 	const page = controlledPage ?? internalPage;
-	const setPage = onPageChange ?? setInternalPage;
+	const requestedPage = React.useRef(controlledPage);
+	const setPage = React.useCallback(
+		(next: number) => {
+			requestedPage.current = next;
+			if (onPageChange) onPageChange(next);
+			else setInternalPage(next);
+		},
+		[onPageChange],
+	);
 	const [selectedIds, setSelectedIds] = React.useState<Set<string>>(() => new Set());
 
 	// Bulk selection is opt-in: the checkbox column + toolbar only render when
@@ -238,6 +246,16 @@ export function ContentList({
 		reportedQuery.current = controlledSearchQuery;
 		setSearchQuery(controlledSearchQuery);
 	}, [controlledSearchQuery]);
+
+	// A page the caller moved on its own (back/forward, a deep link) replaces
+	// the view wholesale, so an edit still waiting out the debounce belongs to
+	// a view that is gone. Re-seeding the input drops the pending timer with it.
+	React.useEffect(() => {
+		if (controlledPage === undefined || controlledPage === requestedPage.current) return;
+		requestedPage.current = controlledPage;
+		reportedQuery.current = controlledSearchQuery;
+		setSearchQuery(controlledSearchQuery);
+	}, [controlledPage, controlledSearchQuery]);
 
 	// Reset page when search changes. Guarded because `setPage` may write to
 	// the URL, and an unconditional reset would push a history entry per
--- a/packages/admin/src/components/ContentList.tsx
+++ b/packages/admin/src/components/ContentList.tsx
@@ -92,6 +92,20 @@ export interface ContentListProps {
 	sort?: ContentListSort;
 	onSortChange?: (sort: ContentListSort) => void;
 	/**
+	 * Controlled pagination (0-based). Pass both to hold the current page
+	 * outside the component — the router keeps it in the URL so the browser's
+	 * back button returns to the page the user left. Without `onPageChange`
+	 * the component keeps its own page state and the pager still works.
+	 */
+	page?: number;
+	onPageChange?: (page: number) => void;
+	/**
+	 * Seeds the search box on mount, for callers that restore the query from
+	 * somewhere durable (the URL). The input stays internally controlled after
+	 * that, so typing isn't lagged by the debounce round-trip.
+	 */
+	initialSearchQuery?: string;
+	/**
 	 * Total rows matching the current filters (ignoring pagination). When
 	 * set, the pagination denominator reflects this stable count instead of
 	 * growing as more API pages are fetched.
@@ -175,6 +189,9 @@ export function ContentList({
 	urlPattern,
 	sort,
 	onSortChange,
+	page: controlledPage,
+	onPageChange,
+	initialSearchQuery = "",
 	total,
 	onSearchChange,
 	statusFilter = "all",
@@ -190,8 +207,10 @@ export function ContentList({
 }: ContentListProps) {
 	const { t } = useLingui();
 	const [activeTab, setActiveTab] = React.useState<ViewTab>("all");
-	const [searchQuery, setSearchQuery] = React.useState("");
-	const [page, setPage] = React.useState(0);
+	const [searchQuery, setSearchQuery] = React.useState(initialSearchQuery);
+	const [internalPage, setInternalPage] = React.useState(0);
+	const page = controlledPage ?? internalPage;
+	const setPage = onPageChange ?? setInternalPage;
 	const [selectedIds, setSelectedIds] = React.useState<Set<string>>(() => new Set());
 
 	// Bulk selection is opt-in: the checkbox column + toolbar only render when
@@ -208,10 +227,12 @@ export function ContentList({
 		if (onSearchChange) onSearchChange(debouncedSearch.trim());
 	}, [debouncedSearch, onSearchChange]);
 
-	// Reset page when search changes
+	// Reset page when search changes. Guarded because `setPage` may write to
+	// the URL, and an unconditional reset would push a history entry per
+	// keystroke.
 	const handleSearchChange = (e: React.ChangeEvent<HTMLInputElement>) => {
 		setSearchQuery(e.target.value);
-		setPage(0);
+		if (page !== 0) setPage(0);
 	};
 
 	const filteredItems = React.useMemo(() => {
--- a/packages/admin/src/components/ContentList.tsx
+++ b/packages/admin/src/components/ContentList.tsx
@@ -100,11 +100,12 @@ export interface ContentListProps {
 	page?: number;
 	onPageChange?: (page: number) => void;
 	/**
-	 * Seeds the search box on mount, for callers that restore the query from
-	 * somewhere durable (the URL). The input stays internally controlled after
-	 * that, so typing isn't lagged by the debounce round-trip.
+	 * The search term the caller holds (the URL, typically). The input keeps
+	 * its own state while typing so keystrokes aren't lagged by the debounce
+	 * round-trip, and re-syncs whenever this changes on its own — a deep link
+	 * or a back-navigation to a different query.
 	 */
-	initialSearchQuery?: string;
+	searchQuery?: string;
 	/**
 	 * Total rows matching the current filters (ignoring pagination). When
 	 * set, the pagination denominator reflects this stable count instead of
@@ -191,7 +192,7 @@ export function ContentList({
 	onSortChange,
 	page: controlledPage,
 	onPageChange,
-	initialSearchQuery = "",
+	searchQuery: controlledSearchQuery = "",
 	total,
 	onSearchChange,
 	statusFilter = "all",
@@ -207,7 +208,7 @@ export function ContentList({
 }: ContentListProps) {
 	const { t } = useLingui();
 	const [activeTab, setActiveTab] = React.useState<ViewTab>("all");
-	const [searchQuery, setSearchQuery] = React.useState(initialSearchQuery);
+	const [searchQuery, setSearchQuery] = React.useState(controlledSearchQuery);
 	const [internalPage, setInternalPage] = React.useState(0);
 	const page = controlledPage ?? internalPage;
 	const setPage = onPageChange ?? setInternalPage;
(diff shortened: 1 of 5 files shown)


MA2153:
Thanks — both comment notes addressed in 0d30ab6b. The route-level block is now the suggested one-liner, and the page-reset paragraph is gone (the sort comment kept). I also trimmed the `handleLocaleChange` comment, which had the same PR-prose problem in the same diff. What is left is limited to non-obvious invariants: the `updateSearch` referential-stability constraint, the mount-time seeded-search guard, `<input type="date">` firing per segment, and the 0-based/1-based page offset.

On the red **Integration Tests** check: it was unrelated to this PR, which touches only `packages/admin`. The failure was `client/comments.test.ts` timing out because its Astro dev server died during Vite pre-bundling with `ENOTEMPTY ... node_modules/.vite/deps`. Every per-suite server symlinks the same donor `node_modules`, so Vite's dep cache is shared across all of them and parallel servers race to re-optimize into it. Filed as #2247 with a suggested fix; it is a sibling of #1604 that the per-suite-fixture fix did not cover. The re-run on this branch is green.

packages/admin/src/router.tsx
```diff
@@ -309,19 +310,82 @@ function DashboardPage() {
 }
 
 // Content list route
+// List view state lives in URL search params so back/forward restore it.
```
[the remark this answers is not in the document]
MA2153: Fixed in 6e0338a4 — dropped the paragraph, kept the `// Content list route` label.

packages/admin/src/router.tsx
```diff
@@ -334,24 +398,108 @@ function ContentListPage() {
...
+				// as a change would clear the very page we just restored.
+				search: (prev) =>
+					(prev.q ?? "") === q ? prev : { ...prev, q: q || undefined, page: undefined },
+				replace: true,
+			});
+		},
+		[navigate, collection],
+	);
 
 	// Filter state (#1288). All are part of the query key so changing any of
```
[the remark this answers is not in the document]
MA2153: Fixed in 6e0338a4 — applied your suggestion. The issue reference is gone, and the second sentence was a duplicate of the search-term comment a few lines up.

packages/admin/src/router.tsx
```diff
@@ -309,19 +310,81 @@ function DashboardPage() {
...
+
+const SORT_FIELDS: ContentListSortField[] = ["title", "status", "locale", "updatedAt"];
+const STATUS_FILTERS = ["published", "draft", "scheduled", "archived"] as const;
+const DATE_FIELDS: ContentDateField[] = ["createdAt", "updatedAt", "publishedAt"];
+
+/** `YYYY-MM-DD`, the shape the date inputs emit. */
+const CALENDAR_DATE_REGEX = /^\d{4}-\d{2}-\d{2}$/;
+
+// Every key is optional: defaults are omitted from the URL, and other routes
+// link here supplying only the params they care about (usually just `locale`).
```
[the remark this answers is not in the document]
MA2153: Trimmed in 655a0bb0, though not quite as suggested.

You're right about the first half — "every key is optional" restates the `?:` on every field, and "defaults are omitted from the URL" restates the `isDefault ? undefined : ...` handlers below. Both gone.

I kept the second clause, because it is not visible from this file. `Dashboard.tsx:236` and `ContentEditor.tsx:635` both link in with `search={{ locale: undefined }}`, so making any key in `ContentListSearch` required breaks them at compile time. Deleting the comment outright would drop the one part a future reader could actually get wrong. It now reads:

```
// Routes link here supplying only the params they care about, so making any
// key required breaks those links.
```

MA2153:
Addressed in 3ee20295.

**Comment cleanup** — removed all seven flagged comments, plus the `dir`/`sort` rationale in `parseContentListSearch` and the "changing a filter restarts the cursor chain" line above `statusFilter` that the review body pointed at.

Two comments in the test file I kept, since they document a footgun rather than the change:

- The one above the first page-indicator assertion, trimmed to two lines. The mocked `ContentEditor` in this file renders the entry title, so `getByText("Post 21")` passes while sitting on the edit screen — a future test author in this file gets this wrong by default, and my first draft did.
- The one explaining the 800ms wait in "restores a page and a search term together". Without it the sleep looks arbitrary.

**The `q` staleness bug is real, and reachable by the back button** — not just deep links. `ContentList` stays mounted across search-param changes, so: page 2 → type a query → back lands on the `?page=2` entry with no `q`, while the search box still shows the term over unfiltered results.

Failing test first (`follows a search term that changes in the URL while the list stays mounted`), then the fix.

I didn't move the debounce up to `ContentListPage`. A URL-controlled input means every keystroke round-trips through `navigate` before it renders, which is exactly the lag the internal state exists to avoid. Instead the list now tracks the term it last reported and adopts the caller's value only when the two differ — an external change syncs the input, the caller echoing back our own query doesn't reset it mid-word. `initialSearchQuery` is renamed to `searchQuery`, since it's no longer mount-only; the prop is new in this PR, so nothing released depends on the old name.

`pnpm lint:json` 0 diagnostics, `pnpm typecheck` clean, `packages/admin` suite 1252 passed.

MA2153:
Ok, at this point the bot is just nitpicking

packages/admin/src/router.tsx
```diff
@@ -309,19 +310,77 @@ function DashboardPage() {
...
+function readCalendarDate(value: unknown): string | undefined {
+	const raw = readString(value);
+	return raw && CALENDAR_DATE_REGEX.test(raw) ? raw : undefined;
+}
+
+export function parseContentListSearch(search: Record<string, unknown>): ContentListSearch {
+	const sort = readOneOf(search.sort, SORT_FIELDS);
+
+	const rawPage = Number(search.page);
+	const page = Number.isInteger(rawPage) && rawPage > 1 ? rawPage : undefined;
```
khoinguyenpham04, line 362: Could we avoid replaying the full cursor chain for deep page URLs? On a fresh load, `?page=500` fetches every preceding API page and retains all rows. Could we bound the automatic loading or use cursor-backed URL state?

packages/admin/src/router.tsx
```diff
@@ -309,19 +310,77 @@ function DashboardPage() {
...
+
+function readOneOf<T extends string>(value: unknown, allowed: readonly T[]): T | undefined {
+	return typeof value === "string" && (allowed as readonly string[]).includes(value)
+		? (value as T)
+		: undefined;
+}
+
+function readCalendarDate(value: unknown): string | undefined {
+	const raw = readString(value);
+	return raw && CALENDAR_DATE_REGEX.test(raw) ? raw : undefined;
```
khoinguyenpham04: Could we also check that this is a real date? Values like 2025-99-99 match the format but are rejected by the API.

packages/admin/src/components/ContentList.tsx
```diff
@@ -204,14 +224,27 @@ export function ContentList({
...
 
-	// Reset page when search changes
+	// Adopt a query the caller changed on its own. Comparing against what we
+	// last reported keeps the caller's echo of our own term from resetting the
+	// input mid-word.
+	React.useEffect(() => {
+		if (controlledSearchQuery === reportedQuery.current) return;
+		reportedQuery.current = controlledSearchQuery;
+		setSearchQuery(controlledSearchQuery);
+	}, [controlledSearchQuery]);
```
khoinguyenpham04, line 248: Could we cancel the pending debounced search when navigating Back? Otherwise a quick Back action can be overwritten when the debounce completes.

khoinguyenpham04 (commented):
Thank you for tackling this. Could you please address these comments.

packages/admin/src/router.tsx
MA2153: Fixed in c6cca1a6 — good catch, `2025-02-31` was the worse one: it passed the regex and reached the API widened to a `2025-03-03T23:59:59.999Z` boundary, so the filter silently covered three extra days rather than erroring.

`readCalendarDate` now round-trips through `Date`, which is what rejects a day the calendar doesn't have — `new Date("2025-02-31")` rolls forward to March 3, so the parse alone isn't enough. Tests cover `2025-99-99`, `2025-13-01`, `2025-02-31`, and `2025-02-29`, plus `2024-02-29` staying accepted.

packages/admin/src/components/ContentList.tsx
MA2153: Fixed in c6cca1a6. Reproduced first, and the failing test is worth recording because the sequence is not obvious:

1. Click Next → `?page=2`.
2. Type in the search box. `handleSearchChange` resets to page 1, which pushes a history entry, so the URL is now bare.
3. Press Back before the 300ms settles → `?page=2` restored.
4. Debounce fires → `?q=Post+3`, and the restored page is gone.

The test asserted `?page=2` and got `?q=Post+3`.

The signal the component was missing is that the page moved without it asking. `setPage` now records what it requested, and a `controlledPage` that arrives differing from that — back/forward, or a deep link — re-seeds the input from the caller's query. Re-seeding restarts `useDebouncedValue`'s timer, so the stale term is dropped rather than cancelled explicitly.

This only covers a nav that moves the page. Back over a status-filter change with a search edit pending still lets the edit land; that needs a general "the URL moved underneath us" signal, which I'd rather not invent here.

packages/admin/src/router.tsx
MA2153: Agreed that this is worth solving, but I don't think it should be solved in this PR — and I don't think bounding the loading is the right shape either. Opened #2255 for it.

Short version of what I found while looking into it. The replay is a symptom of the URL storing an ordinal when we already have an anchor: `encodeCursor` produces `{ orderValue, id }` and `findMany` applies it as a comparison, never a row lookup, so it stays valid even if the anchored row is deleted. `?after=<cursor>` reads `limit` rows at page 2 and page 500 alike, restores in one request on back-navigation, and holds its window steady while other editors save — which the current design and offset pagination both fail to do.

It also collapses the two-tier pagination that causes the replay: the API pages at 100 while the table slices at 20, which is the only reason the auto-chaining effect exists. Match those and one URL page is one API page.

Two things it needs that aren't free — a `before` cursor in `findMany` for the ← button on a cold-loaded deep link, and a decision about the page-number label, since an anchor doesn't know it's the twelfth page. Both are in the Discussion.

Cursor-backed URL state is your other suggestion, so I think we agree on the direction; I'd rather get a maintainer read on the label trade-off before rewriting the list's pagination model.

On this PR specifically: the cost is bounded even today. `totalPages` comes from the server's `total` and `clampedPage` pins the page to it, so `?page=999999` settles onto the last real page after one fetch. What's left is a genuinely large collection reached by a deep bookmark, which is unchanged from what clicking there already costs.

MA2153:
Moved this to draft pending a maintainer call on #2255.

Both of the local findings from the review are fixed in c6cca1a6, each with a failing test first: `readCalendarDate` now rejects dates that fit `YYYY-MM-DD` but never happened, and a search term still inside the debounce no longer overwrites the page a Back press just restored.

What's holding it is the third comment. Chasing it turned up something bigger than the fix it asked for: the deep-page replay is a symptom of this PR storing an ordinal in the URL when the API already hands out a stable anchor. If #2255 lands the anchor direction, `?page=2` stops being the URL contract for this route and becomes `?after=<cursor>` — so merging this as-is would ship a query-param shape that a follow-up immediately changes, and admin URLs people have started bookmarking would break twice instead of once.

The alternatives I weighed and rejected: capping the restorable page in the parser clamps whether or not the rows are already cached, so a legitimate back-navigation from a deep page lands on the wrong one — the same class of bug this PR exists to fix. Bounding the auto-chaining by loaded data avoids that, but it's a real change to the list's pagination model, which is the thing #2255 is asking about.

So the choice is which URL shape this route gets, and that's a maintainer's call rather than something to settle inside a bug-fix PR. Happy to mark it ready again unchanged if the answer is "ship the page number, revisit later" — the reported back-button bug is fixed either way, and the deep-page cost is bounded today by the `total`-based clamp.

PR #2248, opened 2026-07-26
feat(nesting): column widths, draggable rows inside columns, and a collapsed state

bimsonz:
## What does this PR do?

Builds on @louisescher's nesting block prototype from #2049 with the three things that stopped it being usable once I put it in front of an editor: columns could not be given different widths, the blocks inside a column could not be reordered, and a container could not be folded away.

**This is opened as a draft, and is not trying to land ahead of @louisescher.** His prototype commit is the base of this branch and stays attributed to him. It is opened against `main` rather than his branch because CI and a review pass only run on a PR to `main`. If he would rather these commits went onto his branch, or into a PR of his own, say so and I will move them and close this.

Related: https://github.com/emdash-cms/emdash/discussions/2049

### Column widths

Equal columns cannot express a content and sidebar page, which is the most common two column layout there is. `repeat(N, minmax(0, 1fr))` forces 50/50 at two columns, so a sidebar comes out the same width as the main content.

Adds a `widths` attribute with presets: `equal`, `wide-first`, `wide-last`, `narrow-first`, `narrow-last`. The weighted column takes 2fr against 1fr for the rest. Unrecognised values fall back to `equal`, and the existing collapse to a single column under 768px is unaffected. The control is disabled for a flex container, where columns size from their content and the site renderer ignores widths too.

Worth knowing before adding any further attributes: the admin carries its own Portable Text converters in `PortableTextEditor.tsx`, separate from `packages/core/src/content/converters/`. An attribute added to core alone passes core's round trip tests and is then dropped by every real save. That happened here, and the added test asserts a full PT to PM to PT cycle rather than a single direction.

### Blocks inside a column

A block in a column had no drag handle, so it could be typed into but not reordered. TipTap only considers top level nodes unless nested targeting is on, and a block in a column is three levels deep.

**The unit question is the part I am least sure about.** With `nested` on, the default rules resolve the unit *inside* a structure: `listItemFirstChild` and `listWrapperDeprioritize` between them exclude a list's paragraph and its wrapper so the list item wins. That is right for a plain document. For a page built from containers a list is one row the page is made of, so this replaces them with a rule where a row is a child of the document or a child of a column.

That restates what the editor does today with `nested` off, and extends it one level down, so behaviour outside a container does not change. It also keeps a list movable as a block, which taking the defaults does not allow anywhere in the document once `nested` is on. Nothing the defaults guard is lost, because table internals and inline content are never children of the document or of a column; the tests assert that rather than assuming it, with a table cell resolving to the table and a quote's paragraph to the quote.

If you would rather keep your unit, I am happy to rework it. It is a choice about what a page is made of, not a claim that the defaults are wrong.

Two smaller pieces follow from the same decision:

- **The handle's gutter belongs to the row, not the column.** Padding the column leaves the gutter outside every row's box, so a pointer there resolves to the column, which is not a drag target, and the container wins. In use that reads as the handle jumping away exactly as you reach for it. A row that indents its own content, a list or a quote, keeps that indent on top of the gutter, otherwise its markers or rule are drawn into the gutter under the handle.
- **Edge detection is off.** It resolves ambiguity by pointer position, deducting by depth near a node's edge. With rows as the unit the only candidates are the row and its container, and the container has an explicit grab point in its header, so there is nothing left to resolve. Leaving it on excludes a candidate outright at the depth a row sits inside a column, and rows are often short enough that the 12px band covers half of one.

The container keeps a grab point of its own, since hovering it resolves to the deepest row inside it, but it is the header rather than a permanently visible grip.

Two duplicate grips are removed while here. Plugin and HTML blocks each drew their own at `-start-8`, in a gutter that does not exist inside a column, on top of the handle the editor already provides and the node wrapper's own drag target. Happy to split that out if you would rather keep this to nesting.

### Folding a container away

A container is a large open box with no way to close it, so a page built from several cannot be scanned. Collapsed, it is one row reading its own name and what it holds, "2 columns, 9 blocks".

Collapsed is view state rather than a node attribute, so folding a container away is not a change to the document and does not land in a revision. The content is hidden rather than unmounted, because ProseMirror owns that element as the node's contentDOM.

The counts sit outside the translated words rather than in a plural message. A `<Plural>` renders its own ICU source until catalogues are extracted, which looks broken in the editor and in visual regression screenshots, and extracting catalogues is not something a feature PR should carry. The trade is that the summary is not plural agreed for richer locales, which is easy to change once the catalogue exists.

### Not addressed here

- **A container inside a column** is allowed by the schema and the renderer recurses, but I have not driven it in the admin.
- **Drag to reorder has no keyboard path**, here or anywhere else in the editor. Not introduced by this change, but containers make it matter more.
- **The container cannot express a page grid with named regions**, so a sidebar spanning specific rows is out, the way `grid-row: 1 / span 3` would do it. That is a ceiling rather than a bug and is probably worth stating in the RFC.

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] User-visible strings in the admin UI are wrapped for translation. No `messages.po` changes are included.
- [x] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/2049

- [x] This PR includes AI-generated code -- model/tool: Claude Opus 5

## Screenshots / test output

- `@emdash-cms/admin`: 1224 passing across 100 files, 0 failing.
- `emdash`: 4976 passing, 1 failing. The failure is `tests/unit/astro/integration/virtual-modules.test.ts`, which fails identically without this change, so it is not from it.
- `oxlint --type-aware --deny-warnings` clean, `oxfmt --check` clean, `tsgo --noEmit` clean on both packages.

Worth flagging separately: `packages/admin/tests/components/SeoPanel.test.tsx` and `tests/publish-button-locale.test.tsx` each failed during full suite runs, a different test each time, and pass repeatedly in isolation. Both are untouched by this change and look non-deterministic under parallel browser load.

Behaviour was checked by driving the admin in a browser, not only through tests. Two columns measured 296/296 before the widths change and 394/197 after. With the gutter on the row, a pointer at x=801 beside a heading puts the handle at x=781, aligned to that heading and inside the row's own padding; before, the same pointer targeted the container.

### DIFF
--- a/.changeset/nesting-column-widths-and-rows.md
+++ b/.changeset/nesting-column-widths-and-rows.md
@@ -1,0 +1,6 @@
+---
+"@emdash-cms/admin": minor
+"emdash": minor
+---
+
+Nesting blocks gain column width ratios, so a container can express a content and sidebar layout rather than only equal columns. Blocks inside a column become first-class rows that can be reordered by dragging, and a container can be folded away to a one line summary of what it holds.
--- a/packages/admin/src/components/PortableTextEditor.tsx
+++ b/packages/admin/src/components/PortableTextEditor.tsx
@@ -85,6 +85,7 @@ import {
 	DotsSixVertical,
 	CaretDown,
 	type Icon,
+	ColumnsIcon,
 } from "@phosphor-icons/react";
 import { X } from "@phosphor-icons/react";
 import { Extension, type Range } from "@tiptap/core";
@@ -119,6 +120,7 @@ import { HeadingDropdownMenu } from "./editor/HeadingDropdownMenu";
 import { HtmlBlockExtension } from "./editor/HtmlBlockNode";
 import { ImageExtension } from "./editor/ImageNode";
 import { MarkdownLinkExtension } from "./editor/MarkdownLinkExtension";
+import { NestingBlockExtension, NestingColumnExtension } from "./editor/NestingBlockNode";
 import {
 	type PluginBlockDef,
 	PluginBlockExtension,
@@ -243,6 +245,18 @@ function sanitizeGalleryImages(value: unknown, withKeys = false): GalleryImage[]
 // Helpers for safely extracting typed values from ProseMirror attrs (Record<string, any>)
 const attrStr = (v: unknown): string | undefined => (typeof v === "string" && v ? v : undefined);
 const attrNum = (v: unknown): number | undefined => (typeof v === "number" && v ? v : undefined);
+
+// Nesting block layout coercion
+const NESTING_GAPS = ["none", "sm", "md", "lg"] as const;
+const NESTING_ALIGNS = ["start", "center", "end", "stretch"] as const;
+
+function pickNestingGap(v: unknown): (typeof NESTING_GAPS)[number] {
+	return NESTING_GAPS.find((g) => g === v) ?? "md";
+}
+
+function pickNestingAlign(v: unknown): (typeof NESTING_ALIGNS)[number] {
+	return NESTING_ALIGNS.find((a) => a === v) ?? "start";
+}
 
 // ProseMirror to Portable Text converter
 function prosemirrorToPortableText(doc: {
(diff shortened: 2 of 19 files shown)


bimsonz:
Thanks, this was a useful review. All five fixed, plus the comments you did not flag.

**i18n.** Taken, with `#` placeholders to match the other `plural` calls in the admin; `byline-schema.tsx` already nests one inside `t` the same way. I have not committed `messages.po`. AGENTS.md says to keep catalogue changes out of non-translation PRs since `auto-extract.yml` runs on merge, so I ran the extract locally to confirm the message comes out right and then reverted it. Happy to include it if you would rather.

**Comments.** Fixed the four, then went through every comment this branch adds and cut the rest to the same standard, about 120 lines.

One deviation worth flagging. I kept a single line on `_nestedDragOptions`:

```
/** Module level: DragHandle re-registers its plugin if this identity changes. */
```

The same file already carries that note on `computePositionConfig`, so it read as an invariant the file documents rather than narrative. Say the word and it goes.

Also rebased onto main, which resolves the conflict in the two converters (gallery and nesting each added a case, both kept) and clears the query-count diff, which was this branch sitting behind `f81aa684`.

PR #2250, opened 2026-07-27
fix: exclude image/svg+xml from default media upload allowlist

marcusbellamyshaw-cell:
## What does this PR do?

Excludes `image/svg+xml` from the default media upload allowlist
(`GLOBAL_UPLOAD_ALLOWLIST`), which previously admitted it via a bare
`"image/"` prefix match. There's no upload-time content validation for SVG
anywhere in core, so a Contributor+ user could upload an SVG with an
embedded `<script>` tag. The allowlist now enumerates the safe raster types
explicitly (`image/png`, `image/jpeg`, `image/gif`, `image/webp`) instead of
prefix-matching all of `image/*`. Fields that explicitly configure
`image/svg+xml` in their own `allowedMimeTypes` are unaffected — this only
changes the *default* fallback.

Found during a site-level pentest of an Emdash-based site (everybittexas.com)
— flagged as a Low-severity finding since the current serve-time route
already excludes SVG from `SAFE_INLINE_TYPES`, forcing
`Content-Disposition: attachment` + a sandboxed CSP, so it wasn't directly
exploitable through the known serve path. Still worth closing at the
upload-time boundary rather than relying solely on serve-time headers as a
single point of failure. No related issue open upstream — filing directly.

## Type of change

- [x] Bug fix

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

## Screenshots / test output

New test `packages/core/tests/unit/media/media-allowlist.test.ts` — 3 tests,
confirms SVG rejected, PNG/JPEG/GIF/WebP/video/audio/PDF still allowed.
Full `tests/unit/media/**` + the two `media-upload-*` integration tests:
14 files, 151 tests, all passing.


ascorbic (approved):
Good catch

PR #2257, opened 2026-07-28
fix: make FTS5 tokenizer configurable per collection

### LINKED ISSUE #2245: Search: FTS5 tokenizer hardcoded to 'porter unicode61' — full-text search silently returns nothing for non-space-delimited languages (ja/zh/th/km/lo/my)
## Summary

`FTSManager.createFtsTable()` hardcodes `tokenize='porter unicode61'` with no way to override it. This makes full-text search return **zero results for most queries** on content written in languages that don't put spaces between words — Japanese, Chinese, Thai, Khmer, Lao, Burmese. It is not a Japanese-only edge case; it affects every non-spacing script, and the `porter` stemmer half of the setting is English-only and applied unconditionally to every site.

`unicode61` only breaks tokens on non-alphanumeric characters and script boundaries, so a whole sentence of Japanese or Thai collapses into a single token. `escapeQuery()` (packages/core/src/search/query.ts) turns user input into a prefix query `"term"*`, which means a hit only happens when the search term is at the **beginning** of one of those sentence-sized tokens. Search appears to work when someone tests with a word from the start of a title, then silently returns nothing for the same word appearing mid-sentence.

## Real-world impact

Production site on EmDash 0.31.0 / Cloudflare D1. 11 published `news` entries contain 字幕 ("subtitles"):

```sql
SELECT count(*) FROM ec_news WHERE content LIKE '%字幕%';            -- 11
SELECT count(*) FROM _emdash_fts_news WHERE _emdash_fts_news MATCH '字幕';  -- 0
```

The FTS table was built by EmDash itself:

```
CREATE VIRTUAL TABLE "_emdash_fts_news" USING fts5(
    id UNINDEXED, locale UNINDEXED, title, summary, content,
    content='ec_news', content_rowid='rowid', tokenize='porter unicode61')
```

## Reproduction

Minimal, no EmDash needed — stock SQLite with FTS5 (run with `node --experimental-sqlite`, or any sqlite3 build with FTS5):

```js
import { DatabaseSync } from "node:sqlite";
const db = new DatabaseSync(":memory:");
db.exec(`CREATE VIRTUAL TABLE u USING fts5(title, content, tokenize='porter unicode61')`); // what EmDash creates
db.exec(`CREATE VIRTUAL TABLE t USING fts5(title, content, tokenize='trigram')`);          // for comparison

const docs = [
  ["ja", "AI字幕生成の実務ガイド", "放送局における字幕制作のワークフローを解説します。"],
  ["zh", "字幕生成的最佳实践", "我们介绍广播电视台的字幕制作流程。"],
  ["ko", "자막 생성 실무 가이드", "방송국의 자막 제작 워크플로를 설명합니다."],
  ["th", "คู่มือการสร้างคำบรรยาย", "อธิบายขั้นตอนการผลิตคำบรรยายของสถานีโทรทัศน์"],
  ["en", "Subtitle generation guide", "We explain the subtitle production workflow at broadcasters."],
  ["de", "Leitfaden zur Untertitelerstellung", "Wir erklären den Untertitel-Produktionsworkflow bei Sendern."],
];
const iu = db.prepare(`INSERT INTO u VALUES(?,?)`), it = db.prepare(`INSERT INTO t VALUES(?,?)`);
for (const [, ti, co] of docs) { iu.run(ti, co); it.run(ti, co); }

// `"term"*` is exactly the shape escapeQuery() produces
for (const q of ["字幕", "字幕制作", "广播", "자막", "방송국", "คำบรรยาย", "subtitle", "Untertitel"]) {
  const n = (tbl) => db.prepare(`SELECT count(*) c FROM ${tbl} WHERE ${tbl} MATCH ?`).get(`"${q}"*`).c;
  console.log(q, "unicode61:", n("u"), " trigram:", n("t"));
}
```

| lang | query | meaning | `porter unicode61` (current) | `trigram` |
| --- | --- | --- | --- | --- |
| ja | 字幕 | subtitles | 1 — only because the title starts `AI` + script boundary | 0 |
| ja | 字幕制作 | subtitle production | **0** (mid-sentence) | 2 |
| zh | 字幕 | subtitles | 1 — title-initial only | 0 |
| zh | 广播 | broadcast | **0** (mid-sentence) | 0 |
| ko | 자막 | subtitles | 1 | 0 |
| ko | 방송국 | broadcaster | 1 | 1 |
| th | คำบรรยาย | subtitles | **0** | 1 |
| en | subtitle | — | 1 | 1 |
| de | Untertitel | — | 1 | 1 |

Korean survives because Hangul is space-delimited (the prefix operator absorbs the agglutinated particles). Japanese, Chinese and Thai only ever match by accident, when the term lands at the start of a token.

## Root cause

https://github.com/emdash-cms/emdash/blob/main/packages/core/src/search/fts-manager.ts#L98

```ts
CREATE VIRTUAL TABLE IF NOT EXISTS "${ftsTable}" USING fts5(
    ${columns},
    content='${contentTable}',
    content_rowid='rowid',
    tokenize='porter unicode61'
)
```

Same literal in `packages/core/src/database/migrations/039_fix_fts5_triggers.ts#L194`. There is no config path, env var, or schema field that reaches it.

## Why simply switching the default to `trigram` isn't enough

`trigram` gives real substring matching (see the table — it fixes exactly the mid-sentence cases `unicode61` misses), but it cannot match queries shorter than 3 characters. Two-character words are the backbone of Japanese and Chinese vocabulary: 字幕 (subtitles), 放送 (broadcast), 映像 (video), 音声 (audio), 广播, 视频. A blanket switch would trade one class of silent zero-results for another.

So the fix probably isn't a new hardcoded default — it's making the choice reachable.

## Suggested fix

Expose the tokenizer as configuration. There's already a natural home for it: `SearchConfig` (packages/core/src/search/types.ts#L10) is persisted per collection in `_emdash_collections.search_config`, so adding `tokenize?: string` needs no schema migration.

```ts
export interface SearchConfig {
	enabled: boolean;
	weights?: Record<string, number>;
	tokenize?: string; // defaults to "porter unicode61"
}
```

Points worth handling:

1. **A site-level default plus per-collection override.** Sites are often single-language; setting it once should be enough.
2. **Changing it has to drop and recreate the table.** `CREATE VIRTUAL TABLE IF NOT EXISTS` makes the tokenizer permanently sticky once the table exists, so a config change needs `dropFtsTable()` + `rebuildIndex()`, not just a new `CREATE`.
3. **`porter` should be opt-out.** It's an English stemmer; on a German, French or Japanese site it's at best inert and at worst wrong.
4. **Document the limitation either way.** Right now nothing in the docs suggests full-text search behaves differently by language, so the failure mode is silent — searches just return nothing and site owners assume the index is broken or empty.

Happy to send a PR for the `SearchConfig` route if that's the direction you'd take.

## Environment

- emdash 0.31.0
- Cloudflare D1 (production) — reproduced there against the live FTS tables
- Also reproduced on stock SQLite FTS5 via `node:sqlite` (Node 22.22.3)

mvanhorn:
## What does this PR do?

`FTSManager.createFtsTable()` creates every SQLite FTS5 table with `tokenize='porter unicode61'`. Because `unicode61` treats contiguous non-spacing text as a single token and EmDash emits prefix `MATCH` expressions, mid-sentence queries in Japanese, Chinese, Thai, Khmer, Lao, and Burmese silently miss indexed content. The tokenizer is not configurable, so there is no way out of it today.

This adds a backward-compatible optional `tokenize` field to the exported per-collection `SearchConfig`, constrained to the supported built-in configurations (`porter unicode61`, `unicode61`, `trigram`) so user input is never interpolated into FTS DDL unchecked. Omitting it retains the current English-stemming default, so existing collections are unaffected.

The selected tokenizer is threaded through enable, explicit rebuild, schema-triggered rebuild, repair, disable/re-enable persistence, and config parsing, so posting an updated enabled configuration drops, recreates, and repopulates the existing FTS table rather than leaving its tokenizer sticky. The option is exposed through the validated search-enable REST body, with docs on when to choose each tokenizer (Porter's English-only stemming, trigram's three-character minimum).

Closes #2245

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

Two checklist items are deliberately not ticked, with what I actually observed:

- `pnpm typecheck` — fails in my checkout with `TS2307: Cannot find module '@emdash-cms/blocks'` / `'@emdash-cms/plugin-types'`, i.e. unbuilt workspace siblings rather than anything in this diff. Worth a maintainer confirming it is green in CI.
- `pnpm lint` — fails on `demos/simple`, `fixtures/perf-site`, `infra/cache-demo`, `infra/do-solo-demo`, and `packages/registry-client` with `no-redundant-type-constituents`. I verified those same failures reproduce on a pristine `main` checkout, and that `oxlint --type-aware --deny-warnings` is clean when run over only the files this PR changes.

AI was used for assistance.

## Screenshots / test output

Targeted tests for this change:

```
$ pnpm --filter ./packages/core exec vitest run \
    tests/integration/search/tokenizer.test.ts tests/unit/api/search-schema.test.ts

 Test Files  2 passed (2)
      Tests  11 passed (11)
```

### DIFF
--- a/.changeset/fix-search-tokenizers.md
+++ b/.changeset/fix-search-tokenizers.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Adds per-collection search tokenizers so multilingual content can use trigram or non-stemming Unicode search.
--- a/docs/src/content/docs/reference/rest-api.mdx
+++ b/docs/src/content/docs/reference/rest-api.mdx
@@ -750,13 +750,47 @@ GET /_emdash/api/search/suggest?q=hel&limit=5
 
 Returns prefix-matched titles for autocomplete.
 
+### Configure Collection Search
+
+```http
+POST /_emdash/api/search/enable
+Content-Type: application/json
+
+{
+  "collection": "posts",
+  "enabled": true,
+  "tokenize": "trigram",
+  "weights": {
+    "title": 10,
+    "content": 1
+  }
+}
+```
+
+The optional `tokenize` field controls how SQLite FTS5 indexes the collection. Changing it on an
+enabled collection rebuilds and repopulates that collection's search index.
+
+| Value                | When to use |
+| -------------------- | ----------- |
+| `porter unicode61`   | Default. English-language content that benefits from Porter stemming, such as matching related word forms. Porter stemming is English-specific. |
+| `unicode61`          | Languages that use word separators but should not use English stemming. |
+| `trigram`            | Languages with text that is not separated by spaces, including Japanese, Chinese, Thai, Khmer, Lao, and Burmese, or when substring matching is needed. Queries shorter than three Unicode characters return no matches. |
+
+Omitting `tokenize` on a collection without a stored tokenizer uses `porter unicode61`. Disabling
+search preserves the configured tokenizer for the next enable operation.
+
 ### Rebuild Search Index
 
 ```http
 POST /_emdash/api/search/rebuild
+Content-Type: application/json
+
+{
+  "collection": "posts"
+}
 ```
 
-Rebuild FTS index for all or specific collections.
+Rebuilds the FTS index for the specified collection using its stored tokenizer and field weights.
 
 ### Search Stats
 
--- a/packages/core/src/api/schemas/search.ts
+++ b/packages/core/src/api/schemas/search.ts
@@ -1,5 +1,6 @@
 import { z } from "zod";
 
+import { SEARCH_TOKENIZERS } from "../../search/types.js";
 import { localeCode } from "./common.js";
 
 // ---------------------------------------------------------------------------
@@ -37,6 +38,7 @@ export const searchEnableBody = z
 		collection: z.string().min(1),
 		enabled: z.boolean(),
 		weights: z.record(z.string(), z.number()).optional(),
+		tokenize: z.enum(SEARCH_TOKENIZERS).optional(),
 	})
 	.meta({ id: "SearchEnableBody" });
 
--- a/packages/core/src/astro/routes/api/search/enable.ts
+++ b/packages/core/src/astro/routes/api/search/enable.ts
@@ -21,6 +21,7 @@ export const prerender = false;
  * - collection: Collection slug (required)
  * - enabled: boolean (required)
  * - weights: Optional field weights for ranking
+ * - tokenize: Optional FTS5 tokenizer configuration
  */
 export const POST: APIRoute = async ({ request, locals }) => {
 	const { emdash, user } = locals;
@@ -40,7 +41,10 @@ export const POST: APIRoute = async ({ request, locals }) => {
 	try {
 		if (body.enabled) {
 			// Enable search - creates FTS table, triggers, and populates index
-			await ftsManager.enableSearch(body.collection, { weights: body.weights });
+			await ftsManager.enableSearch(body.collection, {
+				weights: body.weights,
+				tokenize: body.tokenize,
+			});
 
 			const stats = await ftsManager.getIndexStats(body.collection);
 
--- a/packages/core/src/astro/routes/api/search/rebuild.ts
+++ b/packages/core/src/astro/routes/api/search/rebuild.ts
@@ -57,7 +57,12 @@ export const POST: APIRoute = async ({ request, locals }) => {
 		}
 
 		// Rebuild the index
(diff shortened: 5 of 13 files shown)


ascorbic (approved):
Thanks

PR #2280, opened 2026-07-29
fix: prefer uncached Hyperdrive after content writes

### LINKED ISSUE #2302: [Cache] Purge is inconsistent for specific types of updates
### Description

Broadly, the Workers Cache and Object Cache work great, but there are specific instances in which they don't consistently apply for edits made in the admin panel.

One way of approaching this is #2297, but another would be looking into the logic around cache clearing to solve the underlying issues:

- Update OG image
- Update the excerpt
- Update images within the body of a post (but no other text)
- Update HTML blocks within the body of a post (but no other text)

_Feels_ like an issue with the object cache, but that's for y'all to determine.

### Steps to reproduce

1. Create an EmDash worker with the Workers Cache / Object Cache enabled.
2. Make a content entry and then make updates to one of the items noted above.
3. Look at the published output.

If you do this a few times you can usually repro.

This _might_ also be solved by #2280, but worth having a dedicated issue in case it isn't (specifically in the excerpt / images / SEO items).

### Environment

- emdash: 0.31.0
- Cloudflare Workers + Hyperdrive + Object cache and workers cache enabled

### Logs / error output

```shell

```

scottbuscemi:
## What does this PR do?

Prevents anonymous public requests from reseeding Workers/Object Cache with stale Hyperdrive query-cache results immediately after content publishes. Content invalidation records a shared last-write timestamp, and Hyperdrive public reads temporarily use the uncached primary binding for a configurable window (60 seconds by default).

Cross-isolate coordination uses the configured distributed Object Cache backend. `preferUncachedAfterWriteMs` should match the cached Hyperdrive configuration's `max_age`.

Closes #2302

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

The i18n and Discussion checklist items are not applicable: this changes no admin UI strings and fixes an existing bug.

- [x] This PR includes AI-generated code — model/tool: Grok 4.5 / OpenCode; GPT-5 Codex

## Screenshots / test output

Non-visual change. Verified after rebasing onto current `main` with:

- `pnpm build`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm format`
- Core Object Cache and middleware tests: 51 passed
- Cloudflare Hyperdrive routing/config tests: 25 passed

---

### Try this PR

**[Open a fresh playground →](https://fix-prefer-uncached-hyperdrive-after-write-emdash-playground.emdash-cms.workers.dev)**

<sub>Tracks `fix/prefer-uncached-hyperdrive-after-write`. Updated automatically when the playground redeploys.</sub>

### DIFF
--- a/packages/cloudflare/src/db/hyperdrive.ts
+++ b/packages/cloudflare/src/db/hyperdrive.ts
@@ -81,7 +81,17 @@ interface HyperdriveConfig {
 	 * authenticated requests and all writes stay on `binding`.
 	 */
 	cachedBinding?: string;
+	/**
+	 * After a content publish, prefer the primary uncached binding for this
+	 * many ms on anonymous public reads. Defaults to 60_000 when
+	 * `cachedBinding` is set (Hyperdrive's default `max_age`); 0 otherwise.
+	 * Set equal to your cached Hyperdrive config's `max_age`.
+	 */
+	preferUncachedAfterWriteMs?: number;
 }
+
+/** Hyperdrive default query-cache max_age when caching is enabled. */
+const DEFAULT_PREFER_UNCACHED_AFTER_WRITE_MS = 60_000;
 
 /**
  * Minimal shape of a Hyperdrive binding. Workers inject `connectionString`
@@ -147,6 +157,8 @@ export interface RequestScopedDbOpts {
 	isWrite: boolean;
 	cookies: CookieJar;
 	url: URL;
+	/** ms-epoch of last content-namespace invalidation; from core object-cache. */
+	lastContentWriteAt?: number;
 }
 
 export interface RequestScopedDb {
@@ -182,6 +194,7 @@ export function createRequestScopedDb(opts: RequestScopedDbOpts): RequestScopedD
 		isAuthenticated: opts.isAuthenticated,
 		isWrite: opts.isWrite,
 		url: opts.url,
+		lastContentWriteAt: opts.lastContentWriteAt,
 	});
 	let binding = getBinding(bindingName);
 	// If the cached binding was selected but isn't present at runtime (e.g.
@@ -242,6 +255,17 @@ function isEmDashInternalPath(url: URL): boolean {
 }
 
 /**
+ * Effective window (ms) to prefer the primary binding after a content write.
+ * Defaults to Hyperdrive's 60s `max_age` when `cachedBinding` is set.
+ */
+function preferUncachedDurationMs(config: HyperdriveConfig): number {
+	if (config.preferUncachedAfterWriteMs !== undefined) {
+		return config.preferUncachedAfterWriteMs;
+	}
+	return config.cachedBinding ? DEFAULT_PREFER_UNCACHED_AFTER_WRITE_MS : 0;
+}
+
+/**
  * Decide which binding a given request should use.
  *
  * The cache-enabled binding (when configured) is used only for **anonymous
@@ -255,12 +279,20 @@ function isEmDashInternalPath(url: URL): boolean {
  *   even an anonymous GET → primary. The setup-status and login-state reads
  *   are anonymous GETs that must see writes made moments earlier; routing them
  *   to the cache would loop the setup wizard and show stale auth state.
+ * - anonymous public reads within `preferUncachedAfterWriteMs` of the last
+ *   content-namespace invalidation → primary, so a post-publish rebuild does
+ *   not reseed edge/object caches from Hyperdrive's still-stale query cache.
  *
  * Pure (no I/O) so the routing rule can be unit-tested directly.
  */
 export function selectBindingName(
 	config: HyperdriveConfig,
-	opts: { isAuthenticated: boolean; isWrite: boolean; url: URL },
+	opts: {
+		isAuthenticated: boolean;
+		isWrite: boolean;
+		url: URL;
+		lastContentWriteAt?: number;
+	},
 ): string {
 	if (
 		config.cachedBinding &&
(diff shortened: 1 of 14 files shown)


packages/core/src/object-cache/index.ts
```diff
@@ -474,13 +530,102 @@ export async function isObjectCacheActive(): Promise<boolean> {
...
+/**
+ * ms-epoch of the last content-namespace invalidation (`content:*`), or `0`
+ * if unknown. Returns `max(local, backend)` so a warm isolate that just
+ * published is immediately correct, and cold isolates learn within their
+ * `revalidate` window after another isolate stamped the backend.
+ */
+export async function getLastContentWriteAt(): Promise<number> {
+	const local = lastContentWrite.value;
+	const now = Date.now();
+	if (local > 0 && now - lastContentWrite.at < holder.config.revalidate) {
```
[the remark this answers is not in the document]
scottbuscemi: Fixed in 20b7135 — confirmed `0` is now cached within the revalidate window (same as missing epochs), with a regression test.

packages/cloudflare/tests/hyperdrive-config.test.ts
```diff
@@ -42,4 +42,23 @@ describe("hyperdrive()", () => {
 			cachedBinding: "HYPERDRIVE_CACHED",
 		});
 	});
+
+	it("omits preferUncachedAfterWriteMs from the descriptor when not provided", () => {
+		const result = hyperdrive({ binding: "HYPERDRIVE", cachedBinding: "HYPERDRIVE_CACHED" });
+		expect(result.config).not.toHaveProperty("preferUncachedAfterWriteMs");
+	});
```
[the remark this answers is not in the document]
scottbuscemi: Done in 20b7135 — removed the config-pin test. Default/omit behavior is covered by the routing tests.

packages/core/src/astro/middleware.ts
```diff
@@ -647,12 +650,14 @@ export const onRequest = defineMiddleware(async (context, next) => {
 				// Even on the anonymous fast path we ask the adapter for a per-request
 				// scoped db. For D1 with read replication this routes anonymous reads
 				// to the nearest replica; for other adapters it's a no-op.
+				const lastContentWriteAt = await getLastContentWriteAt();
```
ascorbic, line 653: This runs on every logged-out request for every adapter, but only Hyperdrive with `cachedBinding` ever consumes the value. On the recommended D1 + KV setup that's a blocking backend read ahead of render once per revalidate window (1s default, 2s timeout worst case) per isolate, when most sites don't need it.

Gate the fetch on the adapter actually wanting it  e.g. a flag on `DatabaseDescriptor` so middleware doesn't need to know adapter config shapes, and skip it entirely when unset. Same applies to the call in the general request path below.

packages/core/src/astro/middleware.ts
```diff
@@ -346,6 +347,7 @@ async function runOutsideRequest<T>(
 ): Promise<T> {
 	const runtime = await getRuntime(config);
 
+	const lastContentWriteAt = await getLastContentWriteAt();
```
ascorbic, line 350: This one can just be dropped: `runOutsideRequest` passes `isWrite: true`, and `selectBindingName` never consults `lastContentWriteAt` on the write path. It's a wasted backend read on every cron tick.

packages/core/src/index.ts
```diff
@@ -199,6 +199,7 @@ export { EmDashStorageError } from "./storage/types.js";
 // Object cache (distributed read-through query cache)
 export {
 	cachedQuery,
+	getLastContentWriteAt,
```
ascorbic, line 202: Agree with the open thread here but there's more: middleware imports this via a relative path, so the export has no consumers - it's permanent public API surface added without needing to. Remove it from the barrel.

packages/cloudflare/tests/hyperdrive-config.test.ts
```diff
@@ -42,4 +42,18 @@ describe("hyperdrive()", () => {
 			cachedBinding: "HYPERDRIVE_CACHED",
 		});
 	});
+
+	it("passes through preferUncachedAfterWriteMs", () => {
```
ascorbic, line 46: On the open thread about this being a config-pin test I half agree, but note the routing tests construct config objects directly and never go through the `hyperdrive()` builder, so this is currently the only test that would catch a typo in the conditional spread. Rather than delete it, replace it with one test through the real chain: `createRequestScopedDb(hyperdrive({...}).config, { ..., lastContentWriteAt: recent })` asserting the primary pool. That tests behaviour and keeps the builder link covered.

packages/cloudflare/src/db/hyperdrive.ts
```diff
@@ -255,19 +279,32 @@ function isEmDashInternalPath(url: URL): boolean {
  *   even an anonymous GET → primary. The setup-status and login-state reads
  *   are anonymous GETs that must see writes made moments earlier; routing them
  *   to the cache would loop the setup wizard and show stale auth state.
+ * - anonymous public reads within `preferUncachedAfterWriteMs` of the last
+ *   content-namespace invalidation → primary, so a post-publish rebuild does
+ *   not reseed edge/object caches from Hyperdrive's still-stale query cache.
  *
  * Pure (no I/O) so the routing rule can be unit-tested directly.
```
ascorbic, line 286: Nit: this doc line is no longer true: the new window check calls `Date.now()` inside. Taking `now` in `opts` would make it pure and make the window tests deterministic instead of wall-clock-relative.

ascorbic (changes requested):

ascorbic:
This is a good additioon, and pretty close to ready. Just a few tidiness and perf changes

ascorbic (approved):
Previous review requests are addressed; exact-head bot review and CI are green.

PR #2281, opened 2026-07-29
Revert "feat: cache purge API (object cache + native Workers Caching)"

scottbuscemi:
## What does this PR do?

Reverts #2275, which was merged before maintainer approval. This removes the unreleased cache-purge APIs and restores `main` to its previous behavior.

This is an exact revert of `e88655417db5753a32d65d5bf5202af145d6ede0`. The unrelated `users:read` change bundled into #2275 is also reverted and can be submitted separately with a reproducing test.

Reverts #2275.

## Type of change

- [x] Bug fix
- [ ] Feature (requires [maintainer-approved Discussion](https://github.com/emdash-cms/emdash/discussions/categories/ideas))
- [ ] Refactor (no behavior change)
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests
- [ ] Chore (dependencies, CI, tooling)

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md)
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (or targeted tests for my change)
- [x] `pnpm format` has been run
- [ ] I have added/updated tests for my changes (if applicable)
- [ ] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable). Do not include `messages.po` changes except in translation PRs — a workflow extracts catalogs on merge to `main`.
- [ ] I have added a [changeset](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#changesets) (if this PR changes a published package)
- [ ] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/...

Tests are not applicable because this removes an unreleased feature and its tests. There are no new user-visible strings. No replacement changeset is needed because #2275 was not released; its pending changeset is removed so unchanged published packages are not bumped. This is not a new feature, so no Discussion is required.

## AI-generated code disclosure

- [ ] This PR includes AI-generated code — model/tool:

No new code was generated; the change is an exact Git-generated revert.

## Screenshots / test output

The full CI suite passes on `70c9dd660031c3bfd005864bb69172bf3522a62a`, including typecheck, lint, unit tests, browser tests, and Node and Cloudflare end-to-end tests. There are no visual changes to demonstrate.

PR #2297, opened 2026-07-30
feat: cache purge API (object cache + native Workers Caching)

scottbuscemi:
## What does this PR do?

Adds a first-party way for admins and sandboxed plugins to clear EmDash caches.

### CMS object cache (KV / memory)
- **Admin API:** `GET` / `POST` `/_emdash/api/admin/cache/object` (`settings:manage`)
  - GET reports whether an object-cache backend is configured
  - POST bumps epochs for fixed namespaces + content collections (or a selected subset)
- **Plugin:** `ctx.cache.getObjectCacheStatus()` / `ctx.cache.purgeObjectCache()`

### Workers Caching (native edge page cache)
- **Admin API:** `GET` / `POST` `/_emdash/api/admin/cache/workers` (`settings:manage`)
  - GET reports whether native `cache.purge` is available
  - POST calls `cache.purge({ purgeEverything: true })` or `cache.purge({ pathPrefixes })`
- **Plugin:** `ctx.cache.getWorkersCacheStatus()` / `ctx.cache.purgeWorkersCache({ pathPrefixes? })`
- Uses platform `cache.purge` via `virtual:emdash/workers-cache` (no zone ID / API token)
- Status `configured: true` when `cache.purge` is a function (production Workers); local workerd may expose `cache` without `purge`

### Shared
- Capability `cache:purge` gates both in-process and through Cloudflare / workerd sandbox bridges
- Block Kit buttons support optional `disabled` and `title` (tooltip)

Re-opened after #2275 was reverted in #2281 so this can go through normal review.

Closes #

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] New features link to an approved Discussion: n/a (EmDash maintainer)

- [x] This PR includes AI-generated code — model/tool: opencode + xAI grok-4.5

## Screenshots / test output

Targeted tests: object-cache + workers-cache handlers/routes, virtual module generator, marketplace `CAPABILITY_LABELS`, blocks disabled+title.

Preferred site setup (see also #2277):

```jsonc
// wrangler.jsonc
{ "cache": { "enabled": true } }
```

```js
// astro.config.mjs
import { cacheCloudflare } from "@astrojs/cloudflare/cache";
cache: { provider: cacheCloudflare() },
```

---

### Try this PR

**[Open a fresh playground →](https://feat-admin-cache-purge-emdash-playground.emdash-cms.workers.dev)**

<sub>Tracks `feat/admin-cache-purge`. Updated automatically when the playground redeploys.</sub>


packages/core/src/api/handlers/object-cache.ts
```diff
@@ -0,0 +1,218 @@
...
+
+export interface ObjectCacheStatus {
+	/** Whether an object-cache backend is configured for this site. */
+	configured: boolean;
+}
+
+export interface ObjectCachePurgeResult {
+	/** Whether an object-cache backend is configured for this site. */
+	configured: boolean;
+	/** @deprecated Use {@link ObjectCachePurgeResult.configured}. */
```
ascorbic, line 54: Why does a new API include a deprecated field?

ascorbic (changes requested):
Adding manual purge support makes sense, but I think this approach needs some work. Is the idea to expose this in admin, or just make it available to plugins?

I'm not comfortable with this Cloudflare-specific code in core, and it's not needed. If we're to add support for purge commands beyond those provided by Astro, they should be behind an EmDash abstraction, not by putting `cloudflare:workers` code in core. This should be moved into the Cloudflare package and the virtual module plumbed in froim there. The longer-term fix is to upstream the "purge everything" support into Astro.

"purge everything" is a good addition, but rather than prefix purging I think we should just expose the Astro's purging API to plugins and the admin UI, with some EMDash-specific sugar for the tag formats. Make the API purging by collection and ID, so we'd purge("posts", "hello-world"), and also expose path purging which is also in the Astro API. This minimises the need for platform-specific code.

ascorbic:
This also makes it clear that we need to deprecate the old `cloudflareCache`, and move everyone to the native CacheW-based one

PR #2308, opened 2026-07-31
fix: materialize plugin-declared storage indexes

edrpls:
## What does this PR do?

Fixes plugin-declared storage indexes never being created — and never being *usable* even if they had been.

The materialization machinery already exists (`plugins/storage-indexes.ts`, the `_plugin_indexes` tracking table from migration 004), but it has **zero production callers**: manifest `storage.indexes` declarations are consumed only as a query-validation allowlist. So `PluginStorageRepository.query()` ordering by a JSON field — e.g. the audit-log plugin's `orderBy: { timestamp: "desc" }` on its dashboard widget — runs as a full-table scan plus temp B-tree sort. Measured on the audited production deployment (Macabro festival site, emdash 0.31.1, 11 MiB `_plugin_storage`): 16.5–20.5 ms of synchronous main-thread work on every admin dashboard load, and `_plugin_indexes` empty.

Two coordinated changes:

1. **Wire the sync into every plugin lifecycle moment.** A new `syncDeclaredStorageIndexes()` (logs per-collection failures, never throws — a missing index is a performance problem, not a correctness one) runs on marketplace install/update, registry install/update, and index drops run on both uninstalls. Configured (in-config) plugins have no install handler at all, so they sync once per process on the scheduler tick — the cron path, never the request path, per the hot-path query rule.

2. **Fix the index shape.** The existing (never-invoked) generator built *partial* indexes (`WHERE plugin_id = 'x' AND collection = 'y'` via literals). Verified empirically: SQLite does not choose a partial index under bound parameters unless ANALYZE has run — and D1 never runs ANALYZE — so even wired up, the old shape would have indexed nothing in production. The new shape is a composite non-partial index `(plugin_id, collection, json_extract(data, '$.field'))`, which SQLite selects in every scenario and which also eliminates the temp B-tree for ORDER BY. Unique-index semantics are preserved (uniqueness stays scoped per plugin+collection via the leading columns). No migration is needed: no deployed database has old-shape indexes, because nothing ever created them — that is this bug.

Postgres note: the composite index accelerates WHERE filters there, but not the `jsonb` ORDER BY (ordering uses `->` for numeric-correct sorting while the index expression uses `->>`; one index expression cannot match both). SQLite/D1 — where the measured problem lives — serves both. Repository `query()` semantics, results, and cursor behavior are unchanged; only the plan changes.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

## Screenshots / test output

Failing first (on `main`, before the fix) — the query-plan regression test reproduces the exact production plan:

```
× serves the repository's ORDER BY from the declared index without a temp B-tree
  + SEARCH _plugin_storage USING INDEX idx_plugin_storage_list (plugin_id=? AND collection=?)
  + USE TEMP B-TREE FOR ORDER BY

× materializes declared storage indexes on install and drops them on uninstall
× materializes indexes for every declared collection and is idempotent
```

After the fix — plugin/marketplace/registry suites all green, and the full `packages/core` suite matches the `main` baseline (the single `virtual-modules.test.ts` failure is pre-existing on a clean checkout in this environment — macOS temp-dir realpath mismatch):

```
Tests  803 passed (803)   (tests/integration/plugins/, tests/unit/plugins/, marketplace + registry handler suites)
```

### DIFF
--- a/.changeset/plugin-storage-indexes.md
+++ b/.changeset/plugin-storage-indexes.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes plugin-declared storage indexes never being created. Indexes declared in a plugin manifest's `storage` section are now materialized on marketplace/registry install and update, dropped on uninstall, and synced for configured plugins on the scheduler tick — so plugin storage queries (like the audit-log dashboard widgets) use an index instead of scanning the whole table. The index shape also changed to a composite that SQLite actually uses on D1, where table statistics are never collected.
--- a/packages/core/src/api/handlers/marketplace.ts
+++ b/packages/core/src/api/handlers/marketplace.ts
@@ -24,6 +24,10 @@ import {
 } from "../../plugins/marketplace.js";
 import type { SandboxRunner } from "../../plugins/sandbox/types.js";
 import { PluginStateRepository } from "../../plugins/state.js";
+import {
+	removeAllPluginIndexes,
+	syncDeclaredStorageIndexes,
+} from "../../plugins/storage-indexes.js";
 import { normalizeCapabilities } from "../../plugins/types.js";
 import type { PluginManifest } from "../../plugins/types.js";
 import { EmDashStorageError } from "../../storage/types.js";
@@ -475,6 +479,9 @@ export async function handleMarketplaceInstall(
 			description: pluginDetail.description ?? undefined,
 		});
 
+		// Materialize declared storage indexes (logs failures, never throws)
+		await syncDeclaredStorageIndexes(db, [bundle.manifest]);
+
 		// Fire-and-forget install stat
 		client.reportInstall(pluginId, version).catch(() => {
 			// Intentional: never fails the install
@@ -713,6 +720,9 @@ export async function handleMarketplaceUpdate(
 			mcpToolsConsent: null,
 		});
 
+		// Sync declared storage indexes — declarations can change between versions
+		await syncDeclaredStorageIndexes(db, [bundle.manifest]);
+
 		// Clean up old bundle from R2 (best-effort)
 		deleteBundleFromR2(storage, pluginId, oldVersion).catch(() => {});
 
@@ -784,6 +794,14 @@ export async function handleMarketplaceUninstall(
 			} catch {
 				// Plugin storage table may not have data for this plugin
 			}
+		}
+
+		// Drop declared storage indexes — the plugin is gone either way, and
+		// orphaned indexes tax every _plugin_storage write
+		try {
+			await removeAllPluginIndexes(db, pluginId);
+		} catch {
+			// Nothing to drop, or tracking table predates the feature
 		}
 
 		// Delete state row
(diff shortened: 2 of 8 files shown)


edrpls:
Addressed in 8f27ad6f:

- **Collection validation bug**: confirmed and fixed exactly as you diagnosed — `validateIdentifier(collection, ...)` is removed from `generateCreateIndexSql()`; the collection only reaches SQL inside the index name, which `sql.ref` quotes. Added a regression test (`accepts collection names that are not SQL identifiers`) that fails on the previous commit with a swallowed `IdentifierError` for `form-submissions` and now passes with the index created and tracked.
- **Comment discipline**: removed all five flagged call-site comments and trimmed the runtime JSDoc. I kept one sentence there — configured plugins have no install handler, so the scheduler tick is their only sync moment — since that's the non-obvious invariant a future reader needs to understand why this method exists at all.

Plugin/marketplace/registry suites: 176 passed.

edrpls:
One more substantive fix in 83f894bb, found by an adversarial review pass over this branch: **the tick sync never covered sandboxed plugins.** Marketplace/registry plugins in normal sandbox mode (the only mode on Cloudflare Workers) load into the `sandboxedPlugins` map and never join `allPipelinePlugins`, so plugins installed *before* this release would never get their indexes — the install/update handlers only cover future installs. Fixed by carrying `storage` declarations in the manifest cache (populated at bundle load, all four loader sites) and including cached manifests in the tick sync. Idempotent under the bypass-mode overlap.

Three smaller review findings I'm noting rather than fixing here, since all are pre-existing properties of the index machinery rather than this PR's wiring:
- Removing a storage collection from a manifest between versions orphans that collection's indexes until full uninstall (`syncStorageIndexes` diffs within a collection, not across).
- `generateIndexName` can collide (`["a_b"]` vs `["a","b"]`), silently dropping one declaration.
- A failed `CREATE UNIQUE INDEX` (e.g. pre-existing duplicate data) is logged-and-swallowed, which silently drops a declared uniqueness guarantee — arguably that one should fail the install, but that's a product decision.

Happy to file these as a follow-up issue or address them here if maintainers prefer.

edrpls:
CI caught that my earlier fix went too far — worth flagging since it touches the security posture.

Removing `validateIdentifier(collection, ...)` outright also removed the injection-shape rejection that `tests/unit/plugins/storage-indexes.test.ts › should reject invalid collection names` guards (`items'; DROP TABLE--`). I verified your reasoning holds — `sql.ref` does neutralize such names, doubling embedded quotes so the identifier stays inert:

```
CREATE INDEX "idx_""; DROP TABLE users; --" ON t(a)
```

so there was no exploitable path. But deleting a security test to make a change pass is the wrong resolution, so fc6d8e6d takes the middle road instead: a new `validateStorageCollectionName` using the *manifest key* charset (`/^[a-zA-Z][a-zA-Z0-9_-]*$/`) rather than SQL-identifier rules. `form-submissions` and `formSubmissions` now index; quote- and semicolon-bearing names are still rejected. Both the pre-existing security test and my kebab-case regression test pass unchanged — 181 tests green across the plugin/marketplace suites, and the full `packages/core` suite matches the main baseline.

ascorbic (approved):
Good catch. Thanks

PR #2314, opened 2026-07-31
fix: guard FTS triggers so only real content changes re-tokenize

edrpls:
## What does this PR do?

> **Stacked on #2313** (`fix/fts-plain-text`) — both PRs rewrite the same trigger-generator SQL, so this branch builds on that one to avoid a guaranteed conflict. Review the last commit only; the earlier commits are #2313.

Fixes FTS sync triggers re-tokenizing the whole document on every row UPDATE, even when nothing searchable changed.

The update trigger had no change detection: any UPDATE deleted and re-inserted the document's full index entry. Metadata-only saves — status flips, scheduling, autosave version bumps — and the publish path's rewrite-identical-values UPDATEs each paid full re-tokenization. Measured on the audited production deployment (Macabro festival site, emdash 0.31.1): **49× CPU on metadata-only saves**, and re-tokenization was **78–89% of a save's WAL bytes** — the dominant replication-volume driver under litestream-style WAL shipping.

The fix adds a `WHEN` guard to the generated update trigger comparing raw column values with null-safe `IS NOT`: the trigger fires only when an indexed field, the row's `locale`, or its trash state (`deleted_at`) actually changed.

- `deleted_at` must stay in the guard or trash/restore stop syncing the index (locked by test).
- Raw-column comparison stays valid change detection for Portable Text fields whose *indexed* values are extracted text (#2313): if the raw JSON didn't change, the extraction didn't either.
- Value comparison (not `UPDATE OF <cols>`) is required because the publish path SETs every data column even when values are unchanged — only comparing values suppresses those re-tokenizations (locked by test).
- No migration of its own: the WHEN guard ships inside #2313's rebuild migration (`064_fts_plain_text`), so existing deployments re-tokenize once during that rebuild instead of twice back to back.

Deliberately out of scope (possible follow-up, would need a Discussion per the performance-PR policy): coalescing the publish path's three separate UPDATEs into one. With the guards in place those UPDATEs no longer re-tokenize unless values actually changed, which removes the write amplification this bug is about.

The failing test observes the FTS `_data` shadow segments byte-for-byte across a metadata-only UPDATE — on the base branch they get rewritten; with the guard they are identical. Companion tests lock the positive paths: searchable-field edits still re-index, publish-shaped identical-value rewrites don't, and trash/restore still add/remove the row.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

## Screenshots / test output

Failing first (on the base branch, before the guard):

```
× does not re-tokenize on a metadata-only update
  → expected FTS _data segments to be unchanged, got rewritten segments
× does not re-tokenize when the publish path rewrites identical data values
```

After the fix — write-amplification suite (the publish case drives repo.publish() for real and fails without the guard) and the full search/migration suites:

```
Tests  124 passed   (tests/integration/search/, all migration suites)
Full packages/core suite: 5081 passed — the single virtual-modules.test.ts failure is
```

### DIFF
--- a/.changeset/fts-trigger-when-guards.md
+++ b/.changeset/fts-trigger-when-guards.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes content saves re-tokenizing the full search index even when nothing searchable changed. Metadata-only saves (status changes, scheduling, autosave version bumps) no longer rewrite the FTS index, cutting save CPU and write-ahead-log volume on large documents. Existing deployments pick up the guarded triggers through the FTS rebuild migration.
--- a/packages/core/src/search/fts-manager.ts
+++ b/packages/core/src/search/fts-manager.ts
@@ -189,7 +189,7 @@ export class FTSManager {
 	 *
 	 * The trigger SQL emitted here MUST stay in lock-step with migration
 	 * `064_fts_plain_text.ts`. If this changes again, add a new migration
-	 * rather than editing that one — migrations are forward-only.
+	 * rather than editing shipped ones — migrations are forward-only.
 	 */
 	private async createTriggers(collectionSlug: string, searchableFields: string[]): Promise<void> {
 		this.validateInputs(collectionSlug, searchableFields);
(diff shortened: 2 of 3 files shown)


ascorbic:
@edrpls Thanks! I've merged 2313, so this can be rebased.

edrpls:
The red Smoke check is the known Google Fonts 404 roulette ([details on #2313](https://github.com/emdash-cms/emdash/pull/2313#issuecomment-5257503136)) — this occurrence was Inter in `templates/blog-cloudflare` ([run](https://github.com/emdash-cms/emdash/actions/runs/31607568222)). Retriggered; everything code-related is green.

ascorbic (approved):
Thanks!

PR #2315, opened 2026-07-31
fix: don't auto-redirect a slug another entry still holds

edrpls:
## What does this PR do?

Fixes a slug rename taking a *different* entry's live page offline.

`createSlugChangeRedirect` builds the auto-redirect purely from the collection's `url_pattern` and the old/new slug — `RedirectRepository.createAutoRedirect` does a literal `urlPattern.replace("{slug}", oldSlug)`. A URL pattern has no locale token, so **every locale variant of an entry generates the same URL**, and slugs are unique per `(slug, locale)` — so a translation legitimately shares its canonical's slug (the standard emdash i18n shape).

Renaming either row therefore emits a 301 whose source is the *other* row's live URL:

- Canonical `es` venue at `cineteca-nacional`, `en` twin at the same slug.
- Rename the twin to `cinematheque` → writes `/sede/cineteca-nacional` → `/sede/cinematheque`.
- The Spanish page is still served at `/sede/cineteca-nacional`, but the redirect middleware is registered `order: "pre"`, so it intercepts before routing and the page can never recover.

The mirror case is just as bad: renaming the canonical while the translation still holds the old slug redirects away from a URL the translation answers on.

The fix skips the auto-redirect when another non-deleted entry in the collection still holds the old slug. This is the same invariant `createAutoRedirect` already enforces in the other direction (it deletes redirects *from* the new URL because "the new URL serves live content again — any redirect from it would shadow the page"). Any surviving row counts, published or not: a draft that publishes later would otherwise be shadowed by a redirect created while it was invisible. Renames that genuinely free the URL still redirect exactly as before.

I chose "is the old slug still taken?" over the narrower "skip when `locale !== defaultLocale`" because the latter fixes only one direction (it leaves the canonical-rename case broken) and would also suppress legitimate redirects for sites that serve translations through Astro's locale routing while keeping one unprefixed `url_pattern`. The occupancy check is correct under either routing model.

Impact worth noting: this is the reason the reporting deployment keeps `url_pattern` NULL on every collection — accepting junk `/<collection>/<slug>` redirects that match no route, rather than enable a feature that can take a live page down.

Found during a measured database audit of a production deployment (emdash 0.31.1).

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

## Screenshots / test output

Both destructive directions fail on `main` (the redirect source is the live canonical URL):

```
× does not redirect the canonical URL when a translation sharing its slug is renamed
  AssertionError: expected [ '/sede/cineteca-nacional' ] to not include '/sede/cineteca-nacional'
× does not redirect a URL still held by a translation when the canonical is renamed
```

A third test locks the positive path — a rename that genuinely frees the URL still produces the redirect — so the fix can't regress into "never redirect".

After the fix:

```
Tests  65 passed   (tests/integration/redirects/)
Tests  193 passed  (redirects + content handlers + i18n + content suites)
Full packages/core suite: 5075 passed — the single virtual-modules.test.ts failure is
```

### DIFF
--- a/.changeset/slug-redirect-locale-aware.md
+++ b/.changeset/slug-redirect-locale-aware.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes renaming a translation taking its canonical page offline. When two locale variants share a slug, renaming either one created a 301 away from a URL the other still serves, making the live page unreachable. Slug-change redirects are now skipped when another entry still holds the old slug.
--- a/packages/core/src/api/handlers/content.ts
+++ b/packages/core/src/api/handlers/content.ts
@@ -376,6 +376,15 @@ async function createSlugChangeRedirect(
 	newSlug: string,
 	contentId: string,
 ): Promise<void> {
+	// A URL pattern has no locale token, so every locale variant of an entry
+	// generates the same URL, and slugs are unique per (slug, locale) — a
+	// translation may still hold the old slug. Redirecting away from a URL
+	// another row still answers on would take that page down: the redirect
+	// middleware runs `order: "pre"`, so routing never gets a chance.
+	// Any surviving row counts, published or not: a draft that publishes later
+	// would otherwise be shadowed by the redirect.
+	if (await slugStillTaken(db, collection, oldSlug, contentId)) return;
+
 	const collectionRow = await db
 		.selectFrom("_emdash_collections")
 		.select("url_pattern")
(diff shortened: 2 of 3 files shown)


edrpls:
Rebased onto current `main` (`ecebade8`) and moved out of draft.

Re-verified the bug is still live on today's `main` before promoting this: with only the test file applied and the guard removed, both destructive directions still reproduce —

```
× does not redirect the canonical URL when a translation sharing its slug is renamed
× does not redirect a URL still held by a translation when the canonical is renamed
```

Nothing on `main` has touched `createSlugChangeRedirect`, `createAutoRedirect`, or the redirect middleware since the branch was cut, so the fix still applies cleanly and the rebase was a fast-forward with no conflicts.

Post-rebase: full `packages/core` suite green at 5095 passing, lint/typecheck/format clean.

ascorbic (approved):
Thanks

PR #2328, opened 2026-08-03
fix(core): index translation_group with locale so translation lookups seek

### LINKED ISSUE #2316: `findTranslations` full-scans content tables: migration 041's locale composites out-prioritise the `translation_group` index
## Summary

`ContentRepository.findTranslations` issues:

```sql
SELECT * FROM "ec_<collection>"
WHERE translation_group = ?
  AND deleted_at IS NULL
ORDER BY locale ASC
```

On SQLite/D1 the planner does **not** use `idx_ec_<collection>_translation_group`.
It uses one of the `(deleted_at, locale, …)` composites added by migration 041,
seeking only on `deleted_at IS NULL` and applying `translation_group` as a residual
filter — so the query reads **every non-deleted row in the table** to return the one
translation group asked for.

On a table of ~12.8K rows this measures as a rows-read / rows-returned ratio of
~12,800:1. The ratio grows linearly with table size.

## Reproduction

Plain SQLite, no EmDash required. Schema and index set mirror
`SchemaRegistry.createContentTable` plus migration 041.

```sql
CREATE TABLE ec_items (
  id TEXT PRIMARY KEY,
  slug TEXT,
  locale TEXT,
  translation_group TEXT,
  deleted_at TEXT,
  created_at TEXT,
  updated_at TEXT
);

-- Index from migration 019 / SchemaRegistry.createContentTable
CREATE INDEX idx_ec_items_translation_group ON ec_items (translation_group);

-- Composites from migration 041
CREATE INDEX idx_ec_items_loc_upd ON ec_items (deleted_at, locale, updated_at DESC, id DESC);
CREATE INDEX idx_ec_items_loc_crt ON ec_items (deleted_at, locale, created_at DESC, id DESC);

WITH RECURSIVE c(n) AS (SELECT 1 UNION ALL SELECT n+1 FROM c WHERE n < 2000)
INSERT INTO ec_items
SELECT 'id'||n, 's'||n,
       CASE n%2 WHEN 0 THEN 'en' ELSE 'de' END,
       'tg'||(n/2), NULL, '2025-01-01', '2025-01-01'
FROM c;

EXPLAIN QUERY PLAN
SELECT * FROM ec_items
WHERE translation_group = 'tg5' AND deleted_at IS NULL
ORDER BY locale ASC;
```

**Actual:**

```
SEARCH ec_items USING INDEX idx_ec_items_loc_crt (deleted_at=?)
```

**Expected:** a seek on `translation_group`, returning the 2 rows of that group.

## Why

Two conditions combine:

1. **No `sqlite_stat1`.** The planner has no selectivity data and cannot tell that
   `translation_group` is highly selective (in a typical table it identifies 1–2 rows
   out of N). On D1 this is not a fixable condition — see "No statistics workaround
   on D1" below.
2. **The filter and sort are an exact prefix of the 041 composites.**
   `idx_…_loc_crt` is `(deleted_at, locale, created_at DESC, id DESC)`. Seeking it on
   `deleted_at IS NULL` returns rows **already ordered by `locale`**, satisfying
   `ORDER BY locale ASC` with no temp B-tree.

The planner therefore trades a full scan for eliminating the sort. Drop either
condition and it recovers:

| Variant | Plan |
| --- | --- |
| As written | `SEARCH … USING INDEX idx_ec_items_loc_crt (deleted_at=?)` |
| Without `ORDER BY locale` | `SEARCH … USING INDEX idx_ec_items_translation_group` |
| Without `deleted_at IS NULL` | `SEARCH … USING INDEX idx_ec_items_translation_group` + temp B-tree |
| After `ANALYZE` (local SQLite only — see below) | `SEARCH … USING INDEX idx_ec_items_translation_group` + temp B-tree |

This looks like a regression introduced alongside the fix for #1219: migration 041's
composites are correct and effective for the locale-scoped **content list**, but they
are also a better prefix match for `findTranslations` than the index intended to
serve it.

## Scope

- Affects **every** `ec_*` content table; severity scales with row count.
- **Independent of locale configuration.** Verified reproducing with two locales,
  with a single locale, with genuinely shared translation groups, and with every row
  in its own group. The plan is chosen from index shape at prepare time — no locale
  value is consulted. This is not a multi-locale-only issue.
- Also reachable from any caller of `findTranslations`, including the admin
  translations panel and the bulk `translation_group IN (…)` sibling query, which has
  the same shape.

## Suggested fix

Widen the translation-group index to cover the sort:

```sql
CREATE INDEX idx_ec_<collection>_translation_group
  ON ec_<collection> (translation_group, locale);
```

This serves both the equality seek and `ORDER BY locale`, so it beats the 041
composites on the planner's own cost terms and needs no statistics:

```
SEARCH ec_items USING INDEX idx_ec_items_tg_locale (translation_group=?)
```

Two places to change:

- `SchemaRegistry.createContentTable` — the standard-index block that currently
  creates the single-column `translation_group` index, so new collections get the
  composite.
- A new forward-only migration to add the composite (and drop the single-column
  index) on existing tables, following the idempotent `IF NOT EXISTS` +
  `listTablesLike` pattern established by 041.

Migration 019 creates the original single-column index in two separate code paths
(Postgres and SQLite branches); both need updating for new installs.

### No statistics workaround on D1

On local SQLite, running `ANALYZE` also restores the correct index choice (it leaves
the temp B-tree sort in place, but avoids the scan). **That workaround is not
available on D1**, which is the deployment target where this bug actually bites:

- D1 exposes `PRAGMA optimize` — the mechanism that runs `ANALYZE` — but documents
  that [D1 PRAGMA statements only apply to the current
  transaction](https://developers.cloudflare.com/d1/sql-api/sql-statements/), so no
  `sqlite_stat1` persists as durable planner state.
- Confirmed on a production D1 database exhibiting this bug: `sqlite_stat1` is absent
  from `sqlite_master` entirely.

So on D1 the planner is **permanently** operating on default guesses, and application
authors have no way to correct it from outside the schema. That makes the index shape
the only available lever, and raises the priority of this fix: affected users cannot
work around it themselves. It also means any EmDash query whose correct plan depends
on statistics is latently broken on D1 — this one is simply the most visible
instance.

## Environment note

Observed on Cloudflare D1. The mechanism is generic SQLite query planning and is not
D1-specific; D1 simply surfaces it clearly via its rows-read metric.

MA2153:
## What does this PR do?

`ContentRepository.findTranslations` reads every non-deleted row of a content table to return one translation group.

The query filters `translation_group = ?` with `deleted_at IS NULL` and sorts `ORDER BY locale ASC`. Seeking migration 041's `(deleted_at, locale, ...)` composites on `deleted_at` alone already returns rows in locale order, so the planner trades a full scan for eliminating the sort — and without `sqlite_stat1` it has no selectivity data to tell it `translation_group` identifies 1–2 rows. D1 never persists `sqlite_stat1`, so affected users cannot correct this from outside the schema; the index shape is the only lever.

Verified before fixing — plan on an unmodified tree:

```
SEARCH ec_post USING INDEX idx_ec_post_loc_crt (deleted_at=?)
```

**The fix:** replace the single-column `translation_group` index with one index per read shape, each matching its read term for term so it wins on the planner's own cost terms without statistics:

- `(deleted_at, translation_group, locale)` — the translation-group reads above.
- `(translation_group, locale)` — lookups that never constrain `deleted_at` (menu and reference resolution). Added in round 3.

Applied in two places:

- `schema/registry.ts` — new collections get both composites directly.
- Migration `055` — creates the replacements *before* dropping the old index, following `052`, because D1 DDL is non-transactional and an interrupted migration must never leave a table with no `translation_group`-leading index.

Migration 019 is deliberately left alone: new installs run 019 → 055 anyway, so rewriting applied migration history would be churn for no gain.

**The batched variant needed more.** The issue suggests `translation_group IN (...)` "has the same shape" and is covered by the same index. It isn't — the planner picks the 041 composite there *even with no `ORDER BY` at all*, so the sort is not the deciding factor. Two things were needed:

- The sort leads with `translation_group`, so the ordering follows the new index past its `deleted_at` equality. Callers group by `translation_group` and depend only on locale order *within* a group, which is unchanged.
- The index leads with `deleted_at`. A `(translation_group, locale)` index alone holds only for the single-group read: the `IN (...)` list multiplies the planner's row estimate, so from five groups onward it falls back to a `deleted_at` composite and reads every non-deleted row again — and the public reference path batches at `SQL_BATCH_SIZE` (50). Caught in review by @khoinguyenpham04.

**Who this actually affects.** The issue frames this around the admin translations panel, which undersells it. Of the four callers, the two hottest are logged-out:

| Caller | Path |
| --- | --- |
| `seo/hreflang.ts` | **Public** — hreflang alternates in `<head>` on every render when i18n is on |
| `query.ts` | **Public** — site-facing translations helper |
| `api/handlers/relations.ts` (batch) | Public + admin — reference-edge resolution |
| `api/handlers/content.ts` | Admin — editor translations panel |

`pnpm query-counts` reports counts and query text unchanged.

Not investigated here: `RelationRepository.findTranslations` and `TaxonomyRepository.findTranslations` are the same query shape against their own tables. Migration 041 only touched `ec_%`, but 036/049 added locale indexes to taxonomies, so a similar trade may exist. Worth its own issue rather than widening this PR.

Closes #2316

## Review follow-up

**Round 3** ([review](https://github.com/emdash-cms/emdash/pull/2328#pullrequestreview-4845825633))

*Menu lookups now scan because they don't filter by `deleted_at`.* Correct, and reproduced against the real path. `resolveContentUrl` reads `WHERE translation_group = ? AND locale = ?` with no `deleted_at` term, so a `deleted_at`-leading index cannot seek it. The single-column index that round 1 dropped had been covering it; without it the planner falls back to `idx_ec_post_locale`:

```
× seeks a menu content reference resolved for 'en'
  Received: "SEARCH ec_post USING INDEX idx_ec_post_locale (locale=?)"
```

That reads every row in the requested locale on a public render — worse than the bug this PR opened against.

Taking the suggestion: both indexes are now created, `(translation_group, locale)` and `(deleted_at, translation_group, locale)`. Neither shape can borrow the other's index — an index leading with `deleted_at` can't seek a lookup that never constrains it, and going the other way the `IN (...)` estimate blowup from round 2 still applies.

The split across `ec_*` call sites:

| Index | Call sites |
| --- | --- |
| `(translation_group, locale)` | `menus/index.ts` locale lookup + any-locale fallback (public), `api/handlers/content.ts` sibling `UPDATE`, `media/usage/content-refresh.ts` sibling ids |
| `(deleted_at, translation_group, locale)` | `ContentRepository.findTranslations`, `.findTranslationsForGroups` |

Regression tests, written failing first, in `content-translations-plan.test.ts`:

- Menu references driven through `getMenuWithDb`, parameterized over the requested-locale hit and the any-locale fallback (`fr` has no `ec_post` row), asserting every emitted `translation_group` query seeks `idx_ec_post_tg_locale`.
- The existing single and batched `findTranslations` plans, repinned to `idx_ec_post_del_tg_locale`.
- The pre-055 upgrade test now drops both new indexes, runs `up()`, and checks both shapes seek afterwards.

`content-translation-index-name.test.ts` asserts the migration leaves *both* composites on the longest creatable Postgres slug — `idx_{table}_del_tg_locale` truncates to `…_del_tg_lo` at 63 bytes, distinct from `…_tg_locale`, and distinct from all 13 other registry index names at that length.

Cost of the extra index: one more B-tree per content table on insert/update. Menu resolution is a logged-out per-render path, so the read side wins that trade.

Menu internals are worth a closer look on their own (the fallback issues a second query, and neither query filters `deleted_at`, so a soft-deleted entry still resolves a link) — out of scope here.

**Round 2** ([review](https://github.com/emdash-cms/emdash/pull/2328#pullrequestreview-4843506642))

- *Batched path still full-scans at realistic batch sizes.* Correct, and reproduced: at 50 groups the plan fell back to `idx_ec_post_deleted_published_id (deleted_at=?)`, or `idx_ec_post_deleted_status` with `publishedOnly`. The index now leads with `deleted_at`, which holds at every batch size and under either filter. The plan test is parameterized over {2, 50} groups × `publishedOnly` on/off.
- *Postgres truncates the old and new index names to the same value for long table names.* The arithmetic holds from a 54-character slug, but the case is unreachable: `SchemaRegistry.createContentTable` already fails on Postgres past a 46-character slug, because `idx_{table}_deleted_updated_id` and `idx_{table}_deleted_status` truncate together first — verified against Postgres 17. Those names date from the first commit, so no such table exists. Added `content-translation-index-name.test.ts`, which pins the swap at the longest creatable slug on both dialects; renaming the index to the long `_translation_group_locale` form makes the Postgres variant fail exactly as described.

The wider naming collision that surfaced here — registry index names capping collection slugs at ~46 rather than the documented 63, with `loc_upd`/`loc_crt` colliding silently from 52 — predates this PR and belongs in its own issue.

**Round 1** ([review](https://github.com/emdash-cms/emdash/pull/2328#pullrequestreview-4841970328))

Both comments were right: the docstring in `content.ts` and the inline comment in `schema/registry.ts` narrated the rejected alternative (`ORDER BY locale` alone) and cited migration 041 as justification. Both are trimmed to the invariant a future reader actually needs; the planner rationale now lives only in the migration 055 docstring.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [x] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable). n/a — no admin UI strings; no `messages.po` changes included.
- [ ] New features link to an approved Discussion — n/a, bug fix against a filed issue.

## Screenshots / test output

Tests were written failing first. `tests/integration/database/content-translations-plan.test.ts` covers the single-group seek, the batched seek, the menu-reference seek, and the migration upgrade path (pre-055 table plans through `loc_crt`; after `up()` both shapes seek and the old index is gone).

Before each fix:

```
× seeks a single translation group through the translation_group index
  Received: "SEARCH ec_post USING INDEX idx_ec_post_loc_crt (deleted_at=?)"
× seeks batched translation groups through the translation_group index
  Received: "SEARCH ec_post USING INDEX idx_ec_post_loc_crt (deleted_at=?)"
× seeks a menu content reference resolved for 'en'
  Received: "SEARCH ec_post USING INDEX idx_ec_post_locale (locale=?)"
```

After:

```
Test Files  395 passed | 1 skipped (396)
     Tests  5082 passed | 3 skipped (5085)
```

Lint 0 diagnostics, typecheck clean, `pnpm query-counts` counts and query text unchanged.

Rounds 1–2 were verified with `EMDASH_TEST_PG` pointed at a local Postgres 17. The round-3 run above is SQLite-only — no Postgres was reachable in that session, so the Postgres legs of the dialect-parity tests are among the skips. The 63-byte name arithmetic for `idx_{table}_del_tg_locale` was checked statically against every registry index name; the parity leg still needs a CI (or local Postgres) run to confirm.


packages/core/tests/integration/database/content-translations-plan.test.ts
```diff
@@ -0,0 +1,136 @@
...
+	expect(contentAccess(plan)).toMatch(
+		/SEARCH ec_post USING (?:COVERING )?INDEX idx_ec_post_tg_locale \(translation_group=\?\)/,
+	);
+	expect(plan).not.toContain("SCAN ec_post");
+	expect(plan).not.toContain("idx_ec_post_loc_crt");
+	expect(plan).not.toContain("idx_ec_post_loc_upd");
+});
+
+it("seeks batched translation groups through the translation_group index", async () => {
+	const items = await repo.findTranslationsForGroups("post", ["tg-0005", "tg-0006"]);
```
khoinguyenpham04: Could we test this with `SQL_BATCH_SIZE` (50 groups), with `publishedOnly` both on and off? I reproduced the current PR and the planner switches back to a `deleted_at` index from five groups onward, so the public 50-edge path still reads every non-deleted row and sorts.

packages/core/src/database/migrations/055_content_translation_group_locale_index.ts
```diff
@@ -0,0 +1,53 @@
...
+ */
+export async function up(db: Kysely<unknown>): Promise<void> {
+	const tableNames = await listTablesLike(db, "ec_%");
+
+	for (const tableName of tableNames) {
+		// D1 DDL is non-transactional: create the replacement before dropping the
+		// old index so an interrupted migration always leaves a
+		// translation_group-leading index in place.
+		await sql`
+			CREATE INDEX IF NOT EXISTS ${sql.ref(`idx_${tableName}_tg_locale`)}
```
khoinguyenpham04, line 42: There's one edge case here. PostgreSQL truncates the old and new index names to the same value for long table names. The create is then skipped and the following drop removes the existing index. 

Could we use a collision-safe name or track this separately?

khoinguyenpham04 (commented):
Thanks for digging into this, the single-group lookup is fixed. The batched public path still full-scans at realistic batch sizes, so could we address that before merging?

packages/core/tests/integration/database/content-translations-plan.test.ts
MA2153: Confirmed, thank you — that was a real hole. Reproduced at `SQL_BATCH_SIZE` (50 groups), both ways:

```
publishedOnly=false → SEARCH ec_post USING INDEX idx_ec_post_deleted_published_id (deleted_at=?)
publishedOnly=true  → SEARCH ec_post USING INDEX idx_ec_post_deleted_status (deleted_at=? AND status=?)
```

The `IN (...)` list multiplies the planner's row estimate for a `translation_group`-leading index, so a `deleted_at` composite wins from five groups onward — the sort isn't the deciding factor in either direction.

Fixed by leading the index with `deleted_at`: `(deleted_at, translation_group, locale)` matches the read term for term, so it holds at every batch size and with either filter. The plan is now `SEARCH ec_post USING INDEX idx_ec_post_tg_locale (deleted_at=? AND translation_group=?)`.

The plan test is parameterized over {2, 50} groups × `publishedOnly` on/off, and asserts the row count as well as the plan so the seek can't be traded for wrong results.

packages/core/src/database/migrations/055_content_translation_group_locale_index.ts
MA2153: The arithmetic is right: both names share the `idx_ec_{slug}_` prefix, so once that prefix reaches 63 bytes — a 54-character slug — they truncate to the same identifier, the create is a no-op, and the drop then takes the surviving index with it.

It isn't reachable, though. `SchemaRegistry.createContentTable` can't create a collection on Postgres past a 46-character slug: `idx_{table}_deleted_updated_id` and `idx_{table}_deleted_status` truncate together first, and the second `CREATE INDEX` errors out. Verified against Postgres 17:

```
len=46  OK
len=47  error: relation "idx_ec_ooo…_deleted_" already exists
len=54  error: relation "idx_ec_ooo…_s" already exists      (status vs slug)
```

`idx_{table}_status` and `idx_{table}_slug` are in the first commit, so no Postgres content table with a slug that long has ever been creatable — there is nothing for migration 055 to find.

Rather than a name change for an unreachable case, I've pinned the behaviour at 46 characters — the longest slug that does exist — in `content-translation-index-name.test.ts`, running on both dialects. It has teeth: renaming the index to the long `_translation_group_locale` form makes the Postgres variant fail in exactly the way you describe, while SQLite stays green.

The underlying issue you've surfaced is real but wider than this PR: registry index names collide under Postgres truncation, capping collection slugs at ~46 rather than the documented 63, and `loc_upd`/`loc_crt` from migration 041 collide from 52 (silently, since those creates are `IF NOT EXISTS`). That's a `SchemaRegistry` naming problem, not a migration-055 one. Happy to file it separately if you'd like.

khoinguyenpham04 (commented):
thanks for the follow-up. i reproduced this with the real migration and public menu path. the batch lookup is fixed, but menu lookups now scan because they don’t filter by `deleted_at`.

could we keep separate indexes for both query shapes, `(translation_group, locale)` for direct lookups and `(deleted_at, translation_group, locale)` for batched reads, and add regression tests for both paths? that seems like the safer long-term fix.

MA2153:
@khoinguyenpham04 Great catch. That's arguably a regression worse than the bug we started with. It appears the `(translation_group, locale)` index is used in more places than `(deleted_at, translation_group, locale)`. Menu internals definitely need to be digged into later, but for now I agree we should just keep the two indexes to keep the scope narrow.

ascorbic (approved):
Thanks!

PR #2331, opened 2026-08-03
fix(core): batch taxonomy term counts under D1's compound-SELECT limit

### LINKED ISSUE #2330: Taxonomy term list 500s when a taxonomy declares more collections than the backend's compound-SELECT limit
## Summary

`fetchVisibleTermCounts` builds one `UNION ALL` branch per declared collection. On a backend with a low compound-`SELECT` ceiling (Cloudflare D1), a taxonomy that declares six or more collections produces SQL the database rejects. The error escapes `fetchVisibleTermCounts`, propagates out of `handleTermList`, and `GET /_emdash/api/taxonomies/:name/terms` returns 500 — so the admin renders an **empty term list** for a taxonomy whose terms are perfectly intact in the database.

The failure is entirely a function of how many collections a taxonomy declares. It is invisible until a site crosses the threshold, then it takes out the whole taxonomy at once.

## Environment

- emdash 0.31.1
- Cloudflare D1 (SQLite)

## Reproduction

1. Register six collections (e.g. `alpha`, `beta`, `gamma`, `delta`, `epsilon`, `zeta`).
2. Create a taxonomy whose `collections` array lists all six.
3. Add any number of terms to it.
4. `GET /_emdash/api/taxonomies/<name>/terms?locale=<locale>`

**Expected:** 200 with the term list.
**Actual:** 500 `TERM_LIST_ERROR`; the admin taxonomy page shows an empty list.

Measured threshold on D1 — same query shape, only the branch count varying:

| `UNION ALL` branches | Result |
| --- | --- |
| 4 | ok |
| 5 | ok |
| 6 | `too many terms in compound SELECT: SQLITE_ERROR [code: 7500]` |

A taxonomy declaring five collections sits exactly on the boundary and breaks as soon as one more collection is added to it.

## Root cause

`src/taxonomies/term-counts.ts` — `runCounts` composes one branch per collection with no bound on the total:

```ts
const branches = collections.map((collection) => collectionBranch(db, taxonomyName, collection));
const union = sql.join(branches, sql` UNION ALL `);
```

SQLite's `SQLITE_MAX_COMPOUND_SELECT` defaults to 500, but D1 configures it far lower — empirically 5. Postgres has no equivalent limit, which is presumably why this hasn't surfaced before.

## Why it presents as an empty list rather than an error

Two layers turn a specific SQL error into a silent blank page:

1. **`fetchVisibleTermCounts` only recovers from missing tables.** The per-collection retry fallback is guarded by `isMissingTableError` (`term-counts.ts:106` and `:118`) and rethrows anything else. A compound-limit error isn't a missing-table error, so the fallback never engages — even though retrying per collection would in fact have worked here, since each individual branch is a single `SELECT`.

2. **`handleTermList` discards the error.** `src/api/handlers/taxonomies.ts:426`:

   ```ts
   } catch {
       return {
           success: false,
           error: { code: "TERM_LIST_ERROR", message: "Failed to list terms" },
       };
   }
   ```

   The original error is not logged or attached, so nothing anywhere records *why* the list failed. Diagnosing this required reconstructing the generated SQL by hand and replaying it against the database.

## Impact

Term counts are a display enhancement, but their failure takes down the entire term list. The affected taxonomies become unmanageable through the admin UI — terms can't be listed, edited, or reordered — while the underlying rows are fine. Taxonomies below the threshold in the same install keep working, which makes the pattern look arbitrary until you count declared collections.

## Suggested fixes

1. **Chunk the union.** Batch collections into groups (conservatively, 4) in `runCounts` and merge the resulting maps — the aggregation is a plain sum over `taxonomy_id`, so partitioning is safe. This removes the ceiling regardless of backend.

2. **Degrade counts instead of failing the list.** A count failure for *any* reason should arguably yield terms without counts rather than no terms. Widening the fallback's error predicate would also help, but the structural fix is that `handleTermList` shouldn't let a counts failure sink the list it decorates.

3. **Preserve the underlying error.** `handleTermList`'s bare `catch` should log or attach the original error. Any of these three changes alone would have made this a five-minute diagnosis.

I'm happy to open a PR for the chunking fix if the approach sounds right.

MA2153:
## What does this PR do?

`fetchVisibleTermCounts` built one `UNION ALL` branch per declared collection. Cloudflare D1 sets `SQLITE_LIMIT_COMPOUND_SELECT` to 5 — SQLite's own default is 500 — so a taxonomy declaring six or more collections produced SQL the backend rejected outright. Because the counts decorate the admin term list, the failure took the whole list down: `GET /_emdash/api/taxonomies/:name/terms` returned 500 and the admin rendered an empty list for a taxonomy whose terms were perfectly intact.

**The ceiling is 5, measured against a live D1** rather than inferred: with the same query shape and only the branch count varying, five `UNION ALL` branches compile and six fail with `too many terms in compound SELECT`. That also settles #895, where the dashboard hit this at 9 arms and was fixed by fan-out without anyone bisecting the real threshold.

The fix splits the collections into batches the backend can carry, runs the batches concurrently, and sums the resulting maps — per-collection sums are commutative, so partitioning cannot change a total. The missing-`ec_*`-table fallback moves per batch, so one absent table only degrades its own batch instead of the whole computation.

**The ceiling belongs to the adapter, not to core.** `compoundSelectLimit(db)` (`database/dialect-helpers.ts`) reads it off the Kysely adapter and returns `null` when the backend declares none — which is every backend except D1, so Node/SQLite, libSQL and Postgres keep counting every collection in a single statement no matter how many there are. `@emdash-cms/cloudflare`'s three D1 dialects (raw binding, session, coalescing) share a `D1Adapter` base that declares `compoundSelectLimit = 5`. The limit is a property of the SQLite build behind the dialect rather than of the SQL flavour, which is why `detectDialect()`'s `sqlite | postgres` answer can't carry it.

On D1 the batch size is the measured 5, not a more conservative 4: **a taxonomy at or below the ceiling still issues exactly one query**, so nothing on the logged-out render path regresses. 4 would push every 5-collection taxonomy from one query to two. The perf fixture's taxonomies declare one collection each, so the query-count snapshots are unchanged.

`handleTermList`'s bare `catch` now logs the original error. Diagnosing this required reconstructing the generated SQL by hand, because nothing anywhere recorded *why* the list failed.

Closes #2330

### Notes on the issue's other two suggestions

Deliberately **not** included, happy to be overruled:

- **Widening the fallback's `isMissingTableError` predicate.** With the ceiling gone, the only errors a wider predicate would newly swallow are genuine ones.
- **Degrading `handleTermList` to counts-free terms on a count failure.** It trades a loud 500 for silently wrong-looking admin data, and the failure that motivated it no longer exists. That reads like a maintainer's call about how much decoration is allowed to fail, not something to slip into a bug fix.

### Testing

TDD: the regression test failed with exactly the reported error at `runCounts` before the fix.

better-sqlite3 uses SQLite's upstream default of 500 and offers no way to lower it, so a query shape D1 rejects runs happily in tests. `setupTestDatabaseWithCompoundSelectLimit(limit)` stands in for either kind of backend: given a number its dialect declares the ceiling the way D1's does and `prepare()` rejects statements past it — where SQLite raises the error too — with D1's error text; given `null` it declares nothing, like better-sqlite3 itself. It also hands back every prepared statement, so a test can assert how many round trips the count took.

Three cases, all at `LIMIT + 1` collections:

- with a declared ceiling, counts aggregate across every collection and `handleTermList` returns them (the reported symptom), in 2 statements;
- with no declared ceiling, the same counts come back in **1** statement;
- a missing `ec_*` table beyond the first batch is still skipped. Mutation-checked by restricting the fallback to the first batch — the test fails as intended.

On the `@emdash-cms/cloudflare` side, each of the three D1 dialects is asserted to declare the ceiling on its adapter — an adapter that drops it silently sends D1 compound SELECTs it rejects.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (or targeted tests for my change) — `packages/core` unit suite (3448 passed), plus the taxonomy, database, utils and loader suites and `packages/cloudflare`'s `tests/db` re-run after the adapter change
- [x] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable) — n/a, no admin UI strings; the added log line is server-side and English-only by convention

## Screenshots / test output

Before the fix, with the D1 ceiling imposed on the test database:

```
FAIL  tests/unit/taxonomies/term-counts.test.ts > visible term counts past the
      compound-SELECT ceiling > aggregates every declared collection when
      there are more than one statement can carry
Error: too many terms in compound SELECT: SQLITE_ERROR
 ❯ runCounts src/taxonomies/term-counts.ts:69:17
 ❯ Module.fetchVisibleTermCounts src/taxonomies/term-counts.ts:104:10
```

After, with the ceiling now read off the adapter — `packages/core` taxonomies (unit + integration) and `packages/cloudflare` `tests/db`:

```
Test Files  10 passed (10)
     Tests  87 passed (87)

Test Files  11 passed (11)
     Tests  176 passed (176)
```

## Review response

**Comment discipline — `#2330` in the test file (40fae20).** Fixed, and the review is right on both counts. The JSDoc block above the new `describe` is deleted outright rather than reworded: it was narrative about how the bug presented, which is exactly what the commit message and this description are for. The `describe` name is now `visible term counts past the compound-SELECT ceiling` — it already said what the block covers without the number. Nothing else in the diff carries an issue reference; the `#581` in `term-counts.ts:2` is pre-existing and untouched.

Targeted suite re-run after the edit: `tests/unit/taxonomies/term-counts.test.ts` — 10 passed.

**D1's ceiling was applied to every database — six collections went from one query to two on Node/SQLite.** Correct, and the measurement matches: batching was unconditional, so every backend paid for a limit only D1 has. The ceiling now comes from the adapter (`compoundSelectLimit(db)`, described above), so a backend that declares nothing runs the counts as a single statement at any collection count — Node/SQLite, libSQL and Postgres included — and only D1 splits. The new test asserts the single statement directly rather than trusting the shape: it fails on the previous commit.

Durable Object SQLite (`DOSqlDialect`, `PreviewDODialect`) deliberately declares nothing. D1 may well share the build, but I have no measurement for a DO the way I do for D1, and guessing here would cost the extra query on a backend that may not need it. Happy to add it if you know the answer.

**Why the changeset is still a single `emdash` patch.** The batching and the adapter lookup both live in core; `@emdash-cms/cloudflare` contributes one declared constant. Since `.changeset/config.json` puts the two packages in the same `fixed` group, listing `emdash` alone already versions cloudflare in lockstep — adding it to the changeset would only copy the same note into a second changelog without changing what anyone installs.

**`compoundSelectLimit` isn't validated — `0`/negative loop forever, `NaN`/fractions build invalid batches (060499c9).** Agreed, and the reasoning holds for every one of those values; the probe now requires a positive integer and throws naming the adapter class otherwise. The ceiling is read off a third-party Kysely adapter by duck-typing, so core has no compile-time guarantee about what it gets, and each malformed value fails silently rather than loudly:

- `0` or negative: `chunks()` never advances `i`, so `fetchVisibleTermCounts` hangs — worse than the 500 this PR set out to fix.
- fractional (`2.5`): `slice()` truncates both bounds, so batch 2 starts at index 2 after batch 1 ended there — an element lands in two batches and its count is added twice. Silently wrong numbers, no error anywhere.
- `NaN`: one empty batch, so the UNION has no terms and the statement is invalid SQL.

Declared-but-not-a-number (`"5"`) throws too rather than falling through to `null`. Silently ignoring a ceiling an adapter meant to declare is exactly the failure this PR exists to remove — it would send D1 the compound SELECTs it rejects, which is the reported bug with an extra step.

Throwing is the right severity here: this is a static property of the dialect, so it is wrong on the first request or never, and it can only be fixed by the person who wrote the adapter. A silent `null` would hide that from them.

New unit test (`tests/unit/database/compound-select-limit.test.ts`) covers `null` for an adapter declaring nothing, pass-through for `5` and `1`, and a throw for `0`, `-1`, `NaN`, `Infinity`, `2.5`, `"5"` and `null` — 8 of its 10 cases fail on the previous commit. `packages/core` taxonomies + database and `packages/cloudflare` `tests/db` re-run clean.

### DIFF
--- a/packages/core/src/database/dialect-helpers.ts
+++ b/packages/core/src/database/dialect-helpers.ts
@@ -38,6 +38,32 @@ export function isPostgres(db: Kysely<any>): boolean {
 }
 
 /**
+ * Declared by an adapter whose backend caps the number of terms in a compound
+ * SELECT (`UNION ALL`, `INTERSECT`, `EXCEPT`). SQLite's own
+ * SQLITE_LIMIT_COMPOUND_SELECT default is 500 — high enough that no query
+ * EmDash builds approaches it — but Cloudflare D1 sets it to 5 and rejects
+ * anything larger with "too many terms in compound SELECT".
+ */
+export interface CompoundSelectLimitedAdapter {
+	readonly compoundSelectLimit: number;
+}
+
+/**
+ * The backend's compound-SELECT ceiling, or null when it has none worth
+ * splitting statements for. Only the adapter knows: the limit is a property of
+ * the SQLite build behind the dialect, not of the SQL flavour, so two "sqlite"
+ * dialects can answer differently.
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- accepts any Kysely instance
+export function compoundSelectLimit(db: Kysely<any>): number | null {
+	const adapter: object = db.getExecutor().adapter;
+	if ("compoundSelectLimit" in adapter && typeof adapter.compoundSelectLimit === "number") {
+		return adapter.compoundSelectLimit;
+	}
+	return null;
+}
+
+/**
  * Default timestamp expression for column defaults.
  * Wrapped in parens for use in CREATE TABLE ... DEFAULT (...).
  *
--- a/packages/core/src/database/dialect-helpers.ts
+++ b/packages/core/src/database/dialect-helpers.ts
@@ -45,6 +45,7 @@ export function isPostgres(db: Kysely<any>): boolean {
  * anything larger with "too many terms in compound SELECT".
  */
 export interface CompoundSelectLimitedAdapter {
+	/** Maximum terms per compound SELECT. Must be a positive integer. */
 	readonly compoundSelectLimit: number;
 }
 
@@ -53,14 +54,25 @@ export interface CompoundSelectLimitedAdapter {
  * splitting statements for. Only the adapter knows: the limit is a property of
  * the SQLite build behind the dialect, not of the SQL flavour, so two "sqlite"
  * dialects can answer differently.
+ *
+ * A declared ceiling must be a positive integer — callers batch by it, and
+ * every other value silently misbehaves rather than failing: 0 and negatives
+ * never advance the batch cursor, fractions overlap batches and double-count,
+ * NaN yields an empty batch. A malformed declaration throws here, where the
+ * message can name the adapter.
  */
 // eslint-disable-next-line @typescript-eslint/no-explicit-any -- accepts any Kysely instance
 export function compoundSelectLimit(db: Kysely<any>): number | null {
 	const adapter: object = db.getExecutor().adapter;
-	if ("compoundSelectLimit" in adapter && typeof adapter.compoundSelectLimit === "number") {
-		return adapter.compoundSelectLimit;
+	if (!("compoundSelectLimit" in adapter)) return null;
+
+	const limit: unknown = adapter.compoundSelectLimit;
+	if (typeof limit !== "number" || !Number.isInteger(limit) || limit < 1) {
+		throw new Error(
+			`${adapter.constructor.name} declares compoundSelectLimit ${String(limit)}; it must be a positive integer.`,
+		);
 	}
-	return null;
+	return limit;
 }
 
 /**
--- a/.changeset/fix-taxonomy-term-counts-compound-select.md
+++ b/.changeset/fix-taxonomy-term-counts-compound-select.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes the taxonomy term list returning a 500 and rendering empty on Cloudflare D1 when a taxonomy declares six or more collections.
--- a/packages/cloudflare/src/db/coalescing-d1.ts
+++ b/packages/cloudflare/src/db/coalescing-d1.ts
@@ -21,7 +21,7 @@ import {
 } from "kysely";
 import type { D1DialectConfig } from "kysely-d1";
 
-import { EmDashD1Dialect } from "./d1-dialect.js";
+import { D1Adapter, EmDashD1Dialect } from "./d1-dialect.js";
 
 /**
  * Statements safe to coalesce: plain SELECTs. Deliberately conservative —
@@ -284,7 +284,7 @@ export class CoalescingD1Driver implements Driver {
  * `executeQuery` calls — that is the whole point — so report `true`.
  * Transactions are rejected by the driver regardless.
  */
-class CoalescingD1Adapter extends SqliteAdapter {
+class CoalescingD1Adapter extends D1Adapter {
 	override get supportsMultipleConnections(): boolean {
 		return true;
 	}
--- a/packages/cloudflare/src/db/d1-dialect.ts
+++ b/packages/cloudflare/src/db/d1-dialect.ts
@@ -6,11 +6,30 @@
  * d1.ts, and without pulling cloudflare:workers into test environments.
  */
 
+import type { CompoundSelectLimitedAdapter } from "emdash";
 import type { DatabaseIntrospector, Kysely } from "kysely";
 import { SqliteAdapter } from "kysely";
 import { D1Dialect } from "kysely-d1";
 
 import { D1Introspector } from "./d1-introspector.js";
+
+/**
+ * Terms D1 allows in one compound SELECT (`UNION ALL`, `INTERSECT`, `EXCEPT`).
+ * D1 sets SQLITE_LIMIT_COMPOUND_SELECT to 5 where SQLite's upstream default is
+ * 500; a sixth branch is rejected with "too many terms in compound SELECT".
+ * Measured against a live D1.
+ */
+export const D1_COMPOUND_SELECT_LIMIT = 5;
+
+/**
+ * Base adapter for every D1-backed dialect. Declares the compound-SELECT
+ * ceiling, which core reads off the adapter to split statements that would
+ * exceed it; a dialect that overrides `createAdapter()` without extending this
+ * silently sends D1 compound SELECTs it rejects.
+ */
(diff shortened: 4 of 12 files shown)


khoinguyenpham04 (commented):
could we avoid applying d1’s five-collection limit to every database? i tested six collections end to end: node/sqlite goes from one query to two on both admin and public pages. d1 needs this split, but other databases do not. 

plz keep the single query where the limit doesn’t apply.

MA2153:
@khoinguyenpham04 done, it is now isolated per adapter. could you re-check?

packages/core/src/database/dialect-helpers.ts
```diff
@@ -37,6 +37,32 @@ export function isPostgres(db: Kysely<any>): boolean {
...
+}
+
+/**
+ * The backend's compound-SELECT ceiling, or null when it has none worth
+ * splitting statements for. Only the adapter knows: the limit is a property of
+ * the SQLite build behind the dialect, not of the SQL flavour, so two "sqlite"
+ * dialects can answer differently.
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- accepts any Kysely instance
+export function compoundSelectLimit(db: Kysely<any>): number | null {
```
khoinguyenpham04, line 65: can we validate `compoundSelectLimit` here? `0` or negative values make `chunks()` loop forever, while `NaN` or fractions can create invalid batches. pls require a positive integer or throw a clear adapter error.

MA2153: fixed

khoinguyenpham04:
Thanks! @MA2153

PR #1482, opened 2026-06-15
fix: read OAuth env from cloudflare workers runtime

richardjoo:
## What does this PR do?

Fixes OAuth login on Cloudflare Workers/Astro v6 by stopping the OAuth routes from reading provider secrets from the removed `Astro.locals.runtime.env` path. Instead, the routes read runtime bindings from `cloudflare:workers` when available and fall back to `import.meta.env` elsewhere.

Closes #

## Type of change

- [x] Bug fix
- [ ] Feature (requires [maintainer-approved Discussion](https://github.com/emdash-cms/emdash/discussions/categories/ideas))
- [ ] Refactor (no behavior change)
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests
- [ ] Chore (dependencies, CI, tooling)

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md)
- [x] `pnpm --filter emdash typecheck` passes
- [ ] `pnpm lint` passes
- [ ] `pnpm test` passes (or targeted tests for my change)
- [x] `pnpm format` has been run
- [ ] I have added/updated tests for my changes (if applicable)
- [ ] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable). Do not include `messages.po` changes except in translation PRs — a workflow extracts catalogs on merge to `main`.
- [ ] I have added a [changeset](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#changesets) (if this PR changes a published package)
- [ ] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/...

Notes:
- This is intentionally narrow and only addresses the OAuth env lookup path.
- The issue was reproduced on a Cloudflare Workers deployment where `/_emdash/api/auth/oauth/github` failed at runtime with: `Astro.locals.runtime.env has been removed in Astro v6`.
- Targeted package validation run: `pnpm --filter emdash typecheck`.

## AI-generated code disclosure

- [x] This PR includes AI-generated code — model/tool: GPT-5.4 / OpenCode

## Screenshots / test output

- Verified the fix by patching the published package in a real Cloudflare deployment and confirming `/_emdash/api/auth/oauth/github` now redirects correctly to GitHub OAuth instead of failing back to the local login page.

PR #2339, opened 2026-08-05
fix(core): resolve OAuth env safely on Cloudflare

### LINKED ISSUE #1736: OAuth login (Google/GitHub) fails with "Failed to start OAuth flow" — route reads Astro.locals.runtime.env, removed in Astro 6+

richardjoo:
## What does this PR do?

Fixes OAuth provider login on Cloudflare Workers when newer Astro runtimes no longer expose a safe `locals.runtime.env` path.

This replaces the older approach in #1482 with the current local fix path that:
- prefers `locals.runtime.env` when it exists and can be read safely
- avoids crashing when that getter throws on newer Astro/Cloudflare combinations
- falls back through `virtual:emdash/env`, then `cloudflare:workers`, then `import.meta.env`
- adds focused regression tests for the helper and both OAuth routes

Closes #1736

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [x] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

- [x] This PR includes AI-generated code — model/tool: OpenCode / GPT-5.4

## Screenshots / test output

- `pnpm exec oxlint --type-aware -f json | jq '.diagnostics | length'` -> `0`
- `pnpm typecheck` passed
- `pnpm --filter emdash exec vitest run tests/unit/astro/oauth-env.test.ts tests/unit/auth/oauth-start-route.test.ts tests/unit/auth/oauth-callback-route.test.ts --testTimeout=30000` passed


ascorbic:
Thanks for this. I think we can skip checking `locals.runtime.env`, because EmDash doesn't support older Astro versions that provide it. Likewise there's never going to be a scenario where `virtual:emdash/env` is unavailable but `cloudflare:workers` is. Simplify it to just use the virtual module

packages/core/src/astro/routes/api/auth/oauth/env.ts
```diff
@@ -0,0 +1,49 @@
...
+
+		const runtime = Reflect.get(locals, "runtime");
+		if (!isRecord(runtime)) return undefined;
+
+		const env = Reflect.get(runtime, "env");
+		return isRecord(env) ? env : undefined;
+	} catch {
+		return undefined;
+	}
+}
```
ascorbic: This is backwards compat for an Astro version that EmDash doesn't support. It should be removed

ascorbic (changes requested):

richardjoo:
Closing this PR as superseded, rather than because the latest implementation or CI is failing.

After synchronizing this branch with current upstream and reviewing the resulting diff, I confirmed that the original OAuth production fix already landed in PR #1845 as commit `ca660c6c` on July 20. That merged change added `virtual:emdash/env`, switched both the OAuth start and callback routes away from `locals.runtime.env`, added the Cloudflare regression coverage, and was released in `emdash@0.33.0`.

The review feedback on this PR was addressed: the unsupported `locals.runtime.env` compatibility path and direct `cloudflare:workers` fallback were removed, the branch was refreshed against current `main`, and the complete current CI suite passes. However, once the upstream implementation from #1845 is used as the merge base, this PR no longer carries the stated user-visible fix. Its remaining production diff primarily extracts the already-shipped inline virtual-env lookup into a helper, while its changeset would announce the same OAuth fix for a second release.

Keeping this open would therefore require more reviewer time and create a duplicate release note without delivering an additional Cloudflare login correction. If centralizing this lookup is still desirable, it should be proposed separately as a narrowly scoped refactor with a description and changeset decision that reflect its actual behavior.

Thank you to everyone who reviewed the iterations and clarified the supported adapter boundary.

PR #2342, opened 2026-08-05
fix(taxonomies): respect active locale in admin surfaces

### LINKED ISSUE #2224: Admin taxonomy navigation and content editor show definitions from every locale
### Description

When a taxonomy has localized definitions, the EmDash admin renders every locale variant instead of resolving one definition for the active content locale.

With a German recipe opened using `?locale=de-de`, the admin displays both the German and English taxonomy definitions:

- `Gänge` and `Courses` both link to `/taxonomies/course`
- `Küchen` and `Cuisines` both link to `/taxonomies/cuisine`

The content editor's taxonomy panel also renders all four sections. Because term requests are scoped to the recipe's `de-de` locale, both `Gänge` and `Courses` display the German term `Hauptgericht`, while both `Küchen` and `Cuisines` display `Deutsch`.

The seed and database contain the expected localized definitions and terms. There is exactly one definition per `(taxonomy name, locale)`, and corresponding locale variants share a translation group.

### Expected behavior

For a `de-de` recipe editor context:

- The navigation should contain one locale-resolved link per taxonomy name.
- Only `Gänge` and `Küchen` should be shown.
- The recipe editor should render one section for `course` and one for `cuisine`.

Switching to the `en-us` recipe should resolve the labels to `Courses` and `Cuisines`.

### Actual behavior

The navigation contains both locale definitions, and the recipe editor renders all four sections. The English-labelled sections display German terms because term requests are correctly scoped to `de-de`, while the taxonomy-definition request is not.

### Validation performed

- `emdash seed seed/seed.json --validate` passes.
- A fresh D1 database contains one `course` and one `cuisine` definition per locale.
- German and English definition variants share their respective translation groups.
- German and English term variants share their respective translation groups.
- Each recipe entry has only the expected `course` and `cuisine` assignments; no duplicate term assignments exist in the database.

This indicates that the duplication is in admin rendering or locale filtering rather than in the seed or stored taxonomy data.

### Initial investigation

The published `0.31.0` admin code appears to fetch taxonomy definitions without a locale parameter, while taxonomy terms and entry terms are fetched with the entry locale.

The taxonomy definitions endpoint already supports `?locale=<locale>`. Scoping the definition query to the active entry locale, and including that locale in the query cache key, appears to be the relevant path for the content editor.

The main admin navigation also renders every taxonomy definition supplied by the manifest, resulting in multiple links with the same taxonomy route.

### Related issue

This is distinct from #1348. Issue #1348 concerns taxonomy assignments drifting between content translations after the initial translation is created. This report concerns localized taxonomy definitions from multiple locales being rendered simultaneously in the admin navigation and content editor.

### Screenshots

**Admin navigation:** Both locale variants are rendered as separate links for the same taxonomy names.

Admin navigation showing localized taxonomy definitions twice

**DE-DE recipe editor:** German and English taxonomy definitions are rendered simultaneously, while both panels resolve the German terms for the active entry locale.

DE-DE recipe editor showing German and English taxonomy definition panels simultaneously

### Steps to reproduce

1. Configure a site with `de-de` as the default locale and `en-us` as an additional locale.
2. Define localized variants of the same taxonomies:
   - `course`: `Gänge` (`de-de`) and `Courses` (`en-us`)
   - `cuisine`: `Küchen` (`de-de`) and `Cuisines` (`en-us`)
3. Link each translated definition and term using `translationOf`.
4. Attach the taxonomies to a localized collection such as `recipes`.
5. Start from a fresh local database and allow the seed to be applied.
6. Open a German recipe in the admin at `/_emdash/admin/content/recipes/<entry-id>?locale=de-de`.
7. Inspect the main admin navigation and the taxonomy panel in the recipe editor.

### Environment

- EmDash: `0.31.0`
- `@emdash-cms/cloudflare`: `0.31.0`
- Astro: `7.1.1`
- Runtime: local Cloudflare Workers development runtime with Miniflare D1
- Database: fresh local D1 initialized from a validated seed
- Browser: Chromium-based Codex in-app browser
- OS: macOS

### Logs / error output

```shell
No error is thrown. This is an incorrect admin UI state.
```

### LINKED ISSUE #2338: Admin sidebar lists a taxonomy once per locale: the manifest reads _emdash_taxonomy_defs unfiltered and drops locale
### Description

The admin's **Manage** sidebar renders one nav entry per `_emdash_taxonomy_defs` row, with no locale filtering and no grouping by `translation_group`. A site with localized taxonomy definitions therefore sees every taxonomy listed once per locale, in one flat list, with nothing indicating the entries are translations of each other.

On our site (11 taxonomies, 2 locales) the sidebar read:

```
Categories
Catégories
Holacracy versions
Versions d'Holacracy
Industries
Secteurs d'activité
Languages
Langues
Locations
Lieux
…
```

22 entries for 11 taxonomies. The scaling is the concern: 11 taxonomies × 10 locales is 110 sidebar rows naming 11 real things, in what is the primary navigation for taxonomy management.

This is the more surprising because the taxonomy *terms* page already handles locales well — `TaxonomyManager.tsx` has a `LocaleSwitcher`, fetches terms per locale, and offers a translate action. It's only the definition nav that is locale-blind.

### Reproduction

1. Create a taxonomy definition in the default locale.
2. `POST /_emdash/api/taxonomies` with `locale: "<other>"` and `translationOf: "<def id>"`.
3. Open `/_emdash/admin`.

Both rows appear as separate top-level entries under **Manage**.

### Cause

The sidebar is fed by the manifest, not by `GET /taxonomies`. In `src/emdash-runtime.ts`:

```ts
// Build taxonomies from database
let manifestTaxonomies: Array<{
    name: string;
    label: string;
    labelSingular?: string;
    hierarchical: boolean;
    collections: string[];
}> = [];
try {
    const rows = await this.db
        .selectFrom("_emdash_taxonomy_defs")
        .selectAll()
        .orderBy("name")
        .execute();
    manifestTaxonomies = rows.map((row) => ({
        name: row.name,
        label: row.label,
        labelSingular: row.label_singular ?? undefined,
        hierarchical: row.hierarchical === 1,
        collections: parseStringArray(row.collections).toSorted(),
    }));
```

Every row, every locale, and the mapped shape drops `locale` and `translation_group`. `Sidebar.tsx` then does `...manifest.taxonomies.map((tax) => ({ to: "/taxonomies/$taxonomy", params: { taxonomy: tax.name } }))` one-to-one — so the admin has nothing to filter or group on even if it wanted to, and two entries route to the same `tax.name`.

Note the manifest hash is computed over `manifestTaxonomies`, so any fix should keep the hash stable-by-content.

### Suggested fix

Either would work, in the runtime rather than the admin:

1. **Emit one entry per taxonomy**, deduped on `translation_group` (or `name`, which is unique per locale and shared across translations), preferring the default locale's label.
2. **Carry `locale` and `translationGroup` into the manifest entry** and let the admin group or filter — this also lets the sidebar show the label in the admin's current UI language, which seems closer to the intent.

### Secondary: definitions have no update or delete endpoint

Related, and what made this awkward to clean up. `src/api/handlers/taxonomies.ts` exports `handleTaxonomyList`, `handleTaxonomyCreate`, `handleTaxonomyDefTranslations` and the full term CRUD — but no update or delete for definitions, and there is no `routes/api/taxonomies/[name]/index.ts`. A localized definition can be created through the API and then never modified or removed through it. Removing ours required direct SQL against D1.

`apply` scripts also can't reconcile a definition whose label changed upstream — they can only report the mismatch.

### Worth noting: localized definitions may be unnecessary

Relevant to how the fix is framed. Terms resolve through `requireTaxonomyDef(db, taxonomyName)` with no locale argument:

```ts
// handleTermList — src/api/handlers/taxonomies.ts
// Definitions are per-locale but terms aren't bound to the def's locale —
// just ensure the taxonomy exists somewhere.
const lookup = await requireTaxonomyDef(db, taxonomyName);
```

`handleTermCreate` says the same. So a localized definition row carries only the taxonomy's own display name in the admin; localized *term* labels work fine against a single source-locale definition, edited through the `LocaleSwitcher` on the terms page.

We deleted our 11 French definition rows and verified no change to the public site: 397 French term rows intact, French facet labels still rendering, taxonomy-filtered routes still 200. That makes the duplicate nav entries close to pure cost, which may argue for the runtime emitting one entry per taxonomy by default.

### Environment

- `emdash` 0.31.0 / `@emdash-cms/admin` 0.31.0, on Astro + Cloudflare Workers (D1 + R2)
- **Checked against the latest published release before filing:** the `manifestTaxonomies` block above is byte-identical in `emdash` 0.31.1 (`latest`) and in the published 1.0.0, so this is not fixed by upgrading.

Possibly adjacent to #1348 (also `locale` semantics in the taxonomy layer), though the mechanism is unrelated — that one is about the `content_taxonomies` junction, this is only about how definitions reach the nav.

fbartolitsch:
### Summary

- Resolves localized taxonomy definitions once per logical taxonomy before rendering admin navigation and editor choices.
- Preserves the active locale when navigating from the admin sidebar into taxonomy management.
- Uses one shared fallback policy: exact active locale, then the configured default locale, then a deterministic definition.
- Scopes visible taxonomy term counts and their cache entries to the resolved content locale.

### Scope

- Admin sidebar taxonomy navigation and content-editor taxonomy choices.
- Runtime manifest taxonomy identity metadata (`id`, `locale`, and `translationGroup`).
- Core taxonomy definition fallback and locale-scoped count queries.
- Regression coverage for sidebar paths, editor choices, manifest metadata, partially translated sites, and term counts.
- Patch changeset for `@emdash-cms/admin` and minor changeset for `emdash`.

### Validation

- `pnpm format` and `pnpm format:check`
- `pnpm build`
- `pnpm typecheck`
- `pnpm lint` and `pnpm lint:quick`
- Admin regression tests: 2 files, 27 tests passed
- Core regression tests: 3 files, 27 tests passed
- SQLite and D1 query-count generation: 20 route/phase entries each, all measured routes returned 200
- Query-count totals unchanged; SQL snapshots only add the expected locale predicate
- `git diff --cached upstream/main --check`

### Risk and rollback

- The resolver groups explicit translations by `translationGroup`; legacy definitions without that metadata fall back to their taxonomy name.
- Incomplete taxonomy translations select the active locale, then the configured default, then the lexically first locale/id while retaining the requested content locale for term rows and counts.
- Public taxonomy helpers now expose per-locale visible counts instead of totals combined across translations; the `emdash` changeset is therefore minor.
- Count queries add a parameterized locale predicate and locale-aware cache key without increasing logged-out query counts.
- No schema migration or data rewrite is included.
- The branch contains a regular merge of `upstream/main`; no history rewrite or force-push was used.
- Rollback is a mainline revert of the merge commit.

### Security/privacy impact

No new external calls, credentials, permissions, or personal-data processing are introduced.

Fixes #2224
Fixes #2338

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable). No new user-visible strings or `messages.po` changes are included.
- [x] New features link to an approved Discussion: N/A — this is a bug fix.

## Screenshots / test output

Targeted automated tests:

```text
Admin: 2 files, 27 tests passed
Core: 3 files, 27 tests passed
```


fbartolitsch:

packages/core/src/api/handlers/taxonomies.ts
```diff
@@ -389,12 +389,12 @@ export async function handleTermList(
 ): Promise<ApiResult<TermListResponse>> {
 	try {
 		// Definitions are per-locale but terms aren't bound to the def's locale —
-		// just ensure the taxonomy exists somewhere.
-		const lookup = await requireTaxonomyDef(db, taxonomyName);
+		// use the active definition for its collection scope.
+		const locale = options.locale ? resolveConfiguredLocale(options.locale) : undefined;
+		const lookup = await requireTaxonomyDef(db, taxonomyName, locale);
```
ascorbic, line 394: This is a hard `where locale =` match with no fallback (`requireTaxonomyDef` filters exactly when a locale is passed), so a site with DE content but taxonomy defs only defined in EN — the normal state of any partially-translated site — now gets `NOT_FOUND` from the term list for DE entries, where it worked before this PR. `handleTermGet` has the milder version of the same problem: the lenient lookup fails, `defCollections` becomes `[]`, and the count silently reads 0.

The client half of this PR already handles this correctly (`resolveTaxonomyDefinitions`: active locale → default → deterministic first). The server needs the same leniency — prefer the exact-locale def, fall back to any row of the group — since the def is only used for its collections scope here. Something like ordering by `locale = ? DESC` and taking the first row instead of filtering.

Worth a test pinning it: term list for a locale that has terms and content but no translated def row.

ascorbic (changes requested):
The direction is right and most of this is well built — the client-side `resolveTaxonomyDefinitions` fallback chain is exactly the right shape, the cache keys correctly grow a locale component, and the dialect-parameterized test extensions are thorough. Two things before merge: a server-side regression for partially-translated sites (inline below), and the changeset undersells a public-site behavior change.

On the changeset: the per-locale counts flow into `getTaxonomyTerms`, which is the public taxonomy widget API — multi-locale sites will see their live term counts change (from cross-locale totals to per-locale). I agree the new counts are the correct ones — the old totals didn't match what a visitor clicking through actually sees — but that's an observable public-site change dressed in admin-fix language. Say it plainly in the changeset, and I'd bump `emdash` to minor.

ascorbic:
The query snapshots can be resolved by accepting main's version: they'll be updated once you push to CI

ascorbic (approved):
Thanks!

fbartolitsch:
Updated this branch with the upstream merge and adjusted the locale fallback tests. Focused taxonomy tests and lint pass locally. Full `pnpm typecheck` currently fails on `main` because of the unrelated missing `restampEntryPivot` method, tracked in #2466. We will rerun full validation once that fix is merged.

PR #2352, opened 2026-08-06
feat(core): add getBylines() for listing a site's bylines

MA2153:
## What does this PR do?

There is no way for site code to enumerate bylines. The public API offers only identifier-keyed lookups — `getByline(id)`, `getBylineBySlug(slug)` — plus `getEntriesByByline()`, all of which need a byline you already have. The only site-wide enumeration, `BylineRepository.findMany()`, is reachable solely through the admin-gated `GET /_emdash/api/admin/bylines`.

So the only thing a frontend can do today is query a content collection and dedupe `entry.data.bylines`. That reaches `_emdash_bylines` through a join on `_emdash_content_bylines` (itself reached through a content query) to read columns sitting on the byline row, and the result is a function of which entries you fetched: bounded by the content query's `limit`/cursor, and silently missing anyone with no resolvable credits — a guest author added ahead of publication, an author whose posts are all draft or scheduled, an author whose entries exist only in another locale.

This adds one export:

```ts
const { items, nextCursor } = await getBylines({ locale, limit, cursor });
```

backed by `BylineRepository.findManyAlphabetical()`, a byline-table-driven read:

- **Media join included.** `LEFT JOIN media` on `avatar_media_id`, the same join `getContentBylinesMany` already performs. Without it every caller rendering avatars does a `MediaRepository.findById` per byline — an N+1 across exactly the page being built.
- **Strict locale, resolved like `getTaxonomyTerms`.** Byline rows are per-locale, so filtering on the resolved locale yields one row per person and needs no translation-group dedupe.
- **No custom-field hydration.** `findMany` calls `withCustomFields` unconditionally, costing extra queries against the EAV tables. A list of names and avatars doesn't need them; `customFields` comes back `{}`, matching the existing `skipHydration` behaviour. An `includeCustomFields: true` flag would be additive later.
- **Alphabetical, cursor-paginated.** `findMany` orders by `created_at DESC`, which is wrong for an author index. Cursor keyed on `(display_name, id)` via the existing `encodeCursor`/`decodeCursor`; returns `{ items, nextCursor? }` per the pagination convention.
- **Wrapped in `requestCached`,** keyed on every argument.

**Cost: one query.** Opt-in, so no existing route's query count moves and the snapshots are unaffected. One additive migration (`068`, a column on `_emdash_bylines` — see the review follow-up below), no change to existing behaviour or exports — not a breaking change.

Deliberately left out, each additive later: `search`/`isGuest` filters, an `includeCustomFields` flag, and adding the media join to `findMany` so the admin list stops N+1-ing avatars (a separate change to an admin path that shouldn't ride along).

### Index behaviour

`_emdash_bylines` has single-column indexes on `display_name` and `locale` (migration `040`), but no composite. The locale filter is indexed; the ordering is not, so each page sorts the locale partition before `LIMIT` applies and the cursor is not a true seek. This is index-availability only — it does not depend on `sqlite_stat1`. At the byline counts real sites carry, sorting the partition is negligible, and a `(locale, display_name_sort)` composite is a forward-only migration whenever that stops being true. That belongs with #1532, which covers the same concern across taxonomy terms, bylines, users, and media.

Ordering moved to `display_name_sort` (below), which carries no index at all, so the plan is unchanged: the planner never used `idx_bylines_display_name` for this query either — the locale filter won, and the sort was already a temp B-tree. `068` deliberately adds no index, to keep the composite question in one place.

### Review follow-up

**Pre-migration databases** (https://github.com/emdash-cms/emdash/pull/2352#discussion_r3743117192). Fixed in 807c6331 — `getBylines()` was the odd one out: `getTaxonomyTerms` returns `[]` and `getBylinesForEntries` further down the same file already catches this, while `getBylines()` propagated the `no such table` error. It now catches `isMissingTableError` and returns `{ items: [] }`. Covered by a new case in `tests/unit/bylines/bylines-query.test.ts` that drops `_emdash_bylines` and asserts the empty result; it fails with the `SqliteError` before the fix.

**Collation** (https://github.com/emdash-cms/emdash/pull/2352#discussion_r3765977657). Fixed in 1dbe2798. Confirmed: ordering on `display_name` used the database's own collation, BINARY on SQLite, so the list read `Adam, Zoe, alice, Álvaro` — and `Øyvind` landed after `Zoe` too, since `ø` doesn't decompose.

Migration `068` adds `_emdash_bylines.display_name_sort`: the display name casefolded, NFD-decomposed with combining marks dropped, and with the Latin letters that carry no mark to drop (`æ œ ø ð þ ß đ ł ħ ı ŋ ŧ ĸ`) expanded (`utils/sort-key.ts`). `create` and `update` write it, and the migration backfills existing rows in batches sized for D1's parameter ceiling, following `056`'s pattern. `findManyAlphabetical` orders on that column and seeks its cursor on it, so the two compare the same value by construction — which is the half that actually breaks pagination if they diverge.

It was numbered `059` when those threads were written; `main` has since claimed `059`–`067`, so successive merges have renumbered it to `068`. Nothing about it changed but the number.

Stored rather than computed in the query, for two reasons: neither dialect can produce this key in SQL (SQLite's `lower()` is ASCII-only, `NOCASE` folds ASCII case and nothing else, and no dialect has `unaccent` available on D1), and a cursor value computed in JS has to compare equal to what SQL sorted. A side effect is that the order no longer depends on the dialect's collation: a Latin-script key is plain ASCII, so SQLite and Postgres agree. Scripts outside Latin keep their letters and still sort by code point — that needs ICU, which D1 doesn't carry, and it's no worse than today.

Covered by `tests/unit/database/migrations/068_byline_display_name_sort.test.ts` (backfill, including a row past the first batch; a re-run against a database that already has the column; `down()`), plus three cases in the repository test: the mixed-case/accent ordering, the same list walked page by page (this is the one that fails if `ORDER BY` and the cursor disagree), and a rename re-sorting the row. The first two fail on `main`'s ordering; the rename case fails if `update` stops writing the key — I checked each by reverting the relevant line.

**Stale request cache** (https://github.com/emdash-cms/emdash/pull/2352#discussion_r3765977661). Fixed in 1dbe2798. Confirmed, and the keys carry locale, limit, and cursor, so a write path can't name the entries it invalidates — hence `clearRequestCachePrefix`, alongside the existing `clearRequestCacheEntry`. `create`, `update`, and `delete` drop the whole `bylines:` namespace, next to the `invalidateBylineObjectCache()` call each already makes. The key is built in one place (`bylines/cache-keys.ts`) so the reader and the write paths can't drift. Covered by a case that reads, creates, renames, and deletes inside one `runWithContext`, asserting the list after each; it returns the stale first page before the fix.

Scoped to the key this PR introduces. `byline-by-slug:`, `taxonomy-terms:`, `menu:`, and `siteSetting:` have the same gap today and none are cleared on write — worth a sweep, but not from this PR.

**Object-cache layer** deferred to a follow-up, as agreed. `getBylines()` stays on `requestCached` only for now.

Implements #2333. No issue to close.

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (or targeted tests for my change) — full `packages/core` suite: 5668 passed, 9 skipped. `packages/registry-verification` fails on my machine for an unrelated environment reason (node choking on the pnpm binary); it fails identically on a clean checkout of `main`.
- [ ] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable) — n/a, no admin UI in this PR
- [x] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/2333

## Screenshots / test output

Tests, all new:

- `tests/unit/database/repositories/byline.test.ts` — alphabetical order with pagination covering every row exactly once; mixed-case and accented names alphabetized, and that same list walked page by page; a rename re-sorting the row; a display-name tie split across a page boundary (the `id` half of the cursor); the media join populating `avatarStorageKey`/alt/blurhash/dominant colour; `locale` returning one row per person; `customFields` staying `{}` while `findById` still reads the seeded value.
- `tests/unit/bylines/bylines-query.test.ts` — a guest author and a draft-only author both appearing in the list (the gap this closes); implicit locale resolving to the configured default rather than returning every locale variant; no fallback when the requested locale has no row; cursor pagination; an empty result on a database missing `_emdash_bylines`; writes during a request invalidating the list read earlier in it.
- `tests/unit/database/migrations/068_byline_display_name_sort.test.ts` — the backfill, a row past the first batch, a re-run against a database that already has the column, and `down()`.
- `tests/integration/byline-list-plan.test.ts` — a query-plan guard following the existing `*-plan.test.ts` pattern (no `ANALYZE`, since D1 never maintains `sqlite_stat1`): locale-index seek with and without a cursor, media reached by primary key, no `SCAN` on either table. Index names are matched by prefix regex so a later locale-leading composite doesn't fail the test on an improvement.

```
|--SEARCH b USING INDEX idx__emdash_bylines_locale (locale=?)
|--SEARCH m USING INDEX sqlite_autoindex_media_1 (id=?) LEFT-JOIN
`--USE TEMP B-TREE FOR ORDER BY
```

I checked the plan test fails on a real regression rather than just passing: dropping `idx__emdash_bylines_locale` fails the two seek assertions, and removing the media join fails the primary-key one.


packages/core/src/bylines/index.ts
```diff
@@ -107,6 +111,53 @@ export async function getBylineBySlug(
...
+		const db = await getDb();
+		const repo = new BylineRepository(db);
+
+		const repoOptions: { locale?: string; limit?: number; cursor?: string } = {};
+		if (locale !== undefined) repoOptions.locale = locale;
+		if (options.limit !== undefined) repoOptions.limit = options.limit;
+		if (options.cursor !== undefined) repoOptions.cursor = options.cursor;
+
+		return repo.findManyAlphabetical(repoOptions);
+	});
```
ascorbic, line 164: On a site that hasn't run the bylines migrations yet this throws, where the equivalent enumeration helpers degrade to empty: `getTaxonomyTerms` returns `[]`, and `getContentBylines` further down this file catches the same case and returns an empty result. A public helper should behave like its siblings — catch `isMissingTableError` (already imported here) and return `{ items: [] }`.

ascorbic (commented):
This looks good. One change before merge. Deferring the object-cache layer to a follow-up is fine.

packages/core/src/bylines/index.ts
MA2153: Done

packages/core/src/database/repositories/byline.ts
```diff
@@ -600,6 +600,86 @@ export class BylineRepository {
...
+				"m.dominant_color as avatar_dominant_color",
+				"b.website_url as website_url",
+				"b.user_id as user_id",
+				"b.is_guest as is_guest",
+				"b.created_at as created_at",
+				"b.updated_at as updated_at",
+				"b.locale as locale",
+				"b.translation_group as translation_group",
+			])
+			.orderBy("b.display_name", "asc")
```
ascorbic: SQLite uses BINARY collation here, so mixed-case and accented names are not alphabetized for users: `Adam, Zoe, alice, Álvaro`. Use a stable normalized/collated sort key, and apply the same semantics to both `ORDER BY` and the cursor predicates.

packages/core/src/bylines/index.ts
```diff
@@ -107,6 +111,58 @@ export async function getBylineBySlug(
...
+ * for (const byline of items) {
+ *   console.log(byline.displayName, byline.avatarStorageKey);
+ * }
+ * ```
+ */
+export async function getBylines(
+	options: { locale?: string; limit?: number; cursor?: string } = {},
+): Promise<FindManyResult<BylineSummary>> {
+	const locale = resolveLocale(options.locale);
+	const cacheKey = `bylines:${locale ?? "*"}:${options.limit ?? "*"}:${options.cursor ?? "*"}`;
```
ascorbic: This cache is never cleared by `BylineRepository.create`, `update`, or `delete`. A `getBylines()` → write → `getBylines()` sequence in one request returns the stale first result. Clear the `bylines:` request-cache namespace from each byline write path.

ascorbic (changes requested):
Two correctness issues block the advertised alphabetical list and fresh same-request reads.

packages/core/src/database/repositories/byline.ts
MA2153: Done. This has resulted in a migration. Currently the sort key implementation is rudimentary but can be expanded later. I think this same surface could also be repurposed for FTS.

packages/core/src/bylines/index.ts
MA2153: Fixed

ascorbic:
I'm afraid It needs a migration renumbering.

MA2153:
Renumbered

PR #1568, opened 2026-06-21
fix(admin): refetch content after save/publish on non-i18n sites (#1557)

marcusbellamyshaw-cell:
## What does this PR do?

Fixes the publish/save controls going stale after editing a published post, as reported in #1557. After saving a title change, the **"Publish changes"** button now appears immediately (previously it only showed after a hard refresh), and after publishing, the button immediately flips to **"Unpublish"** (previously it stayed on "Publish changes").

**Root cause:** the editor reads the content item under a React Query key scoped to the *active* locale — which is `undefined` when i18n is not configured (`router.tsx`, `ContentEditPage`). The save/publish/unpublish/discard/schedule mutations, however, invalidated a key scoped to `rawItem.locale`, which is the database default `"en"` even on non-i18n sites. React Query's partial matcher compares `{ locale: undefined }` against `{ locale: "en" }`, finds no match, and never refetches the item — so the draft-status pointers (`liveRevisionId` / `draftRevisionId`) the buttons key off stay stale until a manual refresh. This is why both reporters see it "since forever": it triggers on any site without i18n enabled.

**Fix:** invalidate by the `["content", collection, id]` prefix (dropping the locale object) in all six content mutations. The `id` already uniquely identifies the row (translations are separate rows), so the locale object was redundant for uniqueness — and a prefix match correctly refetches the item regardless of which locale tab is active.

Closes #1557.

## Type of change

- [x] Bug fix
- [ ] Feature (requires [maintainer-approved Discussion](https://github.com/emdash-cms/emdash/discussions/categories/ideas))
- [ ] Refactor (no behavior change)
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests
- [ ] Chore (dependencies, CI, tooling)

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md)
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (or targeted tests for my change)
- [x] `pnpm format` has been run
- [x] I have added/updated tests for my changes (if applicable)
- [x] User-visible strings in the admin UI are wrapped for translation (if applicable) — no new strings; `messages.po` not touched.
- [x] I have added a [changeset](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#changesets) (if this PR changes a published package)
- [ ] New features link to an approved Discussion — n/a (bug fix)

[...]

Issue #1557, opened 2026-06-20
Bug with publish button when editing posts


mdong1909:
### Description

2 issues with the publishing button:
 
1. After changes are saved, there is no option to publish the changes
2. After refreshing, user can then publish the changes. But the button continues to be active

### Steps to reproduce

https://github.com/user-attachments/assets/3c532989-3404-4c0a-9fe1-6aba6a00fa2f

1. Edit post's title & save
2. No publish button
3. Refresh screen
4. Click publish

### Environment

- emdash: 0.21.0
- Node: v24.17.0
- runtime: cloudflare
- os: mac

### Logs / error output

```shell

```

segmentationfaulter:
yeah, I have been encountering it since forever.

marcusbellamyshaw-cell:
Tracked this down — root cause is a stale React Query cache, and it reproduces on any site **without i18n configured** (which matches both of you seeing it "forever").

The editor reads the post under a query key scoped to the active locale, which is `undefined` when i18n is off. But the save and publish actions invalidated a key scoped to the post's stored `locale` (the DB default `"en"`). The two keys never match, so after a save/publish the editor never refetches the post — the draft-status pointers it uses to decide between **"Publish changes"** and **"Unpublish"** stay stale until you hard-refresh. That's exactly the two symptoms reported:

1. After saving, no "Publish changes" button (the editor still thinks there are no pending changes).
2. After refreshing + publishing, the button stays on "Publish changes" instead of flipping to "Unpublish".

Fix is up in #1568 — it invalidates the content query by `(collection, id)` so it refetches regardless of locale. It includes a regression test that renders the real editor and covers both symptoms (fails on `main`, passes with the fix).

PR #2363, opened 2026-08-07
fix(admin): patch autosave result into the observed content query (#2362)

### LINKED ISSUE #2362: Autosave never surfaces the publish button on sites without i18n (locale-scoped setQueryData key)
### Description

On a site without `i18n` configured, editing a published entry and letting the debounced autosave fire persists the change server-side but never surfaces the "Publish" button. The editor keeps presenting the entry as fully published ("Unpublish Post") until the page is reloaded.

This is the same class of defect as #1557 — and its fix (#1568) is correct — but it left the autosave path untouched, because that path patches the cache instead of invalidating it.

**Root cause.** The editor reads the item with a locale-scoped key (`packages/admin/src/router.tsx`):

```js
queryKey: ["content", collection, id, { locale: activeLocale }]
```

With i18n off, `activeLocale` is `undefined`, so React Query hashes the key as `[..., {}]`.

After a successful autosave, `patchAutosaveQueries` writes the fresh item back with `setQueryData`:

```js
queryClient.setQueryData<ContentItem>(
  locale ? ["content", collection, id, { locale }] : ["content", collection, id],
  savedItem,
);
```

`locale` here is `variables.targetLocale`, i.e. `rawItem?.locale ?? activeLocale` — the DB default `"en"` even on a single-locale site. `setQueryData` matches exactly, so the write lands on `[..., { locale: "en" }]`, a cache entry no component observes. The editor keeps the revision pointers it loaded with, `getDraftStatus()` still returns `"published"` instead of `"published_with_changes"`, and `ContentSettingsPanel` renders "Unpublish Post" rather than "Publish".

Server-side behaviour is correct throughout: `publish()` clears `draft_revision_id`, so the next autosave takes the create-a-new-draft-revision branch in `emdash-runtime.ts` and the entry genuinely is "published with changes". Only the client cache is wrong, which is why a reload fixes it.

**Why it is easy to miss.** Three things blunt the symptom:

1. The other half of the patch — `["revision", draftRevisionId]` — has no locale in its key and always lands, so titles, slugs and field values behave correctly. Only the revision *pointers* go stale.
2. `App.tsx` sets `staleTime: 60000` and leaves `refetchOnWindowFocus` at its default, so a minute later the next window focus refetches and the button appears. It reads as intermittent rather than broken.
3. `router.test.tsx` has a suite named `ContentEditPage – autosave cache patching` running exactly the affected configuration (`i18n: undefined`, item `locale: "en"`), and it passes — because `ContentEditor` is mocked there and the assertions read title and slug, both of which come from the revision patch that works.

### Steps to reproduce

1. Run a site with no `i18n` option passed to the `emdash()` integration (content rows therefore carry the default `locale = "en"`).
2. Open a published entry on a collection with revisions support. The editor shows "Unpublish Post" and no publish button.
3. Edit a field and wait out the 2s autosave debounce without pressing Save. The autosave completes ("Saved").
4. No "Publish" button appears.
5. Reload the page — the button is there, and the entry was "published with changes" server-side all along.

Regression test and a fix are in the PR that follows.

### Environment

- emdash version: reproduced on 0.31.1, present on `main` (776d65f7)
- Node.js version: 22
- Runtime: Cloudflare Workers (D1); not runtime-specific — the defect is entirely in the admin SPA
- OS: macOS

danielmlr:
## What does this PR do?

On a site without `i18n` configured, an autosave persisted the edit but never surfaced the "Publish" button — the editor kept showing the entry as fully published until a reload.

The editor reads the item under a locale-scoped key, and `activeLocale` is `undefined` when i18n is off:

```js
queryKey: ["content", collection, id, { locale: activeLocale }]
```

`patchAutosaveQueries` wrote the saved item back with `setQueryData` under an exact key built from `variables.targetLocale` (`rawItem?.locale ?? activeLocale`), which is the DB default `"en"` even on a single-locale site. Exact matching meant the patch landed on `[..., { locale: "en" }]` — a cache entry nothing observes. The editor kept the revision pointers it loaded with, `getDraftStatus()` returned `"published"` instead of `"published_with_changes"`, and no publish button rendered.

This is the same mismatch #1557 described. #1568 fixed it for every `invalidateQueries` call but not here, because this path deliberately patches the cache instead of invalidating — the autosave must not refetch server state into the form while the user is still typing (#295 / #302). So the fix keeps the patch and only widens its match to the `(collection, id)` prefix, exactly as the mutation invalidations already do. `["content", collection, id, …]` has a single reader, so nothing else is touched.

Server-side behaviour was correct throughout — `publish()` clears `draft_revision_id`, the next autosave creates a new draft revision, and the entry genuinely is "published with changes". Only the client cache was wrong, which is why a reload always fixed it.

**On the test placement.** `router.test.tsx` already has a suite called `ContentEditPage – autosave cache patching` running the affected configuration (`i18n: undefined`, item `locale: "en"`), and it passes on the broken code: it mocks `ContentEditor` and asserts title and slug, which both come from the `["revision", draftRevisionId]` half of the patch — that key has no locale and always landed. The regression therefore needs the real editor, so the new test goes in `publish-button-locale.test.tsx` alongside the #1557 cases. It fails on `main` and passes with this change.

One detail worth flagging for reviewers: the new test registers the PUT handler as `/_emdash/api/content/posts/post_1?locale=en`. Without the query string, `createMockFetch`'s prefix fallback serves the GET handler's response to the PUT, and the test passes for the wrong reason.

Closes #2362

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (or targeted tests for my change) — full `@emdash-cms/admin` suite: 115 files, 1358 tests
- [x] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable) — n/a, no strings changed; no `messages.po` changes included

## Screenshots / test output

Regression test on `main`, fix reverted:

```
 ❯ tests/publish-button-locale.test.tsx:388:4
Caused by: Error: Matcher did not succeed in time.

 Test Files  1 failed (1)
      Tests  1 failed | 2 passed (3)
```

With the fix applied:

```
 Test Files  1 passed (1)
      Tests  3 passed (3)
```

Full admin suite:

```
 Test Files  115 passed (115)
      Tests  1358 passed (1358)
```


danielmlr:

danielmlr:
> The only thing that gives me pause is that the prefix patch updates **every** cached locale variant for the same `(collection, id)` […] if a user switches locales and then autosaves, a previously viewed locale's cache entry could briefly hold the current locale's item until its next refetch.

I looked into whether that state is reachable, and I don't think it is: a given `id` never has more than one locale variant in the cache.

Translations are separate rows with their own IDs (row-per-locale, migration `019_i18n.ts`), and every navigation into the editor pairs an ID with *its own* locale rather than re-reading the same ID under a different one:

- `ContentSettingsPanel.tsx` (opening an existing translation) — `params: { collection, id: tr.id }, search: { locale: tr.locale }`
- `router.tsx`, `createTranslationMutation.onSuccess` — `params: { collection, id: result.id }, search: { locale: result.locale }`
- `LocaleSwitcher` is mounted on the list route, where switching navigates to `/content/$collection` and the list then yields each row's own ID

So `["content", collection, id, { locale }]` only ever gets populated with that row's own locale, and the prefix filter matches exactly the one entry the editor observes. In non-i18n mode it's the single `{ locale: undefined }` entry, which is the case the PR fixes.

Worth noting the prefix is also the more robust option in the one place where the target genuinely diverges from the route: `autosaveMutation` carries `variables.targetId`, which can be a previously edited entry whose in-flight autosave resolves after the user has already navigated to another one (`ContentEditPage` has explicit handling for this — `recordAutosaveCompletion` / `autosaveCompletionSequenceRef`). Patching under the *current* `activeLocale` would miss that entry's key; the prefix hits it.

Happy to narrow it to an exact `{ locale: activeLocale }` key if a maintainer prefers the tighter scope — it fixes the reported bug just as well, at the cost of the stale-target case above.

ascorbic (approved):
I agree: the error state is unreachable. Thanks for this!

PR #2370, opened 2026-08-08
fix(core): let the configured siteUrl steer outbound email links

danielmlr:
## What does this PR do?

`getSiteBaseUrl` — the only origin source for magic-link, invite, signup-confirmation, recovery and comment-notification emails — read only the stored `emdash:site_url` option. That option is written **once**, by the setup wizard (`setIfAbsent`, deliberately, as the Host-spoofing lock), and Admin → Settings writes the different key `site:url`. So a site that completed setup on a throwaway origin (a `*.workers.dev` preview URL, a staging host) kept mailing that origin forever: the operator sets `siteUrl` in the integration options as the docs prescribe ("the single option for the public origin"), every other origin-dependent feature follows `getPublicOrigin` to the new host — and email links still point at the dead one. The only remedy was a manual `UPDATE options …` against the production database. We hit exactly this on a production site.

This PR gives `getSiteBaseUrl` the same precedence as `getPublicOrigin`: configured origin (`config.siteUrl`, then `EMDASH_SITE_URL`/`SITE_URL`) → stored setup origin → request URL (pre-setup only). The request still never overrides a configured or stored value, so the anti-spoofing property is unchanged. As a side effect, deployments with a configured `siteUrl` skip the options query on the email path entirely.

Changes:
- `api/public-url.ts`: extract `getConfiguredOrigin(config)` (config → env, no request fallback); `getPublicOrigin` now uses it — no behavior change there.
- `api/site-url.ts`: accept an optional `config` and apply it first.
- All six call sites pass `emdash.config` (they already have it in scope).
- New test `tests/integration/api/site-url.test.ts` covering the four precedence cases, including a regression guard for the Host-spoofing lock. The first test fails on `main`.

An alternative fix would be re-syncing the stored option at runtime init when `config.siteUrl` is set — happy to rework in that direction if you prefer it; this variant is the smaller diff and matches the documented "single option" contract.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes (core package, `tsc --noEmit`)
- [x] `pnpm lint` passes (`lint:json` → 0 diagnostics)
- [x] `pnpm test` passes (targeted: `tests/integration/api/site-url.test.ts`, `tests/integration/astro/setup-site-url-lock.test.ts`, `tests/integration/auth`, `tests/integration/comments`, `tests/unit/plugins/email-pipeline.test.ts` — 227 tests)
- [ ] User-visible strings in the admin UI are wrapped for translation — n/a, no admin UI change

## Screenshots / test output

```
 Test Files  2 passed (2)
      Tests  8 passed (8)   # site-url.test.ts + setup-site-url-lock.test.ts

 Test Files  10 passed (10)
      Tests  219 passed (219)  # integration/auth, integration/comments, email-pipeline
```

On main, the new precedence test fails with:

```
AssertionError: expected 'https://my-site.workers.dev/_emdash' to be 'https://real.example/_emdash'
```

### DIFF
--- a/.changeset/fix-site-base-url-config.md
+++ b/.changeset/fix-site-base-url-config.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes outbound email links (magic link, invites, signup confirmation, account recovery, comment notifications) ignoring the configured `siteUrl`. The configured origin — `siteUrl` in the integration options or the `EMDASH_SITE_URL`/`SITE_URL` environment variables — now takes precedence over the origin stored during setup, so a site set up on a temporary domain no longer needs a manual database edit to correct its email links.
--- a/packages/core/src/api/public-url.ts
+++ b/packages/core/src/api/public-url.ts
@@ -10,7 +10,7 @@
  */
 
 /** Minimal config shape — avoids importing the full EmDashConfig type tree. */
-interface SiteUrlConfig {
+export interface SiteUrlConfig {
 	siteUrl?: string;
 }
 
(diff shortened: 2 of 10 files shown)


ascorbic (approved):
Thanks!

PR #2374, opened 2026-08-08
fix(core): fold the build into the route cache validator

### LINKED ISSUE #2369: cacheHint.lastModified tracks content only, so a code-only deploy serves 304 and leaves browsers on HTML referencing deleted /_astro/* files
### Description

`CacheHint.lastModified` describes when the *content* changed. Astro's cache layer turns it into the HTTP `Last-Modified` of the *response*. Those are not the same thing: the response also depends on the build, because Astro's asset filenames are content-hashed.

After a deploy that changes only code, `Last-Modified` is unchanged, so a returning visitor's `If-Modified-Since` is answered with **304**. The browser keeps its cached HTML — which references `/_astro/*` files that the new deployment no longer has. On Cloudflare Workers those requests return **404**, and the page renders without CSS and without JavaScript.

Expected: a conditional request made with a pre-deploy validator returns 200, because the response body changed.
Actual: it returns 304, and the visitor is left on HTML whose assets are gone.

What makes this hard to catch is that nothing is logged and every server-side check looks healthy — each page and each current asset returns 200 for a direct request. I hit it on two production sites and could not reproduce it for a day, because I was testing the server rather than the returning visitor.

### Steps to reproduce

1. Deploy an emdash site on Cloudflare Workers with `cache.provider: cacheCloudflare()` and `routeRules`.
2. Load a page. Note its `Last-Modified` and one hashed asset from the HTML.
3. Change code only — do not touch any content. Deploy.
4. Request the old asset. It returns **404**: Workers Assets does not keep files from previous versions.
5. Revalidate the page with the pre-deploy timestamp:

```bash
curl -o /dev/null -w '%{http_code}\n' \
  -H "If-Modified-Since: <pre-deploy Last-Modified>" https://<host>/
```

Expected 200, actual **304** — and the HTML the browser keeps still points at the asset from step 4.

### Environment

- emdash version: 0.32.0 (first observed on 0.31.1)
- astro: 7.2.x, `@astrojs/cloudflare`: 14.2.x, `cache.provider: cacheCloudflare()`
- Node.js version: 26.7.0
- Runtime: Cloudflare Workers (static assets, D1, R2)
- OS: macOS

### Logs / error output

```
Last-Modified of the page:  Fri, 07 Aug 2026 15:52:53 GMT   <- content mtime
latest deployment:          2026-08-07T22:26:49Z            <- code-only deploy

/_astro/BaseLayout.B7gIUk3W.css   -> 404   (previous deployment)
/_astro/BaseLayout.bGLZnLpo.css   -> 200   (current deployment)

curl -H "If-Modified-Since: Fri, 07 Aug 2026 15:52:53 GMT" https://<host>/
  -> 304
```

---

## Root cause

`CacheHint` carries no build or version dimension — `packages/core/src/query.ts:203`:

```ts
export interface CacheHint {
	tags?: string[];
	lastModified?: Date;
}
```

and `lastModified` is the content row's `updated_at` — `packages/core/src/loader.ts:1317`, `:1355`, `:1507`, `:1524`:

```ts
lastModified: row.updated_at ? new Date(rowStr(row, "updated_at")) : undefined,
```

Astro's cache layer emits that verbatim as the response validator (`astro/cache/provider-utils`, `setConditionalHeaders`), via `@astrojs/cloudflare`'s `setHeaders()`.

This only becomes visible once the browser revalidates at all. Without a `Cache-Control` header the browser falls back to RFC 9111 heuristic freshness (roughly 10% of the age of `Last-Modified`) and keeps the stale page for hours without asking — the same breakage, arriving later and even more quietly.

Cloudflare documents this exact failure mode for the *concurrent-versions* case, under [Version affinity](https://developers.cloudflare.com/workers/versions-and-deployments/gradual-deployments/version-affinity/):

> Without version affinity, a user can receive HTML from version A, but when their browser requests `index-a1b2c3d4.js`, that request may be routed to version B — which does not have that file — resulting in a 404 error and a broken page.

The mechanism here is the same, but the second version is the browser cache rather than a concurrently deployed Worker, so version affinity does not cover it.

## An ETag does not close this on Workers

The obvious remedy is a response-derived `ETag`, and `CacheOptions.etag` exists in Astro. I tried it. **Cloudflare removes the `ETag` header from Worker HTML responses.**

- `Astro.cache.options` contained the ETag as the response left the Worker.
- A mirror of the identical value under a custom header name arrived at the client; `etag` did not.
- Independent of compression (`Accept-Encoding: identity` behaves the same) and of validator strength (`W/"…"` and `"…"` both removed).
- Static assets served from the Assets binding keep their ETag, so it is specific to the Worker HTML response.

So on this platform `If-None-Match` is not available, and `Last-Modified` has to carry the whole job.

## What I checked before filing

I went looking for the established pattern, on the assumption that I had configured something wrong. `blog.cloudflare.com` runs on this stack and is maintained by Cloudflare, so it seemed like the best place to learn how this is meant to be handled. It shows the same header shape as my sites: no `Cache-Control` on the HTML, a content-derived `Last-Modified`, `304` on a conditional request, content-hashed `/_astro/*`, and `404` with `no-store` for an `/_astro/*` name outside the current deployment.

I mention it only because it suggests there is no configuration I was missing — and because a blog publishing several times a day is largely shielded from the consequences. Frequent content edits keep `Last-Modified` moving, so the validator invalidates itself and the window between a code deploy and the next content change stays small. On a site whose content is stable for weeks, that window is the normal state.

## Suggested fix

Fold a build or deploy identifier into the value that becomes the response validator, so `Last-Modified` reflects the response rather than only the content. Astro's current implementation accumulates `lastModified` by keeping the later date, so this composes cleanly with per-entry hints.

`packages/core/src/astro/middleware.ts` already post-processes responses (security headers, `Server-Timing`), which seems like a natural home for it.

If that is considered out of scope for core — the hint is arguably correct in describing content, and it is Astro that promotes it to a response header — then documenting the caveat together with the workaround below would already save the next person the debugging. Happy to open a PR for whichever direction you prefer.

## Workaround for anyone hitting this now

Two pieces. First, the browser needs a `Cache-Control` at all: `routeRules` only ever emit `Cloudflare-CDN-Cache-Control`, which Cloudflare consumes and strips, and no zone-level Cache Rule applies because a Worker is zoneless. Second, the validator needs the deploy dimension.

```jsonc
// wrangler.jsonc
"version_metadata": { "binding": "CF_VERSION_METADATA" },
```

```ts
// src/middleware.ts
import { defineMiddleware } from "astro:middleware";

export const onRequest = defineMiddleware(async (context, next) => {
	const response = await next();
	if (context.url.pathname.startsWith("/_emdash/")) return response;
	if (!response.headers.get("content-type")?.startsWith("text/html")) return response;

	if (!response.headers.has("cache-control")) {
		response.headers.set("cache-control", "public, max-age=0, must-revalidate");
	}

	const version = (env as Env).CF_VERSION_METADATA as WorkerVersionMetadata | undefined;
	const cache = context.cache;
	if (!version || !cache?.enabled) return response;

	const deployedAt = new Date(version.timestamp);
	if (Number.isNaN(deployedAt.getTime())) return response;

	// Must go through the cache API rather than response.headers: Astro's cache
	// layer runs after middleware and overwrites Last-Modified from the CacheHint.
	cache.set({ lastModified: deployedAt });

	return response;
});
```

Verified on two sites afterwards: pre-deploy validator → 200 on every page checked, current validator → 304, and every `/_astro/*` reference in the freshly served HTML → 200.

danielmlr:
## What does this PR do?

Closes #2369

`CacheHint.lastModified` carries the content row's `updated_at`, and Astro emits it as the response `Last-Modified`. The response also depends on the build: `/_astro/*` filenames are content-hashed, and a deployment only serves its own. After a deploy that changes only code, the validator is unchanged — a returning visitor's `If-Modified-Since` is answered with **304**, and the browser keeps HTML referencing assets the new deployment no longer has. On Workers those requests 404 and the page renders without CSS or JavaScript. Full reproduction and root-cause analysis in #2369.

The obvious remedy — a response-derived `ETag` — is unavailable on the primary platform: Cloudflare strips `ETag` from Worker HTML responses (verified in the issue), so `Last-Modified` has to carry the build dimension too.

**The fix:** a new `virtual:emdash/build` module exports a build timestamp (captured once per Vite plugin instance, so client/server passes agree), and the middleware folds it into the route cache validator via `cache.set({ lastModified })` before `next()`. Verified against Astro's `AstroCache` semantics:

- Astro keeps the **later** of two dates, so a route's own content hint still wins whenever content is newer than the build.
- Running before `next()` matters: `set()` clears a previous `set(false)`, so applying it after the route would silently undo a route's explicit opt-out. A test pins this ordering.
- A bare `lastModified` never activates caching — Astro only emits headers when `maxAge` or tags are set. Sites that don't use `Astro.cache` see no behavior change.
- Prerendered pages are skipped: the host's static layer manages its own validators (Workers Assets keeps ETags there).

I first tried deriving the deploy time from `CF_VERSION_METADATA`, but that binds core to `cloudflare:workers` and requires a binding no template configures. The build timestamp is adapter-agnostic and fixes the same bug on Node deployments.

**Trade-off, deliberately accepted for minimality:** a pure build timestamp invalidates on *every* deploy, including deploys where the `/_astro/*` hashes did not change — cost is one full 200 instead of a 304 per returning visitor per deploy. A hash over the asset manifest would be more precise but couples core to adapter internals. If you consider core the wrong layer for this altogether — the hint arguably describes content correctly, and it is Astro that promotes it to a response header — say so and I'll turn this into a docs note with the middleware workaround from #2369 instead.

Not included (possible docs follow-up): without any `Cache-Control` on HTML, browsers fall back to RFC 9111 heuristic freshness and may not revalidate for hours; a default header is a behavior change for every site and deserves its own discussion.

**Revised 2026-08-10, after MA2153's review:** merged `main` (picks up #2398), dropped the test case that only exercised its stand-in — the file now has four tests — and documented the rollback limitation in the `applyBuildValidator` doc comment. Details in the review reply. Full `packages/core` unit suite after the merge: 3610 passed, with the same single pre-existing `virtual-modules.test.ts` failure noted in the checklist.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (or targeted tests for my change) — full `packages/core` unit suite: 3532 passed; the one failure (`virtual-modules.test.ts > watches resolved sandbox plugin entries`) also fails on a clean `main` checkout in my environment (macOS tmpdir) and is untouched by this PR
- [ ] User-visible strings in the admin UI are wrapped for translation — n/a, no admin UI changes

- [x] This PR includes AI-generated code — model/tool: Claude Opus 5 (draft), Claude Fable 5 (review pass)

## Screenshots / test output

The reproducing test, on `main` before the fix:

```
FAIL  tests/unit/astro/middleware-cache-validator.test.ts
  × raises a content-only validator to the build time
AssertionError: expected 1786120009000 to be 1786141609000
```

After the fix:

```
✓ tests/unit/astro/middleware-cache-validator.test.ts (4 tests)
✓ tests/unit/astro/integration/virtual-modules.test.ts (20 tests)
```


danielmlr:
Applied the suggestion in `2f978c49` — the block comment is gone. Thanks for the review, and for merging `main` in yesterday.

CI on that SHA is 43/44 green. The one failure is `Browser Tests`:

```
FAIL tests/components/settings/AllowedDomainsSettings.test.tsx
  > AllowedDomainsSettings > updates the default role from the edit dialog
AssertionError: expected "vi.fn()" to be called with arguments: [ 'example.com', { defaultRole: 40 } ]

  1st vi.fn() call:
-     "defaultRole": 40,
+     "defaultRole": 10,

Number of calls: 1
Tests  1 failed | 1383 passed (1384)
```

It appears unrelated to this PR, which touches `packages/core` only. Across the three SHAs this branch has had, `git diff -- packages/admin` is empty in both directions, so that package is identical throughout — yet the same test flips:

| SHA | `Browser Tests` |
| --- | --- |
| `08766d52` | fail |
| `0b9da166` (your `main` merge) | pass |
| `2f978c49` | fail |

The entire delta between the last two is the 10-line comment removal in a core test file.

For what it's worth, the failure mode looks structural rather than incidental. The dialog's role list is Subscriber (10) / Contributor (20) / Author (30) / Editor (40), the fixture domain starts at Author (30), and the expected Editor (40) is exactly **one** step down — but the test focuses the trigger and sends **two** `{ArrowDown}`s before `{Enter}`. So it only passes when exactly one of them is swallowed (presumably the one that opens the popup) *and* the highlight lands on the currently selected item on open. The value that actually arrived, Subscriber (10), is the first item in the list, i.e. the highlight was at the top of the list when `{Enter}` fired.

Both of those are `kumo`/Base UI timing and focus details rather than anything the test means to assert, which would explain why the same commit passes and fails on different runs. Happy to open a separate PR that picks the option directly (`getByRole("option", …)`) instead of navigating blind, if you'd like it fixed rather than just re-run.

Separately: the `EmDashBot review` check did not appear on `2f978c49`. It also didn't on `08766d52`; it did run on `0b9da166`.

I can push an empty commit to re-trigger, but that reshuffles all 44 checks — a re-run of the failing job (or of the bot) would be cheaper if you have a moment.

MA2153 (commented):
I checked the branch out and verified this independently rather than taking the PR body at its word. Everything holds up:

- Removing the `applyBuildValidator(context)` call reproduces your quoted failure exactly (`expected 1786120009000 to be 1786141609000`), and moving the call to after `next()` fails `leaves a route that opts out of caching opted out`. Both tests fail for the right reason — the ordering test in particular is pinning a real constraint, not just describing one.
- The three claims about Astro's semantics match `core/cache/runtime/cache.js` in 7.0.0: later date wins (L35-39), `set()` clears a prior `set(false)` (L22), and a bare `lastModified` never emits headers (`APPLY_HEADERS` returns early at L73, and `IS_ACTIVE` agrees at L81). That last one is what makes running this on every route, including `/_emdash/*`, safe — worth having stated it explicitly.
- `enabled` is `false` on both `NoopAstroCache` and `DisabledAstroCache` and `true` only on the real one, so the `cache?.enabled` guard is right for dev and for the no-provider case.
- Registration is complete — all four touchpoints an existing virtual module has, and the import stays external in the built `dist/astro/middleware.mjs`.
- Locally: 208 `tests/unit/astro` tests pass, lint is 0 diagnostics, typecheck and build are clean.

The issue write-up did most of the review work up front — tracing it to the specific lines and ruling out `ETag` before filing meant there was nothing to reconstruct here. Checking `blog.cloudflare.com` was the detail that settles it: same header shape on Cloudflare's own Astro site, so this is platform default behavior rather than a misconfiguration, which is what makes it core's problem to fix.

Not binding core to `cloudflare:workers` was the right call, and I'd have pushed back if you'd gone the `CF_VERSION_METADATA` route. Adapter-agnostic is what core should be, and it fixes Node deployments for free.

Two things before this goes in — neither is a blocker.

**1. Please add a line about rollbacks.** A build timestamp only moves forward, and `Last-Modified` expresses *newer*, not *different*. Rolling back from version B to version A means the browser holds B's HTML with validator `tB`, version A answers with `max(content, tA)` where `tA < tB`, so the conditional request is 304 and the browser stays on B's HTML referencing `/_astro/*` names that A never had — the reported failure, in the mirror direction.

This isn't a regression (content-only validators fail the same way) and `CF_VERSION_METADATA` wouldn't have saved it either, since `timestamp` is when the *version* was created and a rollback redeploys an existing version. So I'm not asking you to solve it. But the PR currently reads as though the validator now describes the response, and it describes it in one direction only. A sentence in the module comment naming the limitation would stop the next person from concluding the case is closed. Given that Workers rollbacks are a first-class operation, it may deserve its own follow-up issue.

**2. `keeps a content validator newer than the build` only exercises its own mock.** `createCache` reimplements Astro's accumulation rules, so that assertion tests the mock's `max()` rather than anything in `packages/core`. If Astro switched to last-write-wins, the test would stay green while production broke. I confirmed the mock matches 7.0.0 exactly, and `AstroCache` isn't publicly exported (only `./cache/memory` and `./cache/provider-utils`), so there's no clean way to drive the real class — I don't think a deep import into `dist/` is worth it. I'd just drop that case, or reframe it to assert what we actually control: that `set()` receives the build date and receives it before `next()`. The other four are testing real behavior and should stay.

On the flaky `AllowedDomainsSettings` test — your read looks right, and it's clearly unrelated to a `packages/core`-only change. A separate PR selecting the option via `getByRole("option", …)` instead of arrow-key navigation would be welcome. I'm triggering the bot review now. I can't re-run the browser job, though, so clearing that check will need a push — if you're addressing either point above, the resulting commit will pick it up and there's no need for an empty one.

Deferring the `Cache-Control` question to its own discussion was the right instinct — that's a behavior change for every existing site and shouldn't ride along here.

danielmlr:
Both points are in, on top of a `main` merge.

**Dropped `keeps a content validator newer than the build`** rather than reframing it. Your read is right: it only exercised the stand-in, and removing `applyBuildValidator` entirely leaves it green. Asserting that `set()` received the build date would restate the call rather than test behavior, which is the tautological pattern `AGENTS.md` warns about, and the ordering is already pinned by `leaves a route that opts out of caching opted out` — that one fails for a real reason. Four cases remain, and each of them can fail on a regression.

**Named the rollback limitation** in the `applyBuildValidator` doc comment in `packages/core/src/astro/middleware.ts`, which is where the mechanism and its guarantees are already described:

> Only forward moves are covered. `Last-Modified` expresses newer, not different, so after a rollback the earlier build still answers a conditional request with 304 and the browser stays on the newer build's HTML.

**The red check** should clear with this push: the branch now has `main` merged in, which brings #2398. No empty commit, as you suggested.

Re-verified after the deletion, against the merged `main`:

- Removing the `applyBuildValidator(context)` call fails `raises a content-only validator to the build time` with the same assertion as before (`expected 1786120009000 to be 1786141609000`).
- Moving the call after `next()` fails `leaves a route that opts out of caching opted out`.
- `pnpm lint` reports 0 diagnostics, unchanged from the baseline before these edits; `pnpm typecheck` is clean; the `packages/core` unit suite is 3610 passed with the one pre-existing `virtual-modules.test.ts > watches resolved sandbox plugin entries` failure, which is a macOS tmpdir artifact and also fails on a clean `main`.

The PR body still said five tests in that file; it is four now.

On a follow-up for rollbacks: happy to open one, and it would be worth more than a restatement of your paragraph. It would carry a repro over two real deployments plus a `wrangler rollback`, to confirm the part none of us has actually exercised — that the restored version stops serving the newer version's `/_astro/*` names, so the 304 lands the browser on HTML whose assets it can no longer load. The framing I would give it is that `Last-Modified` is ordered while deployments are not, so no timestamp closes this, `CF_VERSION_METADATA` included; a validator that expresses *different* rather than *newer* means an `ETag`, and Cloudflare strips those from Worker HTML. That puts it in the same place as the `Cache-Control` question this PR left alone, so one thread covering both may serve better than two that do not know about each other. Issue or Discussion — your call, and it does not need to hold this up.

ascorbic (approved):
Makes sense. Thanks for this.

PR #2375, opened 2026-08-08
chore(emdash-bot): port to Flue 2.0.3 with Workers traces (bot next-gen 1/5)

ascorbic:
## What does this PR do?

Slice 1 of the bot next-gen stack (design: maintainer-triggered triage/repro bot replacing the Actions-driven gen-1 pipeline). This slice is a pure framework port with no behavior redesign: `infra/emdash-bot` moves from `@flue` 0.4-nightly to **Flue 2.0.3**, with Workers trace collection enabled.

- **Deps:** `@flue/runtime` + `@flue/vite` → 2.0.3; `agents` → ^0.20.1 (aligned with what `@flue/vite` bundles). Root `pnpm-workspace.yaml`: drops the two now-dead nightly `minimumReleaseAgeExclude` pins and repoints the `@flue/runtime>hono` override to 2.0.3. Lockfile changes are confined to the two private `infra/*` importers plus one catalog drift-correction; no published package's resolved tree changes.
- **Flue 2 API port** (all mechanical): tool `run()` returns `{ output }` envelopes; `dispatch()` now resolves on admission so the classifier does `dispatch()` + `read(receipt)` under a single shared 10s budget (admission stall now rejects instead of hanging past GitHub's webhook ack window); `DispatchReceipt.submissionId` rename; `FlueEvent` drops `dispatchId`; `withSandboxDeadlines` rewritten onto the new `createSandbox` factory method — under Flue 2 the old `createSessionEnv` wrapper would have been silently bypassed, so this closes a latent deadline hole.
- **DO migrations:** none needed — verified against the generated deploy config that Flue 2's `Flue<Pascal>Agent` naming produces class names byte-identical to the existing migration v3. Migrations block untouched.
- **Traces:** `observability.traces.enabled` in wrangler + explicit `instrument(createCloudflareTracing())` at app-entry module scope (content-on; public data). Verified against the built runtime that the generated entry's default installer sees the existing instrumentation and no-ops — no double-install.

Groundwork for the stack: slice 2 (state-machine trigger/fix-loop states), slice 3 (@cloudflare/computer execution layer), slice 4 (pkg.pr.new fix loop), slice 5 (evals + Actions cutover).

Closes #

## Type of change

- [x] Chore (dependencies, CI, tooling)

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md)
[...]

PR #2381, opened 2026-08-08
feat(emdash-bot): hybrid computer/sandbox execution layer + investigation skills (bot next-gen 3/5)

ascorbic:
## What does this PR do?

Slice 3 of the bot next-gen stack (on #2376). Replaces emdash-bot's execution plumbing with a hybrid `execEnv` and lands the investigation skills.

- **Hybrid execution:** `@cloudflare/computer` 0.1.1 (pinned exact) provides the isolate surface — durable SQLite VFS on a new `WorkspaceDO` (append-only migration v4), in-VFS isomorphic-git (clone/status/diff without a container), and `just-bash` isolate exec via the Worker Loader. The existing `@cloudflare/sandbox` container keeps the real-Linux work (`pnpm install`, `astro build`, vitest, agent-browser). computer's own container backend requires `computerd`, which Cloudflare has not published (private in their monorepo; verified against the tarball and repo) — the `exec-env.ts` module is the single flip-point when that changes.
- **Stateless VFS→container coherence:** before every container exec, the changed set is re-derived from the VFS via `git status --porcelain -z` and materialized into the container checkout. No in-memory bookkeeping — correctness survives isolate eviction/resume in both directions (adversarial review found the stateful version lost edits across isolate restarts; the stateless design eliminates the class).
- **Skills:** the investigation skill set (index + repro-public/repro-admin/repro-api/diagnose/verify/fix) lands as Flue 2 skill directories — isolate-first/container-on-demand guidance and evidence-carrying verdicts ("could not reproduce" with transcript is a first-class outcome) baked into each.
- **Hardening:** `readArtifact` accepts bare filenames only (traversal/absolute/symlink rejected); every exec is deadline-bounded across both substrates.

Investigation model stays `@cf/moonshotai/kimi-k2.7-code` (verified id); classifier unchanged.

Closes #

## Type of change

> Maintainer-directed: implements the bot next-gen design; no separate Discussion.

## Checklist

- [x] `pnpm typecheck` passes — n/a-with-note: same pre-existing `infra/emdash-bot` red baseline as #2375/#2376; zero new error classes
- [x] `pnpm lint` passes — oxlint `--type-aware --deny-warnings` exit 0
- [x] `pnpm test` passes — unit 111/111, integration 35/35 (workers pool)
- [x] I have added/updated tests for my changes — execEnv seam suite incl. cross-isolate materialization (both restart modes), rename + non-ASCII porcelain parsing, artifact containment
- [x] User-visible strings in the admin UI are wrapped for translation — n/a
- [x] I have added a changeset — n/a, unpublished infra worker
- [x] New features link to an approved Discussion — see note above

- [x] This PR includes AI-generated code — model/tool: Claude Fable 5 (Claude Code; built and adversarially reviewed by Fable subagents, incl. a delta re-review of the coherence fix)

## Screenshots / test output

```
unit:        111 passed (111)
integration:  35 passed (35)
build:       index.js 9.25 MB raw / 2.06 MB gzip (within Workers limits; headroom noted)
```

Known live-validation items deferred to the slice-5 eval harness (cannot be exercised in CI): isolate exec + VFS git clone against real bindings, container attach round-trip, bundle headroom watch.

---

### Try this PR

**[Open a fresh playground →](https://bot-nextgen-03-computer-exec-emdash-playground.emdash-cms.workers.dev)**

<sub>Tracks `bot-nextgen/03-computer-exec`. Updated automatically when the playground redeploys.</sub>


ascorbic:
Addressed in e771ab53: a failed workspace connection is no longer cached as a rejected promise (the next call retries), `wrangler.test.jsonc` now mirrors the production `WorkspaceDO`/`LOADER` bindings and the `experimental` compatibility flag (with `WorkspaceDO` re-exported from the test entry), and the exec-env/cloudflare header comments are trimmed to invariants. The workers-pool integration suite passes against the updated bindings.

The conditional `useSkill(fixSkill)` is left as-is. Flue 2 documents `useSkill` — along with `useTool`, `usePersistentState`, and the event hooks — as conditional/reorderable between renders; hooks are not positionally tracked. `input.mode` is also immutable initialData for the life of a run, so the condition can never change across re-renders. Registering the skill unconditionally would load fix instructions into non-fix investigations.

~ 🤖 Claude Fable 5

ascorbic:
Round-3 changes: `cloneRepo` now probes the existing clone with `git status` and removes an unusable partial clone before re-cloning (with tests for both the reuse and discard paths), and the fix skill is registered for every mode except `diagnose` — the review was right that `repro` ("reproduce and attempt a fix" per machine.ts) needs it.

On the hook-ordering half of the `useSkill` finding: it remains conditional, per the earlier comment — Flue 2 explicitly lists `useSkill` among the conditional/reorderable hooks, and `input.mode` is immutable initialData, so the condition cannot change within a run. Registering it unconditionally would load fix instructions into diagnose-only investigations.

~ 🤖 Claude Fable 5

PR #1386, opened 2026-06-08
[codex] Skip core SEO routes when site routes exist

auggernaut:
## What does this PR do?

Skips EmDash's default root `robots.txt` and `sitemap.xml` route injection when the host Astro site already defines its own matching root route under `src/pages`.

This avoids duplicate static route warnings today and prevents future Astro versions from turning those duplicate root routes into hard errors. Per-collection EmDash sitemaps (`/sitemap-[collection].xml`) continue to be injected.

Closes: N/A

## Type of change

- [x] Bug fix
- [ ] Feature (requires [maintainer-approved Discussion](https://github.com/emdash-cms/emdash/discussions/categories/ideas))
- [ ] Refactor (no behavior change)
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [x] Tests
- [ ] Chore (dependencies, CI, tooling)

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md)
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (or targeted tests for my change)
- [x] `pnpm format` has been run
- [x] I have added/updated tests for my changes (if applicable)
- [x] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable). Do not include `messages.po` changes except in translation PRs — a workflow extracts catalogs on merge to `main`.
- [x] I have added a [changeset](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#changesets) (if this PR changes a published package)
- [ ] New features link to an approved Discussion: N/A, bug fix

## AI-generated code disclosure

- [x] This PR includes AI-generated code — model/tool: Codex GPT-5

## Screenshots / test output

Validation run locally:

```text
pnpm --silent lint:json | jq '.diagnostics | length'
# 0

pnpm lint
# passes

pnpm typecheck
# passes

pnpm --dir packages/core exec vitest run tests/unit/astro/routes.test.ts
# Test Files  1 passed (1)
# Tests  6 passed (6)

pnpm --dir packages/core typecheck
# passes

git diff --check
# passes
```

Before root `pnpm typecheck`, local package artifacts were rebuilt after refreshing `node_modules`:

```text
pnpm --filter @emdash-cms/blocks build
pnpm --filter emdash build
```

PR #2385, opened 2026-08-08
fix(core): let sites override the collection sitemap route

### LINKED ISSUE #1089: Integration option to disable / replace auto-injected SEO routes (robots.txt, sitemap.xml)
## Summary

Request an integration option to control (or fully disable) the SEO routes (`/robots.txt`, `/sitemap.xml`) that the EmDash Astro integration auto-injects. Today the only way to override them is to patch `node_modules/emdash/src/astro/integration/routes.ts` directly — fragile across minor bumps.

## Why we need this

`ar-pay.news` ships its own `/robots.txt` and `/sitemap.xml` from `src/pages/`. Reasons:

1. **Locale-aware sitemap** — the site is i18n'd (en/ar/tr/es) and we emit hreflang alternates in the sitemap. EmDash's default sitemap is single-locale.
2. **News-specific `<lastmod>`** — we order by `COALESCE(NULLIF(wp_post_date, '')::timestamptz, published_at)` so migrated content keeps its real chronology. EmDash's default uses `updated_at`.
3. **`/sitemap-[collection].xml` per-collection split** — we want per-collection sitemap files (articles, authors, etc) so search engines can fetch incremental diffs.
4. **Robots.txt rules tied to deploy env** — we gate `Disallow: /*?p=*` etc. based on `import.meta.env.PROD`.

Without an opt-out, Astro logs a "route collision" warning on every build (currently a soft warning, but the deprecation message says it'll become a hard error in a future Astro). We've been suppressing it with a postinstall patch that rewrites `node_modules/emdash/src/astro/integration/routes.ts` to skip the two `injectRoute(...)` calls and install our own `/sitemap-[collection].xml` pattern instead.

### Option A: explicit `seoRoutes` integration option

```js
emdash({
  seoRoutes: {
    robotsTxt: false,        // don't inject; project provides its own
    sitemapXml: false,
  },
});
```

Simple boolean opt-out per route. Backward-compatible — current behaviour is `seoRoutes: { robotsTxt: true, sitemapXml: true }`.

### Option B: callback-style override

```js
emdash({
  routes: ({ inject, skip }) => {
    skip('sitemap.xml');
    skip('robots.txt');
    inject({ pattern: '/sitemap-[collection].xml', entrypoint: './my-sitemap.xml.ts' });
  },
});
```

More flexibility — projects can mutate the full route list. Heavier API surface.

### Option C: collision-tolerance mode

```js
emdash({
  routes: { allowCollisions: true },
});
```

Don't inject if a user route with the same pattern already exists. Cheapest to implement; relies on Astro's existing route-precedence to favour the user's `src/pages/` entry. Doesn't solve the noisy build-log warning, but does make patching unnecessary.

## Impact of doing nothing

Each EmDash minor bump that touches `routes.ts` silently breaks our suppression patch — string-match drift. We caught this on the 0.5 → 0.12 bump (issue #1085 incidentally was the bigger pain that round, but this was on the same list). Patch script rot is a real maintenance cost for self-hosted EmDash sites that diverge from the default SEO route surface.

## Happy to PR

If you'd accept Option A or C, I can put up a PR. Option B is a larger design conversation.

Tagging this as a follow-on to #1085 for context — same project, same bump round.

danielmlr:
## What does this PR do?

#1386 added a `hasUserDefinedPublicRoute()` guard so a site's own `robots.txt` and `sitemap.xml` route files suppress the injected defaults — but left `/sitemap-[collection].xml` unguarded. It is injected unconditionally, so a site that ships its own `sitemap-[collection].xml.ts` gets Astro's duplicate-route warning ("A dynamic SSR route cannot be defined more than once"), and which handler answers rests on route-precedence luck. Astro has announced the collision will become a hard error in a future version.

Of the three SEO routes this is also the only one that serves content data (every published entry of every SEO-enabled collection, with a public cache header) — so it is the one a site is most likely to need to replace or close, e.g. a non-public deployment behind auth, or the per-collection customisations described in #1089.

This PR applies the same guard to the third route. `existsSync` treats the square brackets as literal path characters, so `hasUserDefinedPublicRoute()` needs no change. Sites without their own file see no difference; sites with one now get a clean override instead of a warning and undefined precedence.

The existing "skips root SEO routes that are defined by the site" test keeps asserting that a site-defined `sitemap.xml.ts` alone does **not** suppress the collection route; the new test covers the collection-specific override. The new test fails on the unpatched source.

Closes #1089 — #1386 implemented its Option C for two of the three routes; this covers the remaining one, which the issue lists as use case 3. With it, all three injected SEO routes can be replaced by shipping a route file, with no `node_modules` patching. (An explicit config option, the issue's Option A, would still be a reasonable follow-up, but is a design decision beyond this fix.)

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (or targeted tests for my change) — `tests/unit/astro/routes.test.ts`, 9/9
- [ ] User-visible strings in the admin UI are wrapped for translation (n/a — no admin UI change)

- [x] This PR includes AI-generated code — model/tool: Claude Fable 5 (drafted and reviewed in separate sessions)

## Screenshots / test output

New test against the unpatched source (proves it can fail):

```
FAIL tests/unit/astro/routes.test.ts > core media route injection > skips the collection sitemap route when the site defines its own
Tests  1 failed | 8 passed (9)
```

With the fix:

```
Tests  9 passed (9)
```

### DIFF
--- a/.changeset/sitemap-collection-route-override.md
+++ b/.changeset/sitemap-collection-route-override.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Skips the default `/sitemap-[collection].xml` route injection when the host site defines its own, matching the existing behaviour for `robots.txt` and `sitemap.xml`.
--- a/packages/core/src/astro/integration/routes.ts
+++ b/packages/core/src/astro/integration/routes.ts
@@ -842,10 +842,12 @@ export function injectCoreRoutes(
 		});
 	}
 
-	injectRoute({
-		pattern: "/sitemap-[collection].xml",
-		entrypoint: resolveRoute("sitemap-[collection].xml.ts"),
-	});
+	if (!options.srcDir || !hasUserDefinedPublicRoute(options.srcDir, "sitemap-[collection].xml")) {
+		injectRoute({
+			pattern: "/sitemap-[collection].xml",
+			entrypoint: resolveRoute("sitemap-[collection].xml.ts"),
+		});
+	}
 
 	if (!options.srcDir || !hasUserDefinedPublicRoute(options.srcDir, "robots.txt")) {
 		injectRoute({
--- a/packages/core/tests/unit/astro/routes.test.ts
+++ b/packages/core/tests/unit/astro/routes.test.ts
@@ -111,6 +111,21 @@ describe("core media route injection", () => {
 		);
 	});
 
+	it("skips the collection sitemap route when the site defines its own", async () => {
+		await withTempSrcDir(
+			{
+				"pages/sitemap-[collection].xml.ts": "export const GET = () => new Response('');",
+			},
+			(srcDir) => {
+				const routes = collectRoutePatterns(srcDir);
+
+				expect(routes).not.toContain("/sitemap-[collection].xml");
+				expect(routes).toContain("/sitemap.xml");
+				expect(routes).toContain("/robots.txt");
+			},
+		);
+	});
+
 	it("detects index route files for root public route overrides", async () => {
 		await withTempSrcDir(
 			{


MA2153 (approved):

ascorbic (approved):
Thanks

PR #2388, opened 2026-08-08
fix(flue-review): cap tar entry size and honor pax linkpath overrides

ascorbic:
## What does this PR do?

Closes the two remaining findings from the review of #2386 on the flue-review untar parser:

- **Entry size cap.** The parser buffers one entry's content at a time in Durable Object memory and previously trusted the declared `size` field unconditionally, so a malformed or hostile header could trigger an unbounded allocation. Entries larger than 64 MiB (or with a non-numeric/negative size field) are now rejected.
- **pax `linkpath` support.** Only the pax `path` keyword was parsed, so a symlink whose target exceeds the 100-byte ustar `linkname` field would have been silently truncated. The parser now honors pax `linkpath` and the GNU `K` long-linkname record, and applies the same escape validation to overridden targets.

Also fixes a one-line `noUncheckedIndexedAccess` violation in the R2 instrumentation that surfaced when running `tsc --noEmit` against freshly generated worker types.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

Changeset and i18n are n/a: this touches only the private `infra/flue-review` worker, no published package or admin UI. Typecheck was run for `infra/flue-review` (`tsc --noEmit`); tests are the flue-review vitest suite (49 passing, including new cases for the size cap, pax `linkpath` override, escaping `linkpath` targets, and GNU `K` records).

## Screenshots / test output

```
 Test Files  5 passed (5)
      Tests  49 passed (49)
```

---

### Try this PR

**[Open a fresh playground →](https://fix-flue-review-untar-size-cap-linkpath-emdash-playground.emdash-cms.workers.dev)**

<sub>Tracks `fix/flue-review-untar-size-cap-linkpath`. Updated automatically when the playground redeploys.</sub>

### DIFF
--- a/infra/flue-review/.flue/lib/untar.ts
+++ b/infra/flue-review/.flue/lib/untar.ts
@@ -13,6 +13,8 @@ export interface UntarTarget {
 /** Ceiling on a single entry's declared size; content is buffered in DO memory. */
 const MAX_ENTRY_SIZE = 64 * 1024 * 1024;
 
+const OCTAL_FIELD = /^[0-7]*$/;
+
 function stripRoot(name: string): string | undefined {
 	const slash = name.indexOf("/");
 	if (slash === -1) return undefined;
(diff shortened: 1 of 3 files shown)


ascorbic:
Both review rounds addressed in 5347166b. The size field is now validated as strictly octal before parsing — Copilot's point was right that `parseInt("10x", 8)` silently parses the prefix and desyncs the stream — and there's a test iterating malformed fields (`10x`, `size!`, `-0000001`), which also covers the non-numeric/negative claim from the PR description that emdashbot noted was untested. The unreachable `Number.isFinite`/negative branch is gone since the strict check subsumes it.

On the R2 instrumentation `!`: it's the fix that keeps `pnpm typecheck` green for this worker against freshly generated types, as noted in the description — without it the checklist claim would be false. Kept here rather than a separate PR since it's one character in the same package.

~ 🤖 Claude Fable 5

PR #2390, opened 2026-08-09
fix(core): admin UI fails to load on Windows in dev mode

dashimu:
## What does this PR do?

Fixes the admin UI failing to load in dev mode on Windows — it gets stuck on "Loading EmDash..." forever, with console errors like `Could not resolve "babel-plugin-macros"` and later `process is not defined`.

Root cause: `linguiMacroPlugin` (in `packages/core/src/astro/integration/vite-config.ts`) is a dev-only Vite plugin that compiles away `@lingui/core/macro` calls in admin source files before they reach the browser. It matches files via `id.startsWith(adminSourcePath)`, where `adminSourcePath` comes straight out of `path.resolve()`. On Windows, `resolve()` returns backslash-separated paths, but Vite always normalizes the `id`/`importer` it passes to plugin hooks to forward slashes — even on Windows. So the `startsWith` check is silently always `false` there, the plugin's hooks never fire, and raw `@lingui/core/macro` imports ship to the browser. Vite's dependency optimizer then tries to pre-bundle that module as a real dependency, which fails at runtime (it's only meant to be consumed by a macro loader).

Three fixes, all in the same plugin:

1. Normalize `adminSourcePath` to forward slashes before comparing against Vite's ids/importers (`replaceAll("\\", "/")`, not `path.sep`-driven, so it's also correct — and testable — when a Windows-style path is passed on a non-Windows host).
2. Convert the dynamic `import()` of `@babel/core` to a `file://` URL via `pathToFileURL()` — Node's ESM loader on Windows rejects a raw drive-letter path for dynamic `import()`.
3. Resolve `@lingui/babel-plugin-lingui-macro` to an absolute path via admin's own `node_modules` (same `adminRequire` used for `@babel/core`), instead of passing the bare specifier to Babel, which resolves plugin names starting from the transformed file's directory — outside admin's own dependency tree when EmDash is aliased into a demo/template app.

Verified manually end to end on Windows 11: seeded and ran both `demos/simple` and `templates/marketing`, confirmed the admin dashboard, welcome modal, and content list all render and hydrate correctly after the fix (previously stuck on the loading spinner in both).

Closes #

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] New features link to an approved Discussion: N/A — bug fix

## Screenshots / test output

New regression tests in `packages/core/tests/unit/astro/vite-config.test.ts` (`linguiMacroPlugin Windows path-separator handling`) reproduce the mismatch with a synthetic Windows-style path, independent of the host OS running the test — they fail against the pre-fix `linguiMacroPlugin` (in fact fail to even import it, since it wasn't exported) and pass after the fix.

```
✓ linguiMacroPlugin Windows path-separator handling > compiles away @lingui macro calls in admin source files
✓ linguiMacroPlugin Windows path-separator handling > redirects locale catalog imports from admin source to dist/locales
✓ linguiMacroPlugin Windows path-separator handling > does not match files outside admin source
```

Note: this test file has 5 pre-existing, unrelated failures on Windows (`createViteConfig admin aliasing` / `use-sync-external-store shim aliasing` — `TypeError: File URL path must be absolute`, from a synthetic `file:///workspace/emdash-site/` URL that isn't valid on Windows without a drive letter). Left untouched per the "no drive-by fixes" policy; happy to open a separate issue/PR for it if useful.

### DIFF
--- a/.changeset/fix-admin-windows-lingui-macro.md
+++ b/.changeset/fix-admin-windows-lingui-macro.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes the admin UI failing to load in dev mode on Windows (stuck on "Loading EmDash..." with a `babel-plugin-macros` / `process is not defined` console error). The Lingui macro compiler that runs against admin source in local-monorepo dev never matched any files on Windows, so `@lingui/core/macro` imports shipped uncompiled to the browser instead of being transformed away.
--- a/packages/core/src/astro/integration/vite-config.ts
+++ b/packages/core/src/astro/integration/vite-config.ts
@@ -8,7 +8,7 @@
 import { existsSync } from "node:fs";
 import { createRequire } from "node:module";
 import { dirname, isAbsolute, relative, resolve } from "node:path";
-import { fileURLToPath } from "node:url";
+import { fileURLToPath, pathToFileURL } from "node:url";
 
 import type { AstroConfig } from "astro";
 import type { Plugin } from "vite";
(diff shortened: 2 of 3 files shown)


dashimu:

dashimu:
recheck

PR #2402, opened 2026-08-09
fix(admin): localize the editor word-count footer

danielmlr:
## What does this PR do?

The word-count footer under the Portable Text editor ("1 word · 5 characters · 1 min read") stays English in every locale. Its strings are hard-coded and never reach the translation catalogs. The plural handling is also hand-rolled (`words === 1 ? "word" : "words"`), which breaks for locales with plural rules beyond one/other.

The cause: the footer in `packages/admin/src/components/PortableTextEditor.tsx` renders raw string literals instead of going through Lingui. CONTRIBUTING.md (§ Internationalization) requires all user-facing admin strings to be localized.

The fix wraps the three segments in Lingui `plural()` macros so they reach the catalogs. Each metric keeps its own `<span>` and stays a separate message with a single number slot, matching the existing `# item`/`# items` catalog entries. No `messages.po` changes are included; the extraction workflow adds the new entries on merge.

**Test.** `PortableTextEditor.footer.test.tsx` binds a German test catalog for exactly these three messages and asserts the footer renders the translations, including the singular branch. The test fails on `main` and passes with the fix. (The `msg` macro in the test hashes to the same message ids as the macros in the component. `EditorFooter` is exported as `_EditorFooter`, following the file's existing `_`-prefixed test exports.)

No screenshots: every locale falls back to the English source strings until the new entries are extracted and translated, so a before/after screenshot would show no difference. The browser test covers the change.

Drafted and reviewed with Claude Fable 5 (separate sessions).

## Revisions since this PR was opened

- `ddab38c2` — applies the two review suggestions: the footer test resets the i18n locale in `afterAll`, and the comment on the test-only export is gone.
- `5617449c` — removes a duplicated `_EditorFooter` export that broke the typecheck.
- `7006eebb` — two changes after review feedback. The reading-time segment now uses `plural()` like the other two metrics; a flat `t` string kept translators from branching on number. English keeps the same wording in both branches ("# min read"); the German test catalog uses two different wordings ("1 Minute Lesezeit" / "# Min. Lesezeit"). And `EditorFooter` now subscribes to locale changes via the runtime `useLingui` hook from `@lingui/react`. Without the subscription, the footer kept showing the previous locale after a language switch: on `i18n.activate`, `I18nProvider` swaps its context value but passes `children` through by reference, so React re-renders only components that read `LinguiContext` — and a footer built purely from `@lingui/core/macro` calls reads the global `i18n` instance instead. (The macro of the same name from `@lingui/react/macro` cannot be called bare; it has to bind a variable.)

Verified on `7006eebb`: admin suite 116/116 files and 1386/1386 tests, `pnpm lint:json` 0 diagnostics, `pnpm typecheck` exit 0, `pnpm format` clean.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes (exit 0 at `7006eebb`)
- [x] `pnpm lint` passes (`lint:json` 0 diagnostics at `7006eebb`)
- [x] `pnpm test` passes (or targeted tests for my change) (admin suite at `7006eebb`: 116 files, 1386 tests green, including the 2 new ones; new test verified red on `main` before the fix)
- [x] `pnpm format` has been run (no residual changes)
- [x] I have added a [changeset](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#changesets) (patch: `emdash`, `@emdash-cms/admin`)
- [ ] New features link to an approved Discussion: n/a

## Screenshots / test output

Covered by the browser test (see above); no visual change until the new catalog entries are extracted and translated.

### DIFF
--- a/packages/admin/src/components/PortableTextEditor.tsx
+++ b/packages/admin/src/components/PortableTextEditor.tsx
@@ -41,7 +41,7 @@ import {
 import { CSS } from "@dnd-kit/utilities";
 import type { Element } from "@emdash-cms/blocks";
 import type { MessageDescriptor } from "@lingui/core";
-import { msg } from "@lingui/core/macro";
+import { msg, plural } from "@lingui/core/macro";
 import { useLingui } from "@lingui/react/macro";
 import {
 	TextB,
@@ -2486,20 +2486,20 @@ function EditorFooter({ editor }: { editor: Editor }) {
 		},
 	});
 
+	const { t } = useLingui();
 	const readingTime = calculateReadingTime(text);
 
 	return (
 		<div className="border-t px-4 py-2 flex items-center gap-4 text-xs text-kumo-subtle">
-			<span>
-				{words} {words === 1 ? "word" : "words"}
-			</span>
-			<span>
-				{characters} {characters === 1 ? "character" : "characters"}
-			</span>
-			<span>{readingTime} min read</span>
+			<span>{plural(words, { one: "# word", other: "# words" })}</span>
+			<span>{plural(characters, { one: "# character", other: "# characters" })}</span>
+			<span>{t`${readingTime} min read`}</span>
 		</div>
 	);
 }
+
+// Exported for unit testing
+export { EditorFooter as _EditorFooter };
 
 /** Focus mode state for the editor */
 export type FocusMode = "normal" | "spotlight";
--- a/packages/admin/src/components/PortableTextEditor.tsx
+++ b/packages/admin/src/components/PortableTextEditor.tsx
@@ -2498,7 +2498,7 @@ function EditorFooter({ editor }: { editor: Editor }) {
 	);
 }
 
-// Exported for unit testing
+export { EditorFooter as _EditorFooter };
 export { EditorFooter as _EditorFooter };
 
 /** Focus mode state for the editor */
--- a/packages/admin/src/components/PortableTextEditor.tsx
+++ b/packages/admin/src/components/PortableTextEditor.tsx
@@ -42,6 +42,7 @@ import { CSS } from "@dnd-kit/utilities";
 import type { Element } from "@emdash-cms/blocks";
 import type { MessageDescriptor } from "@lingui/core";
 import { msg, plural } from "@lingui/core/macro";
+import { useLingui as useLinguiContext } from "@lingui/react";
 import { useLingui } from "@lingui/react/macro";
 import {
 	TextB,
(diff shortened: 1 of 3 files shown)


packages/admin/src/components/PortableTextEditor.tsx
```diff
@@ -2486,21 +2486,21 @@ function EditorFooter({ editor }: { editor: Editor }) {
...
-			</span>
-			<span>{readingTime} min read</span>
+			<span>{plural(words, { one: "# word", other: "# words" })}</span>
+			<span>{plural(characters, { one: "# character", other: "# characters" })}</span>
+			<span>{t`${readingTime} min read`}</span>
 		</div>
 	);
 }
 
+// Exported for unit testing
```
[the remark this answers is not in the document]
danielmlr: Heads-up on this one: the suggestion anchored to the comment line only, so applying it replaced the comment with a second copy of the export line — `ddab38c2` had `export { EditorFooter as _EditorFooter };` duplicated, which fails typecheck. Deduped in 5617449c; lint, typecheck, and the footer test are green locally. The i18n-reset suggestion applied cleanly.

packages/admin/src/components/PortableTextEditor.tsx
```diff
@@ -2486,21 +2486,20 @@ function EditorFooter({ editor }: { editor: Editor }) {
...
-			<span>
-				{words} {words === 1 ? "word" : "words"}
-			</span>
-			<span>
-				{characters} {characters === 1 ? "character" : "characters"}
-			</span>
-			<span>{readingTime} min read</span>
+			<span>{plural(words, { one: "# word", other: "# words" })}</span>
+			<span>{plural(characters, { one: "# character", other: "# characters" })}</span>
+			<span>{t`${readingTime} min read`}</span>
```
MA2153: Shouldn't reading time be plural like the rest?

MA2153 (commented):
Looks good, just a small note.

packages/admin/src/components/PortableTextEditor.tsx
danielmlr: Good catch, thank you! You're right — English "min" happens to be invariant, which is why this slipped through as a flat string, but that shape locks translators out of number branching entirely. Locales that spell the unit out (German "1 Minute" / "5 Minuten", and anything with more plural categories than one/other) need the ICU plural form.

Switched it to `plural(readingTime, { one: "# min read", other: "# min read" })` in 7006eebb — identical branches in English, same pattern as the `Scheduled` label on the dashboard. The test catalog now gives the reading time distinct singular/plural translations so the branch is actually exercised.

MA2153 (approved):

ascorbic (approved):
Thanks!

PR #2413, opened 2026-08-10
fix(core): isolate Vite cache for integration test servers

### LINKED ISSUE #2247: Integration tests flaky: parallel dev servers race on the shared Vite dep cache (ENOTEMPTY on node_modules/.vite/deps)
## Summary

The **Integration Tests** CI job (`vitest run --config vitest.integration.config.ts`) fails intermittently when a suite's Astro dev server dies during Vite dependency pre-bundling:

```
19:22:08 [vite] Re-optimizing dependencies because vite config has changed
ENOTEMPTY: directory not empty, rmdir '/home/runner/work/emdash/emdash/packages/core/tests/integration/.servers/srv-U2YBeb/node_modules/.vite/deps'

Error: Server at http://localhost:4396/_emdash/api/setup/dev-bypass did not start within 90000ms
```

The server never binds, `waitForServer` spins for the full 90s, and the suite fails. This is an infra/harness flake, **not** a product or test-logic bug — in the failing run 75 tests passed and only `client/comments.test.ts` (all 15 of its tests) went down, on a PR that touches only `packages/admin`.

Seen on PR #2246 — [job log](https://github.com/emdash-cms/emdash/actions/runs/30216638396/job/89831895064). Re-running the same job on the same branch passed, which is the usual flake tell.

This is a sibling of #1604 (same file, same parallelism, different shared resource) that its fix did not cover — see below.

## Root cause

Every per-suite server gets its own copied fixture root, but they all **share one `node_modules`**, and Vite's dependency-optimizer cache lives inside it.

`packages/core/tests/integration/server.ts`:

```ts
const DONOR_NODE_MODULES = resolve(import.meta.dirname, "../../../../demos/simple/node_modules"); // :34

const workDir = mkdtempSync(join(SERVERS_BASE, "srv-"));                    // :156
cpSync(FIXTURE_DIR, workDir, { recursive: true, filter: ... });             // :159
symlinkSync(DONOR_NODE_MODULES, join(workDir, "node_modules"));             // :169
```

Vite's `cacheDir` defaults to `<root>/node_modules/.vite`. Since `<root>/node_modules` is a **symlink to the donor**, every server's cache dir resolves to the same physical directory:

```
demos/simple/node_modules/.vite/deps
```

Vitest runs integration test files in parallel by default (`vitest.integration.config.ts` sets no `pool`/`fileParallelism` options), so several servers boot at once. Each one has a different project root and a different env (`EMDASH_TEST_DB`, `EMDASH_TEST_UPLOADS`), so each independently decides *"Re-optimizing dependencies because vite config has changed"* and tries to wipe and rewrite `deps/`. When server A's `rmdir` lands while server B is populating the same directory, A gets `ENOTEMPTY` and exits.

It's timing-dependent — whoever wins the directory first survives — which is why it only flakes sometimes, and why suite identity varies between runs.

### Why #1604's fix doesn't cover this

#1604 was the same class of bug (parallel suites contending on shared fixture state) and was fixed with option 2: give each suite its own fixture root. That isolated Astro's `<root>/.astro/dev.json` lock, and `client/concurrent-servers.test.ts` guards it.

But the fix deliberately kept `node_modules` shared via symlink (it's large, and pnpm workspace resolution maintains it), so everything Vite keeps *inside* `node_modules` stayed shared. The Astro lock moved out of contention; the Vite dep cache did not.

## Suggested fixes (pick one)

1. **Give each server its own Vite cache dir.** Set `cacheDir` in the fixture's `astro.config.mjs` from an env var the harness passes per server, e.g. `vite: { cacheDir: process.env.EMDASH_TEST_VITE_CACHE || undefined }`, with `createTestServer` setting it to `join(workDir, ".vite-cache")`. Smallest change, preserves parallelism, and matches the existing per-server env pattern (`EMDASH_TEST_DB`, `EMDASH_TEST_UPLOADS`). Note the dir must sit inside `workDir` so it is torn down with the rest of the copy.
2. **Copy `node_modules/.vite` into the per-suite root** instead of inheriting it through the symlink, so the optimizer starts from a warm but private cache.
3. **Run integration suites serially** (`fileParallelism: false`). Removes this whole class of flake permanently but costs wall-clock time, since each suite boots its own dev server in `beforeAll`.

Option 1 is the least invasive and keeps the parallelism #1604's fix was chosen to preserve.

## Acceptance

- Integration Tests passes reliably across repeated runs.
- Per the repo's TDD-for-bugs rule, a reproducing condition: `client/concurrent-servers.test.ts` already boots two servers from one fixture in parallel — extending it to assert the two servers resolve **distinct** `cacheDir` paths (or forcing a re-optimize in both and asserting both come up) would cover this the same way that test covers the Astro lock.

bferanmi806-sketch:
## What does this PR do?

Fixes #2247

Each integration test server already receives its own copied `workDir`, which isolates Astro's project lock. The fixture intentionally shares `node_modules` through a symlink to the workspace dependency tree so server startup remains fast and parallel.

Vite followed that symlink and inherited one physical `node_modules/.vite` dependency-optimizer cache. Parallel servers could therefore rewrite the same cache concurrently, producing `ENOTEMPTY` failures in `node_modules/.vite/deps` and making the integration suite flaky.

This change assigns each server a private `.vite-cache` directory under its own `workDir`, passes it through the test-only `EMDASH_TEST_VITE_CACHE` environment variable, and maps it to Astro/Vite's `cacheDir`. Shared `node_modules` and parallelism are preserved. The existing cleanup removes the cache with the normal per-server `workDir`.

The regression test requires each server's private `.vite-cache` directory to exist and contain actual optimizer/cache file output. This catches fallback to the shared `node_modules/.vite` cache without relying on cwd inequality, while retaining the concurrent server startup and seeded-content checks.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (or targeted tests for my change) ? focused concurrent-server regression and integration/CLI coverage pass
- [ ] `pnpm format` has been run ? the three changed files were formatted and checked directly with Oxfmt and Prettier; the repository-wide format baseline was not rewritten
- [ ] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n). Do not include `messages.po` changes except in translation PRs ? n/a, no user-facing strings
- [ ] I have added a [changeset](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#changesets) (if this PR changes a published package) ? n/a, test/CI-only change under the current contribution policy
- [ ] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/... ? n/a, bug fix

- [x] This PR includes AI-generated code ? model/tool: OpenCode + GPT-5.6 Luna; Codex adversarial review + GPT-5.6 Sol

### RED -> GREEN

- RED: before the fix, the concurrent regression observed both servers resolving to the same physical Vite cache path; the shared-cache repro also triggered the Astro concurrent-server race path.
- GREEN: the focused race loop passed 5/5 before shipping, and the final WSL run against the synchronized validation setup passed 1 test in 56.6 seconds.
- The prior supported cold run took about 56.6 seconds; 30s/60s were not supported by that measurement. The test now declares an explicit 90-second timeout on this slow `it(...)` only; 90s matches the helper startup budget while reducing the previous 120s allowance; global Vitest timeout behavior is unchanged.
- This validates the reported cache race path but does not claim that one successful run proves all CI flakiness is impossible.

### Validation

- `pnpm typecheck`: passed across the workspace packages.
- `pnpm lint`: passed.
- `pnpm lint:quick`: passed with 0 diagnostics across 2,310 files.
- Focused Oxfmt check on the three changed files: passed.
- Focused Prettier check on the three changed files: passed.
- `git diff --check`: passed.
- `pnpm install --frozen-lockfile`: passed; the lockfile remained unchanged.
- `pnpm --filter emdash build`: passed in 64 seconds.
- Focused WSL integration run using `vitest.integration.config.ts` passed 1 test in 56.6 seconds; the current test timeout is 90 seconds (the recorded run used `--testTimeout 120000`).
- Earlier focused integration/CLI validation: 8 files and 90 tests passed; the concurrent race loop passed 5/5.
- The full workspace `pnpm build` was attempted but exceeded the local five-minute tool ceiling while building `packages/core`; it reported no build error before the timeout, and the narrower core build passed.
- Native Windows official integration setup remains blocked by the pre-existing `execFile("pnpm", ...)` `spawn pnpm ENOENT` issue and Windows symlink permissions, so the real-server regression was run under WSL/Linux.
- No changeset or unrelated generated files are included.


bferanmi806-sketch:

bferanmi806-sketch:
recheck

packages/core/tests/integration/client/concurrent-servers.test.ts
```diff
@@ -48,6 +51,12 @@ describe("Concurrent dev servers (#1604)", () => {
 		// Distinct roots -- the fix gives each server its own copied fixture.
 		expect(a.cwd).not.toBe(b.cwd);
 
+		// The Vite dependency optimizer must not follow both node_modules symlinks
+		// back to the same physical cache directory.
+		const viteCacheA = realpathSync(join(a.cwd, ".vite-cache"));
+		const viteCacheB = realpathSync(join(b.cwd, ".vite-cache"));
+		expect(viteCacheA).not.toBe(viteCacheB);
```
MA2153: This seems to be a tautology. Two lines above, the test already asserts a's cwd and b's cwd are different. What actually makes the test fail appears to be `realpathSync` throwing. You can instead try something like:
```diff
-import { realpathSync } from "node:fs";
+import { existsSync, readdirSync } from "node:fs";
@@
-             // The Vite dependency optimizer must not follow both node_modules symlinks
-             // back to the same physical cache directory.
-             const viteCacheA = realpathSync(join(a.cwd, ".vite-cache"));
-             const viteCacheB = realpathSync(join(b.cwd, ".vite-cache"));
-             expect(viteCacheA).not.toBe(viteCacheB);
+             // Vite must write the dependency optimizer's output into each server's own
+             // cache dir, not follow the shared node_modules symlink to one physical copy.
+             for (const cwd of [a.cwd, b.cwd]) {
+                     const cacheDir = join(cwd, ".vite-cache");
+                     expect(existsSync(cacheDir), `${cacheDir} was never created`).toBe(true);
+                     expect(readdirSync(cacheDir)).toContain("deps");
+             }
```

packages/core/tests/integration/client/concurrent-servers.test.ts
```diff
@@ -58,5 +67,5 @@ describe("Concurrent dev servers (#1604)", () => {
 			expect(slugs).toContain("posts");
 			expect(slugs).toContain("pages");
 		}
-	});
+	}, 120_000);
```
MA2153, line 73: This might be too generous. I am not sure what's the appropriate number to set it to, but other tests seem to run comfortably with a 30s timeout.

MA2153 (commented):
Thank you for tackling this! Just a couple notes on the regression test.

packages/core/tests/integration/client/concurrent-servers.test.ts
```diff
@@ -58,5 +70,5 @@ describe("Concurrent dev servers (#1604)", () => {
 			expect(slugs).toContain("posts");
 			expect(slugs).toContain("pages");
 		}
-	});
+	}, 90_000);
```
ascorbic: `createTestServer` itself allows 90 seconds just for the server to bind, then still performs setup and seeding. With this identical 90-second Vitest limit, a server that binds near the helper deadline—or exhausts it—makes Vitest abort before setup/error reporting and cleanup can finish, potentially leaving the child alive. Set this above the helper budget (for example, restore 120 seconds), or pass shorter `timeout` values to these calls so the outer limit retains cleanup headroom.

ascorbic (changes requested):
The outer timeout must exceed the helper's startup-only budget so accepted and failure paths can finish setup, report errors, and clean up.

bferanmi806-sketch:
Restored the outer timeout to 120s while keeping the helper at 90s, so cleanup has headroom. Focused test and lint/format checks pass thanks for catching this.

ascorbic (approved):
Thanks!

PR #2431, opened 2026-08-12
feat(core): add single-taxonomy endpoints for CRUD parity

MA2153:
## What does this PR do?

Taxonomy *terms* had full CRUD, but the *taxonomy definition* itself only supported list + create. This fills in the missing operations so taxonomies match menus, content, and bylines.

**New endpoints**

| Method   | Path                                          | Perm                | Handler                              |
| -------- | --------------------------------------------- | ------------------- | ------------------------------------ |
| `GET`    | `/_emdash/api/taxonomies/{name}`              | `taxonomies:read`   | `handleTaxonomyGet` *(new)*          |
| `PUT`    | `/_emdash/api/taxonomies/{name}`              | `taxonomies:manage` | `handleTaxonomyUpdate` *(new)*       |
| `DELETE` | `/_emdash/api/taxonomies/{name}`              | `taxonomies:manage` | `handleTaxonomyDelete` *(new)*       |
| `GET`    | `/_emdash/api/taxonomies/{name}/translations` | `taxonomies:read`   | `handleTaxonomyDefTranslations` *(already existed, unexposed)* |

Purely additive — no schema changes, no migration, no change to any existing response. Also documents the new paths in the OpenAPI document and adds the four `TAXONOMY_*` error codes to `ErrorCode` (including `TAXONOMY_TRANSLATIONS_ERROR`, which the translations handler already emitted).

### Locale semantics

`GET` and `PUT` address **one locale's definition row** via `?locale=`, falling back to the lowest-locale match when omitted — the same rule `handleMenuGet` uses. `label`, `labelSingular`, `hierarchical`, and `collections` all live on the per-locale row, and the public read path (`getTaxonomyDef`) resolves them per locale, so writing the whole `translation_group` would silently overwrite other locales.

`PUT` is `.strict()`: `name` and `locale` both identify the row being written, so an attempt to change either is a 400 rather than a silently dropped field. Renaming is disallowed because terms are keyed on `name`.

### Delete semantics

`DELETE` removes the taxonomy outright: every locale's definition, every term under that name in every locale, and those terms' `content_taxonomies` assignments. Destructive and unguarded, matching collection delete (and per the Discussion, no "non-empty" guard).

**One deviation from the Discussion**, flagged deliberately: it specced `handleTaxonomyDelete(db, name, { locale? })`, but the delete takes no locale. Terms are keyed on the taxonomy `name`, not on a definition row, so a locale-scoped delete would either leave the taxonomy half-removed (a definition with no terms) or accept a `locale` that doesn't actually narrow the blast radius. The Discussion's own prose — "remove the def, all terms in `taxonomies` for that taxonomy, and their `content_taxonomies` rows" — is unconditional, so the parameter is dropped rather than made misleading on a destructive endpoint. Happy to add per-locale definition removal as a separate operation if that's wanted.

Closes #1346

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes — `packages/core` clean. `packages/cloudflare` reports four pre-existing `sandbox` import errors that reproduce on a clean `main` checkout (stale `packages/core/dist`); untouched by this PR.
- [x] `pnpm lint` passes — `pnpm lint:json` reports 0 diagnostics
- [x] `pnpm test` passes — full `packages/core` suite: 5491 passed, 4 skipped
- [ ] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable) — n/a, no admin UI in this PR
- [x] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/1346

## Screenshots / test output

New suite: `packages/core/tests/unit/taxonomies/taxonomy-crud.test.ts`, 24 tests under `describeEachDialect` (SQLite locally; Postgres in CI), covering get/update/delete, the delete cascade, per-locale update scoping, orphan-collection filtering, route wiring, and permission denial. A separate block runs the delete against a database with `PRAGMA foreign_keys = ON` — production enables it, the in-memory test DBs don't, and the delete has to survive `taxonomies_parent_fk` (`ON DELETE SET NULL`) firing mid-statement when a parent and its children go in one delete.

Both delete-cascade and locale-scoping assertions were mutation-checked (removing the `content_taxonomies` cleanup fails 3 tests; broadening the update from `id` to `name` fails 1), so they're not passing vacuously.

Verified end-to-end against `demos/simple` with a rebuilt `packages/core`, since route registration in `astro/integration/routes.ts` isn't covered by unit tests:

```
GET    /_emdash/api/taxonomies/category                → 200 { taxonomy: {...} }
GET    /_emdash/api/taxonomies/nope                    → 404 NOT_FOUND
GET    /_emdash/api/taxonomies/category/translations   → 200 { translationGroup, translations: [...] }
GET    /_emdash/api/taxonomies/category/terms          → 200 (existing route unaffected)
PUT    /_emdash/api/taxonomies/category (no CSRF hdr)  → 403 CSRF_REJECTED
PUT    /_emdash/api/taxonomies/category                → 200 { taxonomy: { label: "Topics", ... } }
PUT    {"name":"topics"}                               → 400 VALIDATION_ERROR  Unrecognized key: "name"
PUT    {"collections":["ghost"]}                       → 400 VALIDATION_ERROR  Unknown collection(s): ghost
DELETE /_emdash/api/taxonomies/mood                    → 200 { deleted: true }
DELETE /_emdash/api/taxonomies/mood (again)            → 404 NOT_FOUND
```

---

## Review response (07169101)

**Test-file module comment — fixed.** Removed. The `describe` name plus the explicit `sqlite.pragma("foreign_keys = ON")` line carry the meaning without the narrative.

**Unused `handleTermCreate` import — not applicable.** It is used, at `taxonomy-crud.test.ts:306`, to create the Spanish term that the "removes every locale of the definition and of its terms" test asserts is cascaded away. Applying the suggested import block would break the build; `pnpm lint:json` also reports 0 diagnostics, so no unused-import rule is firing. Left as is.

Re-verified after the change: 24/24 tests pass, lint clean, formatted.


ascorbic (approved):
Thanks!

PR #2449, opened 2026-08-13
fix(cloudflare): honour fit and position in the image endpoint

### LINKED ISSUE #2228: Cloudflare image endpoint ignores Astro fit/position params for EmDash media transforms
## Summary

EmDash's Cloudflare `_image` endpoint wrapper (`@emdash-cms/cloudflare/src/image-endpoint.ts`) parses `w`, `h`, `f`, and `q` from Astro's image transform URLs but **ignores `fit` and `position`**. As a result, square/avatar crops served through the EmDash media pipeline don't get the same crop semantics as Astro's stock Cloudflare endpoint, and there's no way to opt into Cloudflare `gravity: "face"` for face-aware cropping.

## Background

PR #1438 added responsive `srcset` for EmDash media via `astro:assets`. PR #1549 wrapped Astro's image endpoint so EmDash media bytes load from storage (Access-safe) instead of over HTTP. PR #2037 added a default quality of 85.

Astro's image service generates `/_image` URLs like:

```
/_image?href=/_emdash/api/media/file/{key}&w=32&h=32&q=60&f=webp&fit=cover&position=center
```

Astro's stock Cloudflare binding endpoint (`@astrojs/cloudflare/dist/utils/image-binding-transform.js`) already forwards `fit` to `ImagesBinding.transform()`:

```js
.transform({
  width: ...,
  height: ...,
  fit: url.searchParams.get("fit")
})
```

But EmDash's Cloudflare wrapper (`@emdash-cms/cloudflare/src/image-endpoint.ts`) only parses `w/h/f/q` via `parseTransformParams` and builds its own `ImageTransform` object without `fit` or `position`:

```ts
const transform: ImageTransform = {};
if (width) transform.width = width;
if (height) transform.height = height;
// no fit, no gravity
```

So when Astro requests a square `32×32` avatar with `fit=cover`, the Cloudflare Images binding receives `width=32, height=32` but no `fit` — it defaults to `scale-down` (or the binding's default), which preserves aspect ratio and doesn't crop to the requested dimensions.

## Impact

- **Avatars and square crops** rendered from EmDash media on Cloudflare don't crop correctly — they get letterboxed/scale-down instead of `cover`-cropped.
- **DPR variance**: DPR 1 screens fetch the `32w` candidate (32×32, no `fit`), DPR 2 screens fetch `64w` (64×64, no `fit`), producing inconsistent crop/quality across devices.
- **No face gravity**: Cloudflare's `ImagesBinding` supports `gravity: "face"` for face-aware cropping, but there's no EmDash config to enable it. The `images?: boolean` integration option has no sub-config.
- **Astro `position=center` is always emitted** for constrained images, so even if `gravity` were configurable, it would need to take precedence over the default `position=center` to take effect.

### 1. Forward `fit` from `_image` URLs

Extend `parseTransformParams` in `emdash/src/media/image-endpoint.ts` to parse `fit` and `position`, and forward `fit` to `ImagesBinding.transform()` in `@emdash-cms/cloudflare/src/image-endpoint.ts`.

Astro's `ImageFit` values: `fill | contain | cover | none | scale-down`. Cloudflare's `ImageTransform.fit` accepts: `scale-down | contain | pad | squeeze | cover | crop`. Most map directly; `fill` → `squeeze`, `none` → omit.

### 2. Map `position` to Cloudflare `gravity`

Astro's `position` is a free string mirroring CSS `object-position`. Map only the subset Cloudflare supports: `center` → `center`, `top` → `top`, `bottom` → `bottom`, `left` → `left`, `right` → `right`. Ignore unsupported values (fall back to binding default).

### 3. Add configurable defaults via `images` option

Extend `images?: boolean` to `images?: boolean | ImageConfig` where:

```ts
interface ImageConfig {
  /** Default gravity for all EmDash media transforms. */
  gravity?: "face" | "center" | "auto" | "entropy" | "left" | "right" | "top" | "bottom";
  /** Default quality override (explicit `?q=` still wins). */
  quality?: number;
}
```

Preserve `images: false` (opt-out) and `images: true` (current default behavior).

When `gravity: "face"` is configured, it should override Astro's default `position=center` (since Astro always emits `position=center` for constrained images, a configured gravity would never take effect otherwise). An explicit non-`center` `position` from the URL should still take precedence over the configured gravity.

### 4. Tests

- Unit test: `parseTransformParams` parses `fit` and `position`
- Unit test: Cloudflare endpoint forwards `fit` to `ImagesBinding.transform()`
- Unit test: configured `gravity: "face"` overrides default `position=center`
- Unit test: explicit `position=top` overrides configured `gravity`

## Related

- #1438 — responsive srcset for media via astro:assets
- #1549 — load transformed images from storage, not via HTTP request
- #2037 — default image transform quality to 85
- #1390 / #1494 — earlier direct media transform approaches (closed)

afonsojramos:
## What does this PR do?

The Cloudflare image endpoint parsed only `w/h/f/q` and built its own transform, so the `fit` Astro asks for never reached the Images binding. A request for a square cover-crop arrived as width and height alone, and the binding fell back to its default fit: the image was scaled down inside the box and letterboxed instead of cropped to fill it. Astro's stock Cloudflare endpoint forwards `fit`, so identical markup cropped correctly on Node and silently did not on Cloudflare.

`fit` and `position` are now parsed alongside the existing params and mapped onto the binding's vocabulary, which only partly overlaps Astro's:

| Astro | Images binding | Why |
| --- | --- | --- |
| `cover`, `contain`, `scale-down` | same | Names line up |
| `fill` | `squeeze` | Both distort to fill the box |
| `inside` | `contain` | Not in `ImageFit`, but Astro's sharp service accepts it |
| `outside` | *dropped* | Nothing in the binding resizes to exceed the box without cropping |
| `position: north/south/east/west` | `top/bottom/right/left` | sharp's compass names for the same edges |
| `position: centre` / `attention` | `center` / `auto` | sharp spellings |
| `position: left top` | *dropped* | Names a corner the binding has no keyword for |

Values with no binding equivalent are dropped rather than swapped for something that would crop, so they keep the behaviour they had before this mapping existed instead of gaining a confidently wrong one. Both params are advisory: an unrecognised value is dropped instead of failing the request, because `ImageFit` is open-ended (`string & {}`) and each backend supports a different subset.

Closes #2228

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] User-visible strings in the admin UI are wrapped for translation — n/a, no UI strings

## Screenshots / test output

The regression test reproduces the bug. Reverting only the adapter mapping and re-running:

```
❯ tests/image-endpoint-fit.test.ts (7 tests | 5 failed)
AssertionError: expected { width: 32, height: 32 } to match object { width: 32, height: 32, fit: 'cover' }
AssertionError: expected undefined to be 'top'
```

With the fix, `pnpm test:unit` is green, including 323 tests in `@emdash-cms/cloudflare`.

## Two things worth a maintainer's eye

**1. `@emdash-cms/cloudflare` was not in `test:unit`.** CI runs only that script, so none of the package's 25 test files have been running — a regression test added here would never have executed. I added the package to the filter; all 323 tests pass. Happy to split that into its own PR if you would rather keep this diff to the endpoint.

**2. An adjacent divergence I deliberately did not change.** Astro only defaults `fit` to `cover` when a `layout` is set (`validateOptions`: `if (options.layout && options.width && options.height) options.fit ??= "cover"`). EmDash's own components pass `layout="constrained"`, so the path this issue describes does emit `fit` and is fixed here. But a bare `<Image src={emdashMedia} width={400} height={400} />` with `image.layout` unset emits no `fit` at all, and then sharp still crops (its own default is `cover`) while the binding scale-downs. Defaulting `fit` to `cover` here when both dimensions are present would align the two, at the cost of cropping requests that never asked to be cropped. That felt like a behaviour change rather than this bug, so I left it — happy to add it if you'd prefer the platforms match.

### DIFF
--- a/.changeset/cf-image-endpoint-fit.md
+++ b/.changeset/cf-image-endpoint-fit.md
@@ -3,4 +3,4 @@
 "emdash": patch
 ---
 
-Fixes cropping for EmDash media on Cloudflare. Images asked for a `cover` crop — square avatars, fixed-ratio thumbnails — came back scaled down and letterboxed inside the requested box instead of filling it, because the endpoint never passed the requested fit to the Images binding. `fit` and `position` are now honoured, so a crop crops and its focal side is respected.
+Fixes cropping for EmDash media on Cloudflare. An image asked to fill a fixed box — a square avatar, a fixed-ratio thumbnail — came back scaled down and letterboxed inside it, because the endpoint never passed the requested fit to the Images binding. `fit` and `position` are now honoured, so a crop crops and its focal side is respected.
--- a/.changeset/cf-image-endpoint-fit.md
+++ b/.changeset/cf-image-endpoint-fit.md
@@ -1,0 +1,6 @@
+---
+"@emdash-cms/cloudflare": patch
+"emdash": patch
+---
+
+Fixes cropping for EmDash media on Cloudflare. Images asked for a `cover` crop — square avatars, fixed-ratio thumbnails — came back scaled down and letterboxed inside the requested box instead of filling it, because the endpoint never passed the requested fit to the Images binding. `fit` and `position` are now honoured, so a crop crops and its focal side is respected.
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
 		"typecheck:templates": "pnpm run --workspace-concurrency=1 --filter {./templates/*} typecheck",
 		"check": "pnpm run typecheck && pnpm run --filter {./packages/*} check",
 		"test": "pnpm run --filter {./packages/*} test",
-		"test:unit": "pnpm run --filter emdash --filter @emdash-cms/auth --filter @emdash-cms/blocks --filter @emdash-cms/gutenberg-to-portable-text --filter @emdash-cms/marketplace --filter @emdash-cms/plugin-cli --filter @emdash-cms/plugin-forms --filter @emdash-cms/plugin-types --filter @emdash-cms/registry-client --filter @emdash-cms/registry-lexicons test",
+		"test:unit": "pnpm run --filter emdash --filter @emdash-cms/auth --filter @emdash-cms/blocks --filter @emdash-cms/cloudflare --filter @emdash-cms/gutenberg-to-portable-text --filter @emdash-cms/marketplace --filter @emdash-cms/plugin-cli --filter @emdash-cms/plugin-forms --filter @emdash-cms/plugin-types --filter @emdash-cms/registry-client --filter @emdash-cms/registry-lexicons test",
 		"test:browser": "pnpm run --filter @emdash-cms/admin test",
 		"test:e2e": "playwright test",
 		"test:e2e:ui": "playwright test --ui",
--- a/packages/cloudflare/src/image-endpoint.ts
+++ b/packages/cloudflare/src/image-endpoint.ts
@@ -21,6 +21,7 @@ import {
 	originalMediaHeaders,
 	parseTransformParams,
 	resolveTransformQuality,
+	type ImageTransformFit,
 	type ImageTransformFormat,
 } from "emdash/media/image-endpoint";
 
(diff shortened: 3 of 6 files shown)


afonsojramos:
Thanks both — pushed in 6c86e35.

**Comment discipline** (@emdashbot, 3 threads): applied all three suggestions. You're right that those blocks argued for the mapping instead of describing it; the rationale belongs in this PR, which is where it now lives. I trimmed the same narrative style out of the `parseTransformParams` docstring and dropped an inline comment in the endpoint for the same reason.

**Changeset grammar** (@Copilot): reworded to "An image asked to fill a fixed box — a square avatar, a fixed-ratio thumbnail — came back scaled down and letterboxed inside it".

**Whitespace-only `position`** (@Copilot): now normalised at parse time — `params.get("position")?.trim()`, empty becomes `undefined`.

**Optional keys in the returned `options`** (@Copilot): not changed, deliberately. The object literal already returned `height` unconditionally before this PR, so the shape is unchanged in kind, and the existing tests pass as written — `toEqual` treats an `undefined`-valued key as equivalent to an absent one, which is why `parse("w=800")` still matches `{ width: 800, format: "webp", quality: undefined }`. Rebuilding the literal incrementally would be a refactor of untouched code, so it seemed better left out of a bug fix. Happy to do it if you'd prefer the shape tightened.

---

One unrelated thing worth flagging: `tests/integration/mcp/taxonomy.test.ts > paginates correctly when multiple terms share the same label` fails in a full `packages/core` run and passes in isolation. It reproduces on a clean `main` with this branch stashed — same assertion, same `1 failed | 5585 passed | 5 skipped` — so it is pre-existing and not from this PR. The dropped entry is a `shared-*` slug, so it looks like pagination over terms sharing a label depends on a tiebreak that isn't stable under load. Happy to open a separate issue if that's useful.

PR #2462, opened 2026-08-13
fix(core): make lingui macro transform match Vite's normalized paths on Windows

### LINKED ISSUE #1862: React admin UI silently hangs on "Loading EmDash..." — createRoot().render() runs but produces zero DOM output, zero errors, zero network activity
### Description

After resolving several earlier build-time errors (missing renderer registration, an AWS SDK/tslib duplication causing a `Cannot destructure property '__extends' of 'e(...).default'` TypeError, and a "Yjs was already imported" duplicate-module warning — the last two fixed by adding `dedupe: ["tslib", "yjs"]` to both `vite.resolve` and `vite.ssr.resolve`), the admin panel now builds and deploys cleanly with no console errors, but the UI never progresses past the "Loading EmDash..." spinner.

Diagnostics performed:
• The <astro-island> element for the admin component is present in server-rendered HTML with client="only" and the correct ssr attribute (confirmed via raw HTML parse, ruling out an HTML-generation bug).
• The hydrator lifecycle runs: this.hydrator(this)(this.Component, ...) is invoked, followed immediately by removeAttribute("ssr") and dispatch of astro:hydrate — this happens regardless of render success, so absence of the ssr attribute post-load is expected, not diagnostic.
• Manually invoking the renderer function directly confirms it reaches createRoot(e); root.render(u) (client-only path) without throwing.
• After render is invoked, element.childElementCount remains 0 permanently, no new entries ever appear in performance.getEntriesByType('resource'), zero XHR/fetch requests fire, and zero console errors/warnings/logs of any kind are produced.
• No source map is available for the admin bundle chunk (e.g. PluginRegistry.*.js), so it isn't possible to trace further into the minified vendored component tree from the consuming app.

This looks like either an error being silently swallowed by an error boundary inside the vendored admin component tree, or a suspended/never-resolving promise with no observable network or timer activity to explain what it's waiting on.

Could you help point to: (1) whether there's a known issue with React 19 / Astro 7 / client="only" islands for the admin UI, (2) any required environment variables/secrets the admin app expects at runtime that might cause a silent early return, or (3) how to obtain/enable source maps for the admin bundle so we can trace the render further? Happy to provide a minimal reproduction repo if useful.

### Steps to reproduce

Step 1: Deploy emdash 0.28.0 on Astro 7 + @astrojs/cloudflare 14 + @astrojs/react 6 + react/react-dom 19, targeting Cloudflare Workers with D1 + R2 bindings.
Step 2: Navigate to /_emdash/admin (redirects to /_emdash/admin/setup).
Step 3: Observe the "Loading EmDash..." spinner, which never resolves.
Step 4: Open devtools: no console errors/warnings, and no network requests to any API endpoint ever fire.

### Environment

• emdash version: 0.28.0
• @emdash-cms/cloudflare version: 0.28.0
• astro version: 7.0.0
• @astrojs/cloudflare version: 14.0.0
• @astrojs/react version: 6.0.0
• react / react-dom version: 19.0.0
• Runtime: Cloudflare Workers (D1 + R2 bindings)
• OS: N/A (deployed serverless, tested from macOS client browser)

### Logs / error output

```shell
No errors are produced in the browser console or in Cloudflare Worker logs. The renderer call completes without throwing but produces zero DOM output. No source map is available for the admin bundle (e.g. PluginRegistry.*.js) to trace further into the vendored component tree.
```

marcusbellamyshaw-cell:
## What does this PR do?

On Windows, the admin UI can hang forever on "Loading EmDash…" in local dev with no visible error — no console error, no failed network request, nothing. This reproduces reliably in this environment via the documented `pnpm dev` contributor workflow (aliasing `@emdash-cms/admin` to source for HMR).

Root cause, found via live reproduction with browser devtools:

The dev-mode `linguiMacroPlugin` in `packages/core/src/astro/integration/vite-config.ts` gates its Babel transform on `id.startsWith(adminSourcePath)`. `adminSourcePath` is built with `path.resolve`, which on Windows returns a backslash-separated path in whatever casing the process happened to launch with. Vite, however, normalizes module `id`s it passes to plugin hooks to forward slashes and to the file's true on-disk casing. Whenever those two don't literally match — which is easy to hit (e.g. any case difference between how a shell was invoked and the real directory casing) — the `startsWith` check silently fails for *every* admin source file, so the Lingui macro compilation step never runs. Raw, uncompiled `@lingui/react/macro` / `@lingui/core/macro` imports then ship straight to the browser. Lingui's macro runtime throws by design when it isn't run through the Babel transform, and Astro's `astro-island` hydration error handler swallows that throw to the console with no user-visible fallback — so the page is left stuck on its static pre-hydration "Loading EmDash…" placeholder forever.

This normalizes `adminSourcePath` to match Vite's own convention (`realpathSync` for casing, forward slashes for separators), so the comparison actually works.

Fixing that exposed a second, previously-unreached bug in the same function: once the transform actually runs, it does `await import(babelCorePath)` with a raw filesystem path resolved via `require.resolve`. Node's ESM loader rejects that on Windows with `ERR_UNSUPPORTED_ESM_URL_SCHEME` (absolute paths need a `file://` URL). Fixed by wrapping the resolved path with `pathToFileURL(...).href`.

I believe this is the underlying cause of #1862 ("React admin UI silently hangs on 'Loading EmDash...' — createRoot().render() runs but produces zero DOM output, zero errors, zero network activity") — that symptom (including "zero errors" for a reporter who likely didn't check the console) matches exactly what I observed before this fix.

Verified by reproducing live: before the fix, the admin UI hangs indefinitely with a `[astro-island] Error hydrating ... Could not resolve "babel-plugin-macros"` console error; after the fix, the dashboard loads and renders normally with real API calls succeeding, no console errors.

Closes #1862

## Type of change

- [x] Bug fix

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] User-visible strings in the admin UI are wrapped for translation (if applicable) — n/a, no user-visible strings touched

## Screenshots / test output

Before fix (hung, console error):
```
[astro-island] Error hydrating /@fs/.../packages/core/dist/astro/routes/PluginRegistry.mjs Error: Could not resolve "babel-plugin-macros" imported by "@lingui/babel-plugin-lingui-macro". Is it installed?
```

After fix: admin dashboard renders normally, `/_emdash/api/manifest`, `/_emdash/api/auth/me`, `/_emdash/api/dashboard` all return 200, zero console errors.


marcusbellamyshaw-cell:
Still active — all checks green (fixes #1862), just waiting on maintainer review.

PR #111, opened 2026-04-02
feat: add repeater field type for structured repeating data

mvanhorn:
## What does this PR do?

Add a `repeater` field type that stores structured repeating data as a JSON array. This fills the biggest architectural gap between EmDash and WordPress+ACF / Payload CMS, where repeater/array fields are among the most-used features.

Closes #110

Prior Discussion: https://github.com/emdash-cms/emdash/discussions/334

### Why this matters

EmDash only supports flat field types. There's no way to model repeating structures like social links, FAQ items, team members, or pricing tiers without using raw JSON (no validation, no UI guidance) or creating separate collections with references (clunky for simple structures).

- Payload CMS (41.5K stars) has array fields with sub-field definitions
- WordPress + ACF has repeater as one of its most popular Pro features
- Strapi has repeatable components

### Changes

**Core type system** (`packages/core`):
- Add `"repeater"` to `FieldType` union, `FIELD_TYPES` array, and `FIELD_TYPE_TO_COLUMN` (maps to `JSON`)
- Add `RepeaterSubField` interface and `REPEATER_SUB_FIELD_TYPES` constant
- Add `subFields`, `minItems`, `maxItems` to `FieldValidation` and Zod schemas
- Add `"repeater"` to `FIELD_TYPE_TO_KIND` map and plugin manifest schema

**Admin UI** (`packages/admin`):
- Add "Repeater" to the field type picker in `FieldEditor.tsx` with sub-field builder UI
- Create `RepeaterField.tsx`: collapsible item cards, drag-and-drop reordering via dnd-kit (same pattern as Widgets.tsx), add/remove items, sub-field inputs for each type
- Wire `case "repeater"` into `ContentEditor.tsx`'s `FieldRenderer` switch

### Storage model

Repeater data stored as JSON array in a `JSON` column (same as `portableText` and `multiSelect`):

```json
[
  { "platform": "Twitter", "url": "https://x.com/example" },
  { "platform": "GitHub", "url": "https://github.com/example" }
]
```

Sub-field definitions live in `validation.subFields`. Sub-field types restricted to simple types (string, text, number, integer, boolean, datetime, select) to prevent nesting.

## Type of change

- [ ] Bug fix
- [x] Feature (requires [approved Discussion](https://github.com/emdash-cms/emdash/discussions/categories/ideas))
- [ ] Refactor (no behavior change)
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests
- [ ] Chore (dependencies, CI, tooling)

## Checklist

[...]

PR #2469, opened 2026-08-14
fix(core): generate row types for repeater fields

helio-cf:
## What does this PR do?

Type generation emitted `unknown` for every `repeater` field, so consuming code could not index or iterate a repeater and had to declare row shapes by hand and cast to them.

`fieldTypeToTypeScript` had no `repeater` case and fell through to `default: return "unknown"`, even though the zod side derives a precise row schema from the same `validation.subFields` metadata. This PR adds the missing case, so a repeater emits an inline row type built from its sub-fields.

Given a `products` collection with a `specs` repeater (`name` required, `value` optional) and a `gallery` repeater (`photo` required, `caption` optional):

**Before**

```ts
export interface Product {
  specs?: unknown;
  gallery?: unknown;
}
```

**After**

```ts
export interface Product {
  specs?: { name: string; value?: string | null }[];
  gallery?: { photo: { id: string; src?: string; alt?: string; /* …media literal… */ }; caption?: string | null }[];
}
```

### Why this was a gap rather than a decision

The neighbouring `case "json": return "unknown"` is deliberate — `json` is untyped by design. `repeater` was only sharing that fallback by accident:

- The emitter's switch dates from `43fcb9a1` and predates the field type.
- #111 added `repeater` without touching `zod-generator.ts` at all, so neither the zod case nor the TS case arrived with the feature.
- #2458 added the zod case (`generateRepeaterRowSchema`) and left the TS side as it was.

Type generation was the last consumer of `validation.subFields` that discarded it — media-usage extraction (`content-fields.ts`, `projection-fingerprint.ts`) and the admin `FieldEditor` all read it.

### Why the casts were the real cost

`as` silences the compiler. When a sub-field is renamed or retyped in the schema, the generated type stays `unknown`, the hand-written cast keeps asserting the old shape, and `astro check` still reports 0 errors. The component then reads a property that no longer exists, gets `undefined`, and renders empty markup instead of failing — so a rename becomes a silently blank section in production rather than a build error.

### Design decisions worth reviewing

- **A sub-field that is not required is typed `T | null`**, because `generateRepeaterRowSchema` applies `.nullish()` to it, so `null` is a legal stored value. A bare `T` would trade an honest `unknown` for an unsound type.
- **`subFields` absent or empty still emits `unknown`.** A repeater with no declared rows is schema-valid (`subFields: z.array(...).min(1).optional()`), and `{}[]` would be falsely permissive.
- **No index signature**, despite rows being `.passthrough()` at runtime. TS object types are not sealed for reads, and an index signature would make misspelled property access legal again — defeating the purpose of the change.
- **`fieldTypeToTypeScript` now takes a structural subset of `Field`** (`{ type; validation? }`) so a sub-field can reuse the mapping without a type assertion. Sub-field types are a subset of field types, so a `select` sub-field enumerates its options and an `image` sub-field emits the same media literal as a top-level image field, by delegation rather than duplication.
- Sub-field slugs match `/^[a-z][a-z0-9_]*$/`, so they are valid bare identifiers and need no quoting in the emitted literal.

### Why `minor` and not `patch`

No runtime behaviour changes, but this narrows `unknown` to a concrete type, which revokes `unknown`'s universal-assertion privilege. `x as SomethingUnrelated` and assignment to a hand-written row type whose optional members are not nullable will now error. Consumers do not opt in either: the writer runs in the `astro:server:setup` hook and rewrites `emdash-env.d.ts` whenever content differs, so breakage lands on the next dev-server start or CI `astro check`.

This follows the 0.23.0 precedent (#1349), where regenerating interface names was a **minor** with "users should regenerate and update any direct interface references". The 0.10.0 precedent that aligned generated `image`/`file` types was a **patch**, but that change was purely additive — new optional properties could not break existing access.

The changeset tells consumers to regenerate types and warns that casts which have drifted from the schema may now report errors.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes — one pre-existing error on `main`, unrelated to this change; see notes below
- [x] `pnpm lint` passes

Not applicable: no admin UI strings, and this is a bug fix rather than a feature.

- [x] This PR includes AI-generated code — model/tool: Claude Opus 5 (`anthropic/claude-opus-5`) via OpenCode

## Screenshots / test output

Eight tests added to `packages/core/tests/unit/schema/zod-generator.test.ts`, written before the fix. Six failed against unfixed source; the two asserting the `unknown` fallback passed, as they should:

```
× builds a row object array from subFields
× types a sub-field that is not required as nullable
× enumerates the options of a select sub-field
× emits the media literal for an image sub-field
× maps every allowed sub-field type
× keeps a required repeater non-optional
✓ falls back to unknown when subFields is absent
✓ falls back to unknown when subFields is empty

Tests  6 failed | 29 passed (35)
```

After the fix:

```
Test Files  1 passed (1)
     Tests  35 passed (35)
```

Full `packages/core` suite on Node 22: **455/459 files, 5677 tests passing**.


helio-cf:
recheck

helio-cf:
`  1) [chromium] › e2e/tests/invite-flow.spec.ts:227:2 › Full invite flow with passkey registration › invited user appears in the users list `

this test failing but my changes dont really touch any of this? 🤔

MA2153:
Our CI pipeline has been a bit too flaky recently 😬

packages/core/src/schema/zod-generator.ts
```diff
@@ -450,6 +453,27 @@ function fieldTypeToTypeScript(field: Field): string {
...
+			const members = subFields.map((subField) => {
+				const type = fieldTypeToTypeScript({
+					type: subField.type,
+					validation: subField.options ? { options: subField.options } : undefined,
+				});
+				// `generateRepeaterRowSchema` applies `.nullish()` to a sub-field that
+				// is not required, so `null` is a legal stored value.
+				return subField.required
+					? `${subField.slug}: ${type}`
+					: `${subField.slug}?: ${type} | null`;
```
MA2153: Sub-field slugs go in as bare identifiers, but the `/^[a-z][a-z0-9_]*$/` guarantee only holds on the admin API route (`repeaterSubFieldSchema`, `api/schemas/schema.ts:59`). Seed files bypass it: `SeedField.validation` is `Record<string, unknown>` (`seed/types.ts:111`), `validateSeed` checks pattern and uniqueness for top-level field slugs only, and `apply.ts:224` hands `validation` straight to `registry.createField`, which stringifies it without looking. Rows written before that route schema existed are unchecked too.

Ran `generateTypeScript` on this branch with such a field:

```
specs?: { first-name: string }[];            // slug "first-name"
specs?: { a: any }[] | { evil: string }[];   // slug 'a: any }[] | { evil'
```

The first is TS1131, so `emdash-env.d.ts` stops parsing and every collection loses its types. On `main` both emit `unknown`. It also doesn't self-heal: the integration rewrites the file on each dev server start whenever content differs (`astro/integration/index.ts:660`), so fixing it by hand doesn't stick, and the committed file is what `astro check` reads in CI.

`JSON.stringify(subField.slug)` as the member name covers any string. Falling back to `unknown` when a slug fails the pattern works too.

packages/core/src/schema/zod-generator.ts
```diff
@@ -450,6 +453,27 @@ function fieldTypeToTypeScript(field: Field): string {
 			}
 			return "string[]";
 
+		case "repeater": {
+			const subFields = field.validation?.subFields;
+			// A repeater carrying no sub-fields describes no rows, so `{}[]` would
+			// be falsely permissive.
+			if (!subFields || subFields.length === 0) return "unknown";
+
+			const members = subFields.map((subField) => {
```
MA2153: Duplicate sub-field slugs give TS2300, and this one is reachable from the admin without a seed file. `FieldEditor` derives each sub-field slug from its label with no uniqueness check (`FieldEditor.tsx:583`), and `subFields: z.array(repeaterSubFieldSchema).min(1)` has no uniqueness refinement. Two sub-fields labelled "Name" both get `name`:

```
specs?: { name: string; name: number }[];   // TS2300 x2
```

`generateRepeaterRowSchema` tolerates this silently, since `shape[subField.slug] = schema` is last wins, so the emitter is the only thing that breaks. Deduping last wins here would match the zod side. Rejecting duplicates in the API schema is the better fix but a wider change.

packages/core/src/schema/zod-generator.ts
```diff
@@ -450,6 +453,27 @@ function fieldTypeToTypeScript(field: Field): string {
 			}
 			return "string[]";
 
+		case "repeater": {
+			const subFields = field.validation?.subFields;
+			// A repeater carrying no sub-fields describes no rows, so `{}[]` would
+			// be falsely permissive.
+			if (!subFields || subFields.length === 0) return "unknown";
```
MA2153: Minor: this assumes an array, but on the seed/registry path `validation` is unvalidated JSON. `subFields: {}` gets past the length check and throws in the `.map` below. That degrades to a logged typegen warning rather than a broken file, so much milder than the other two, but the other consumer of this data guards it with `Array.isArray` (`media/usage/content-fields.ts:117`).

MA2153 (commented):
Approach looks right: the missing case reads as a gap rather than a decision, the nullability matches `generateRepeaterRowSchema`, and delegating sub-fields back through the same mapper is the right call. Checked out the branch, tests pass.

Three things worth fixing first, all in the new `repeater` case, left inline. Two of them turn a previously inert `unknown` into a generated file that doesn't compile.

Smaller notes:

* `"maps every allowed sub-field type"` doesn't actually cover `select` or `image` (both covered separately, fine) and won't fail when a new entry is added to `REPEATER_SUB_FIELD_TYPES`. Driving it from that const would make it a real guard. A `select` sub-field with no `options` is untested.
* The `` `{}[]` would be falsely permissive `` comment is justification for a decision, which the comment rules exclude. The `.nullish()` one is worth keeping, it's a cross-file invariant a reader would otherwise get wrong.
* Pre-existing and out of scope, but the sub-field path inherits it: `select` options are interpolated into string literals unescaped, and unlike slugs they have no character constraint even on the API path.

MA2153 (approved):
Thanks!

helio-cf:
> Thanks!

no prob! and thanks for reviewing, that array assumpption shouldnt have made it through, idk how i missed it, noticed itd also allow strings to pass through

ascorbic (approved):

PR #2526, opened 2026-08-17
fix: skip admin Font component when fonts are disabled

### LINKED ISSUE #2527: fonts: false breaks the entire admin UI — admin.astro renders <Font> unconditionally, causing a silent empty response
### Summary

Setting `fonts: false` in the `emdash()` integration config — documented as *"disable font injection entirely and use system fonts"* — completely breaks the admin UI instead of falling back to system fonts. Every route under `/_emdash/admin/*` (dashboard, login, the first-run setup wizard) returns `HTTP 200` with a **completely empty body**, with no error surfaced anywhere visible to the user or in application logs.

### Root cause

`packages/core/src/astro/routes/admin.astro` unconditionally renders:

```astro
<Font cssVariable="--font-emdash" />
```

When `fonts: false` is passed to `emdash()`, the integration sets `emdashFonts = []` and injects that empty array as Astro's `fonts` config, so no font family is ever registered for `--font-emdash`. Astro's `<Font>` component throws `AstroError(FontFamilyNotFound)` when it can't find a matching registered family — and since this happens partway through streaming the SSR response, the error is thrown *after* the 200 status and headers have already been sent to the client. The stream just ends with zero bytes; there's no way to downgrade to a 500 at that point, and nothing logs the underlying error in a way that's easy to find.

### Repro steps

1. `emdash({ ..., fonts: false })` in `astro.config.mjs`.
2. Build and deploy (or `astro dev`).
3. Visit `/_emdash/admin` (or any `/_emdash/admin/*` route).
4. Response is `200 text/html` with a 0-byte body. No visible error.

### Fix

Opened in #2526 — guards the `<Font>` render on Astro's actual font registry instead of assuming a family is always available.

### Secondary note (optional, separate from the fix above)

Independent of this specific bug: an SSR component throwing mid-stream degrading silently to "200 + empty body" is a rough failure mode in general — worth considering whether the admin shell route (or Astro itself) could catch stream errors and at least log a clear diagnostic. This bug took real effort to trace precisely because nothing in the response or logs pointed at the cause.

### Environment

- Astro 7.2.2, `@astrojs/cloudflare` adapter (Hyperdrive/Postgres), but the bug is adapter-agnostic — it's in `admin.astro` itself.

timmillwood:
## What does this PR do?

Fixes `admin.astro` (the shell that serves `/_emdash/admin/*` - dashboard, login, and the setup wizard) unconditionally rendering `<Font cssVariable="--font-emdash" />` regardless of the `fonts` config option.

When `fonts: false` is passed to `emdash()` - documented as disabling font injection entirely and falling back to system fonts - no font family is ever registered for `--font-emdash`. Astro's `<Font>` component throws `AstroError(FontFamilyNotFound)` when that happens, and because the error surfaces mid-stream (after the response has already started), the entire admin app degrades to a 200 response with a completely empty body. No error is visible to the user or obviously logged anywhere.

Fix: check Astro's own font registry (`componentDataByCssVariable` from the fonts virtual module - the same source `Font.astro` itself checks) before rendering `<Font>`. Note that `Astro.locals.emdash.config.fonts` isn't a viable check here - `fonts` isn't included in the `serializableConfig` object that backs it, so that field is always `undefined`.

Closes #2527

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

## Screenshots / test output

New regression test (`packages/core/tests/repro/admin-fonts-disabled.render.test.ts`) confirmed failing against the pre-fix code with:

```
FontFamilyNotFound: No data was found for the `"--font-emdash"` family passed to the `<Font>` component.
```

and passing after the fix. `pnpm typecheck`, `pnpm lint`, and `pnpm test:repro` all pass.

### DIFF
--- a/packages/core/src/astro/routes/admin.astro
+++ b/packages/core/src/astro/routes/admin.astro
@@ -13,6 +13,10 @@ import adminStylesUrl from "@emdash-cms/admin/styles.css?url";
 // (relative imports resolve to absolute paths which break client hydration)
 import AdminWrapper from "emdash/routes/PluginRegistry";
 import { Font } from "astro:assets";
+// EmDashConfig.fonts isn't part of serializableConfig (see integration/index.ts),
+// so Astro.locals.emdash.config.fonts is never available here - check Astro's own
+// font registry instead of guessing from EmDash's config.
+import { componentDataByCssVariable } from "virtual:astro:assets/fonts/internal";
 
 export const prerender = false;
 
@@ -84,7 +88,13 @@ const faviconType = adminConfig?.favicon ? undefined : siteFaviconType;
 			})();
 		</script>
 		<link rel="stylesheet" href={adminStylesUrl} />
-		<Font cssVariable="--font-emdash" />
+		{/* fonts: false disables font injection and never registers --font-emdash
+		with Astro's Font API, so <Font> throws FontFamilyNotFound mid-stream -
+		after the 200 response has already started, producing an empty body.
+		Skip it and let the admin CSS's system-font fallback stack apply instead. */}
+		{componentDataByCssVariable.has("--font-emdash") && (
+			<Font cssVariable="--font-emdash" />
+		)}
 		{favicon ? (
 			<link rel="icon" href={favicon} type={faviconType} />
 		) : (
--- a/.changeset/fix-admin-fonts-disabled-empty-response.md
+++ b/.changeset/fix-admin-fonts-disabled-empty-response.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes the admin UI (dashboard, login, and the setup wizard) returning a completely empty response when `fonts: false` is set in the `emdash()` integration config. Previously, disabling fonts left the admin shell rendering a `<Font>` reference to a family that was never registered, which threw partway through the response and produced a 200 with no body.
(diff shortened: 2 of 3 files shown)


timmillwood:

packages/core/src/astro/routes/admin.astro
```diff
@@ -84,7 +88,13 @@ const faviconType = adminConfig?.favicon ? undefined : siteFaviconType;
...
 		</script>
 		<link rel="stylesheet" href={adminStylesUrl} />
-		<Font cssVariable="--font-emdash" />
+		{/* fonts: false disables font injection and never registers --font-emdash
+		with Astro's Font API, so <Font> throws FontFamilyNotFound mid-stream -
+		after the 200 response has already started, producing an empty body.
+		Skip it and let the admin CSS's system-font fallback stack apply instead. */}
+		{componentDataByCssVariable.has("--font-emdash") && (
+			<Font cssVariable="--font-emdash" />
+		)}
```
ascorbic, line 97: Use a ternary rather than boolean short circuit

ascorbic (commented):
Good fix, thanks. Can you address the comments and we can get this in.

swissky:
@timmillwood are you still on this? The review points are small, so to get it in before it goes stale I'll open a follow-up PR that keeps your commit (you stay the author) and adds the review fixes on top. Happy to close mine if you'd rather finish it here.

swissky:
Follow-up with your commit plus the review fixes: #3359.

ascorbic:
Closing as superseded by #3359

PR #2609, opened 2026-08-23
feat(core): add dark mode variants to image fields

danielmlr:
## What does this PR do?

Editors can give an image field a second image for dark color schemes, and `<Image>` from `emdash/ui` shows the one that matches the visitor's scheme. This is part 1 of Discussion #1877 (the featured-image slot). Part 2, images inside the block editor, is deferred. See Details.

The slot is opt-in per field through the `darkVariant` widget option. Fields without the option, and values without a variant, render exactly as before.

What the diff does:

- Field value: `ImageValue` gains an optional `darkVariant` in the same shape as the primary media item (`fields/types.ts`, `fields/image.ts`, `schema/zod-generator.ts`, `schema/types.ts`). It stops being a deprecated alias of `MediaValue` and becomes an interface that extends it, so every `MediaValue` is still a valid `ImageValue`. The Zod schema and the generated types accept the variant.
- Save path: `emdash-runtime.ts` normalizes the variant like the primary image (bare id, dimensions, `storageKey`). `normalizeMediaValue` rebuilds values from a whitelist and would drop the key otherwise. Image and file fields share one normalization loop; only image fields take the variant-aware path.
- Render: `EmDashImage.astro` renders a second ``. The primary image gets the class `emdash-image--light`, the variant `emdash-image--dark`. A global style block hides the one that does not match. Both images share the primary image's `alt`, size overrides and loading attributes; an `id` stays on the primary image and the variant gets it with a `--dark` suffix. A `darkVariant` prop accepts an explicit value for sites that store the dark image in another field.
- Admin: `ImageFieldRenderer` shows **Add dark mode variant** below the selected image when the field enables it. A field still holding a legacy string URL is upgraded to object form so it can carry a variant. `FieldEditor` gets a **Dark mode variant** switch for image fields. `ContentEditor` passes the option through.
- Media usage index: `extractor.ts` records the variant as a second occurrence under `<path>.darkVariant`, for image fields and repeater image sub-fields.
- Docs: a new guide `guides/dark-mode` covers the `<html>` class convention, enabling the slot, rendering, loading and overriding the selectors. `reference/field-types` and `guides/media-library` link to it. Changeset for `emdash` and `@emdash-cms/admin`.

Tests: `tests/repro/image-dark-variant.render.test.ts` renders the component through the Astro container and pins the two-image markup, shared alt, per-image placeholder, loading attributes, the `id` suffix and the explicit prop. `e2e/tests/image-dark-variant.spec.ts` checks computed `display` in Chromium for four combinations of `<html>` class and `prefers-color-scheme`, and that the hidden variant is not requested. `tests/integration/content/image-dark-variant-normalize.test.ts` saves content through the runtime on both dialects, including legacy string primaries with and without a variant, and a file field. Admin tests cover the slot, the picker flow, removal, the legacy upgrade and the switch. `usage-extractor`, `zod-generator` and `fields/image` tests cover extraction, schema and typegen.

### Details

Theme convention: a `dark` or `light` class on `<html>` wins, otherwise `prefers-color-scheme` decides. The templates with a theme switcher and the emdashcms.com hero already work this way. A `<picture>` with `<source media>` cannot follow a class, so both images are rendered and CSS decides. The shipped rules only set `display: none` and wrap the `<html>` part in `:where()`, so a site with another convention (for example `data-theme`) overrides them with its own selectors. The guide shows the four rules.

Loading: both images are lazy by default. A lazy image hidden with `display: none` is not fetched, so a visitor downloads one variant. With `priority` both images are eager and both download, because the server cannot know the scheme. I chose this over a lazy variant so that dark-scheme visitors get the same LCP on a priority hero.

Legacy string values: a field can still hold a bare URL string, which has nowhere to put a variant. The admin upgrades such a value to `{ id: "", src: url }` when the slot needs it. Normalizing that as an object would count it as local media, strip `src` and leave an empty value behind, so `normalizeImageValue` normalizes a primary with no id and no provider from its `src` string instead. That reuses the classification a bare string already gets: internal media URLs and local ids resolve through the provider, everything else stays external. The upgrade outlives the variant — an editor can add one and remove it again — so the shape alone decides, not the presence of `darkVariant`. That makes `{ id: "", src }` an implicit "still a string" marker between admin and runtime. The alternative, classifying the string in the admin, downgrades internal media URLs to `provider: "external"` and loses the provider linkage, so the classification stays in the runtime.

Scope: the slot is opt-in, and the "require both variants" request from @dimitropoulos fits later as a second option value or a validation rule. Repeater image sub-fields keep the variant on save but get no UI, because sub-fields have no widget options. Part 2 (block editor) waits, because several open PRs touch the portable-text image block. It can reuse the same `darkVariant` shape and CSS. The media usage extraction schema version stays at 1: no stored content can carry a variant before this release, so there is nothing to re-extract.

I verified the admin in English and Arabic on `demos/simple`: switch, slot, picker title, card and buttons render correctly in RTL.

### Revisions

- `c02c81c6` — the e2e page existed only in the Node fixture, so the Cloudflare shard had no page to load and every assertion in the spec failed. The page is now in `e2e/fixture-cloudflare` as well, like `text-alignment.astro`. The spec passes on both targets locally.
- `c5d2a249` — the media usage extractor records the `darkVariant` of image fields and repeater image sub-fields (review finding on `extractor.ts:39`). The new test fails without the change. The 30 media-usage integration suites under `tests/integration/database/` (340 tests) pass unchanged.
- `e115a9c9` — `id` was spread into both elements, so `<Image id="hero">` emitted two elements with `id="hero"`. The dark element now carries `hero--dark`; an image without a variant keeps its `id` unchanged. Two render tests pin both cases.
- `53bc3411` — the dark mode slot was gated on an object value, so a field holding a legacy string URL offered no variant at all, and this reverses that gate. The mechanism and the alternative I rejected are in Details above. Two integration tests cover an external URL string and a local id string, both with a variant; the admin test that asserted the old gate now asserts the upgrade.
- `2fdb4df1` — the normalization loop covers image and file fields, and the whole loop had moved to the image normalizer, so a file value carrying a `darkVariant` key survived the whitelist rebuild. Only image fields take that path now. The new integration test fails without the change.
- `e5a12a0b` — the string fallback above was reachable only while a `darkVariant` was present, so an editor who added a variant to a legacy string and removed it again saved `{ id: "", src: url }` as local media and lost the image. The shape decides now. Two integration tests cover the removal and a local id stub, both red without the change; an admin test pins the value the renderer emits after removal.
- Rebase onto `main` (`1717d31b`, 61 commits). Two files conflicted. `ImageFieldRenderer.tsx` was additive: the focal-point `objectPosition` from `main` and the dark mode slot both stand. `EmDashImage.astro` needed a decision: `main` folds the focal point into one top-level `generatedStyle` for a single `imgProps` object, while this branch resolves each image separately through `resolveImage`. The focal point moved inside `resolveImage`, so the variant gets its own `object-position` instead of inheriting the primary's, and the returned `placeholderStyle` became `generatedStyle` because it now carries both. `tests/repro/focal-point-render.render.test.ts` and `tests/repro/image-dark-variant.render.test.ts` pass together.
- `233a41ab` — the rebase put the dark mode slot under a rule it was never written against: every `object-cover` preview in `ImageFieldRenderer` honours its media value's focal point. The dark preview now does too, with distinct coordinates on the two values so the test fails if both previews read the primary.
- The e2e suite ran locally on both targets at `233a41ab`: Node 281 passed / 31 skipped, Cloudflare 280 passed / 32 skipped. The one extra skip is the `cloudflare`-gated case in `i18n.spec.ts`. The earlier `DEV_BYPASS_ERROR` came from `global-setup` polling `setup/dev-bypass`, which mints a new token per call; #2762 fixed it and the rebase picked it up.

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/1877

- [x] This PR includes AI-generated code — model/tool: Claude Fable 5 (review pass: Claude Opus 5)

## Screenshots / test output

```
packages/core  pnpm vitest run                          507 files passed, 6177 tests passed
packages/core  vitest --config vitest.repro.config.ts   8 files, 45 tests passed
packages/admin pnpm vitest run                          139 files, 1766 tests passed
pnpm test:e2e (Node)                                    281 passed, 31 skipped
pnpm test:e2e (Cloudflare)                              280 passed, 32 skipped
pnpm test:e2e image-dark-variant                        7 passed on both targets
packages/core  tests/unit/media/usage-extractor            9 passed
pnpm typecheck / pnpm lint / pnpm format / docs build   clean
```


packages/core/src/components/EmDashImage.astro
```diff
@@ -170,56 +189,102 @@ if (img) {
...
+		sizes: r.sizes,
+		width: r.width,
+		height: r.height,
+		alt: finalAlt,
+		loading: priority ? "eager" : "lazy",
+		fetchpriority: priority ? "high" : undefined,
+		decoding: "async",
+		style: r.placeholderStyle || undefined,
+		class: ["emdash-image-media", variantClass, className].filter(Boolean).join(" "),
+		...attrs,
```
ascorbic, line 257: If there's an `id` set on the image, this will mean there are duplicate elements with the same ID. I'd suggest generating an ID ending in `--dark` or soemthing like that for the dark variant

packages/admin/src/components/ImageFieldRenderer.tsx
```diff
@@ -104,13 +128,26 @@ export function ImageFieldRenderer({
 			blurhash: item.blurhash ?? metaString(item.meta, "blurhash"),
 			dominantColor: item.dominantColor ?? metaString(item.meta, "dominantColor"),
 			meta: isLocalProvider ? { ...item.meta, storageKey: item.storageKey } : item.meta,
-		});
+		};
+
+		if (pickerTarget === "darkVariant") {
+			if (objectValue) onChange({ ...objectValue, darkVariant: selected });
```
ascorbic, line 145: This will silently discard legacy string images. Instead I'd upgrade the legacy string image into an object, and then apply the dark mode to that

packages/core/src/emdash-runtime.ts
```diff
@@ -4056,7 +4057,7 @@ export class EmDashRuntime {
 			if (value == null) continue;
 
 			try {
-				const normalized = await normalizeMediaValue(value, getProvider);
+				const normalized = await normalizeImageValue(value, getProvider);
```
ascorbic: This applies the dark mode handling to files as well as images

ascorbic (changes requested):
This is a good addition, but there are a few things that need fixing

danielmlr:
All three are pushed, but the legacy string one needs your call on the shape.

Upgrading to an object is not enough on its own: `{ id: "", src: url }` counts as local media on save, so `normalizeMediaValue` strips `src`, finds no id and stores `{ provider: "local" }` — the silent discard becomes a deleted image. The runtime now classifies a primary with no id and no provider from its `src` instead, on every path: an editor can add a variant and remove it again, and that stub takes the same route. `{ id: "", src }` is therefore a standing "still a string" marker between admin and runtime, and that is the part I would like your opinion on. Classifying in the admin loses the provider linkage on internal URLs, but I will move it there if you prefer.

The other two are as you described: the dark element carries `<id>--dark`, and only image fields take the variant-aware normalizer. Each has a test that fails without the change.

packages/core/src/components/EmDashImage.astro
danielmlr: Fixed in `ac50ec8e` — the dark element carries `<id>--dark`.

packages/admin/src/components/ImageFieldRenderer.tsx
danielmlr: Pushed in `7421ee8d` and `133b08a4`. The shape needs your call; details in the comment on the PR.

packages/core/src/emdash-runtime.ts
danielmlr: Fixed in `d20672de` — only image fields take the variant-aware normalizer.

packages/admin/src/components/ImageFieldRenderer.tsx
danielmlr: One addition here that your review predates: `main` gained focal points while this sat, so the dark mode preview now honours its own like the primary preview beside it, with distinct coordinates in the test so it fails if both read the primary. The rebase and the `EmDashImage` side of it are under Revisions in the description.

That leaves the shape question from [my comment above](https://github.com/emdash-cms/emdash/pull/2609#issuecomment-5416165381) as the only thing waiting on you: whether `{ id: "", src }` may stand as the "still a string" marker between admin and runtime, or whether the classification moves into the admin and gives up the provider linkage on internal URLs. Either is fine by me, and it is cheaper to settle before 1.0 than after.

ascorbic (approved):
Thanks!

PR #2612, opened 2026-08-23
fix(core): restore content bylines table left staged by an interrupted rebuild

### LINKED ISSUE #2241: Migration 040 can leave `_emdash_content_bylines_new` behind on D1
On a fresh Cloudflare deploy (emdash 0.31.1) every content query came back empty
while the data was intact. `_emdash_content_bylines` did not exist —
`_emdash_content_bylines_new` did.

```
sqlite_master  : _emdash_content_bylines_new  (0 rows), no _emdash_content_bylines
_emdash_collections : posts (source=seed)
_emdash_fields      : 5
ec_posts            : 1 row, status=published, deleted_at NULL
_emdash_migrations  : recorded up to 053
```

`040_byline_i18n.ts` drops the old table (L89) then renames the staged one (L90).
Here the drop landed and the rename didn't, but the migration is still recorded as
applied, so it never self-heals.

Every `loadEntry` / `loadCollection` folds a bylines subquery over
`_emdash_content_bylines`, so the statement throws and the error is swallowed —
the site reports "no entries" rather than failing. Symptoms were:

- public pages rendered the empty state
- `/rss.xml` returned 0 items
- admin showed `Collection 'posts' not found`
- **`/sitemap-posts.xml` kept listing the post correctly**, because it doesn't
  hydrate bylines. That mismatch was what isolated it to the query layer rather
  than the data.

Running the exact `loadEntry` SQL by hand against the same D1 returned the row,
which ruled out the data and the bindings.

Fix that resolved it, no redeploy needed:

```sql
ALTER TABLE _emdash_content_bylines_new RENAME TO _emdash_content_bylines;
```

Possibly the same no-op `acquireMigrationLock` on D1 described in #857, but the
failure mode differs from #762: collections *were* seeded here, so the usual
"empty `_emdash_collections`" signature doesn't apply.

Environment: emdash 0.31.1, @emdash-cms/cloudflare 0.31.1, astro 7.1.3,
wrangler 4.114.0, Cloudflare Workers + D1 (free plan), single region APAC.

danielmlr:
## What does this PR do?

A SQLite or D1 site can report no entries while every row is intact. Public pages render the empty state, `/rss.xml` lists nothing, and the admin reports the collection as not found. In that state the byline credits table exists only as `_emdash_content_bylines_new`, left behind by an interrupted run of migration 040, and `_emdash_content_bylines` is missing. Every loader query (`loadEntry`, `loadCollection`) then fails on its bylines subquery. Running the migrations again does not repair it.

The cause: 040 rebuilds the table on SQLite in four steps. It creates a staged copy, copies the rows, drops the old table, and renames the copy. If the run stops between the drop and the rename, the retry reads `PRAGMA foreign_key_list` on a table that no longer exists. SQLite returns no rows for a missing table, so the retry treats the table as already rebuilt and skips the rename. The runner records 040 as applied and the later migrations run. The staged copy stays behind.

Migration 071 renames `_emdash_content_bylines_new` back to `_emdash_content_bylines` when the live table is missing and the staged copy exists. It then re-creates the two indexes that the drop removed. On a healthy database it changes nothing. On Postgres it changes nothing either, because the 040 path there alters the table in place. It can restart after any of its statements. 040 itself stays unchanged because published migrations are immutable. Covered by integration tests in both dialects.

Closes #2241

### Details

Numbering: #2584 also adds a migration numbered 071. Whichever of the two merges second needs a renumber. Nothing else in this change depends on the number.

Stale copy: if both tables exist, 071 leaves `_emdash_content_bylines_new` alone. The live table is authoritative in that state, and the migration deletes nothing.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable). Do not include `messages.po` changes except in translation PRs — a workflow extracts catalogs on merge to `main`. (n/a: no admin strings)

## Screenshots / test output

```
packages/core/tests/integration/database/restore-content-bylines-table-migration.test.ts
  — healthy table and credits untouched (both dialects; Postgres only in CI); staged copy renamed back and indexes restored;
    second run after the rename completes; runner retry from 040 recovers the table; stale copy next to a live table left alone
  with up() stubbed to a no-op: 3 of 5 fail (rename, re-run, runner retry)

packages/core  vitest run tests/integration/database tests/database tests/unit/database tests/unit/astro
               105 files, 1080 tests passed
packages/cloudflare  vitest run tests/db   19 files, 249 tests passed
pnpm typecheck / lint / format             clean
```

### DIFF
--- a/.changeset/restore-content-bylines-table.md
+++ b/.changeset/restore-content-bylines-table.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes SQLite and D1 sites where an interrupted upgrade left the byline credits table staged as `_emdash_content_bylines_new`, so pages, feeds and the admin reported no entries although the content was intact. The next migration run restores the table and the stored credits.
--- a/packages/core/src/database/migrations/071_restore_content_bylines_table.ts
+++ b/packages/core/src/database/migrations/071_restore_content_bylines_table.ts
@@ -1,0 +1,43 @@
+import type { Kysely } from "kysely";
+import { sql } from "kysely";
+
+import { isSqlite, tableExists } from "../dialect-helpers.js";
+
+/**
+ * Restore `_emdash_content_bylines` where migration 040's SQLite rebuild
+ * stopped between dropping the old table and renaming the staged copy.
+ *
+ * 040 guards its rebuild on `PRAGMA foreign_key_list`, which returns no rows
+ * for a missing table, so a retry after that partial run skips the rename and
+ * leaves `_emdash_content_bylines_new` behind with the credits inside. The
+ * indexes went with the dropped table.
+ *
+ * The Postgres path of 040 alters the table in place, so this is a no-op there.
+ */
+export async function up(db: Kysely<unknown>): Promise<void> {
+	if (!isSqlite(db)) return;
+
+	if (!(await tableExists(db, "_emdash_content_bylines"))) {
+		if (!(await tableExists(db, "_emdash_content_bylines_new"))) return;
+		await sql`ALTER TABLE _emdash_content_bylines_new RENAME TO _emdash_content_bylines`.execute(
+			db,
+		);
+	}
+
+	await db.schema
+		.createIndex("idx_content_bylines_content")
+		.ifNotExists()
+		.on("_emdash_content_bylines")
+		.columns(["collection_slug", "content_id", "sort_order"])
+		.execute();
+	await db.schema
+		.createIndex("idx_content_bylines_byline")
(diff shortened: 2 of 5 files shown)


ascorbic (approved):
Thanks

Issue #2417, opened 2026-08-11
Field constraints are invisible until the server rejects the save, and autosave then retries the doomed request indefinitely


virafb:
### Description

A collection field can declare `validation` constraints (`maxLength`, `min`, `minItems`, `pattern`). These are enforced server-side by the generated Zod schema, but nothing in the admin surfaces them to the person typing. The editor's first and only signal is a failed save.

Since #1838 that failure message is at least specific — the toast reads `excerpt: Too big: expected string to have <=160 characters` rather than the old generic "Invalid request data", which is a real improvement. But it arrives after the fact, names the field by slug rather than by its label, uses validator-internal phrasing, and — the substantive part of this report — **autosave keeps retrying a request that can never succeed**.

In one editing session on 0.32.0 we recorded **443 failed `PUT`s to the same entry over 23.4 minutes**, a median of 3.0 seconds apart (~19/min), every one a 400 with the identical body. Each produced its own toast. The author had added a few words to a summary field capped at 160 characters; nothing in the UI had indicated a cap existed.

### Steps to reproduce

1. Declare a text field with `validation: { maxLength: 160 }` in a collection.
2. Open an entry in the admin and type past 160 characters in that field.
3. Observe: no counter, no `maxlength`, no help text — nothing marks the limit while typing.
4. Wait. Autosave fires, fails 400, and toasts `excerpt: Too big: expected string to have <=160 characters`.
5. Leave the tab open. Autosave retries every ~3 seconds indefinitely, toasting each time, with no back-off and no terminal state.

### Observed

- The constraint is unknowable from the UI until it is violated.
- The message identifies the field by **slug** (`excerpt`), not by the label the editor actually sees ("Summary").
- The message is the raw validator string ("Too big: expected string to have <=160 characters") rather than a sentence aimed at an author.
- Autosave treats a permanent 400 as retryable. 443 attempts, 23.4 minutes, no escalation, no stop, no back-off.

### Expected

- Constraints declared in the schema are visible while editing — at minimum a character counter or `maxlength` on length-limited fields, and help text derived from `validation` rather than hand-written into the label.
- Validation errors name the field by its label.
- Autosave treats a 4xx other than 429 as terminal: stop retrying, keep the buffer dirty, surface the problem once, and retry only when the content changes again.

### Why this matters beyond one field

[...]

PR #2586, opened 2026-08-21
feat(media): add flat folder UI

khoinguyenpham04:
## What does this PR do?

Adds the focused admin UI for the flat Media Library folder foundation in #2584. Editors can create, rename, browse, and delete folders; permitted users can move local media through the Media Details **Location** control or by dragging a grid card or list row onto a visible folder; deleting a folder returns its media to the Main library without changing IDs, URLs, storage keys, or usage records.

The interface follows the stable Strapi Media Library baseline while using native Kumo components: compact folder rows, explicit link and edit actions, Back and breadcrumb navigation, one mixed folder/media table in list mode, bounded loading/retry/load-more states, responsive mobile actions, and direction-aware folder names. Folder navigation preserves relevant search, filter, provider, view, page-size, focus, and scroll state.

This is PR2 of the folders sequence and is stacked directly on #2584. It intentionally excludes nesting, counts, bulk selection or moves, upload-to-folder behavior, Media Picker changes, provider behavior, CLI commands, MCP commands, and plugin changes. Local uploads continue to enter the Main library.

Stack:

1. #2582 — numbered Media Library pagination
2. #2584 — flat folder API foundation
3. This PR — flat folder admin UI

Related media roadmap Discussions:

- https://github.com/emdash-cms/emdash/discussions/990
- https://github.com/emdash-cms/emdash/discussions/1655

A folder-specific maintainer-approved Discussion was not found. Code review can proceed, but this feature must not merge until the folder scope is approved.

## Type of change

- [ ] Bug fix
- [x] Feature (requires [maintainer-approved Discussion](https://github.com/emdash-cms/emdash/discussions/categories/ideas))
- [ ] Refactor (no behavior change)
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests
- [ ] Chore (dependencies, CI, tooling)

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md)
- [x] Typecheck passes for the changed admin package
- [x] Type-aware lint passes with 0 diagnostics
- [x] Targeted component, router, API, database, and browser tests pass
- [x] All changed TypeScript and documentation files pass their formatters
- [x] I have added/updated tests for my changes
- [x] User-visible admin strings are wrapped for translation, RTL-safe, and no `messages.po` files are included
- [x] I have added and reviewed `emdash` and `@emdash-cms/admin` minor changesets
[...]

PR #2614, opened 2026-08-23
fix(admin): stop autosave from retrying a payload the server rejected

danielmlr:
## What does this PR do?

Stops autosave from resending a payload the server has already rejected. A 400 currently makes the editor resend the same body every two seconds until the tab closes, each attempt adding another failed-save toast; #2417 records 443 identical PUTs for one entry over 23 minutes.

The editor schedules an autosave whenever the form is dirty and nothing is in flight, and a failed one leaves it dirty and clears the flag, so the next timer starts at once.

`isTerminalRequestError` reads the status off the `ApiResponseError` that `throwResponseError` already throws. A 4xx is terminal unless it is in `RETRYABLE_CLIENT_ERROR_STATUSES` — 408, 421, 425, 429 — which judge the connection or the rate limit, not the payload. A terminal error advances an entry-scoped `autosaveRejectionToken`, and no autosave is scheduled while the form still matches the rejected state. Manual Save keeps working and the next content change autosaves again; 5xx and network failures still retry. 409 and 401 stay terminal: a conflict repeats with the same body, and the admin client has no session refresh.

Part of #2417

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (n/a: no new strings)

- [x] This PR includes AI-generated code — model/tool: Claude Fable 5, Claude Opus 5

## Screenshots / test output

```
packages/admin/tests/components/ContentEditor.test.tsx — after a rejection token, no further autosave within 10 s; Save stays enabled; the next content change autosaves again
packages/admin/tests/router.test.tsx — a 400 on the autosave PUT advances the rejection token for that entry; a 500 does not
packages/admin/tests/lib/api-client.test.ts — 400, 401, 404, 409 and 413 are terminal, 408, 421, 425, 429, 500, 504 and a network TypeError are not
e2e/tests/autosave.spec.ts — real server: one 400 PUT, no further PUT in the following 7 s, the corrected value saves with 200

pnpm --filter @emdash-cms/admin test    139 files, 1761 passed
pnpm exec playwright test               275 passed, 31 skipped (node)
red/green: with packages/admin/src at main, the editor test sees 2 autosave calls instead of 1, the router test a token of 0, api-client.test.ts no longer loads, and the e2e spec records 4 PUTs with 400 inside the 7 s window instead of 1
pnpm typecheck / oxlint --type-aware --deny-warnings / oxfmt --check / prettier --check    clean, 0 diagnostics
```

### DIFF
--- a/packages/admin/src/lib/api/client.ts
+++ b/packages/admin/src/lib/api/client.ts
@@ -58,6 +58,16 @@ function formatValidationIssues(error: Record<string, unknown>): string | undefi
 	return messages.length > 0 ? messages.join("; ") : undefined;
 }
 
+/** Whether retrying the same request unchanged can never succeed. */
+export function isTerminalRequestError(error: unknown): boolean {
+	return (
+		error instanceof ApiResponseError &&
+		error.status >= 400 &&
+		error.status < 500 &&
+		error.status !== 429
+	);
+}
+
 /**
  * Throw an error with the message from the API response body if available,
  * falling back to a generic message. All API error responses use the shape
--- a/packages/admin/src/lib/api/client.ts
+++ b/packages/admin/src/lib/api/client.ts
@@ -58,13 +58,23 @@ function formatValidationIssues(error: Record<string, unknown>): string | undefi
 	return messages.length > 0 ? messages.join("; ") : undefined;
 }
 
+/**
+ * Client errors that pass no verdict on the request body, so resending it
+ * unchanged can still succeed. Every other 4xx repeats its verdict on every
+ * attempt.
+ */
+const RETRYABLE_CLIENT_ERROR_STATUSES: ReadonlySet<number> = new Set([
+	408, // Request Timeout: the server gave up waiting for the request
+	421, // Misdirected Request: another connection can be routed correctly
+	425, // Too Early: sent as TLS early data, replayable after the handshake
+	429, // Too Many Requests: succeeds once the rate limit window has passed
+]);
+
 /** Whether retrying the same request unchanged can never succeed. */
 export function isTerminalRequestError(error: unknown): boolean {
+	if (!(error instanceof ApiResponseError)) return false;
 	return (
-		error instanceof ApiResponseError &&
-		error.status >= 400 &&
-		error.status < 500 &&
-		error.status !== 429
+		error.status >= 400 && error.status < 500 && !RETRYABLE_CLIENT_ERROR_STATUSES.has(error.status)
 	);
 }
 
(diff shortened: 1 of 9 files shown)


packages/admin/src/lib/api/client.ts
```diff
@@ -46,6 +46,24 @@ function formatValidationIssues(error: Record<string, unknown>): string | undefi
...
+		super(message);
+		this.name = "ApiError";
+		this.status = status;
+	}
+}
+
+/** Whether retrying the same request unchanged can never succeed. */
+export function isTerminalRequestError(error: unknown): boolean {
+	return (
+		error instanceof ApiError && error.status >= 400 && error.status < 500 && error.status !== 429
```
ascorbic: 429 isn't the only retryable response. e.g. 408 and 425 should be retried. Can you create a list of retryable responses and exclude them.

ascorbic (changes requested):

packages/admin/src/lib/api/client.ts
danielmlr: Done. `RETRYABLE_CLIENT_ERROR_STATUSES` in `packages/admin/src/lib/api/client.ts` now holds 408, 421, 425 and 429, and `isTerminalRequestError` excludes that list instead of special-casing 429.

I put 421 in it for the same reason as your two: it judges the connection rather than the body, and RFC 9110 says the client may retry over a different one.

409 and 401 I left terminal. A conflict repeats with the same body, and the admin client has no session refresh, so a resend cannot establish one. If you would rather have either of them retried, say which and I will move it.

ascorbic:
Thanks. Could you resolve the conflicts, then we can get this merged

danielmlr:
Rebased — the branch sits directly on `1717d31b` with nothing behind it, and the conflicts are gone.

The conflict was #2586 landing `ApiResponseError` with the status on it. Resolving it dropped the `ApiError` class this PR had added: `isTerminalRequestError` now narrows on `ApiResponseError`, and the status test goes with it, since `preserves status, code, and details on API response errors` already covers that. That is the only code change since the commit that made the retryable statuses a list.

409 and 401 are still terminal, as in my earlier reply. Say the word if you want either of them in `RETRYABLE_CLIENT_ERROR_STATUSES` and I will push it; otherwise this is ready from my side.

ascorbic (approved):
Thanks!

PR #2674, opened 2026-08-25
feat(blocks,admin): portable_text and block_list Block Kit fields + Duplicate on plugin block cards

simone-coelho:
## What does this PR do?

Adds two structured Block Kit elements — `portable_text` and `block_list` — and a Duplicate action on plugin block cards, so plugin blocks can carry rich text and nested registered-block content that stays fully editable in the admin editor.

- **`portable_text`**: the plugin block dialog mounts the existing `PortableTextEditor` for a Portable Text block array. Structure round-trips intact (blocks, spans, marks, markDefs, keys) and is never flattened to a string. The previous best option — a multiline `text_input` over stored Portable Text — showed an empty box and replaced the array with a plain string on save.
- **`block_list`**: an ordered list of registered plugin blocks (`{ _type, _key, ...fields }` items). Authors add from the same registered catalog the document editor offers, edit through the chosen block's own Block Kit form, duplicate, delete, drag to reorder, and move an item to a sibling list when one form holds several lists (including lists nested in `repeater` rows — e.g. one list per column). `allowed_types` / `min_items` / `max_items` bound the list.
- **Duplicate** on plugin block cards inserts an identical copy after the original, beside Edit and Delete.

Both elements are additive `Element` union members with validation mirroring their peers, are accepted as `repeater` sub-fields, and return `null` from the runtime `renderElement` exactly like `repeater`. No existing block or stored value changes behavior.

Discussion: https://github.com/emdash-cms/emdash/discussions/2673 (opened alongside this PR — maintainer approval pending; happy to adjust the design there).

Closes #

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes (`@emdash-cms/blocks` and `@emdash-cms/admin`; `registry-verification` needs built siblings and is untouched)
- [x] `pnpm lint` passes (`lint:quick` clean)
- [x] `pnpm test` passes (blocks 120/120; full admin browser suite 124 files / 1523 tests; new focused suite `tests/editor/block-kit-structured-fields.test.tsx` covers all three behaviors)
- [x] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/2673 (approval pending — opened with this PR)

## Screenshots / test output

```
blocks:  Test Files  4 passed (4) · Tests  120 passed (120)
admin:   Test Files  124 passed (124) · Tests  1523 passed (1523)
new:     tests/editor/block-kit-structured-fields.test.tsx  5 passed (5)
```

### DIFF
--- a/.changeset/structured-block-kit-fields.md
+++ b/.changeset/structured-block-kit-fields.md
@@ -23,6 +23,10 @@ An ordered list of registered plugin blocks. In the dialog, authors add an item 
 
 Both element types are also accepted as `repeater` sub-fields, so a repeater row can carry rich text or its own ordered block list.
 
+#### `hidden` on plugin block definitions
+
+A plugin block definition may declare `hidden: true`: the block stays registered — existing blocks of the type render and edit exactly as before — but document-level insert surfaces (the slash menu, and a `block_list` picker without an `allowed_types` list) no longer offer it. A `block_list` whose `allowed_types` names the type still offers it, so a plugin can ship child-only blocks that authors reach exclusively inside the parent blocks that declare them.
+
 #### Duplicate on plugin block cards
 
 Plugin block cards in the document editor gain a Duplicate action beside Edit and Delete that inserts a copy of the block, with identical data, immediately after the original.
--- a/.changeset/structured-block-kit-fields.md
+++ b/.changeset/structured-block-kit-fields.md
@@ -1,0 +1,28 @@
+---
+"@emdash-cms/blocks": minor
+"@emdash-cms/admin": minor
+---
+
+Adds two structured Block Kit elements so plugin blocks can carry rich text and nested block content that stays fully editable in the admin editor, plus a Duplicate action on plugin block cards.
+
+#### `portable_text`
+
+A rich-text field whose value is a Portable Text block array. The plugin block dialog renders it as a nested instance of the standard Portable Text editor, so authors get the familiar writing surface and the stored value keeps its complete structure — blocks, spans, marks, markDefs, and keys. The value is never flattened to a string. Like `repeater`, the runtime `renderElement` returns `null`; the parent block's own component renders the persisted value.
+
+```ts
+{ type: "portable_text", action_id: "content", label: "Content" }
+```
+
+#### `block_list`
+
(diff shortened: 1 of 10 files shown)


simone-coelho:
Added one more piece to the same feature (c96e80c): an optional `hidden` flag on plugin block definitions. A hidden block stays fully registered — existing blocks render and edit unchanged, and `block_list` items resolve their defs — but document-level insert surfaces (the slash menu, and a `block_list` picker without `allowed_types`) no longer offer it. A `block_list` whose `allowed_types` names the type still does, so a plugin can ship child-only blocks reachable exclusively inside the parents that declare them. Covered by three focused tests in the same test file; changeset updated.

simone-coelho:
Second follow-up commit (0ca416a), both found by driving the real dialog against an imported page rather than in unit tests:

- **portable_text**: a child block's dialog seeds its form values in an effect after first render, so the uncontrolled nested Portable Text editor mounted empty and one Save replaced the stored paragraphs with only the newly typed text. The field now re-seeds a changed incoming value until the author types (the same prop-sync the repeater already applies to its rows); once dirty the author's document is never replaced from outside. The child dialog's `initialValues` are also memoized so the seed effect doesn't re-run every render.
- **repeater**: the sub-field change handler built the next rows from its render closure, so the two writes of a move between two rows' block lists (target row, then source row) let the second erase the first and the moved block vanished. Handlers now read the latest rows through a ref.
- `visiblePluginBlocks` moved into PortableTextEditor.tsx so suites that mock PluginBlockNode with partial factories keep importing.

Focused tests cover both (10 in the structured-fields file); the full admin suite is green (124 files / 1528 tests).

simone-coelho:
Rebased onto current `main` (`4eca8e3`), still as the same three commits.

- One conflict, in the plugin block dialog sizing. I kept the `xl` dialog from #3108, which already covers blocks with fields, so this PR no longer changes the sizing.
- The inventory test added in #3236 lists every Block Kit element type. I added `portable_text` and `block_list` there as `authoring-only`, the same as `repeater`. Without that, `@emdash-cms/plugin-test` fails typecheck.

Typecheck, lint and format checks pass across the repo. Tests pass for `@emdash-cms/blocks` (163), `@emdash-cms/admin` (182 files, 2480 tests, browser) and `@emdash-cms/plugin-test` (39). We still use this downstream. I'm happy to split it if that makes review easier.

simone-coelho:

PR #2699, opened 2026-08-27
fix(plugin-storage): seek on the sort key when paginating with orderBy

### LINKED ISSUE #2691: plugin storage: cursor pagination returns duplicate pages with orderBy desc, and limit is silently capped at 100
`PluginStorageRepository.query()` builds the cursor predicate hardcoded
ascending (`(created_at, id) > cursor`) regardless of `orderBy`
direction, and clamps `limit` with `Math.min(options.limit ?? 50, 100)`
without surfacing that. With `orderBy: { createdAt: "desc" }` page 2
matches exactly the rows page 1 already returned (everything newer than
the cursor) and older rows are unreachable — silent data loss for any
plugin that paginates newest-first (we hit this exporting form
submissions to CSV: 150 rows → 199 rows with 99 duplicates, oldest 50
missing). Suggested fix: flip the cursor comparison with the sort
direction (row-value `<` for desc), and either document or error on
limits above the cap.

---
*Found while building a multilingual client-site starter on EmDash 0.30.0 (Astro 7, Node standalone).*

hossein-webdev:
## What does this PR do?

`PluginStorageRepository.query()` sorted results by an expression over the `data` blob but stepped the cursor through the `created_at` **column**, so the ORDER BY and the seek predicate were reading two different keys. Paging with `orderBy: { field: "desc" }` re-returned the rows page one had already yielded and never reached the older ones.

**The bug is broader than the issue reports.** #2691 describes it as a `desc` problem, but ascending is affected too — it just hides whenever the sort field happens to agree with insertion order. A test that inserts records so `seq` runs *opposite* to `created_at` and then pages `orderBy: { seq: "asc" }` returns 5 of 12 rows, exactly like the `desc` case. Anyone whose records were written in sorted order would never have seen it.

The seek now compares the same expressions the ORDER BY uses, recomputed for the cursor row by id:

```sql
(sort_expr, id) > ((SELECT sort_expr FROM _plugin_storage WHERE ... AND id = ?), ?)
```

Comparing an expression against itself is what keeps this dialect-neutral. No cursor value is bound as a literal, so neither SQLite nor Postgres has to coerce a JSON value to text to make the comparison, and the numeric-vs-lexical ordering `pluginDataOrderExpr` establishes is preserved on both. `id` is appended to the ORDER BY as a tiebreaker so a page boundary cannot fall inside a group of equal sort values.

Mixing directions across several `orderBy` fields cannot be expressed as a single row-value comparison, so that combination now throws `StorageQueryError` when a cursor is supplied, rather than silently returning wrong pages.

### Two things I deliberately did not change

- **The `limit` clamp of 100**, also raised in the issue. Raising it or throwing above it changes behaviour for existing callers, which is a judgement call for you rather than a bug fix. Say the word and I will follow up.
- **The default (no `orderBy`) path.** Its `(created_at, id) > (...)` predicate is already correct and is left byte-for-byte alone, so the common case keeps its current plan and cost.

Closes #2691

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] User-visible strings in the admin UI are wrapped for translation — n/a, no admin UI strings. No `messages.po` changes included.
- [x] I have added and reviewed the user-facing [changeset](https://github.com/emdash-cms/emdash/blob/main/.changeset/README.md) (`patch` for `emdash`)

## Screenshots / test output

Five new cases in `tests/unit/plugins/plugin-storage-pagination.test.ts`, each draining every page and asserting no duplicates and nothing missing. Against `main` three fail:

```
× returns every row exactly once ordering by an indexed field descending
    AssertionError: expected 5 to be 12
× orders descending pages newest-first across the whole walk
    AssertionError: expected [ 'doc-11', 'doc-10', 'doc-09', …(6) ] to deeply equal […(9)]
× returns every row exactly once when sort order is uncorrelated with insert order
    AssertionError: expected 5 to be 12
```

With the fix, all five pass, and the existing suite is unaffected:

```
Test Files  31 passed (31)
     Tests  647 passed (647)
```

`pnpm lint` (`oxlint --type-aware --deny-warnings`) and `pnpm typecheck` are clean.

### One caveat worth your attention

I could not run the **Postgres** parity tests locally (no Docker on this machine, `EMDASH_TEST_PG` unset), so SQLite is the only dialect I verified directly. That is precisely why the seek is written as an expression-vs-itself comparison rather than binding the cursor value with a cast — it introduces no dialect-specific SQL and reuses `jsonOrderExtract` unchanged for both operands. Row-value comparison and scalar subqueries are supported by both dialects. Please still run CI's Postgres job over it; if it disagrees I will fix it promptly.


hossein-webdev:
Thanks — both were worth catching, and the NULL one was a real hole rather than a documentation gap, so I fixed it rather than taking the "documented limitation" option.

**2. NULL sort values.** You were right, and it was worse than "rows after a NULL-valued cursor". I measured it: a 10-row collection where half the documents omit the sorted key returned **3 rows ascending and 5 descending** on my branch. Checking `main` for comparison, it returns **8 unique out of 11 emitted** ascending (i.e. duplicates *and* missing rows), so this was already broken before my change — but my row-value seek converted duplicates into a harder truncation, and I did not want to trade one failure mode for another.

The root cause is that NULL has no agreed position: SQLite sorts it first, Postgres last, and every comparison against it is UNKNOWN. So I gave it an explicit one. Each sort field now contributes a null rank ahead of its value, in both the ORDER BY and the seek:

```
ORDER BY (case when <expr> is null then 1 else 0 end) <dir>, <expr> <dir>, id <dir>
```

and the seek is lexicographic over those ranked fields, with the both-NULL case falling through to `id`:

```
(rank <op> cursorRank) OR (rank = cursorRank AND rank = 0 AND expr <op> cursorExpr)
```

Every operand is non-NULL, so nothing evaluates to UNKNOWN, and it needs no dialect-specific NULL syntax. Making the comparison lexicographic also means it stays correct for more than one `orderBy` field, not just the single-field case.

This does change where those rows sort — NULLs last ascending, first descending, instead of each dialect's default. That is a visible behaviour change for collections whose documents omit an indexed field, so the changeset now calls it out explicitly.

**1. Mixed-direction error.** Covered now.

Both new tests fail against `main`; the file is **5 failing / 2 passing** there and 7/7 on the branch. Full suite `31 files, 649 tests` green, `oxlint --type-aware --deny-warnings` and `tsgo --noEmit` clean.

Postgres is still unverified locally (no Docker, `EMDASH_TEST_PG` unset) — that constraint is exactly why the null rank is a plain `CASE` and the seek compares expressions to themselves rather than using `NULLS FIRST/LAST` or `IS NOT DISTINCT FROM`. Please still let CI's Postgres job have it.

The `limit` clamp of 100 I've left alone as before — happy to follow up separately if you want it raised or made to throw.

ascorbic (approved):
Thanks!

PR #2743, opened 2026-08-27
feat(release-service): add delegated release foundations

ascorbic:
## What does this PR do?

Adds the foundation for the delegated release service: the Worker shell, exact create-only PDS conformance harness, confidential OAuth metadata, compact JWE envelope encryption, safe unexpected-error logging, and CI support for dependent stack PR bases.

This is PR 1 of 7. Its three focused commits are retained for review. The stack merges as one unit and no intermediate branch is a deployable service version.

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] I have added and reviewed the user-facing changeset for the published CLI OAuth fix

## Screenshots / test output

The complete stack passes formatting, strict lint, package/application typechecks and tests. The public-client G0 lifecycle passed against the npmX-hosted Bluesky PDS and Cirrus; deployment-only confidential-client checks remain pending.


ascorbic:
## 🔍 Adversarial review — stack #2766, layer 2 of 8

_Automated deep review of this layer's diff (`feat/blob-hosted-registry-artifacts...feat/drs-review-01-foundation`). Findings ranked most-severe first; confidence is **CONFIRMED** (full code path read) or **PLAUSIBLE** (strong suspicion, path partially read). Nothing was auto-fixed._

PR #2743 adds a new private Cloudflare Worker (`apps/release-service`) with three GET routes (client metadata, JWKS, health), fail-closed configuration parsing (origin, deployment ID, redirect URIs, ES256 assertion keyset, encryption keyring), a compact-JWE envelope-encryption module (A256GCMKW + A256GCM, versioned keyring, context digest bound in the protected header), and redacted error logging. It adds an exact-scope PDS conformance probe runner to `@emdash-cms/registry-client` (`./internal/conformance` export), a hidden `pds-conformance` command to the published CLI, reworks CLI OAuth session resume/revoke to persist and reuse the original loopback client registration (the "published CLI OAuth fix", with a changeset), a citty 0.1→0.2 bump, a spec/evidence doc, and a one-line CI change so checks run on stacked PRs. Overall the crypto and worker shell are carefully written with genuinely strong tests (real jose crypto, header tampering, fail-closed config), but the layer ships with its own test suite red, the conformance harness has a soundness hole, and the OAuth fix has a design flaw in what it persists. Verdict: solid engineering undermined by a rebase-staleness break at exactly the thing the PR claims ("exact scope"), which a later stack commit quietly fixes.

### Findings

- **[high]** `apps/release-service/test/config.test.ts:12`, `test/worker.test.ts:29` — **the layer's own tests fail.** Both assert `scope: "atproto repo:com.emdashcms.experimental.package.release?action=create"`, but `getDelegatedReleasePermission()` at this commit (inherited from base #2765) returns that string plus ` blob:application/gzip blob:image/*`, and `config.ts` puts `permission.scope` into the client metadata verbatim. Verified empirically by running the suite at this ref in an isolated clone: 2 failed / 60 passed. A later stack commit (`35c901a9 fix(release-service): request artifact blob scopes`) repairs the expectations, but this layer as-is is red, and the PR's "exact create-only" framing is false at this layer. **CONFIRMED** (empirically).
- **[high]** CI never sees the failure: **release-service is excluded from CI at this layer.** Root `test:unit` and `typecheck` filters (`package.json:8,13`) enumerate packages plus `@emdash-cms/aggregator`/`@emdash-cms/labeler` but not `@emdash-cms/release-service`; only the top stack layer adds it to ci.yml. So the very PR that enables CI on stacked PRs ships a Worker whose failing tests and typecheck run in no CI job. **CONFIRMED.**
- **[medium]** `packages/registry-client/src/conformance/index.ts:74` (`expectedDenial`) — **any 4xx counts as a scope denial.** A 401 from a token that expires between the create probe and the deny probes, or a 429 rate limit, makes `release-update`/`release-delete`/`profile-create`/`unrelated-create` all "pass" without the PDS ever demonstrating scope enforcement — a non-conforming PDS can produce a passing report. The error code (`ScopeMissingError`/`InsufficientScope`, both recorded in the doc's evidence) is captured but never checked. The tests cover 5xx→error but not 401/429. **CONFIRMED** semantics, **PLAUSIBLE** false-pass scenario.
- **[medium]** `packages/plugin-cli/src/oauth.ts:735-740` — **registration stores the *granted* scope, but the refresh token is bound to the *requested* scope.** atcute builds the loopback `client_id` from `metadata.scope` verbatim (`@atcute/oauth-types` `buildLoopbackClientId`); authorize used the requested scope (DEFAULT_CLI_SCOPE / `options.scope` / legacy fallback), while `clients.json` records `storedSession.tokenSet.scope` (the AS's token-response scope). Whenever granted ≠ requested — partial grants are a state this codebase explicitly anticipates (`publish.ts:298` `missingBlobScopes` on the active grant), and `buildLoopbackClientId` even drops the scope param entirely when it equals the default `atproto` — resume/revoke reconstruct a *different* client_id and refresh fails again, the exact bug the changeset claims to fix. Store the client's `metadata.scope` instead. **CONFIRMED** code path; trigger depends on AS behavior.
- **[medium]** `packages/plugin-cli/src/oauth.ts:772-775` (`getClientRegistration`) — **upgrading the published CLI hard-invalidates every existing session immediately.** `resumeSession` throws "sign in again" for any pre-fix session (no `clients.json` entry) even when the access token is still valid and no refresh is needed. The changeset's "Sign in again before their access token expires" implies a grace window that does not exist. Acceptable as a fix, but the behavior and the release note disagree. **CONFIRMED.**
- **[low]** `packages/plugin-cli/src/oauth.ts:838-846` — non-strict `revokeSession` for pre-fix sessions silently skips server-side revocation entirely (deletes local state, `serverRevoked: false`); the old code at least attempted `client.revoke`. Mitigated by the old attempt likely failing at the AS anyway. **CONFIRMED.**
- **[low]** `apps/release-service/src/crypto/encryption.ts:4` — A256GCMKW wraps each CEK under the long-lived master key with a random 96-bit IV; nothing enforces rotation before the NIST SP 800-38D ~2^32-wrap bound. Deterministic A256KW would remove nonce management from the master key. **PLAUSIBLE**, long-horizon.
- **[low]** `apps/release-service/src/index.ts:31-41` — unexpected-error logs carry no error identity at all (`{name:"UnhandledError"}`, no constructor name, no `EncryptionError.code`). Redaction is correct (test proves the message can't leak), but production 500s are undiagnosable. **CONFIRMED** behavior.

### PR description vs code

- "Exact create-only PDS conformance" / "exact-scope OAuth metadata": at this layer the delegated permission (and therefore the served client metadata and the CLI's conformance authorize request) includes `blob:application/gzip blob:image/*`. The spec doc (`docs/technical-specs/delegated-release-service-g0-conformance.md`) contradicts the code in the same way: it records granted scope as create-only, and states "the browser consent screen must show only the active release collection's create authority" — untrue as of this rebase. Its 2026-08-24/25 evidence was gathered against the pre-blob-scope permission but is presented as passing the current "exact scope" column.
- Changeset wording vs. behavior (grace window), per the medium finding above.
- "CI support for dependent stack PR bases" is accurate and safe (pull_request event, read-only permissions, pinned actions, `persist-credentials: false`, no pull_request_target) — but it does not extend to the new Worker, whose checks exist in no CI job until the top layer.
- Verified non-issues worth stating: `wrangler.jsonc`'s top-level `secrets.required` is valid in the bundled wrangler schema (secrets are real bindings, not vars); citty 0.2 still yields `boolean|undefined` for boolean args (the `args.json === true` change is typing-only); JWKS route provably never serves `d`; error redaction genuinely holds on every catch path traced.

### Test-coverage gaps

- Conformance harness: no test that a 401/429 on a deny probe is rejected (only 5xx is tested); no `com.atproto.repo.applyWrites` update/delete probe; no probes at all for the granted blob scopes (`uploadBlob` allowed for gzip/image, denied for other MIME types).
- `oauth-registration.test.ts` mocks fetch with blanket 200s and never asserts that the client_id sent to the token/revocation endpoint matches the authorize-time client_id — the central invariant of the fix — so the granted-vs-requested scope defect is invisible to it; the granted≠requested case has no test.
- No test covers `resumeSession` for a session whose registration scope differs from the stored token scope, nor `needsRotation` throwing on malformed input (interface says boolean).

_Verification environment: full test runs at ref `bdebf5db` in an isolated clone (release-service: 2 failed/60 passed; plugin-cli 404 passed; registry-client 86 passed; release-service typecheck clean). The shared working tree was not modified._

~ 🤖 Fable

PR #2744, opened 2026-08-27
feat(release-service): add publisher authority and admission

ascorbic:
## What does this PR do?

Adds publisher identity and confidential OAuth custody, encrypted retained delegation, application sessions, serialized refresh, GitHub Actions OIDC verification, publisher-owned workload policy, intent admission, idempotent package/version reservations, and generation-bound publication operations.

This is PR 2 of 7. It keeps the five custody, compatibility, log-redaction, and admission commits separate inside the PR. The stack merges as one unit and this branch is not deployable by itself.

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] I have added and reviewed the user-facing changeset (not applicable; this layer changes a private application)

## Screenshots / test output

OAuth custody/callback, publisher Durable Object, session, OIDC claim-substitution, workload-policy, intent-state, reservation, replay, and publication-operation tests pass in workerd.


ascorbic:
## 🔍 Adversarial review — stack #2766, layer 3 of 8

_Automated deep review of this layer's diff (`feat/drs-review-01-foundation...feat/drs-review-02-authority-admission`). Findings ranked most-severe first; confidence is **CONFIRMED** (full code path read) or **PLAUSIBLE** (strong suspicion, path partially read). Nothing was auto-fixed._

At this layer the PR adds: (a) a `PublisherDurableObject` (one per publisher DID, SQLite-backed) holding OAuth transaction state, a single encrypted delegation row with CAS state-versioning, hashed publisher app sessions with epoch-based mass revocation, workload policies, release intents with reservations/idempotency, and lease-based publication operations; (b) OAuth custody stores bridging `@atcute/oauth-node-client` to the DO, with layer-01 JWE envelope encryption context-bound to `{purpose, table, primaryKey, ownerDid}` so ciphertext cannot be replayed across publishers, identity-flow sessions kept request-local in memory, delegation-flow refresh serialized through a DO lease; (c) three HTTP routes (identity authorize, delegation authorize, callback) with `__Host-` cookies, double-submit CSRF, state-to-cookie binding; (d) GitHub Actions OIDC verification and policy evaluation — which are **not wired to any route at this layer** (used only by tests). Overall quality is high: real-DO tests (no mocked security boundary), fail-closed validation everywhere, hashed-at-rest tokens, genuine refresh serialization (verified against the library: `SessionGetter.get` wraps every read in `requestLock` → `beginDelegationRefresh` lease with token+generation+version binding, all writes in `transactionSync`). The substantive defects are in state-machine/lifecycle coherence, not in crypto or authentication.

### Findings

- **[high]** `apps/release-service/src/publisher-do/intent-state.ts:345-364,467-469` + `publication-operation.ts:140-232,234-307` — expired-but-non-terminal intents lose their version reservation while the publication path ignores intent expiry, so two intents can drive the same package+version concurrently. `create()` deletes a reservation whose intent has `expires_at <= now` and is not terminal/published — including intents in `publishing`/`reconciling`. But `beginPublicationOperation`/`completePublicationOperation` never check intent expiry (only `state === 'publishing'` + generation), so the first intent can still complete its PDS write while a second intent, admitted after the steal, publishes the same version. Additionally, an expired intent in `publishing`/`reconciling` is permanently wedged in `transition()`: every `toState !== 'expired'` is blocked by the expiry guard (line 467) and `'expired'` is absent from `ALLOWED_TRANSITIONS['publishing'|'reconciling']` (lines 39-40) — so post-recovery reconciliation of an expired intent can never record a final state. Wedge + reservation release: **CONFIRMED**; end-to-end double publish depends on later layers' reconciliation: **PLAUSIBLE**.
- **[medium]** `apps/release-service/src/publisher-do/publisher-do.ts:1110-1128` — expired-row GC is alarm-starved: `#scheduleNextAlarm` sets an alarm only from `publication_operations` deadlines and deletes it otherwise, yet `alarm()` is the only place `oauth_states` and `intent_idempotency` expired rows are purged. A publisher DO with no publication activity accumulates rows forever, and `putOAuthState` is reachable **unauthenticated** via `POST /v1/publisher/session/authorize` (the Origin/`x-emdash-request` check at `src/oauth/routes.ts:103-110` is trivially satisfiable by non-browser clients), so anyone can grow a victim DID's DO storage without bound. **CONFIRMED.**
- **[medium]** `apps/release-service/src/oauth/custody.ts:41-42,567-599` — abandoned refresh lease blocks all delegation use with no possible wait-out: waiters poll for `REFRESH_LOCK_TIMEOUT_MS` (30s) but a crashed holder's lease lasts `REFRESH_LEASE_MS` (60s), so every waiter is guaranteed `OAUTH_REFRESH_LOCK_TIMEOUT` during the tail ~30s. Compounding it, the library acquires this lock on **every** session read (verified in `@atcute/oauth-node-client` `session-getter.js:133`), not just on refresh, so each restore costs a lease-acquire write + generation bump and all reads for a publisher serialize. **CONFIRMED** mechanics; self-heals after lease expiry.
- **[medium]** `apps/release-service/src/publisher-do/workload-policy.ts:156-166` vs `src/workload/policy.ts:29` — `workflow_ref` matching is case-inconsistent: `put()` lowercases `repository` and compares repo names case-insensitively, but stores `workflowRef` verbatim, and `evaluateWorkloadPolicy` exact-compares it against GitHub's canonical-case claim. A publisher who registers `myorg/myrepo/.github/workflows/release.yml@...` for a repo canonically cased `MyOrg/MyRepo` gets `WORKLOAD_WORKFLOW_MISMATCH` on every admission. Fail-closed (availability, not security). **CONFIRMED** inconsistency; whether it bites depends on how later layers source the value.
- **[low]** `apps/release-service/src/workload/policy.ts:17-43` — policy cannot constrain `job_workflow_ref`, `event_name`, or `actor`: a pinned caller workflow invoking an arbitrary reusable workflow, or triggered via `pull_request_target`, passes evaluation. The `workflow_ref`-belongs-to-repo check in `github-oidc.ts:80` prevents cross-repo *caller* spoofing, but the reusable-workflow dimension is normalized, digested, and then ignored. Design gap; publisher controls the pinned workflow's content. **PLAUSIBLE.**
- **[low]** No OIDC `jti` replay ledger anywhere at this layer: within the 10-minute `maxTokenAge` window a captured token can be presented repeatedly; `digestWorkloadIdempotencyIdentity` deliberately excludes `jti`, so replays with different package/version/idempotency keys mint distinct intents. Depends on whether a later layer records `tokenId`; judged as-is: **PLAUSIBLE.**
- **[low]** `apps/release-service/src/oauth/custody.ts:457-458,486-495,546-557` — the `#preserveNextDelete` protocol guards a call sequence the real library never makes: `CachedGetter.getStored` (verified) swallows `store.get` errors and skips `deleteStored` when `storedValue` is `undefined`, so the "delete after failed get" it defends against cannot occur via the library; meanwhile the sticky flag will silently no-op the next *legitimate* `delete()` (i.e., a revocation) on the same store instance after any failed get. Fragile cross-call state; currently unreachable harm: **CONFIRMED** behavior.
- **[low]** `apps/release-service/src/oauth/routes.ts:289-292` — identity-flow tokens (scope `atproto`, and a `refresh_token` when the PDS issues one) are abandoned in request memory without upstream revocation after the app session is minted; every login leaves a live grant on the user's PDS. **CONFIRMED.**
- **[low]** `apps/release-service/src/oauth/custody.ts:526-529` — with an active delegation, a second delegation authorization always fails CAS at the *end* of the OAuth dance (the library then revokes the just-issued tokens upstream — verified in `oauth-client.js` callback + `onStoreError`), and no revoke route exists at this layer, so consent rotation is a dead-end returning a generic `OAUTH_CALLBACK_INVALID`. Safe but a guaranteed UX trap. **CONFIRMED.**

### PR description vs code

- "GitHub Actions OIDC verification", "workload policy", and "intent admission" ship as **unwired code**: no HTTP route calls `verifyGitHubActionsToken`, `evaluateWorkloadPolicy`, or `createIntent` — they exist only as DO RPC methods and exported functions exercised by tests. The only live HTTP surface added is the two authorize routes and the callback.
- "Serialized refresh" is accurate and stronger than claimed (lease + generation + hashed token + CAS on state version, all inside `transactionSync`).
- "Idempotent package/version reservations" are **per-publisher-DO only**; nothing at this layer prevents two different publishers reserving the same package name (presumably a later product-layer concern, but the wording overstates this layer).

### Test-coverage gaps

- No test of intent expiry while in `publishing`/`reconciling` (would have caught the high finding); `publisher-intent-state.test.ts:145` only covers expiry in `received`.
- No test that expired `oauth_states`/`intent_idempotency` rows are ever purged absent publication activity (would have caught the alarm starvation).
- No custody-level test of an abandoned refresh lease vs the 30s waiter timeout (DO-level lease takeover is tested; the custody polling mismatch is not).
- No case-variant `workflow_ref` test between policy storage and evaluation.
- OIDC tests omit alg-confusion (e.g., an HS256/ES256 token), `aud`-as-array, and any reusable-workflow *policy* scenario — the `job_workflow_ref` tests only cover claim shape, not that policy ignores it.
- The real `OAuthClient.restore()`/refresh path (requestLock through the library against a fake PDS) is never exercised — refresh serialization is only tested by calling `custody.requestLock` directly with the lock name hard-coded in the test.

No repo-convention violations found; tests run against real DO storage and real envelope encryption, which is exactly right for this code.

~ 🤖 Fable

PR #2746, opened 2026-08-27
feat(registry): verify delegated releases independently

ascorbic:
## What does this PR do?

Adds the isolated verifier Worker, shared artifact/manifest/provenance verification, authoritative direct-PDS record reads, release Workflow verification steps, and independent EmDash installer enforcement. Moderation labels remain metadata-only visibility inputs bound to exact signed profile or release CIDs; they never attest to plugin code.

This is PR 4 of 7. It retains the main verification commit and the artifact-proxy checksum compatibility follow-up separately. The stack merges as one unit and this branch is not deployable by itself.

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] User-visible admin strings are wrapped for translation; no `messages.po` files are included
- [x] I have added and reviewed the user-facing changesets for core, admin, registry client, and registry verification

## Screenshots / test output

Core direct-PDS install conformance, admin, registry-client, registry-verification Node/workerd, release-service verification, and verifier Worker suites pass.


ascorbic:
## 🔍 Adversarial review — stack #2766, layer 5 of 8

_Automated deep review of this layer's diff (`feat/drs-review-03-approvals...feat/drs-review-04-verification`). Findings ranked most-severe first; confidence is **CONFIRMED** (full code path read) or **PLAUSIBLE** (strong suspicion, path partially read). Nothing was auto-fixed._

It adds an isolated `apps/release-verifier` Worker (no bindings, no secrets; RPC-only) that fetches an artifact + Sigstore provenance under DoH-based SSRF guards and verifies checksum, bundle shape, and SLSA subject digests (now multi-algorithm via `artifactDigests`); a Cloudflare Workflow (`ReleaseIntentWorkflow`) that reads the publisher's PDS three times, persists six idempotent verification steps in the publisher DO (digest-bound, conflict-on-divergence), evaluates policy/access-diff, and drives the intent state machine (whose `transitionIntent` correctly replays same-digest transitions); a `DirectPdsClient` in `registry-client` that fetches records as CAR proofs via `com.atproto.sync.getRecord` and genuinely verifies MST inclusion + commit signature against the DID document key; and a rewritten core install/update path that reads authoritative records from the publisher PDS (`readAuthoritativePackageRelease` over `ssrfSafeFetch` httpsOnly), rejects aggregator URI/CID drift (`AGGREGATOR_RECORD_MISMATCH`), verifies artifact multihash + bundle + canonical declaredAccess equality + provenance (real `GitHubProvenanceVerifier` in production), binds consent to exact profile/release CIDs (`RECORD_CONSENT_REQUIRED`/`RECORD_VERIFICATION_DRIFT`), and adds a `verifyOnly` preview endpoint with proper `plugins:manage` authorization. Moderation labels are CID-bound and block-only. Overall quality is high — checks are ordered fail-closed, the cross-implementation conformance fixture is real (exercised against both the verifier Worker and the core installer with a fake-PDS repo signing real proofs, including key-substitution and tampered-CAR cases), and **no verification bypass was found**. The real defects are fail-closed operational breaks and two service-vs-installer divergences.

### Findings

- **[high]** `apps/release-service/src/verification/pds.ts:228,257` — `listPackageReleases` filters by `rkeyStart`/`rkeyEnd`, parameters that no longer exist in `com.atproto.repo.listRecords` (verified against the current lexicon: only `limit`, `cursor`, `reverse`); real PDSes ignore unknown params and return the whole collection, and `releaseVersion()` **throws** `RELEASE_LIST_INVALID` for any record outside the `{slug}:` prefix instead of filtering. Failure scenario: any publisher with a second package (or any non-semver release rkey) can never complete delegated-release verification — every intent dies in the `authoritative-records` step. Fails closed, so no bypass, but it breaks the feature's core flow on real PDSes; the test (`verification-pds.test.ts`) asserts the dead params are echoed and returns only in-range records, encoding the wrong assumption. **CONFIRMED.**
- **[medium]** `apps/release-service/src/verification/evaluate.ts:339,299` + `apps/release-verifier/src/verify.ts:167,246` — the verifier reports the **post-redirect final URL** (`resource.value.url`), while `evaluateVerifiedRelease` requires it to equal the signed record's URL exactly (`ARTIFACT_RECORD_MISMATCH`), and `reportBackedVerifier` does the same for provenance (`PROVENANCE_UNVERIFIABLE`). `fetchVerifiedResource` follows up to 3 redirects, so the redirect-following is dead weight: any artifact or provenance URL that redirects — e.g. GitHub release-asset URLs, the natural home for GitHub-provenance artifacts — invalidates the intent. Fails closed. **CONFIRMED** behavior (possibly intended strictness, but then redirects should be disabled at fetch, not converted into a confusing mismatch).
- **[medium]** `packages/registry-verification/src/records.ts:287` — the installer path calls `verifier.verify()` without `artifactDigests`, so core install/update only matches **sha256** SLSA subjects, while the release-verifier Worker supplies sha256/384/512 candidates. A release whose attestation subject carries only a sha384/512 digest passes delegated service verification but fails independent installer verification on every EmDash instance ("Artifact digest mismatch") — a publish/install split-brain the shared conformance fixture doesn't cover (its subject is sha256). **CONFIRMED** divergence.
- **[medium]** `packages/core/src/astro/routes/api/admin/plugins/registry/artifact.ts:305` (commit 5185cc8f) — the follow-up commit dropped the deliberate `!descriptor.blobCid` skip and now verifies the **raw-blob multihash** against the bytes returned by the record-scoped cache's `/img/{preset}` URL, which the aggregator lexicon itself describes as "image representations" (i.e. transformed renditions). If the external Cumulus cache resizes/re-encodes (as avatar/banner/thumbnail presets exist to do), every blob-backed icon/banner/screenshot 502s with `ARTIFACT_CHECKSUM_MISMATCH`. The proxy test mocks the cache returning byte-identical originals, so the suite cannot detect this. **PLAUSIBLE** (depends on external cache behavior; the removed guard came from the layer below, suggesting the transform case was previously accounted for). On the masquerade question: the retained bare-hex path here is image-proxy-only; the install/update path strictly rejects hex (`INVALID_MULTIHASH`), and hex is still SHA-256, so no weaker binding is reachable for installable code.
- **[low]** `packages/admin/src/components/RegistryPluginDetail.tsx:148` + `packages/admin/src/lib/api/registry.ts` (`resolveDidToHandle`) — publisher handle re-verification blocks only on `status === "invalid"`; network/DID-method failures return `"missing"`, which installs fine, and the server never verifies handle↔DID correspondence at all. The changeset's impersonation defense is advisory client-side UI; an aggregator that maps a coveted handle to an attacker DID is only caught if bidirectional resolution *provably* fails rather than merely erroring. **CONFIRMED** behavior (arguably deliberate fail-open on "couldn't determine", but worth an explicit decision).
- **[low]** `apps/release-service/src/workflows/release-intent.ts:1446-1463` (isolated-verifier step) and step-conflict paths — `NonRetryableError` on invalid verifier input or `VERIFICATION_STEP_CONFLICT` (PDS content changed between step retries) terminates the workflow without transitioning the intent, leaving it in `verifying` until expiry rather than `invalid` with a reason code. Operational only. **CONFIRMED.**
- **[low]** `packages/core/src/api/handlers/registry.ts` (`registryArtifactError` default branch) — artifact *fetch* failures (network/host errors from `fetchReleaseArtifact`) now map to `INVALID_BUNDLE`, mislabeling transient network problems as bundle corruption for the admin's code→message mapping; previously they surfaced as a generic install error. **CONFIRMED**, cosmetic/error-semantics.

### PR description vs code

- "Isolated verifier Worker" is accurate in the no-shared-state sense (zero bindings/secrets, RPC only), but its inputs (URL, checksum, repository) are computed by the release service from its own PDS read — the verifier is an isolated *executor* that cannot detect a compromised release service feeding it consistent-but-wrong inputs; the genuine independent backstop is the EmDash installer path, which does re-verify everything from authoritative sources (claim holds there).
- "Authoritative direct-PDS record reads": the core installer verifies MST proofs and commit signatures (`verifyRecord` over CARs); the **release service** reads plain `com.atproto.repo.getRecord` JSON and trusts the PDS-asserted CID with no proof or CID recomputation (`apps/release-service/src/verification/pds.ts` `parseRecord`). Self-consistent under a PDS-as-authority model, but weaker than the stack's proof language; upper layers must not treat those CIDs as proof-verified.
- Moderation-label claims check out: `hasCurrentRecordLabel` binds `uri`+`cid` exactly, respects `neg`, blocks only (metadata-only). One nit: a malformed `exp` makes a yank label *not* current, i.e. fails open for a blocking label.

### Test-coverage gaps

- `verification-pds.test.ts` is tautological about `rkeyStart`/`rkeyEnd` — no test with a PDS that ignores them or a second package in the collection.
- The artifact-proxy blob test mocks the image cache returning original bytes, hiding the transform-vs-checksum contradiction.
- No test that a redirecting artifact/provenance URL survives service-side verification (it doesn't).
- No installer-side test for a provenance subject using sha384/512 only (the divergence above).
- The workflow "ready" path uses an echo-stub verifier in miniflare, so it proves state-machine wiring only — acceptable because the verifier Worker and the core installer both run the *same* shared conformance fixture against a fake PDS with real signed proofs, which is the opposite of tautological and the strongest test asset in the PR.

~ 🤖 Fable

PR #2748, opened 2026-08-27
feat(release-service): add operations and recovery

ascorbic:
## What does this PR do?

Adds the sharded identity directory, per-publisher abuse limits, encrypted R2 archive and fail-safe restore, encryption rotation controls, Analytics Engine metrics, archive Workflow, and the operational runbook. The follow-up commit makes approval expiry authoritative and restore preparation replay-safe after partial progress.

This is PR 6 of 7. The implementation and recovery-safety follow-up remain separate commits. The stack merges as one unit and this branch is not deployable by itself.

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] User-visible strings use Lingui and logical RTL-safe layout
- [x] I have added and reviewed the user-facing changeset for the typed operator client

## Screenshots / test output

Directory, rate-limit, archive/restore, encryption operations, metrics, alarm expiry, replay safety, Workflow, and typed-client tests pass.


ascorbic:
## 🔍 Adversarial review — stack #2766, layer 7 of 8

_Automated deep review of this layer's diff (`feat/drs-review-05-publication-product...feat/drs-review-06-operations`). Findings ranked most-severe first; confidence is **CONFIRMED** (full code path read) or **PLAUSIBLE** (strong suspicion, path partially read). Nothing was auto-fixed._

Layer 06 adds, on top of the layer-05 release service: a 256-shard non-authoritative identity directory DO registered from OAuth callbacks and listed via an Access-viewer route; a per-publisher-DO fixed-window rate limiter (workload/repository/publisher scopes) consumed in the submit-intent route after token verification and policy evaluation; encrypted R2 publisher archives (JWE pages context-bound to deploymentId/objectClass/table/`archiveId:page`/ownerDid, create-only R2 writes with decrypt-and-compare replay) plus plaintext "sanitized" audit export objects; a prepare-then-page-by-page restore state machine in the publisher DO (atomic wipe+marker, digest-bound page replay, forced suspension, delegation stripped to `reauthorization_required`, non-terminal intents forced to `failed`); CAS-based encryption-record rotation endpoints for publisher and approver ciphertexts; Secrets-Store-aware config loading; Analytics Engine metrics; an archive Workflow; operator UI panels; a runbook; and typed operator client methods with a changeset. The two headline recovery claims hold: intent/approval expiry is enforced at transition time inside the authoritative DO (`intent-state.ts:523` rejects any transition of an expired intent except to `expired`; the new alarm sweep is only cleanup), and restore preparation/pages are replay-safe (wipe + `operations_restore` marker in one `transactionSync`; page data + `operations_restore_pages` digest marker in one transaction). Overall quality is high — validation is obsessive, CAS discipline is consistent, and authorization on every new route is admin/viewer-gated with audience-scoped Access JWTs — but there are real gaps between the runbook's privacy claims and the audit-export code, and the restore state machine has unrecoverable dead-ends.

### Findings

- **[medium]** `apps/release-service/src/backup/routes.ts:427-460` — **Audit export objects are written to R2 unencrypted and contain the full audit event rows plus the raw `publisherDid`**, contradicting the runbook's invariant ("Audit export objects contain only the sanitized `audit_events.public_payload` contract"). `writeAuditObject` serializes `{version, publisherDid, events}` where each event carries `actorIdentity` (Access operator subjects/emails from `#appendAudit`), `subject`, and `reasonCode` — the plaintext DID also defeats the `ownerHash` pseudonymization used for the R2 key. The test only asserts the ciphertext/private-payload strings are absent, not the DID or actor identities (it asserts DID absence for snapshots but not audit objects). Failure scenario: anyone with R2 bucket read access reads operator identities and publisher DIDs the encrypted-snapshot design was built to protect. **CONFIRMED.**
- **[medium]** `apps/release-service/src/publisher-do/publisher-do.ts:1096-1122` (`prepareOperationsRestore`) + `operations-restore.ts` — **a restore that cannot complete is a permanent dead-end.** Once `operations_restore` is `prepared`/`restoring`, `prepare` for any other archive returns `RESTORE_CONFLICT`, and there is no abort/reset operation anywhere. If an archive page object is corrupted or deleted after the manifest check (manifest decrypts, page N doesn't — `readEncryptedObject` 404/409 forever), the shard can never be restored from any archive, contradicting the runbook's "Runbooks never require direct Durable Object SQLite edits." Relatedly, a *completed* restore of an archive can never be re-applied (prepare replays by archiveId+pages; encryption context pins objects to the archiveId), so each archive is one-shot — undocumented. **CONFIRMED** mechanism; **PLAUSIBLE** trigger.
- **[medium]** `apps/release-service/src/backup/routes.ts:447-459` — **after restoring into genuinely lost DO storage, audit sequences restart at 1** (fresh `sqlite_sequence`; `AUTOINCREMENT` only protects the same storage), so the next archive of that publisher writes audit keys like `audit/{hash}/…0001-…0100.json` that collide with the pre-loss export objects with different content → `ARCHIVE_OPERATION_FAILED` 409, permanently, once page boundaries align (guaranteed at ≥100 events since the first page is always 1–100). The restored publisher becomes un-archivable — and therefore un-restorable next time. **PLAUSIBLE** (only after true storage loss, which is exactly the scenario this feature exists for).
- **[low/medium]** `apps/release-service/src/backup/routes.ts:279` (`INTENT_PAGE_SIZE = 4`) vs `crypto/encryption.ts` `MAX_PLAINTEXT_BYTES = 1MB` — four maximal intents (16KB identity + 64KB release + 64KB stateData each, doubled by JSON string escaping when re-embedded in the snapshot) can exceed the 1MB encrypt cap; `encrypt` throws → deterministic `ARCHIVE_OPERATION_FAILED` on that page on every retry → that publisher can never be archived. **PLAUSIBLE.**
- **[low]** `apps/release-service/src/workflows/release-intent.ts:417-425` — `writeOperationsMetric` throws `TypeError` for any `outcome` not matching `DIMENSION_PATTERN`, and it is called with the verifier's error code *before* the failure report is persisted; an out-of-pattern or empty code would turn a terminal "verification failed" into workflow step retries and an errored workflow. Today the verifier emits only `VERIFIER_INPUT_INVALID`/`VERIFIER_INTERNAL_ERROR` (both pass), so this is a fragility, not a live bug — but the RPC boundary types the code as arbitrary `string`. **PLAUSIBLE.**
- **[low]** `apps/release-service/src/operations/encryption-routes.ts:1993` / `OperatorPage.tsx` — the rotation `complete` flag is per-page (`nextCursor === null && raced === 0`); races on earlier pages are forgotten once the cursor advances, so the final page (and the UI "Verified" badge, which keeps only the last result) can read complete while earlier raced records remain under the old key. The runbook's mandatory second full scan mitigates this, but an operator trusting the badge and executing step 8 (retire old key) strands retained OAuth ciphertext under a removed key → `ENCRYPTION_OPERATION_FAILED` / forced reauthorization. **CONFIRMED** behavior, procedural mitigation exists.
- **[low]** `apps/release-service/src/publisher-do/publisher-do.ts` — `#scheduleNextAlarm` includes `oauth_states`, `publisher_sessions`, and `intent_rate_idempotency` expiries, but nothing arms the alarm when only those rows are written (`putOAuthState`, `createPublisherSession`, `consumeIntentRateLimit` don't schedule); a publisher DO that only ever does OAuth flows accumulates expired rows until an intent or publication event first schedules an alarm. Correctness is protected by lazy expiry checks at read; housekeeping only. **CONFIRMED.**
- **[low]** `apps/release-service/src/publisher-do/publisher-do.ts:456-467` and `approver-do/schema.ts` — the layer changes `CREATE TABLE IF NOT EXISTS` DDL in place (adds `encryption_purpose NOT NULL` to `oauth_states`; adds `'access'` to the approver `audit_events` CHECK). `IF NOT EXISTS` never migrates existing DO storage, so any pre-existing DO would fail inserts. Benign only because the runbook explicitly states the service must not be deployed before the full stack lands; if any earlier layer was ever deployed, this breaks OAuth on existing shards. **CONFIRMED** mechanism, moot under the stated deployment constraint.

Verified non-findings worth relaying: rate-limit attribution is the authenticated OIDC identity (repository.id/workflow.ref from the verified token) and quota is consumed only after the victim publisher's own policy authorizes the workload, so a third party cannot burn a victim's budget; limiter errors fail closed (500); replays don't double-count. Restore cannot silently roll back newer live state — it requires admin role, service-control suspension, exact DID confirmation, and an explicit destructive prepare, and restored authority is neutralized (empty delegation ciphertext, `reauthorization_required`, workloads inactive, non-terminal intents failed; restored reservations for terminal intents are lazily reaped by `create()`). Archive objects are tamper/substitution-proof via the JWE context digest (deploymentId, table, `archiveId:page`, ownerDid, key version in AAD-equivalent header). Rotation is CAS-per-record with fail-closed decryption on missing retained keys, and every ciphertext class currently stored (delegation, publisher oauth_states, approver identity_transactions) is enumerated by the rotation cursors. Metrics writes are non-blocking, dimension-validated, and use hashes not DIDs.

### PR description vs code

- The runbook's "Audit export objects contain only the sanitized `audit_events.public_payload` contract" is false — full rows plus plaintext publisherDid are exported (finding 1).
- "Fail-safe restore" is accurate for authority handling but overstated for recoverability: no abort path for a wedged restore, and archives are one-shot restorable — neither is documented.
- The changeset exists and accurately describes the operator-client additions.
- Everything else claimed (sharded directory, per-publisher limits, rotation, workflow, metrics, runbook, typed client) is present and matches.

### Test-coverage gaps

- No negative restore tests at all: page-digest conflict (`RESTORE_CONFLICT`), out-of-order pages, `RESTORE_NOT_EMPTY`, missing/corrupt page mid-restore (the dead-end), or a tampered/substituted R2 object failing decrypt.
- Audit export objects are never asserted free of `publisherDid`/actor identities (they aren't free of them).
- No test of the archive Workflow's retry-after-content-changed 409 path, nor of archive page size vs the 1MB encryption cap.
- Rotation tests don't cover a race on a non-final page followed by a "complete" final page.
- Rate-limit tests don't cover fixed-window boundary bursts or limiter-error fail-closed behavior.
- The new admin routes are all tested by calling handlers directly, so `handleRequest`-level enforcement (Access audience + CSRF headers) for these specific paths is untested at this layer.

~ 🤖 Fable

PR #2765, opened 2026-08-28
feat(registry): host release artifacts as publisher blobs

ascorbic:
## What does this PR do?

Makes the publisher's PDS the default host for sandboxed plugin release artifacts. `emdash-plugin publish` now builds and validates the bundle, uploads the gzip package and declared PNG/JPEG/WebP listing images through `com.atproto.repo.uploadBlob`, and writes CID-bound checksums into the release record. `--url` remains available for externally hosted package bundles; the HTTP PUT `--artifact-base-url` flow is removed.

The release Lexicon accepts blob or URL artifact sources, applies slot-specific MIME and size constraints, and changes `auth` to an open union. Shared verification rejects source-less artifacts, blob/checksum mismatches, gated artifacts without a supported auth method, unsafe fetch targets, and corrupt bytes. Resolution tries typed record-scoped artifact caches, the publisher PDS, then the explicit URL.

Aggregator release envelopes now expose typed `artifactCaches` instead of `mirrors`. Raw and image cache routes bind both the exact release-record CID and blob CID. The admin image proxy uses Cumulus image presets, while package installation and labeler assessment use checksum-verifiable raw bytes.

Related work:

- RFC 0001: https://github.com/emdash-cms/emdash/pull/694
- RFC 0002: https://github.com/emdash-cms/emdash/pull/1870
- Cumulus exact-record route: https://github.com/ascorbic/cumulus/issues/2

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable). No new admin UI strings are introduced.
- [x] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/296

## Screenshots / test output

No visual UI changes.

Verified locally:

- `pnpm typecheck`
- `pnpm lint:json | jq '.diagnostics | length'` → `0`
- Plugin CLI: 21 files, 397 tests
- Registry client: 5 files, 82 tests
- Registry Lexicons: 2 files, 30 tests
- Registry verification: 7 files, 105 Node tests; 3 files, 37 Workerd tests; packed-output check
- Core registry handlers/proxy: 2 files, 38 tests
- Aggregator read API: 36 tests under Workerd
- Labeler canonical assessment: 11 tests under Workerd
- Admin registry artifact extraction: 13 tests
- `pnpm --dir docs build`


ascorbic:
## 🔍 Adversarial review — stack #2766, layer 1 of 8

_Automated deep review of this layer's diff (`origin/main...feat/blob-hosted-registry-artifacts`). Findings ranked most-severe first; confidence is **CONFIRMED** (full code path read) or **PLAUSIBLE** (strong suspicion, path partially read). Nothing was auto-fixed._

The PR moves plugin release artifacts from publisher-hosted URLs to atproto blobs on the publisher's PDS. Verified against code: `emdash-plugin publish` now builds the bundle itself, checks OAuth blob scopes, uploads gzip + PNG/JPEG/WebP images via `uploadBlob` (with client-side caps 256 KiB / 1 MiB and blob-CID-vs-bytes cross-checks), and writes CID-bound checksums; `--url` remains and `--artifact-base-url` is gone. A new `fetchReleaseArtifact` in registry-verification tries record-scoped caches → publisher PDS (PLC/did:web resolution, https-only, no creds) → declared URL, sha256-verifying fetched bytes against the checksum at every raw source, and derives/enforces checksum-from-blob-CID equality before any fetch. Record-level verification (`verifyPackageReleaseRecords`) rejects source-less artifacts, blob/checksum mismatches, and fails closed on any `auth` value or `requiresAuth` artifact (the open union with zero variants means every auth is unknown). The core install/update handlers delegate fetching to this machinery over the existing DNS-pinned SSRF transport; decompression stays capped (256 KiB / 32 entries). The aggregator envelope renames `mirrors` → required `artifactCaches` and always advertises `https://cdn.em-da.sh`. Overall quality is high — the checksum/CID binding, fail-closed auth, SSRF layering, and CLI validation are genuinely thorough — but the admin image proxy quietly drops its byte-integrity guarantee for blob-carrying artifacts, and there are several compat sharp edges.

### Findings

- **[high]** `packages/core/src/astro/routes/api/admin/plugins/registry/artifact.ts:305,415-439` — Checksum verification of fetched bytes is skipped for *any* artifact that carries a blob ref (`!descriptor.blobCid && !(await verifyChecksum(...))`), including when the bytes come from the publisher's arbitrary external `url`. Scenario: a malicious publisher declares an icon with a self-consistent blob+checksum *plus* a `url` pointing at their server; against an aggregator that advertises no `artifactCaches` (any honest aggregator not running a Cumulus deployment — the field is unsigned operational data), `resolveArtifact` leaves `descriptor.url` as the external URL, and the proxy serves whatever PNG/JPEG/WebP bytes come back — different per request, never matching the approved record — with no `ARTIFACT_CHECKSUM_MISMATCH`. The skip is justifiable only for cache-transformed variants; for raw URL fetches verification was possible and enforced before this PR. Residual mitigations: content-type allowlist, CSP sandbox/nosniff/attachment, admin-auth, no-store. **CONFIRMED.**
- **[medium]** `packages/plugin-cli/src/commands/publish.ts:292-305` + `src/oauth.ts:96,620-627` — Publishers whose AS rejects granular scopes are permanently bricked with a misleading loop: login falls back to `atproto transition:generic` (which *does* permit blob uploads), `missingBlobScopes` then reports both blob scopes missing, and the remediation ("logout, login again") reproduces the same legacy grant forever. Any manifest with listing images (even with `--url`) hits this. **CONFIRMED** code path.
- **[medium]** `packages/registry-lexicons/src/generated/.../aggregator/defs.ts` (`artifactCaches` non-optional) + @atcute/client output validation — atcute runtime-validates XRPC outputs (`ClientValidationError('output')`, surfaced as `AGGREGATOR_RESPONSE_INVALID` in `handlers/registry.ts:996-1002`), so a CMS on this release pointed at a not-yet-upgraded aggregator (returns `mirrors`, no `artifactCaches`) fails *every* release lookup — browse/install/update all hard-broken rather than degrading. Making the field optional client-side would have given a graceful window. The changeset notes the rename but not the lockstep/hard-fail. **CONFIRMED** (test fixtures had to add `artifactCaches: []` to keep passing).
- **[medium]** `packages/registry-verification/src/records.ts:387-392`, `src/artifact.ts` (`unsupportedAuth`) → `packages/core/src/api/handlers/registry.ts:1037/1523` — Publisher-controlled `auth.hint` is used verbatim, unbounded, as the error *message* (`Artifact retrieval failed (AUTH_METHOD_UNSUPPORTED): <hint>`) which the install/update handlers return to the admin client — an attacker-text injection channel into trusted UI error surfaces; `details.hintUrl` is never scheme-validated (a `javascript:` URL passes through the report contract; nothing renders it at this layer, but later layers inherit it unvalidated). **CONFIRMED.**
- **[medium-low]** lexicon blob constraints vs runtime — `maxSize`/`accept` validate only the publisher-*declared* `size`/`mimeType` fields (atcute `blobSize`/`blobAccept`); fetched bytes are never compared to the declared size. A publisher declaring `size: 6` on a 500 KiB gzip blob installs fine (real cap is the generic 512 KiB `MAX_ARTIFACT_BYTES`); images pass up to the proxy's 5 MiB despite the lexicon's 1 MiB. Bounded, but the "slot-specific size constraints" are effectively honest-CLI-only. **CONFIRMED.**
- **[low]** `artifact.ts` route (`resolveArtifact`) — Blob-backed images have no publisher-PDS fallback: blob-only artifact + cache-less aggregator → 404; cache advertised but down → 502 (the declared `url`, if any, is overwritten by the cache URL). Install path falls back to the PDS; the image proxy never does. **CONFIRMED.**
- **[low]** `packages/registry-verification/src/records.ts:324-360` — `validateArtifacts`/`unsupportedAuth` inspect only the four known slots; `x-`-prefixed custom artifacts (explicitly allowed by the lexicon) pass record verification with missing sources, mismatched blob CIDs, or `requiresAuth` set. Nothing fetches them today. **CONFIRMED** at this layer.
- **[low]** `handlers/registry.ts` `fetchArtifact` + registry-verification `fetch.ts` — Dev regression: install/update artifact fetches now go through `fetchVerifiedResource`, which rejects `http:` and localhost/IP-literal hosts unconditionally; the old path allowed `http://localhost` artifacts in dev (`assertSafeArtifactUrl` DEV branch, still used only by the image proxy). Local E2E installs from a localhost-hosted tarball/PDS break. **CONFIRMED.**
- **[low]** `packages/plugin-cli/src/commands/publish.ts` — `--artifact-base-url` was removed without an explicit rejection; citty does not error on unknown flags, so old publish scripts silently switch to PDS-hosted images instead of failing loudly (the changeset/docs do cover the migration). **PLAUSIBLE** (citty default behavior not exercised in-repo).

### PR description vs code

- "slot-specific MIME and size constraints" — enforced against self-declared blob metadata and in the CLI only; no verifier compares fetched bytes to the declared blob size or MIME (the proxy checks the *response* content-type, the install path never checks MIME).
- "Resolution order: caches → publisher PDS → explicit URL" — true for install/update only; the admin image proxy skips the PDS entirely and drops the URL fallback once a cache is advertised.
- "Raw and image cache routes bind both the exact release-record CID and blob CID" — the binding lives only in the request URL; for image routes the response bytes are accepted unverified (finding 1).
- The lexicon checksum doc says clients SHOULD support sha2-512 and blake3; the implementation supports sha2-256 only and rejects the rest (correctly fail-closed, but the SHOULD is unimplemented).
- Changeset says nothing about old EmDash sites being unable to install blob-only releases ("Release record is missing artifact url or checksum" on pre-upgrade cores), nor about the client/aggregator lockstep hard-fail.

### Test-coverage gaps

- No test that a cache returning 200 with *wrong bytes* falls through to the PDS — only HTTP-503 fallthrough is pinned; the poisoned-cache-with-content case is the security-relevant one.
- Nothing pins the proxy's checksum-skip semantics for blob artifacts, and the blob+url/no-cache unverified path (finding 1) is untested — proxy fixtures always advertise the Cumulus cache.
- No oversized-artifact test through `fetchReleaseArtifact` (maxBytes), and no declared-blob-size-vs-actual-bytes test anywhere.
- The aggregator test "does not represent the cache descriptor as admission for a gated blob" asserts the descriptor *is* advertised for a `requiresAuth` blob — the name claims the opposite of its assertion; it pins nothing.
- No fetch-layer test for `requiresAuth: true` with `auth` absent (only the combined case), and no test that a `transition:generic` grant can(not) publish — which would have caught finding 2.

~ 🤖 Fable

PR #2768, opened 2026-08-28
fix(core): keep Cloudflare dev maintenance running

### LINKED ISSUE #2767: Cloudflare `astro dev` cancels the selected NodeCronScheduler timer after the response
Thanks for fixing #1635 in #1636 — I confirmed that 0.35.0 now generates a non-null `createScheduler` factory under `@astrojs/cloudflare` during `astro dev`. I think there is one more Cloudflare dev lifecycle edge after that fix.

## Environment

- `emdash` 0.35.0
- `@emdash-cms/cloudflare` 0.35.0
- `astro` 7.1.3
- `@astrojs/cloudflare` 14.1.3
- `wrangler` 4.112.0
- Node 22.23.2
- Linux

## Reproduction

1. Mount a native plugin that declares a unique plugin-storage index.
2. Run the site with the Cloudflare adapter via:
   ```sh
   astro dev --host 127.0.0.1 --port 4323
   ```
3. Hit an EmDash route so the runtime initializes.
4. Wait longer than the Node scheduler's 60-second maximum interval.
5. Inspect `options['system:scheduler:last_completed_at']` and the declared index in `sqlite_master`.

## Observed

The generated virtual module is correct:

```json
{"createScheduler":"function"}
```

Instrumenting the real scheduler without changing its timing shows:

```text
[proof:scheduler] start
[proof:scheduler] arm
[proof:scheduler] next due resolved null
[proof:scheduler] timer scheduled 60000
```

After more than 65 seconds there is no `executeTick()`, no scheduler heartbeat, and no declared plugin index.

A small control isolates the lifecycle boundary:

- If the response remains pending, the same generated scheduler factory fires normally (`next-due → tick → recover → cleanup`).
- If a route schedules a raw one-second `setTimeout` and immediately returns, that detached callback never fires.
- In this environment `setTimeout` returns a number, so the scheduler's Node-only `unref()` branch is not involved.

That suggests the Cloudflare Astro dev runtime treats these timers as request-scoped. Selecting `NodeCronScheduler` fixes the null-factory problem from #1635, but its detached 60-second callback still cannot drive maintenance after the initializing request completes.

## Expected

Local Cloudflare dev should run the EmDash maintenance path it selects, so the scheduler heartbeat and plugin-declared indexes appear within 60 seconds (matching the documented automatic index provisioning behavior).

## Actual

The scheduler arms successfully but its detached timer never executes. Plugin storage therefore remains fail-closed because its declared unique indexes are never materialized.

Current `main` still appears to select `NodeCronScheduler` for Cloudflare `serve`: https://github.com/emdash-cms/emdash/blob/688257ff63f761fd07531a91233f0c49c3bb1c6b/packages/core/src/astro/integration/virtual-modules.ts#L525-L558

Plugin storage docs: https://docs.emdashcms.com/plugins/creating-plugins/storage/

Happy to test a patch or provide a smaller fixture if useful.

saariuslystoned:
## What does this PR do?

Fixes Cloudflare `astro dev` maintenance stopping after the request that initializes EmDash.

This is separate from #1635/#1636. Those changes fixed scheduler-factory generation during Cloudflare development. With the compatible latest stable stack, the generated `NodeCronScheduler` still arms successfully, but workerd cancels its detached timer when the response completes.

During Cloudflare development, this change keeps the cadence in Astro's long-lived Node/Vite host. Each tick uses Vite's resolved server origin to POST to an EmDash-owned route that is injected only for Cloudflare `astro dev`; the route runs `runScheduledTasks()` directly inside workerd. It does not fabricate a Cron event or invoke the application's `scheduled()` handler, so custom `generalCron` values and unrelated scheduled jobs are unaffected.

Node and Bun scheduling are unchanged. Production Cloudflare continues to use its configured Cron Trigger and Worker `scheduled()` handler.

The lifecycle control was reproduced without EmDash or private application code on Node 22.23.2, Astro 7.2.9, `@astrojs/cloudflare` 14.2.5, and Wrangler 4.127.0:

- A raw one-second timer did not fire after an immediate response.
- The same timer fired while the request remained pending.
- Wrangler's local scheduled-event endpoint invoked the Worker's `scheduled()` handler.

Closes #2767

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (not applicable; no admin UI strings changed)
- [ ] New features link to an approved Discussion: not applicable; this is a bug fix

- [x] This PR includes AI-generated code — model/tool: OpenAI Codex; OpenClaw autoreview (`gpt-5.6-sol`)

## Screenshots / test output

```text
Focused repair coverage:       43 passed
Full core suite:            6,169 passed, 9 skipped
Cloudflare package suite:      386 passed, 2 skipped
pnpm typecheck:                passed across 34 packages
pnpm lint:                     passed
pnpm format:check:             passed
pnpm --filter emdash build:    passed
```

Four CLI harness checks in the full core run could not spawn `/bin/sh` in the managed sandbox (`EPERM`); all other core tests passed.

OpenClaw reviewed exact head `8ef91383b64ac342b7f0e1dc854680ed741d847a` against base `688257ff63f761fd07531a91233f0c49c3bb1c6b`: completed, clean, 0 findings.

### DIFF
--- a/packages/core/src/astro/integration/cloudflare-dev-scheduler.ts
+++ b/packages/core/src/astro/integration/cloudflare-dev-scheduler.ts
@@ -1,0 +1,68 @@
+import type { Server } from "node:http";
+
+import type { AstroIntegrationLogger } from "astro";
+
+const DEFAULT_INTERVAL_MS = 60_000;
+const GENERAL_CRON = "* * * * *";
+
+interface DevServer {
+	httpServer: Pick<Server, "address" | "once"> | null;
+}
+
+interface SchedulerOptions {
+	intervalMs?: number;
+	fetch?: typeof fetch;
+}
+
+export function startCloudflareDevScheduler(
+	server: DevServer,
+	logger: Pick<AstroIntegrationLogger, "warn">,
+	options: SchedulerOptions = {},
+): void {
+	const httpServer = server.httpServer;
+	if (!httpServer) return;
+
+	const intervalMs = options.intervalMs ?? DEFAULT_INTERVAL_MS;
+	const fetchScheduled = options.fetch ?? fetch;
+	let stopped = false;
+	let timer: ReturnType<typeof setTimeout> | undefined;
+
+	const schedule = () => {
+		if (stopped) return;
+		timer = setTimeout(() => void run(), intervalMs);
+		if (typeof timer === "object" && "unref" in timer) timer.unref();
+	};
+
+	const run = async () => {
+		try {
+			const address = httpServer.address();
+			if (!address || typeof address === "string") {
+				logger.warn("Cloudflare dev scheduler could not resolve the local server port.");
+				return;
+			}
+
+			const url = new URL("/cdn-cgi/handler/scheduled", `http://localhost:${address.port}`);
+			url.searchParams.set("cron", GENERAL_CRON);
+			url.searchParams.set("format", "json");
+			const response = await fetchScheduled(url);
+			if (!response.ok) {
+				logger.warn(
+					`Cloudflare dev scheduler request failed with status ${response.status}. ` +
+						"Verify that the Worker entrypoint exports its scheduled handler.",
+				);
+			}
+		} catch (error) {
+			logger.warn(
+				`Cloudflare dev scheduler request failed: ${error instanceof Error ? error.message : String(error)}`,
+			);
+		} finally {
+			schedule();
+		}
+	};
+
+	httpServer.once("listening", schedule);
+	httpServer.once("close", () => {
+		stopped = true;
+		if (timer !== undefined) clearTimeout(timer);
+	});
+}
--- a/packages/core/src/astro/integration/cloudflare-dev-scheduler.ts
+++ b/packages/core/src/astro/integration/cloudflare-dev-scheduler.ts
@@ -3,10 +3,10 @@ import type { Server } from "node:http";
 import type { AstroIntegrationLogger } from "astro";
 
 const DEFAULT_INTERVAL_MS = 60_000;
-const GENERAL_CRON = "* * * * *";
 
 interface DevServer {
-	httpServer: Pick<Server, "address" | "once"> | null;
+	httpServer: Pick<Server, "once"> | null;
+	resolvedUrls: { local: string[]; network: string[] } | null;
 }
 
 interface SchedulerOptions {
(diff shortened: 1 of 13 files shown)


packages/core/src/astro/integration/cloudflare-dev-scheduler.ts
```diff
@@ -0,0 +1,68 @@
...
+
+	const run = async () => {
+		try {
+			const address = httpServer.address();
+			if (!address || typeof address === "string") {
+				logger.warn("Cloudflare dev scheduler could not resolve the local server port.");
+				return;
+			}
+
+			const url = new URL("/cdn-cgi/handler/scheduled", `http://localhost:${address.port}`);
```
ascorbic: You can't hard-code localhost here. It might be set to something else: the host is configurable, and TLS can be enabled in Vite. Please use Vite’s resolved server origin instead of reconstructing one from the port.

packages/core/src/astro/integration/cloudflare-dev-scheduler.ts
```diff
@@ -0,0 +1,68 @@
...
+			url.searchParams.set("cron", GENERAL_CRON);
```
ascorbic: This cannot safely use the default `* * * * *` with the project’s application-wide `scheduled()` handler. A supported custom setup such as `createScheduledHandler({ generalCron: "*/5 * * * *" })` rejects every one of these events before `runScheduledTasks()`. Conversely, a custom Worker that dispatches another job on `* * * * *` will run that job every minute merely because EmDash is installed. Could the host timer call an EmDash-owned, dev-only bridge that invokes `runScheduledTasks()` inside workerd instead of fabricating a project-wide Cron event?

ascorbic (changes requested):
Thanks. It'll be good to fix this. It just needs a couple of changes before it can be merged

packages/core/src/astro/integration/cloudflare-dev-scheduler.ts
saariuslystoned: Fixed in `8ef91383`. The host now builds the bridge URL from Vite's resolved server origin (`server.resolvedUrls.local[0]`, with the resolved network origin as fallback) instead of reconstructing localhost from the port. Focused coverage preserves `https://dev.example.test:7443` end to end and confirms the request uses that origin. The final focused set passes 43/43, and exact-head OpenClaw review is clean with 0 findings.

packages/core/src/astro/integration/cloudflare-dev-scheduler.ts
saariuslystoned: Fixed in `8ef91383`. I removed the fabricated `* * * * *` event and the `/cdn-cgi/handler/scheduled` dispatch. Cloudflare `astro dev` now conditionally injects an EmDash-owned POST route whose handler directly awaits `runScheduledTasks()` inside workerd. Focused coverage proves EmDash maintenance runs while the application-wide scheduled handler is not invoked, and proves the route is absent outside the Cloudflare-dev path. The final focused set passes 43/43; exact-head OpenClaw review is clean with 0 findings.

PR #2770, opened 2026-08-28
fix(db): survive an idle-client error and bound the Postgres connect wait

### LINKED ISSUE #2752: `createDialect` builds a `pg.Pool` with no connect or idle timeout and no `error` listener, so a database failover can hang requests and crash the process
### Summary

The Postgres dialect creates its pool with five of node-postgres's settings and
omits two that matter in a hosted environment, plus the `error` listener that
node-postgres's own documentation says a pool needs.

Two consequences, one slow and one abrupt:

- With no `connectionTimeoutMillis`, a request that cannot reach the database
  waits on the operating system's TCP timeout rather than on anything the
  application chose. That is minutes on a default Linux configuration, on a
  request the user is watching.
- With no `error` handler on the pool, an error raised on an **idle** pooled
  client — which is what a managed-database failover produces — is an
  unhandled `EventEmitter` `'error'` event. Node's documented behaviour for
  that is to throw, and in a server process that means the process exits.

The second one is the sharp edge. node-postgres removes the bad client from the
pool by itself, so the pool recovers; the only thing standing between a routine
failover and a crashed process is a listener that is never attached.

Verified against `emdash@0.32.0`, which we run, and re-checked unchanged in
`0.34.0` and `0.35.0`, the current `latest`. Also present in the deprecated
`1.0.0` published in April, so this is long-standing rather than recent.

### Where

`src/db/postgres.ts`, `createDialect`, `0.35.0` verbatim:

```ts
export function createDialect(config: PostgresConfig): PostgresDialect {
    const pool = new Pool({
        connectionString: config.connectionString,
        host: config.host,
        port: config.port,
        database: config.database,
        user: config.user,
        password: config.password,
        ssl: config.ssl,
        min: config.pool?.min ?? 0,
        max: config.pool?.max ?? 10,
    });

    return new FailFastPostgresDialect({ pool });
}
```

`config.pool` already exists and already carries `min` and `max`, so the
configuration seam for the two timeouts is present — the values are simply not
read. No `pool.on("error", …)` is attached anywhere in the file.

node-postgres's documentation is explicit about the listener:

> The pool will emit an `error` event on behalf of any idle clients it contains
> if a backend error or network partition happens… It is important you add an
> event listener to the pool to catch errors. Just like other event emitters,
> if a pool emits an `error` event and no listeners are added node will emit an
> uncaught error and potentially crash your node process.

### Reproducing

The idle-client case, without needing a real failover:

1. Start EmDash against a Postgres instance and serve one request, so the pool
   holds at least one idle client.
2. Terminate that client from the server side —
   `SELECT pg_terminate_backend(pid) FROM pg_stat_activity WHERE application_name = …`,
   or restart the database.
3. The pool emits `error` with no listener attached.

The connect-timeout case:

1. Point the configuration at a host that drops packets rather than refusing
   them — a firewalled address, not a closed port; a closed port fails fast and
   does not show the problem.
2. Request a page. The request hangs until the operating system gives up.

### Why it matters

Both cases are ordinary operations rather than exotic failures. A managed
Postgres service fails over during routine maintenance, and the application
gets exactly the idle-client error above. A network partition between an
application subnet and a database subnet is a firewall rule away.

The failure modes are also poorly shaped for diagnosis. A crash from an
unhandled emitter error carries a stack from inside the pool, with nothing
naming the database or the failover. A hung request looks like a slow page.

### How we work around it

We patch `createDialect` to pass `config.pool?.connectionTimeoutMillis` and
`config.pool?.idleTimeoutMillis` through — `undefined` deliberately, when unset,
so node-postgres's own defaults are preserved and the change is inert for anyone
who has not configured them — and to attach an `error` listener that logs a
structured line and lets the pool self-heal.

The patch is about fifteen lines and re-carried on every upgrade. It would be
better as a default.

### Suggested fix

1. Read `connectionTimeoutMillis` and `idleTimeoutMillis` from `config.pool`,
   passing them through untouched so an unset value keeps node-postgres's
   default. Same shape as `min` and `max` already use.
2. Attach a `pool.on("error", …)` that logs and does nothing else. The pool
   already discards the failed client; the listener exists so its absence does
   not end the process.
3. Consider whether (2) wants to be an option at all. We cannot construct a case
   where crashing on an idle-client error is the behaviour someone wants, so we
   would suggest it is simply always attached.

Happy to send a PR if you would like a particular shape — particularly on
whether the log line should go through an existing logger rather than
`console.error`.

virafb:
## What does this PR do?

Fixes the two faults reported in #2752, both in `createDialect` in `packages/core/src/db/postgres.ts`.

**A failover can end the process.** node-postgres emits `error` on the pool for a fault raised on an *idle* client, which is what a managed-database failover produces. No listener is attached anywhere in the file, and an EventEmitter `error` event with no listener makes Node throw — in a server process, that is an exit. The pool itself is fine: pg discards the bad client and carries on. So the listener is the whole fix, and logging is the whole handler.

**A request that cannot reach the database waits on the OS.** With no `connectionTimeoutMillis`, the wait is the operating system's TCP timeout rather than a bound the application chose — minutes on a default Linux configuration, on a request a user is watching.

`config.pool` already carried `min` and `max`, so the configuration seam existed and the values were simply never read. This adds `connectionTimeoutMillis` and `idleTimeoutMillis` alongside them.

**Both stay unset by default**, which leaves node-postgres's own defaults in place. Nobody who has not asked for a timeout sees any change in behaviour.

Closes #2752

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [ ] `pnpm typecheck` passes — see the note below; one **pre-existing** error remains, unrelated to this change
- [x] `pnpm lint` passes
- [ ] User-visible strings in the admin UI are wrapped for translation (if applicable) — not applicable, no admin UI strings
- [ ] New features link to an approved Discussion — not applicable, this is a bug fix

### Note on `pnpm typecheck`

It reports one error in `packages/core/src/utils/slugify.ts`:

```
src/utils/slugify.ts(1,25): error TS2307: Cannot find module '@emdash-cms/admin/slugify' or its corresponding type declarations.
```

That file is not touched by this PR. I checked out pristine `main` (`688257f`) after a full `pnpm install && pnpm build` and ran the identical command: the same single error appears. So it is pre-existing in my environment rather than introduced here, and this change adds no new type errors. Flagging it rather than leaving a reviewer to wonder — happy to be told it is environment-specific on my side.

- [x] This PR includes AI-generated code — model/tool: Claude Opus 5 (Claude Code). Authored and verified with the runs shown below; the submitter is responsible for correctness.

## Screenshots / test output

New tests: `packages/core/tests/unit/db/postgres-pool.test.ts`, five cases covering both timeouts passing through, both staying `undefined` when unset, the existing `min`/`max` defaults surviving, the listener being attached, and an idle-client error being swallowed rather than reaching the process.

**With the fix:**

```
 Test Files  1 passed (1)
      Tests  5 passed (5)
```

**Same tests against unmodified `createDialect`,** to show they fail for the right reason:

```
 × passes connect and idle timeouts through to pg
 × attaches an error listener to the pool
 × swallows an idle-client error instead of letting it reach the process

 Test Files  1 failed (1)
      Tests  3 failed | 2 passed (5)
```

Three of the five fail. The two that still pass assert behaviour that already worked — the existing `min`/`max` defaults, and that an unset option stays absent — so they are guards against this change altering something it should not, not evidence of the bug.

**No regression in the surrounding area** (`tests/unit/db` and `tests/unit/migrations`):

```
 Test Files  10 passed (10)
      Tests  78 passed (78)
```

`oxlint --type-aware --deny-warnings`, `prettier --check` and `oxfmt --check` are all clean on the changed files.

### DIFF
--- a/.changeset/postgres-pool-timeouts-and-error-listener.md
+++ b/.changeset/postgres-pool-timeouts-and-error-listener.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes two Postgres pool faults that could hang requests or end the server process. An error raised on an idle pooled client — what a managed-database failover produces — was an unhandled event, which Node turns into an uncaught error that exits the process; the pool now logs it and keeps running, because node-postgres discards the bad client on its own. The `database` adapter's `pool` option also accepts `connectionTimeoutMillis` and `idleTimeoutMillis`, so a request that cannot reach the database fails on a bound you choose instead of waiting on the operating system's TCP timeout. Both are unset by default and node-postgres's own defaults still apply.
--- a/packages/core/src/db/adapters.ts
+++ b/packages/core/src/db/adapters.ts
@@ -199,7 +199,21 @@ export interface PostgresConfig {
 	user?: string;
 	password?: string;
 	ssl?: boolean;
-	pool?: { min?: number; max?: number };
+	pool?: {
+		min?: number;
+		max?: number;
+		/**
+		 * How long to wait for a connection before failing, in milliseconds.
+		 * Left unset, a request that cannot reach the database waits on the
+		 * operating system's TCP timeout instead. Passed through to `pg`.
+		 */
+		connectionTimeoutMillis?: number;
+		/**
+		 * How long an idle client stays in the pool before being closed, in
+		 * milliseconds. Passed through to `pg`.
+		 */
+		idleTimeoutMillis?: number;
+	};
 	migrationConnectionStringEnv?: string;
 }
 
--- a/packages/core/src/db/postgres.ts
+++ b/packages/core/src/db/postgres.ts
@@ -25,6 +25,19 @@ export function createDialect(config: PostgresConfig): PostgresDialect {
 		ssl: config.ssl,
 		min: config.pool?.min ?? 0,
 		max: config.pool?.max ?? 10,
+		// Left undefined when unset, which preserves node-postgres's own
+		// defaults rather than imposing new ones.
+		connectionTimeoutMillis: config.pool?.connectionTimeoutMillis,
+		idleTimeoutMillis: config.pool?.idleTimeoutMillis,
+	});
+
+	// node-postgres emits `error` on the pool for a fault raised on an IDLE
+	// client — a managed-database failover is the common cause. That is an
+	// EventEmitter `error` event, so with no listener attached Node throws and
+	// the server process exits. The pool itself recovers: pg discards the bad
+	// client. Logging is therefore the whole handler.
+	pool.on("error", (error) => {
+		console.error("[emdash] idle Postgres client error; the pool discarded it", error);
 	});
 
 	// Fail-fast migration locking instead of Kysely's blocking advisory
--- a/packages/core/tests/unit/db/postgres-pool.test.ts
+++ b/packages/core/tests/unit/db/postgres-pool.test.ts
@@ -1,0 +1,94 @@
+import { describe, expect, it, vi } from "vitest";
+
+const poolInstances: Array<{
+	options: Record<string, unknown>;
+	listeners: Map<string, Array<(...args: unknown[]) => void>>;
+}> = [];
+
+vi.mock("pg", () => ({
+	Pool: class {
+		options: Record<string, unknown>;
+		listeners = new Map<string, Array<(...args: unknown[]) => void>>();
+
+		constructor(options: Record<string, unknown>) {
+			this.options = options;
+			poolInstances.push(this);
+		}
+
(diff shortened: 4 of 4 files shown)


virafb:
Closing this one. We are not able to sign the CLA, so this pull request cannot be merged as it stands.

The fault is described in full in #2752 — why an idle-client error can end the process, and why the connect wait needs a bound. Please take the fix from there if it is useful. No attribution needed.

Thanks for the review, and sorry for the noise.

Issue #2774, opened 2026-08-28
`export-seed` writes log lines to stdout, migrates the database it exports, and emits `$ref` values that `seed` cannot resolve


yet2come:
## Summary

Three independent problems in `emdash export-seed`, found while validating a documented backup procedure by round-tripping `export-seed` → `seed` in an isolated environment. Each is reproducible on its own; together they make the round trip silently lossy.

1. **Log output goes to stdout**, so redirecting the command to a file produces invalid JSON. Exit code is `0` and stderr is empty.
2. **The command runs migrations on the database it is exporting**, so an "export" writes to its source.
3. **`reference` field values are exported as `$ref:<source-row-id>`** while every other entity is exported by seed id, so `seed` cannot resolve them and the reference is silently dropped.

**Affected version:** `emdash@0.30.0`

Related: #2055 covers `export-seed` being local-only and `wrangler d1 export --remote` being blocked by FTS5. This report is about the local `export-seed` → `seed` path itself, which is the fallback that issue leaves standing.

## 1. Log output on stdout corrupts a redirected export

`src/cli/commands/export-seed.ts` (`dist/cli/index.mjs:1223`) prints the resolved path with `consola.info`, which writes to stdout, and the export query triggers a kysely deprecation warning that also reaches stdout.

```console
$ npx emdash export-seed -d ./data.db --with-content > backup.json
$ head -3 backup.json
ℹ Database: /path/to/data.db
orderBy(array) is deprecated, use multiple orderBy calls instead.
{
$ echo $?
0
```

stderr is empty — `2>&1 >/dev/null` produces no output. The JSON payload itself is written with `console.log(output)`, so the log lines and the data share one stream.

The deprecation line comes from `kysely/dist/parser/order-by-parser.js` and is triggered by `orderBy` being called with an array inside the export path, so it appears on every run rather than only in unusual configurations.

The failure surfaces only at restore time:

```console
$ npx emdash seed backup.json -d ./restored.db
 ERROR  Failed to parse seed file: Unexpected token 'ℹ', "ℹ Database"... is not valid JSON
```

**Expected:** informational output goes to stderr (or is suppressed when stdout is not a TTY), so that `> file` yields parseable JSON. As a smaller point, the internal `orderBy(array)` call could be updated so the deprecation warning stops firing.

## 2. `export-seed` runs migrations on the database it exports

```js
// dist/cli/index.mjs:1221-1231
async run({ args }) {
[...]

PR #2776, opened 2026-08-28
fix(cli): keep export-seed's stdout parseable and its references resolvable

yet2come:
## What does this PR do?

Fixes two of the three problems reported in #2774 — the two that are unambiguous bugs with one obvious correct behaviour. The third (`export-seed` running migrations against the database it is exporting) is deliberately left out: fixing it means choosing between failing on an out-of-date schema and adding an opt-in flag, and that call is yours to make.

Refs #2774

### 1. stdout carried more than the seed document

`emdash export-seed > seed.json` produced a file that is not JSON. The document is written with `console.log`, but the resolved database path went to the same stream via `consola.info`, and `orderBy(["name", "locale"])` made kysely print its array-form deprecation notice there too. The command still exited `0` with an empty stderr, so nothing signalled the corruption until `emdash seed` rejected the file at restore time.

The path now goes to stderr — the same thing `emdash migrate --json`, `auth` and `secrets` already do — and the ordering calls are chained so kysely stops warning at all, rather than warning somewhere quieter.

### 3. `reference` values were exported as source row ids

Every other entity in the file is keyed by a derived seed id (`groups:soumu`), but reference *values* kept the source database's ULID. `emdash seed` assigns fresh ids on insert, so the lookup misses and the literal `$ref:01M0…` string is written into the column. The restore reports success; the reference is simply gone wherever it was rendered.

A reference is now named by the seed id this export assigns its target. Two things follow:

- The row-id → seed-id map has to be complete before any entry's data is converted, because a reference can point into a collection the loop has not reached yet. Reading and converting are therefore separate passes. This re-indentation is most of the diff — `git diff -w` is much smaller than the line count suggests.
- Emitting the right id is not sufficient on its own. `applySeed` fills its seed-id map as it walks the file, so a `$ref` into a collection further down still resolves to nothing. Collections are now ordered by their reference targets, depth-first, with the existing order as the tie-break. A cycle has no valid order, so those collections stay where they were rather than failing the export.

A reference whose target sits outside the export (`--with-content events` when the target lives in `groups`) keeps the row id. Nothing in the file can resolve it either way, and minting a seed id for an entry that is not in the file would only make the dangling reference harder to spot.

### Not addressed here

- Problem 2 in the issue: `export-seed` runs `runMigrations` on the database it exports.
- `applySeed` resolving `$ref` in a single forward pass. Ordering the export sidesteps it for exported files, but a hand-written seed with a forward reference still silently stores the literal string, and [the seed docs](https://docs.emdashcms.com/themes/seed-files/) don't mention the ordering constraint. Happy to file that separately if you want it tightened.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

Two changesets, one per problem, so each lands as its own CHANGELOG entry. No admin-UI strings are touched.

- [x] This PR includes AI-generated code — model/tool: Claude Opus 5 via Claude Code

## Screenshots / test output

Both tests were written first and watched fail against `main`.

`tests/integration/cli/export-seed-stdout.test.ts` runs the built binary, because the stream a value lands on is the thing under test and only a real process has real streams to separate. Against `main`:

```
 × redirects to a file that parses as JSON
   → "SyntaxError: Unexpected token 'o', \"orderBy(ar\"... is not valid JSON"
 × reports the database path on stderr, not stdout
   → expected '' to contain '…/data.db'
 × keeps kysely's orderBy deprecation notice off stdout
```

`tests/unit/seed/export-seed-references.test.ts` against `main`:

```
 × exports a reference as the target's seed id, not its row id
   Expected: "$ref:groups:soumu"
   Received: "$ref:01M1524FNY8AWFKQ8ZHEHYB1VA"
 × resolves the reference to the restored row on export → apply
   Expected: "01M1524FRFF0590BC2Z9V1FY88"
   Received: "$ref:01M1524FQ8JZ8S5JM9ZN2DF66J"
```

That second failure is the whole defect in one line: the restored row holds the string, not a reference.

The issue's own reproduction, after the change:

```console
$ npx emdash export-seed -d ./data.db --with-content all > backup.json
$ head -1 backup.json
{
$ node -e "JSON.parse(require('fs').readFileSync('backup.json','utf8'))" && echo ok
ok
```

Full runs on this branch:

```
pnpm typecheck                                          ✓
pnpm lint                                               ✓
pnpm format:check                                       ✓
pnpm --filter emdash exec vitest run                     6171 passed | 9 skipped
pnpm --filter emdash exec vitest run --config vitest.integration.config.ts
                                                         100 passed
```

### DIFF
--- a/.changeset/export-seed-resolvable-refs.md
+++ b/.changeset/export-seed-resolvable-refs.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes `emdash export-seed --with-content` so `reference` field values survive a round trip through `emdash seed`. The export now names a reference's target by the seed id it assigns that entry, and writes a referenced collection before the collection pointing at it. Previously the export emitted the source database's row id, which the restored database does not carry: the literal `$ref:<row-id>` string was stored in the column, the restore reported success, and the reference was lost wherever it was rendered.
--- a/.changeset/export-seed-stdout-only-json.md
+++ b/.changeset/export-seed-stdout-only-json.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes `emdash export-seed` so its progress line goes to stderr and kysely's `orderBy` deprecation notice is no longer triggered, allowing `emdash export-seed > seed.json` to write a file that parses as JSON. Previously the redirected file began with `ℹ Database: …` and `orderBy(array) is deprecated…`, the command still exited `0` with an empty stderr, and the corruption surfaced only when `emdash seed` rejected the file at restore time.
--- a/packages/core/src/cli/commands/export-seed.ts
+++ b/packages/core/src/cli/commands/export-seed.ts
@@ -75,7 +75,10 @@ export const exportSeedCommand = defineCommand({
 
 		// Connect to database
 		const dbPath = resolve(cwd, args.database);
-		consola.info(`Database: ${dbPath}`);
+		// The seed document is this command's stdout payload, so anything else
+		// written there corrupts `emdash export-seed > seed.json`. Diagnostics
+		// go to stderr, where a redirect leaves them visible.
+		process.stderr.write(`Database: ${dbPath}\n`);
 
 		const db = createDatabase({ url: `file:${dbPath}` });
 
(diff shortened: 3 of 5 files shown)


yet2come:

ascorbic (approved):
Thanks!

PR #2796, opened 2026-08-30
chore: gitignore .claude/worktrees/ so an agent worktree is not committable

runnane:
Adds `.claude/worktrees/` to `.gitignore`.

Claude Code's built-in `EnterWorktree` creates agent worktrees at `<repo>/.claude/worktrees/` and its only setting is `worktree.baseRef` — there is **no base-path setting**. Without this ignore line such a worktree is permanently visible to git, and `git add -A` stages it as a gitlink:

```
$ git worktree add .claude/worktrees/probe -b probe
$ git status --porcelain
?? .claude/
$ git add -A
warning: adding embedded git repository: .claude/worktrees/probe
```

Precondition for **RCP-1065**, which loosens the userspace bundle's placement rule so an agent may use that built-in location rather than passing an explicit `path` and hitting a permission-root relocation prompt on every run. Measured 2026-08-30, only 7 of the 26 checkouts on the workstation carried this line.

**Verified:** `git check-ignore -v .claude/worktrees/probe` resolves to the new rule, and `git status --porcelain` is clean with a probe worktree present — it reported `?? .claude/` before.


runnane:
Closing immediately — opened in error against the upstream parent. This was intended for the fork runnane/emdash only; it is local agent-tooling hygiene of no relevance to this project. Apologies for the noise.

PR #2810, opened 2026-08-31
feat(forms): secure private attachments

masonjames:
## Summary
- store native form uploads as private media while preserving public-media defaults
- require editor/admin authorization for private downloads and return non-disclosing 404s
- validate PDF, JPEG, PNG, and MP4 signatures plus per-file/count/total limits
- retain submissions when object cleanup fails so deletion and retention cleanup are retryable

## Security invariants
- anonymous and insufficient-role callers cannot read private attachments
- private responses use `private, no-store` and `X-Content-Type-Options: nosniff`
- private files are excluded from public listing, deduplication, optimization, feeds, and SEO paths

## Verification
- core private-media/routes/plugin tests: 65 passed
- forms typecheck: passed
- forms validation/upload tests: 7 passed
- `git diff --check`: passed

Current upstream `main` has three unrelated registry-client type errors in the full core typecheck; the focused core and forms checks above pass.

### DIFF
--- a/.changeset/private-form-attachments.md
+++ b/.changeset/private-form-attachments.md
@@ -1,0 +1,8 @@
+---
+"emdash": minor
+"@emdash-cms/plugin-forms": minor
+"@emdash-cms/cloudflare": patch
+"@emdash-cms/sandbox-workerd": patch
+---
+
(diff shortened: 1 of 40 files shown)


masonjames:
Closing by project decision. Atmospheric Pro will consume this work from Mason's custom fork while we evaluate a smaller standalone forms plugin/solution; no upstream merge or package publication is requested.

PR #2812, opened 2026-08-31
fix(core): add includeCounts opt-out to getTerm

### LINKED ISSUE #2811: getTerm() has no includeCounts opt-out, so category/tag archive pages run the visible-count aggregate on every render
### Description

#2219 added an `includeCounts` option to `getTaxonomyTerms()` so callers that don't render counts can skip the visible-count aggregate. `getTerm()` never got the same option. It takes the base `TaxonomyQueryOptions` (`packages/core/src/taxonomies/index.ts`), and `loadTerm` always calls `getVisibleTermCounts` -> `fetchVisibleTermCounts` (`packages/core/src/taxonomies/term-counts.ts`). There is no way to fetch one term without aggregating counts for the whole taxonomy.

The shipped archive pages only need the label. `templates/blog`, `templates/starter`, `templates/blog-cloudflare` and `templates/starter-cloudflare` all render `category/[slug].astro` and `tag/[slug].astro` with `getTerm(...)`, display `term.label`/`term.slug`, and show `filteredPosts.length` (not `term.count`). So the aggregate runs and is discarded on every uncached render.

The aggregate (from `term-counts.ts`, one branch per collection):

```sql
SELECT taxonomy_id, SUM(count) AS count FROM (
  SELECT ct.taxonomy_id AS taxonomy_id, COUNT(*) AS count
  FROM content_taxonomies AS ct
  INNER JOIN "ec_posts" AS e ON e.id = ct.entry_id
  WHERE ct.collection = ?
    AND ct.taxonomy_id IN (SELECT translation_group FROM taxonomies WHERE name = ?)
    AND (e.status = 'published' OR (e.status = 'scheduled' AND e.scheduled_at <= ...))
    AND e.deleted_at IS NULL
  GROUP BY ct.taxonomy_id
) AS per_collection GROUP BY taxonomy_id
```

Its cost is O(total assignment rows) per call. #2238 removed the near-quadratic factor, but the aggregate still runs once per term-page render.

Observed on a migrated production site (Cloudflare D1): 1,542 posts, 2,271 taxonomy terms (2,219 tags), 7,736 assignments, so 2,271 tag/category archive URLs. On 0.30, when crawler traffic hit those archive pages cold, this aggregate became the most expensive query on the database by far: ~4,800 executions, 99% of D1 runtime, and ~15.5B rows read over ~3 days. On 0.30 the aggregate had several triggers, and later fixes removed most of them: #2219 stopped the layout prefetch and widgets from demanding counts, and #2238 removed the near-quadratic query plan. The one trigger those fixes did not touch is `getTerm` on the term-archive pages, which still runs the aggregate once per render.

Expected: `getTerm()` should let a caller skip the count, the way `getTaxonomyTerms()` does. Give it the same `includeCounts` option: accept `TaxonomyTermsOptions`, and when `includeCounts` is false load the term row and children without `getVisibleTermCounts` (count omitted), default `true` so existing callers are unchanged. The shipped archive pages then pass `includeCounts: false`. Happy to open a PR with the failing test and the query-count snapshot deltas.

Image

### Steps to reproduce

Using the repo's own query-count harness on `main`:

1. Run `pnpm query-counts --target sqlite --skip-build`.
2. In `scripts/query-counts.queries.sqlite.json`, the `GET /category/development` and `GET /tag/webdev` routes (cold and warm) each record the `per_collection ... GROUP BY taxonomy_id` aggregate once.
3. Those pages — `fixtures/perf-site/src/pages/category/[slug].astro` and `tag/[slug].astro` — call `getTerm(...)` and render only `term.label`/`term.slug`, never `term.count`. So that aggregate is `getTerm`'s, computed and thrown away. Passing `includeCounts: false` on those calls removes it (1 -> 0 on those routes).

On a real site: add a `category/[slug].astro` (or `tag/[slug].astro`) page that calls `getTerm(name, slug)` for the label only, seed a taxonomy with many terms and assignments, and render an uncached `/category/<slug>`. The `SELECT ... SUM(count) ... AS per_collection` aggregate fires on every render, because `loadTerm` always calls `getVisibleTermCounts`.

### Environment

- emdash version: 0.35.0 (behavior present on `main`)
- Node.js version: 24
- Runtime: Cloudflare Workers + D1 (the query is engine-agnostic; it also reproduces on the SQLite query-count harness)
- OS: macOS (dev) / Cloudflare Workers (prod)

### Logs / error output

```shell

```

iNerdStack:
## What does this PR do?

`getTerm()` always computes a term's visible-entry count, and unlike `getTaxonomyTerms()` (#2219) it had no way to opt out: `loadTerm` always ran `getVisibleTermCounts`, an aggregate over the whole `content_taxonomies` pivot for every counted collection. A term-archive page that renders only the label still paid that aggregate on every render.

`getTerm()` now takes the same `includeCounts` option (default `true`, so existing callers are unchanged). With `includeCounts: false` it loads the term row and its children without the aggregate and omits `count`. The term row is cached under the taxonomy epoch, separate from the count, the way `getTaxonomyTerms` already caches its list. The built-in `category/[slug]` and `tag/[slug]` pages in the blog and starter templates render `filteredPosts.length` and never `term.count`, so they pass `includeCounts: false`.

Counts are unchanged for callers that want them.

Closes #2811

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [x] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (targeted taxonomy tests, and the full `emdash` unit suite)
- [x] User-visible strings in the admin UI are wrapped for translation — n/a, no admin UI strings

## Screenshots / test output

```
$ pnpm --filter emdash exec vitest run tests/unit/taxonomies/term-count-demand.test.ts
  Test Files  1 passed (1)
       Tests  7 passed (7)

$ pnpm --filter emdash test
  Test Files  506 passed | 2 skipped (508)
       Tests  6169 passed | 9 skipped (6178)

$ pnpm typecheck   # clean
$ pnpm lint        # 0 diagnostics
$ pnpm format      # applied
```

Reverting only the source change (keeping the new tests) fails the opt-out case on `main`, confirming the tests reproduce the bug:

```
× does not aggregate counts for a getTerm caller that opts out
```


iNerdStack:

iNerdStack:
Removed the template comments in 8d6c3d2. On the snapshots: this PR leaves the perf-site benchmark fixture on the default (counts) path, so the committed snapshots are unchanged — the opt-out saving is covered by the new cases in `term-count-demand.test.ts`.

ascorbic (approved):
Thanks

PR #2818, opened 2026-08-31
fix(plugin-cli): compare the schema, not the checkout's line endings

### LINKED ISSUE #2809: `plugin-cli` schema-drift test fails on any CRLF checkout, and its error message sends you the wrong way
### Summary

`packages/plugin-cli/tests/schema-drift.test.ts` compares the committed
`schemas/emdash-plugin.schema.json` against a freshly generated string **byte-for-byte**. The
generated side is built with `"\n"`; the committed side is read with `readFile(path, "utf8")`.
Neither side is newline-normalised.

The repository ships **no `.gitattributes`**, so a clone on a machine configured with
`core.autocrlf=true` gets CRLF on disk, and this test then fails on a clean checkout with
nothing edited. **The schema's content is not involved in the failure.**

All line numbers, byte counts and commands below are from
`1717d31b351164a5f78e95fe004ee582c7c50f40`.

### Why the comparison cannot succeed — this is determined, not just observed

Three facts, each checkable from the repository alone:

1. **The committed blob contains no CR bytes at all.** `packages/plugin-cli/schemas/emdash-plugin.schema.json`
   is **16,143 bytes with 654 LF and 0 CR** in git's object store. The CRLF never comes from
   the repository — it is introduced by checkout conversion.
2. **A clean clone puts CRLF on disk.** With `core.autocrlf=true` and no `.gitattributes` to
   override it, the same file lands as **16,797 bytes with 654 CRLF and 0 bare LF**, and
   `git status` is clean.
3. **The generated side can never contain a raw CR byte.** `scripts/gen-schema.ts:56` builds
   ``const serialised = `${JSON.stringify(document, null, "\t")}\n`;`` and the test mirrors it.
   `JSON.stringify` emits a CR inside a string value as the two characters `\` and `r`, never as
   a raw `0x0D`, and the only newline the template adds is LF.

So on a CRLF checkout one side has 654 CR bytes and the other has none, and
`committed !== regenerated` is **guaranteed — regardless of what the schema contains, and
regardless of whether it is actually in sync.** The test reports schema drift for a condition
that has nothing to do with schema drift.

**This is not a Windows-specific bug.** Nothing in the mechanism refers to the platform: a
Windows checkout with `core.autocrlf=input` passes, and a Linux checkout that has CRLF on disk
fails. The variable is the line endings in the working copy, not the OS.

### Reproducing from a clean clone

```console
$ git -c core.autocrlf=true clone --depth 1 https://github.com/emdash-cms/emdash.git
$ cd emdash && git log -1 --format=%H
1717d31b351164a5f78e95fe004ee582c7c50f40

$ ls .gitattributes
ls: cannot access '.gitattributes': No such file or directory

$ git status --porcelain          # clean; nothing edited
$                                  

# schema on disk
bytes 16797   CRLF 654   bare LF 0
```

There is **no `.gitattributes` anywhere in the repository** — the full recursive tree at that
commit is 4,760 entries (untruncated) and contains no file of that name at the root or under
`packages/plugin-cli/`. The conversion is repo-wide rather than specific to this file; the test
source itself lands with 60 CRLF.

### The error message points away from the cause

```
schemas/emdash-plugin.schema.json is out of date with the Zod schema.
Run: pnpm --filter @emdash-cms/plugin-cli gen-schema
Then commit the result.
```

The schema is not out of date, and following the instruction leads into a loop where **each
step looks like progress**. Measured on the clean clone above:

1. **The generator rewrites the file with LF** — `gen-schema.ts:59` is
   `await writeFile(OUT_PATH, serialised, "utf8")` with the LF-terminated string from line 56.
   The file goes from 16,797 bytes to 16,143. **The test now passes**, so the message appears to
   have been correct.
2. **`git status` reports the file as modified while `git diff` shows no changed lines at all** —
   `--numstat` and `--stat` are both empty, because git normalises on comparison. Git also
   prints `warning: in the working copy of '...', LF will be replaced by CRLF the next time Git
   touches it`, which is the actual explanation, but it arrives as a warning rather than as the
   test failure's reason. A contributor is looking at a modified file with an empty diff.
3. **The next checkout converts it back.** `git checkout -- <file>` restores 16,797 bytes with
   654 CRLF, and the test fails again.

The same mechanism makes `pnpm build` appear to dirty the working tree: `plugin-cli`'s build
script is `node --run gen-schema && tsdown`, so every build writes LF over a CRLF file.

### Why CI has never caught it

Every job in `.github/workflows/ci.yml` runs on `ubuntu-latest` — **11 of 11, with no Windows
or macOS runner** — and the workflow triggers on pushes to `main` and on pull requests. The
`Tests` job (`pnpm test:unit`) is where `schema-drift` runs, and on a Linux checkout the file is
LF, so it passes. The failure is only visible to contributors whose working copy has CRLF.

### Suggested fix

Add a `.gitattributes` at the repository root:

```gitattributes
*.json text eol=lf
```

or, if a repo-wide rule is too broad:

```gitattributes
packages/plugin-cli/schemas/*.json text eol=lf
```

**This is zero content churn.** On a clean clone of the commit above, adding that
`.gitattributes` and running `git add --renormalize .` stages **no content change to any
tracked file** — the index already holds LF everywhere, so the only thing being changed is the
checkout policy. Existing clones need `git add --renormalize .` or a fresh checkout to pick it
up.

This is preferable to asking contributors to set `core.autocrlf`, because it controls the actual
variable for everyone rather than relying on each contributor's local git configuration.

**Optionally, as defence in depth**, normalise inside the assertion so the test measures schema
drift rather than checkout policy:

```ts
const norm = (s: string) => s.replace(/\r\n/g, "\n");
if (norm(committed) !== norm(regenerated)) { /* ... */ }
```

Both is best: the `.gitattributes` fixes the cause, and the normalisation stops the test
reporting a misleading reason if it ever recurs.

### What was not measured here

- **The test was not executed against this commit.** Doing so needs a full workspace install,
  which was deliberately not run. It *was* observed failing on a CRLF checkout and passing on an
  LF checkout of a fork of this repository. **The argument above does not depend on that
  observation** — it rests on the committed blob's bytes, the checkout's bytes, and the
  generator's source, all of which are stated above and independently checkable.
- **`core.autocrlf=true` being the Git for Windows installer default is general knowledge and is
  not verified here.** What was verified is that it was set at the system level
  (`/etc/gitconfig`) on the machine used, and that a clone made with it produces the CRLF
  working copy described above.

### Not a duplicate

Open issues and pull requests were searched for `gitattributes` (0 results), `crlf` (0), `line
endings` (1, unrelated — #1623), and `schema-drift` (6, all unrelated). Nothing open proposes
this.

### Related, same package — happy to file separately

`packages/plugin-cli/tests/init-scaffold.test.ts` has an unrelated POSIX assumption, still
present at the same commit and visible in the source (these were not run against upstream
either):

- **Line 66** strips a `` `${targetDir}/` `` prefix from the paths `scaffold()` returns, but
  `src/init/scaffold.ts:97` and `:115` build those paths with `join(absDir, file)`. On Windows
  the separator is `\`, so the prefix never matches, nothing is stripped, and the
  `fileSet.has("package.json")` assertions cannot succeed. **The production code is correct here;
  only the test is wrong.**
- **Lines 51 and 57** disagree for the same reason: `beforeEach` builds `targetDir = join(root,
  "plugin")` while `afterEach` recovers the parent with `targetDir.replace(/\/plugin$/, "")`.
  On Windows that regex never matches, so `root` stays equal to `targetDir` and the `mkdtemp`
  parent is left behind once per test — which the comment on line 55 says it is specifically
  trying to avoid.

jmirchandani:
## What does this PR do?

`packages/plugin-cli/tests/schema-drift.test.ts` compares the committed `schemas/emdash-plugin.schema.json` against a freshly generated string byte-for-byte. The generated side is always LF (`JSON.stringify(document, null, "\t") + "\n"`) and the committed side is read with `readFile(path, "utf8")`, so the comparison measures the working copy's line endings as well as the schema. The repository ships no `.gitattributes`, so a clone on a machine configured with `core.autocrlf=true` gets CRLF and this test fails on a clean checkout with nothing edited. The committed blob has no CR bytes at all — checkout conversion introduces them — so the schema's content is not involved in the failure.

This adds a `.gitattributes` pinning `*.json` to LF, which fixes the cause for everyone rather than asking each contributor to set `core.autocrlf`. On a clean clone, `git add --renormalize .` stages no content change to any tracked file: the index already holds LF, so the only thing that changes is what git writes on checkout.

It also normalises newlines inside the assertion, as defence in depth, so the test measures schema drift rather than checkout policy. That matters because of what the current message does to a reader: it reports the schema as out of date and points at `gen-schema`, which rewrites the file with LF so the test passes, leaves `git status` reporting a modified file whose `git diff` is empty, and then fails again on the next checkout. Each step looks like progress. The same mechanism makes `pnpm build` appear to dirty the working tree, since `build` runs `gen-schema`.

The second test is the failing test for this bug. `.gitattributes` alone cannot be covered by one — it affects checkout, not runtime — so the reproduction is the mechanism: a CRLF copy of the committed schema must not be reported as drift. It fails without the normalisation and passes with it.

CI has not caught this because every job in `ci.yml` runs on `ubuntu-latest`, where the checkout is LF.

No changeset: this changes no published package’s behaviour — a `.gitattributes` and a test.

Closes #2809

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] I have added and reviewed the user-facing [changeset](https://github.com/emdash-cms/emdash/blob/main/.changeset/README.md) (if this PR changes a published package)

- [x] This PR includes AI-generated code — model/tool: Claude Opus 5, via Claude Code

## Screenshots / test output

`pnpm typecheck` and `pnpm lint` pass on this branch, and `pnpm format` leaves both changed files untouched. The schema-drift tests pass, and the new one fails without the normalisation:

```console
$ pnpm --filter @emdash-cms/plugin-cli exec vitest run tests/schema-drift.test.ts
 Test Files  1 passed (1)
      Tests  2 passed (2)

# with the normalisation removed from the new test
 × reports drift in the schema, not in the checkout's line endings
 Test Files  1 failed (1)
      Tests  1 failed | 1 passed (2)
```

One unrelated test in this package fails on Windows both with and without this change: `tests/init-scaffold.test.ts` strips a `${targetDir}/` prefix from paths that `scaffold()` builds with `join`, so nothing is stripped when the separator is a backslash. It passes on Linux, where CI runs. That is the second defect described in #2809 and is not touched here.

With this branch applied the loop closes: the schema is checked out as LF, `gen-schema` writes LF, and the tree stays clean.

```console
$ git checkout -- packages/plugin-cli/schemas/emdash-plugin.schema.json
# 16143 bytes on disk, 0 CRLF
$ pnpm --filter @emdash-cms/plugin-cli run gen-schema
$ git status --porcelain packages/plugin-cli/schemas/emdash-plugin.schema.json
$
```

Before the change, measured on a clean clone of `main` at `1717d31b351164a5f78e95fe004ee582c7c50f40`:

```console
$ git -c core.autocrlf=true clone --depth 1 https://github.com/emdash-cms/emdash.git
$ ls .gitattributes
ls: cannot access '.gitattributes': No such file or directory
$ git status --porcelain          # clean; nothing edited
$

# schemas/emdash-plugin.schema.json on disk
bytes 16797   CRLF 654   bare LF 0
# the same blob in git's object store
bytes 16143   LF 654   CR 0
```

Writing the file with LF, which is what the generator emits:

```console
$ git status --porcelain
 M packages/plugin-cli/schemas/emdash-plugin.schema.json
$ git diff --numstat
warning: in the working copy of '...', LF will be replaced by CRLF the next time Git touches it
$ git checkout -- packages/plugin-cli/schemas/emdash-plugin.schema.json
# back to 16797 bytes with 654 CRLF, and the failure returns
```


jmirchandani:

ascorbic (approved):
Thanks

PR #2852, opened 2026-09-02
fix(core): keep the media-usage cleanup scan bounded when the lease is lost

### LINKED ISSUE #2844: Media-usage cleanup lease guard inside WHERE EXISTS turns a LIMITed scan into a full-table scan
### Description

`MediaUsageRepository.findMediaUsageCleanupCandidates` folds the cleanup lease check into the `WHERE` clause of the candidate scan, as a `WHERE ... EXISTS (SELECT 1 FROM _emdash_media_usage_cleanup ...)` alongside a keyset cursor and `LIMIT`:

```sql
select "u"."id", ..., "writer"."expires_at"
from "_emdash_media_usage" as "u"
left join "_emdash_media_usage_sources" as "s" on "s"."source_key" = "u"."source_key"
left join "_emdash_media_usage_generation_writes" as "writer"
  on "writer"."source_key" = "u"."source_key" and "writer"."generation" = "u"."generation"
where "u"."created_at" < ?
  and EXISTS (
    SELECT 1 FROM _emdash_media_usage_cleanup AS cleanup
    WHERE cleanup.task_key = 'projection_gc'
      AND cleanup.lease_token = ?
      AND "cleanup"."lease_expires_at" > strftime('%Y-%m-%dT%H:%M:%fZ', 'now')
  )
  and ("u"."created_at" > ? or ("u"."created_at" = ? and "u"."id" > ?))
order by "u"."created_at" asc, "u"."id" asc
limit ?
```

The lease predicate is constant with respect to `u` — it is either true for every row or false for every row. But because it lives inside the `WHERE` of a `LIMIT`ed join, **when it is false no row can ever satisfy the predicate, the `LIMIT` is never reached, and the query scans the entire `created_at` range through all three joins before returning nothing.**

Measured against a `_emdash_media_usage` table of 166,110 rows, varying only the lease predicate:

| Variant | Rows read | Rows returned |
| --- | ---: | ---: |
| Lease `EXISTS` true | **751** | 250 |
| `EXISTS` clause removed entirely | 501 | 250 |
| Lease `EXISTS` **false** | **496,341** | 0 |

That is a ~660× cost cliff on a query whose entire design intent is to be bounded. The `LIMIT 250` and the keyset cursor provide no protection at all in the failure case, and the cost scales with total table size — so it grows precisely on the installations that need the GC most.

The same lease-guard pattern is applied to the delete statements in this lane (`deleteOrphanOccurrencesOlderThan`, `deleteStaleGenerationsOlderThan`, `deleteAbandonedGenerationsOlderThan`), which look to have the same exposure.

**Suggested fix:** verify the lease in its own statement before issuing the scan, and drop the `EXISTS` from the scan's predicate. A lease that has been lost should short-circuit the sweep in the caller — cheaply, and without touching the data tables — rather than being expressed as a row-level filter the planner has to evaluate inside a scan it cannot terminate early.

Independently: a constant-false guard turning a bounded query into a full-table scan is a sharp edge worth avoiding wherever this pattern appears, not only in this lane.

### Steps to reproduce

1. Deploy an EmDash site on Cloudflare Workers + D1 with a large `_emdash_media_usage` table (~150K+ occurrence rows).
2. Enable a Cron Trigger so the media-usage cleanup lane runs.
3. Run the candidate-scan query with a `lease_token` that does not match the stored lease (or with an expired `lease_expires_at`).
4. Compare `meta.rows_read` against the same query with a matching lease.

The cost difference is visible directly in the D1 GraphQL Analytics API (`d1QueriesAdaptiveGroups`): the same normalized query text reports a few hundred rows read per execution when the lease holds, and hundreds of thousands when it does not.

### Environment

- emdash version: 0.35.0
- `_emdash_media_usage`: 166,110 rows; `_emdash_media_usage_sources`: 49,859 rows

### Logs / error output

```shell
# Lease EXISTS true
rows_read: 751        rows_returned: 250

# Lease EXISTS false — identical SQL, identical cursor, identical LIMIT 250
rows_read: 496341     rows_returned: 0

# Plan is "correct" in both cases; nothing in it signals the cliff
SEARCH u USING COVERING INDEX idx__emdash_media_usage_cleanup_scan (created_at>? AND created_at<?)
SEARCH s USING INDEX sqlite_autoindex__emdash_media_usage_sources_1 (source_key=?) LEFT-JOIN
SEARCH writer USING INDEX sqlite_autoindex__emdash_media_usage_generation_writes_2 (source_key=? AND generation=?) LEFT-JOIN
SEARCH cleanup EXISTS USING COVERING INDEX sqlite_autoindex__emdash_media_usage_cleanup_1 (task_key=?)
```

MA2153:
## What does this PR do?

The media-usage cleanup lane folds its lease check into the `WHERE` of the candidate scan as `EXISTS (SELECT 1 FROM _emdash_media_usage_cleanup ...)`. That predicate is uncorrelated — true for every candidate row or false for every candidate row. Inside a `LIMIT`ed keyset scan a false guard means no row can ever satisfy the predicate, the limit is never reached, and the scan walks the entire `created_at` range through all three joins before returning nothing. On the reporter's 166,110-row `_emdash_media_usage` table that is 496,341 rows read versus 751 for the identical scan with the lease held, and the cost scales with table size, so it grows on exactly the installations that need the GC most.

This PR verifies the lease in its own statement and drops the `EXISTS` from every lease-gated scan:

- `findMediaUsageCleanupCandidates`
- the fallback scans in `deleteOrphanOccurrencesOlderThan`, `deleteStaleGenerationsOlderThan`, `deleteAbandonedGenerationsOlderThan`
- `deleteExpiredGenerationWriteLeases`

The mutating statements keep `activeCleanupLeaseExpression()` unchanged. They are what actually fences a displaced owner, and their `id IN (...)` / `lease_token IN (...)` bound means a constant-false guard costs an indexed probe per batch, not a table scan. On PostgreSQL the new check takes the same `FOR UPDATE` row lock the in-statement guard did, so the lane's lock ordering against source writes is unchanged.

Closes #2844

### Two things reviewers should weigh

1. **`MAX_CLEANUP_STATEMENTS_PER_TICK` moves from 14 to 16.** The check is one single-row read at each of the two scan sites a full tick reaches. That is a deliberate budget increase — two indexed single-row reads per cron tick in exchange for removing the cliff. The alternative that holds the budget at 14 is to drop the lease consultation from the scans entirely: the claim already proves the lease at tick start, the 5s time budget stops a suspended tick, and the mutations re-check. A fenced tick would then still burn one *bounded* (~750-row) scan, which is what the issue asked to avoid, so this PR follows the issue's suggested fix. Happy to switch if you prefer the flat budget.
2. **The check lives in the repository, per call, not only in `cleanupMediaUsage`.** That costs the second statement but means a future caller cannot reintroduce the pattern by forgetting the gate.

The issue's closing note — that this shape is worth avoiding wherever it appears, not only in this lane — is not addressed here. That looks like a separate audit rather than a drive-by.

### Updates from review

- The candidate scan resolves to `MediaUsageCleanupCandidate[] | null`. `[]` means the scan ran and found nothing, which retires the sweep window and clears the cursor; `null` means it was never issued and the caller leaves the cursor alone. Previously a fenced tick returned `[]` and relied on `completeMediaUsageCleanup` being lease-fenced to reject the resulting `sweepComplete: true`.
- Both gates now go through one `admitCleanupScan` helper — budget, lease row, budget again — shared by the candidate scan and the four bounded delete scans. The lease read is a statement, so a tick suspended across it could resume with its 5s budget spent and still issue the scan.
- The PostgreSQL plan test no longer hand-writes the scan SQL (which still carried the removed `EXISTS`). It compiles the statement the repository actually issues and `EXPLAIN`s that, so the two cannot drift again.

`MAX_CLEANUP_STATEMENTS_PER_TICK` is unchanged at 16 — no new statements.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [x] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable). Do not include `messages.po` changes except in translation PRs — a workflow extracts catalogs on merge to `main`. — n/a, no admin UI strings

## Screenshots / test output

New regression test in `media-usage-cleanup.test.ts`. It fails on `main`, where the fenced scan reads all three data tables:

```
AssertionError: expected [ '_emdash_media_usage', …(6) ] to deeply equal [ '_emdash_media_usage_cleanup' ]

- Expected
+ Received

  [
-   "_emdash_media_usage_cleanup",
+   "_emdash_media_usage",
+   "_emdash_media_usage_sources",
+   "s",
+   "u",
+   "_emdash_media_usage_generation_writes",
+   "writer",
+   "cleanup",
  ]
```

With the fix, both cleanup suites pass on SQLite and on a real PostgreSQL 17, so the dialect-specific lock-ordering tests actually executed:

```
$ EMDASH_TEST_PG=... vitest run tests/integration/database/media-usage-cleanup.test.ts \
    tests/integration/database/media-usage-cleanup-plan.test.ts

 Test Files  2 passed (2)
      Tests  58 passed (58)
```

`tests/integration/database` and `tests/unit/database` are green on both dialects apart from two 5s vitest timeouts under parallel load (`media-usage-projection-admission-measurement` and the PostgreSQL plan test); both pass in isolation in ~2s and neither is related to this change.

https://claude.ai/code/session_01RvCS2j4PWyCjtZX4pWLzT5

### DIFF
--- a/.changeset/bounded-media-usage-gc-scan.md
+++ b/.changeset/bounded-media-usage-gc-scan.md
@@ -2,4 +2,4 @@
 "emdash": patch
 ---
 
-Fixes the scheduled media-usage cleanup so a tick that no longer holds the cleanup lease stops immediately instead of scanning the whole occurrence table. On large sites the guarded scan previously read every occurrence in the sweep window — hundreds of thousands of rows on a table its `LIMIT` was meant to keep at a few hundred — and returned nothing. Sites on Cloudflare D1 will see the corresponding rows-read spikes disappear from cron ticks.
+Fixes scheduled media-usage cleanup reading far more rows than its batch size on large sites. A cleanup run that had lost its lease scanned the whole occurrence table before returning nothing, so cron ticks could spike into the hundreds of thousands of rows read. Sites on Cloudflare D1 will see those spikes disappear.
--- a/.changeset/bounded-media-usage-gc-scan.md
+++ b/.changeset/bounded-media-usage-gc-scan.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes the scheduled media-usage cleanup so a tick that no longer holds the cleanup lease stops immediately instead of scanning the whole occurrence table. On large sites the guarded scan previously read every occurrence in the sweep window — hundreds of thousands of rows on a table its `LIMIT` was meant to keep at a few hundred — and returned nothing. Sites on Cloudflare D1 will see the corresponding rows-read spikes disappear from cron ticks.
--- a/packages/core/src/database/repositories/media-usage.ts
+++ b/packages/core/src/database/repositories/media-usage.ts
@@ -1807,6 +1807,8 @@ export class MediaUsageRepository {
 		limit: number;
 		cleanupLease?: MediaUsageCleanupLease;
 	}): Promise<MediaUsageCleanupCandidate[]> {
+		if (input.cleanupLease && !(await this.holdsCleanupLease(input.cleanupLease))) return [];
+
 		let query = this.db
 			.selectFrom("_emdash_media_usage as u")
 			.leftJoin("_emdash_media_usage_sources as s", "s.source_key", "u.source_key")
@@ -1828,9 +1830,6 @@ export class MediaUsageRepository {
 			.orderBy("u.created_at", "asc")
 			.orderBy("u.id", "asc")
 			.limit(Math.max(0, Math.floor(input.limit)));
-		if (input.cleanupLease) {
-			query = query.where(this.activeCleanupLeaseExpression(input.cleanupLease));
-		}
 
 		if (input.cursor) {
 			query = query.where((eb) =>
@@ -1929,8 +1928,9 @@ export class MediaUsageRepository {
 			);
 		}
 		if (!canIssueCleanupStatement(options.canIssueStatement)) return 0;
+		if (options.cleanupLease && !(await this.holdsCleanupLease(options.cleanupLease))) return 0;
 
-		let query = this.db
+		const rows = await this.db
 			.selectFrom("_emdash_media_usage as u")
 			.leftJoin("_emdash_media_usage_sources as s", (join) =>
 				join.onRef("s.source_key", "=", "u.source_key"),
@@ -1946,11 +1946,8 @@ export class MediaUsageRepository {
 			.where(this.noActiveGenerationWriteExpression("u"))
 			.orderBy("u.created_at", "asc")
 			.orderBy("u.id", "asc")
-			.limit(batchLimit);
-		if (options.cleanupLease) {
-			query = query.where(this.activeCleanupLeaseExpression(options.cleanupLease));
-		}
-		const rows = await query.execute();
+			.limit(batchLimit)
+			.execute();
 
 		return this.deleteOrphanCandidateIds(
 			rows.map((row) => row.id),
@@ -1976,8 +1973,9 @@ export class MediaUsageRepository {
 			);
 		}
 		if (!canIssueCleanupStatement(options.canIssueStatement)) return 0;
+		if (options.cleanupLease && !(await this.holdsCleanupLease(options.cleanupLease))) return 0;
 
-		let query = this.db
+		const rows = await this.db
 			.selectFrom("_emdash_media_usage as u")
 			.innerJoin("_emdash_media_usage_sources as s", (join) =>
 				join.onRef("s.source_key", "=", "u.source_key"),
@@ -1994,11 +1992,8 @@ export class MediaUsageRepository {
 			.where(this.noActiveGenerationWriteExpression("u"))
 			.orderBy("u.created_at", "asc")
 			.orderBy("u.id", "asc")
-			.limit(batchLimit);
-		if (options.cleanupLease) {
-			query = query.where(this.activeCleanupLeaseExpression(options.cleanupLease));
-		}
-		const rows = await query.execute();
+			.limit(batchLimit)
+			.execute();
 
 		return this.deleteStaleCandidateIds(
 			rows.map((row) => row.id),
@@ -2024,8 +2019,9 @@ export class MediaUsageRepository {
 			);
 		}
 		if (!canIssueCleanupStatement(options.canIssueStatement)) return 0;
+		if (options.cleanupLease && !(await this.holdsCleanupLease(options.cleanupLease))) return 0;
 
-		let query = this.db
+		const rows = await this.db
 			.selectFrom("_emdash_media_usage as u")
 			.innerJoin("_emdash_media_usage_sources as s", (join) =>
 				join.onRef("s.source_key", "=", "u.source_key"),
@@ -2042,11 +2038,8 @@ export class MediaUsageRepository {
 			.where(this.noActiveGenerationWriteExpression("u"))
 			.orderBy("u.created_at", "asc")
 			.orderBy("u.id", "asc")
-			.limit(batchLimit);
-		if (options.cleanupLease) {
-			query = query.where(this.activeCleanupLeaseExpression(options.cleanupLease));
-		}
-		const rows = await query.execute();
+			.limit(batchLimit)
+			.execute();
 
 		return this.deleteAbandonedCandidateIds(
 			rows.map((row) => row.id),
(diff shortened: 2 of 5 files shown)


packages/core/src/database/repositories/media-usage.ts
```diff
@@ -1807,6 +1807,8 @@ export class MediaUsageRepository {
 		limit: number;
 		cleanupLease?: MediaUsageCleanupLease;
 	}): Promise<MediaUsageCleanupCandidate[]> {
+		if (input.cleanupLease && !(await this.holdsCleanupLease(input.cleanupLease))) return [];
+
```
khoinguyenpham04: The lease query can use the rest of the 5 secs budget, but the data scan still starts afterward. I'd suggest checking the budget again and returning a separate `skipped` result, since `[]` would look like a completed scan.

packages/core/tests/integration/database/media-usage-cleanup-plan.test.ts
khoinguyenpham04: can we also update the PostgreSQL plan test below to use the new query? It still includes the removed `EXISTS` check, so it is testing the old query.

.changeset/bounded-media-usage-gc-scan.md
```diff
@@ -0,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes the scheduled media-usage cleanup so a tick that no longer holds the cleanup lease stops immediately instead of scanning the whole occurrence table. On large sites the guarded scan previously read every occurrence in the sweep window — hundreds of thousands of rows on a table its `LIMIT` was meant to keep at a few hundred — and returned nothing. Sites on Cloudflare D1 will see the corresponding rows-read spikes disappear from cron ticks.
```
khoinguyenpham04: could we simplify this and focus on the user-facing outcome please? Changesets are user-facing docs and should avoid implementation details.

khoinguyenpham04 (commented):
thanks for the PR1 everything looks good, but just a few things to change :)

packages/core/src/database/repositories/media-usage.ts
MA2153: Done, and you were right that `[]` was the wrong signal — it was worse than ambiguous. With an empty array, `cleanupMediaUsage` computes `sweepComplete = true` (no candidates, no incomplete targets), which nulls the cursor and `scan_before_at`. On the lost-lease path that never lands because `completeMediaUsageCleanup` is itself fenced by the lease token, so the tick reports `skipped` — but that is the fence saving us, not the return value being honest. On a budget-exhausted path there is no fence: the lease is still valid, the completion write succeeds, and the tick would retire a sweep window it never scanned.

So `findMediaUsageCleanupCandidates` now resolves to `MediaUsageCleanupCandidate[] | null`, matching the `null`-means-skipped idiom `claimMediaUsageCleanup` already uses. `null` means the scan was never issued and the caller leaves the cursor alone; `[]` keeps its old meaning of "scanned, nothing there".

On the budget itself: both gates now go through one `admitCleanupScan` helper that checks the budget, reads the lease row, then re-checks the budget, shared by the candidate scan and the four bounded delete scans. `cleanup.ts` passes `canIssueStatement` into the scan instead of gating the call from outside, so no statement is issued either way. Regression test in `media-usage-cleanup.test.ts`: a Kysely plugin burns the full 5s while the lease row is being read, and the scan must return `null` having read only `_emdash_media_usage_cleanup`. It fails without the re-check.

No change to `MAX_CLEANUP_STATEMENTS_PER_TICK` — still 16, no new statements.

packages/core/tests/integration/database/media-usage-cleanup-plan.test.ts
MA2153: Good catch — that test was explaining a query the code no longer issues.

Rather than delete the `EXISTS` from the hand-written SQL, I removed the hand-written SQL: the test now runs `findMediaUsageCleanupCandidates` through a Kysely plugin that compiles each statement it issues, then `EXPLAIN (COSTS OFF)`s the compiled scan. The duplicated SQL is what made the drift invisible, so this closes the class rather than the instance. Same assertions as before — `idx__emdash_media_usage_cleanup_scan` and no `Sort` — and the lease-row setup went away with the predicate that needed it.

Verified against a real PostgreSQL 17, so the pg-only test actually executed:

```
$ EMDASH_TEST_PG=... vitest run tests/integration/database/media-usage-cleanup.test.ts \
    tests/integration/database/media-usage-cleanup-plan.test.ts
 Test Files  2 passed (2)
      Tests  60 passed (60)
```

`tests/integration/database` and `tests/unit/database` are green on both dialects too — 1174 passed, and the two timeouts I flagged in the description passed this run as well.

.changeset/bounded-media-usage-gc-scan.md
MA2153: Rewritten — dropped the lease/`LIMIT` mechanics and led with the symptom:

> Fixes scheduled media-usage cleanup reading far more rows than its batch size on large sites. A cleanup run that had lost its lease scanned the whole occurrence table before returning nothing, so cron ticks could spike into the hundreds of thousands of rows read. Sites on Cloudflare D1 will see those spikes disappear.

I kept the "lost its lease" clause because it is the condition that decides whether an upgrader was affected at all — without it there is no way to tell whether this explains the spikes on their dashboard. Happy to cut it if you'd rather it stay purely about the outcome.

khoinguyenpham04 (approved):
Thanks!

PR #2879, opened 2026-09-02
test(core): run core migrations and queries against real D1

### LINKED ISSUE #1689: Add real-D1 coverage for core migrations and queries
## Goal

Core migration and query behavior should be tested against real D1 before 1.0.

## Scope

Extend the existing D1 test patterns to packages/core so migration and query bugs caused by D1-specific behavior are caught before release.

## Acceptance criteria

- packages/core has a real-D1 test path for migrations.
- Representative query helpers run against real D1.
- The setup works in CI or has a documented opt-in path with clear coverage expectations.
- Tests cover at least the classes of D1-specific bugs seen in previous production issues.
- The test utilities make it hard to accidentally run only sqlite/postgres for D1-sensitive code.

danielmlr:
## What does this PR do?

Runs the core migration chain and the loader's entry read against a real D1 database in CI, so a regression that only D1 produces fails a check instead of a deploy. D1 enforces foreign keys and ignores `PRAGMA foreign_keys = OFF`, auto-commits each DDL statement, and caps tables and result sets at 100 columns. The unit tests simulate the first two on SQLite; its cap is 2000 columns, so the third has no cover that can fail.

The cases go into the workerd pool `packages/core` already has, which never ran in CI: a full run and a no-op second run, the 036 and 040 rebuilds, a retry after 016 stopped partway with no bookkeeping row, and the entry read at the widest collection that fits D1's result-set cap. `test:workerd` becomes a CI job. The live-D1 suite stays the opt-in smoke test, unchanged. `D1 Tests` stays a check of its own rather than joining the `Tests` rollup: branch protection can require a stable-named job directly, as it does Smoke Tests and Browser Tests. Whether a new suite gates every merge from its first run is that call. The retry cases cover 016 and 036, which guard their statements for a rerun; 25 others fail the same replay today and are a separate discussion.

Closes #1689

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [x] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable). Do not include `messages.po` changes except in translation PRs — a workflow extracts catalogs on merge to `main`. (n/a: no admin UI change)
- [ ] I have added and reviewed the user-facing [changeset](https://github.com/emdash-cms/emdash/blob/main/.changeset/README.md) (if this PR changes a published package) (n/a: tests and CI; the one source edit exports an existing function unchanged)
- [ ] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/... (n/a: tests, for an issue on the 1.0 milestone)
- [ ] I have included screenshots below if this PR changes the UI (n/a: no UI change)

- [x] This PR includes AI-generated code — model/tool: Claude Opus 5, Claude Fable 5.1

## Screenshots / test output

Not applicable. `pnpm --filter emdash exec vitest run --config vitest.workerd.config.ts`: 7 files, 26 tests passed. With the `IF NOT EXISTS` guard removed from 016's first statement, both retry cases fail with `table "_emdash_api_tokens" already exists`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -211,6 +211,25 @@ jobs:
       - run: pnpm install --frozen-lockfile
       - run: pnpm --filter emdash exec vitest run --config vitest.integration.config.ts
 
+  test-workerd:
+    name: D1 Tests
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: false
+      - uses: pnpm/action-setup@0e279bb959325dab635dd2c09392533439d90093 # v6.0.8
+      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: 22
+          cache: pnpm
+      - run: pnpm install --frozen-lockfile
+      # The suite imports the Cloudflare package, which resolves `emdash`
+      # through its published entry points rather than source.
+      - run: pnpm run --filter emdash... build
+      - run: pnpm --filter emdash exec vitest run --config vitest.workerd.config.ts
+
   test-browser:
     name: Browser Tests
     runs-on: ubuntu-latest
--- a/packages/core/src/database/migrations/runner.ts
+++ b/packages/core/src/database/migrations/runner.ts
@@ -211,7 +211,7 @@ export interface MigrationOptions {
 	raceWaitMs?: number;
 }
 
-function createMigrator(db: Kysely<Database>, options?: MigrationOptions): Migrator {
+export function createMigrator(db: Kysely<Database>, options?: MigrationOptions): Migrator {
 	return new Migrator({
 		db,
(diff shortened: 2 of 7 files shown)


danielmlr:
Four of the five are in `ff374f49`. The Kysely instance now lives in `beforeAll`, one per file, destroyed in `afterAll`. `runWithContext` passes `editMode: false`. The column-cap probe uses the real `_emdash_seo` columns and matches on `collection` and `content_id` the way the loader's own subquery does; it's still five alias columns on top of `c.*`, so D1 still rejects it. And `d1-schema.ts` now checks the catalog names before they reach `sql.raw` — I used a local pattern rather than `validateIdentifier()`, which rejects the leading underscore every `_emdash_*` table has.

I've left `D1 Tests` out of the `Tests` rollup. It has a stable check name, so branch protection can require it directly, the way it already does Smoke Tests and Browser Tests. Folding it in would also put a new suite in front of every merge from its first run. If you'd rather have it in the rollup, that's two lines.

ascorbic (approved):
Thanks

PR #2885, opened 2026-09-03
fix(core): join avatar media on the single-row byline finders

### LINKED ISSUE #2613: Byline finders return `avatarMediaId` but no storage key, so a byline profile page can't render an avatar
## Summary

A site that renders a byline profile page (`/creators/<slug>`, `/authors/<slug>`, …) has no way to show the byline's avatar. The finder that page is built on — `getBylineBySlug()` / `BylineRepository.findBySlug()` — returns `avatarMediaId` but leaves `avatarStorageKey` null, and there is no public API that turns a media id into a URL. The same avatar renders fine on a content page, because the content-credit hydration path does join `media`.

The result is that the avatar is visible everywhere a byline appears *as a credit on an entry*, and impossible on the page dedicated to that byline.

## Why there's no workaround in site code

- `Astro.locals.emdash.getPublicMediaUrl` takes a **storage key**, not a media id (`packages/core/src/astro/types.ts:502`).
- The public entry point (`packages/core/src/index.ts`) exports `MediaItem` / `CreateMediaInput` as types and `normalizeMediaValue` / `generatePlaceholder`, but no media-by-id lookup. `MediaRepository` isn't reachable from site code.

So a site holding `avatarMediaId` has nothing to call. The only escapes are reaching into internals or keeping a side table of byline → storage key, which is what we ended up doing.

## Current behaviour is deliberate

`findById` / `findBySlug` / `findByUserId` / `findMany` all `selectFrom("_emdash_bylines").selectAll()` with no media join (`packages/core/src/database/repositories/byline.ts:518`), and a test pins that:

```ts
// findById/findBySlug don't join media — the field is null there, and
// callers should rely on the content-credit hydration path for it.
const byId = await bylineRepo.findById(created.id);
expect(byId?.avatarMediaId).toBe("media-avatar-2");
expect(byId?.avatarStorageKey).toBeNull();
```
`packages/core/tests/unit/database/repositories/byline.test.ts:275`

"Rely on the content-credit hydration path" works for a credit rendered next to an entry. It doesn't work for a page keyed on the byline itself — there is no entry to hydrate through, and loading one of the byline's credited entries just to steal the avatar off it is a query for data the page already identified.

The avoided N+1 that motivates the current shape doesn't apply here either: a profile page resolves exactly one byline.

## Reproduce

1. Create a byline with an uploaded avatar.
2. In a site page: `const b = await getBylineBySlug(slug, { locale: Astro.currentLocale })`.
3. `b.avatarMediaId` is set; `b.avatarStorageKey`, `avatarAlt`, `avatarBlurhash`, `avatarDominantColor` are all null — no input for `<Image>` or `getPublicMediaUrl`.

Verified against `upstream/main` @ 353ff4fd.

## Possible directions

Any one of these closes it; listed roughly in order of how contained the change is.

1. **Opt-in join on the single-row finders** — `findById(id, { withAvatar: true })` / `findBySlug(slug, { withAvatar: true })`, reusing the same `LEFT JOIN media` and `BylineRowWithAvatar` reshape the hydration path already has, and surfaced through `getByline` / `getBylineBySlug` options. Callers that don't ask keep today's query cost, and the existing test's assertion stays true for the default path.
2. **Always join on the single-row finders**, leaving `findMany` alone. One extra LEFT JOIN on a by-id/by-slug lookup, and the "avatar fields are populated on some paths but not others" asymmetry disappears for everything except list queries.
3. **Export a public media lookup** (`getMedia(id)` or a `getPublicMediaUrlById`), which also helps any other field that hands site code a bare media id.

Happy to open a PR for (1) if that's the shape you'd prefer.

MA2153:
## What does this PR do?

A site rendering a byline profile page (`/authors/<slug>`, `/creators/<slug>`, …) had no way to show the byline's avatar. `getByline` / `getBylineBySlug` — and the `BylineRepository` finders behind them — returned `avatarMediaId` but left `avatarStorageKey`, `avatarAlt`, `avatarBlurhash`, and `avatarDominantColor` null, and there is no public API that turns a media id into a URL (`getPublicMediaUrl` takes a storage key). So the avatar rendered everywhere a byline appeared as a credit on an entry, and nowhere on the page dedicated to that byline.

This is **option 2** from the issue, per @ascorbic's comment: always join on the single-row finders, leaving `findMany` alone.

- Adds a private `selectBylineWithAvatar()` base select — every byline column plus the avatar's media columns, folded in by the same `LEFT JOIN media` the credit-hydration queries (`getContentBylines` / `getContentBylinesMany` / `findByUserIds`) already use.
- `findById`, `findBySlug`, and `findByUserId` now build on it, so `getByline` / `getBylineBySlug` need no change — they pass the repository result straight through.
- `findMany` is untouched: byline list pages keep their single-table scan, and the join can't fan out per listed row.
- Updates the doc comments in `repositories/types.ts` and `api/schemas/bylines.ts` that described the old asymmetry.

A profile page can now do:

```astro
---
import { getBylineBySlug } from "emdash";

const byline = await getBylineBySlug(Astro.params.slug, { locale: Astro.currentLocale });
const avatar = byline?.avatarStorageKey
	? Astro.locals.emdash.getPublicMediaUrl(byline.avatarStorageKey)
	: null;
---
{avatar && <img src={avatar} alt={byline.avatarAlt ?? byline.displayName} />}
```

**Query cost is unchanged.** The avatar arrives as a `LEFT JOIN` on a lookup the page already issues, not a second round trip, so no logged-out route gains a query. No route in `scripts/query-counts.queries.*.json` calls these finders, so the snapshots are unaffected.

Closes #2613

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes — clean for this change. `pnpm lint:json` reports one diagnostic in `apps/release-service/src/ui/PublisherPage.tsx`; it reproduces on unmodified `main` and is unrelated to these files.
- [x] `pnpm test` passes — targeted: the full `packages/core` unit + integration suite (6145 passed, 5 skipped).
- [x] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable) — n/a, no admin UI strings change.
- [x] New features link to an approved Discussion — n/a, this is a bug fix against an existing issue.
- [x] I have included screenshots below if this PR changes the UI — n/a, no UI change.

## Screenshots / test output

On the tests: the previous test pinned the old behaviour ("leaves avatar storage key null on the plain byline finders"). It is replaced with one asserting the new contract, written and run against unmodified source first, where it failed on `expected null to be 'media-avatar-2.png'`. It covers all three single-row finders, `null` (not `undefined`) for a byline with no avatar, and that `findMany` still leaves `avatarStorageKey` null.

```
 Test Files  511 passed | 1 skipped (512)
      Tests  6145 passed | 5 skipped (6150)
```

https://claude.ai/code/session_01MdmEcFqHNk8gU4yPfVoDBk

### DIFF
--- a/.changeset/byline-finder-avatar-join.md
+++ b/.changeset/byline-finder-avatar-join.md
@@ -1,0 +1,19 @@
+---
+"emdash": patch
+---
+
+Fixes byline profile pages having no way to render the byline's avatar ([#2613](https://github.com/emdash-cms/emdash/issues/2613)). `getByline`, `getBylineBySlug`, and the underlying single-row `BylineRepository` finders now resolve the avatar's media row in the same query, so `avatarStorageKey`, `avatarAlt`, `avatarBlurhash`, and `avatarDominantColor` are populated alongside `avatarMediaId`:
+
+```astro
+---
+import { getBylineBySlug } from "emdash";
+
+const byline = await getBylineBySlug(Astro.params.slug, { locale: Astro.currentLocale });
+const avatar = byline?.avatarStorageKey
+	? Astro.locals.emdash.getPublicMediaUrl(byline.avatarStorageKey)
+	: null;
+---
+{avatar && <img src={avatar} alt={byline.avatarAlt ?? byline.displayName} />}
+```
+
+Previously these fields were populated only when a byline was hydrated as a credit on a content entry, so a page keyed on the byline itself — `/authors/<slug>` and the like — held a bare media id with no public API to turn it into a URL. Nothing else changes: the lookup still costs one query (the avatar is a `LEFT JOIN`, not a second round trip), and `findMany` still skips the join, so byline list pages are unaffected.
--- a/packages/core/src/api/schemas/bylines.ts
+++ b/packages/core/src/api/schemas/bylines.ts
@@ -13,9 +13,9 @@ export const bylineSummarySchema = z
 		bio: z.string().nullable(),
 		avatarMediaId: z.string().nullable(),
 		/**
-		 * Avatar media storage key + alt, folded in by the media join during
-		 * content byline hydration. Null on the plain byline finders, which
-		 * don't join media.
+		 * Avatar media storage key + alt, folded in by the media join on the
+		 * content byline hydration path and the single-row byline finders.
+		 * Null on `findMany`, which doesn't join media.
 		 */
 		avatarStorageKey: z.string().nullish(),
 		avatarAlt: z.string().nullish(),
--- a/packages/core/src/database/repositories/byline.ts
+++ b/packages/core/src/database/repositories/byline.ts
@@ -30,9 +30,9 @@ type BylineRow = Selectable<BylineTable>;
 
 /**
  * A byline row optionally augmented with the avatar's media columns, folded in
- * by the `LEFT JOIN media` in the content-credit hydration queries. The plain
- * `selectAll()` finders produce rows without these keys, so they're optional
- * and `rowToByline` defaults them to null.
+ * by the `LEFT JOIN media` in the credit-hydration and single-row finder
+ * queries. `findMany` selects without the join, so the keys are optional and
+ * `rowToByline` defaults them to null.
  */
 type BylineRowWithAvatar = BylineRow & {
 	avatar_storage_key?: string | null;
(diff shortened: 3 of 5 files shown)


ascorbic (approved):
Thanks!

PR #2890, opened 2026-09-03
fix(core): index scheduled sweep on (deleted_at, scheduled_at)

### LINKED ISSUE #2843: Scheduled-publish sweep scans every collection on every tick; partial scheduled_at index unused
### Description

On every scheduled-maintenance tick, the scheduled-publishing sweep issues one query per collection:

```sql
SELECT * FROM "ec_article"
WHERE scheduled_at IS NOT NULL AND scheduled_at <= ? AND deleted_at IS NULL
ORDER BY scheduled_at ASC LIMIT ?
```

Two problems compound:

**1. The partial index built for this query is not used.** The migrations create exactly the right index:

```sql
CREATE INDEX "idx_ec_article_scheduled" ON "ec_article" (scheduled_at) WHERE scheduled_at IS NOT NULL
```

but SQLite plans the query against the low-selectivity `deleted_at` index instead, then sorts:

```
SEARCH ec_article USING INDEX idx_ec_article_deleted_status (deleted_at=?)
USE TEMP B-TREE FOR ORDER BY
```

Measured on a collection with 7,753 live rows: **7,755 rows read, 0 rows returned.**

**2. The sweep runs unconditionally, per collection, per tick** — even when no content in the database is scheduled at all. On a site where no entry has `scheduled_at` set, every tick reads every live content row across every collection to publish nothing. With six collections totalling ~13,000 live rows and a one-minute Cron Trigger, that is ~18.7M D1 rows read per day, entirely wasted.

This is D1-billable read volume that scales with total content size and tick frequency, not with the amount of scheduled work.

Two independent fixes are possible, and they stack:

- **Guard the scan.** Skip the per-collection query when nothing is scheduled — e.g. a single cheap probe against the partial index up front, or tracking a "next scheduled at" watermark that the sweep can compare against before touching any collection.
- **Get the partial index used.** Ordering by an indexed column with a `LIMIT` should be an index walk. Dropping the redundant `scheduled_at IS NOT NULL` predicate (already implied by `scheduled_at <= ?`) may be enough to let the planner match the partial index, but the reliable fix is to not issue the query at all when there is no work.

Note that this is the second of the two queries reported in #2211. That issue was closed by #2407, which fixed the revision-history scan only — the `scheduled_at` scan was never addressed. Filing separately rather than reopening.

### Steps to reproduce

1. Deploy an EmDash site on Cloudflare Workers + D1 with several content collections holding a few thousand entries each.
2. Leave `scheduled_at` unset on every entry (no scheduled publishing in use).
3. Enable a Cron Trigger per the [Scheduled Publishing](https://docs.emdashcms.com/deployment/cloudflare/#scheduled-publishing) guide.
4. Observe D1 "Rows read" climb steadily with no scheduled content and no traffic.
5. Confirm per query with the GraphQL Analytics API (`d1QueriesAdaptiveGroups`, ordered by `sum_rowsRead_DESC`): each collection's `scheduled_at` query shows `rowsRead ≈ live row count` and `rowsReturned = 0` on every tick.

### Environment

- emdash version: 0.35.0
- Cron Trigger: `* * * * *`

### Logs / error output

```shell
# EXPLAIN QUERY PLAN, measured against remote D1
SEARCH ec_article USING INDEX idx_ec_article_deleted_status (deleted_at=?)
USE TEMP B-TREE FOR ORDER BY

# Actual cost, same query, same database
rows_read: 7755
rows_returned: 0

# Index that exists and is not chosen
CREATE INDEX "idx_ec_article_scheduled" ON "ec_article" (scheduled_at) WHERE scheduled_at IS NOT NULL

# Rows with scheduled_at set, all collections
0
```

MA2153:
## What does this PR do?

The scheduled-publishing sweep runs one query per collection per scheduler tick:

```sql
SELECT * FROM "ec_article"
WHERE scheduled_at IS NOT NULL AND scheduled_at <= ? AND deleted_at IS NULL
ORDER BY scheduled_at ASC LIMIT ?
```

A stats-blind planner cannot seek the `scheduled_at`-only partial index and satisfy `deleted_at IS NULL` at the same time, so it takes a `deleted_at=?` equality on whichever `deleted_at`-leading composite it finds first, reads every live row in the table, and sorts the result in a temp b-tree. Sites with no scheduled content at all still paid their entire content size in reads on every tick, which on D1 is directly billable.

Migration `074` replaces `idx_{table}_scheduled` with `idx_{table}_del_sched` on `(deleted_at, scheduled_at)`, keeping the `WHERE scheduled_at IS NOT NULL` predicate that migration `030` widened so the index stays proportional to scheduled content rather than to the table. Leading with `deleted_at` lets one index serve the equality, the range, and the `ORDER BY` together. `findReadyToPublish`'s SQL is unchanged. `SchemaRegistry` is updated to give newly created tables the same shape.

Measured with `EXPLAIN QUERY PLAN` on SQLite without `ANALYZE`, matching D1's stats-blind planner:

| | Plan |
| --- | --- |
| Before | `SEARCH ec_post USING INDEX idx_ec_post_loc_crt (deleted_at=?)` + `USE TEMP B-TREE FOR ORDER BY` |
| After | `SEARCH ec_post USING INDEX idx_ec_post_del_sched (deleted_at=? AND scheduled_at>? AND scheduled_at<?)` |

`countScheduled`, which backs admin scheduled-content counts, reads the same shape and picks up the new index too.

### Two notes on the issue as filed

**The suggested query fix does not work.** #2843 proposes dropping the redundant `scheduled_at IS NOT NULL` predicate to let the planner match the partial index. I measured this: the plan is byte-for-byte unchanged. `deleted_at IS NULL` is what pulls the planner off the partial index, not the `IS NOT NULL` term. Adding `deleted_at IS NULL` to the partial index's own `WHERE` clause does not help either. Only the composite key does.

**The guard/watermark is deliberately not implemented.** The issue proposes two stacking fixes. With the index corrected, a tick that finds nothing due costs a seek per collection instead of a full scan, so the remaining cost is round trips, not read volume. A "next scheduled at" watermark would buy those back at the price of durable state to invalidate on every schedule, unschedule, and publish write — which also cuts against the "prefer the batch query to a 'has any' probe" guidance in `CLAUDE.md`. Happy to add it in a follow-up if maintainers disagree.

Closes #2843

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [x] Performance improvement
- [x] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

Admin localization, a feature Discussion, and screenshots are not applicable: this changes a background database index and adds no UI or user-facing strings.

## Screenshots / test output

Test coverage added in `packages/core/tests/integration/database/scheduled-publish-plan.test.ts`:

- The sweep seeks through `idx_{table}_del_sched` with no scan and no temp b-tree. The assertion matches the plan shape rather than naming a rival index, so it cannot be satisfied by the planner merely switching to a different `deleted_at` composite.
- Due content still comes back oldest-first with soft-deleted and not-yet-due rows excluded.
- A pre-074 table is migrated and its plan flips, asserted before and after `migration074.up()`.

Verification:

- Full `packages/core` suite: 6267 passed, 9 skipped. The single unhandled error is a pre-existing tiptap/jsdom `document is not defined` in `inline-portable-text-media-picker.test.ts`, unrelated to this change and present on `main`.
- `pnpm typecheck` clean. `pnpm lint` reports one diagnostic in `apps/release-service/src/ui/PublisherPage.tsx`, which is pre-existing and unrelated.

**Not verified by me:** Postgres parity. No local Postgres was available and `EMDASH_TEST_PG` was unset, so `dialect-compat.test.ts` skipped. The DDL is valid Postgres and that suite picks up migration `074` automatically via `MIGRATION_NAMES`, so CI should cover it — but I have not run it. Worth a reviewer confirming the composite does not regress a Postgres deployment, which has real statistics and may have planned the original query differently.

https://claude.ai/code/session_01Whic7zpyBYDDVwBAmY7TF4

---

## Review response (review 5100896550)

**`AGENTS.md` index list — fixed.** Valid catch; the canonical list still described the old single-column partial. Updated in place as `(deleted_at, scheduled_at)` (partial, `WHERE scheduled_at IS NOT NULL`) rather than the suggested trailing sentence, which keeps the enumeration's existing shape.

**Migration and registry comments — trimmed, not to the suggested length.** Removed the parts that were PR narrative: the cost analysis, the restatement of the full query, and the rejected-alternative discussion. Kept why `deleted_at` must lead and why the partial predicate stays.

The suggested six-line version states the mechanism but drops the failure mode, and the failure mode is the thing a future reader gets wrong — a `scheduled_at`-only index looks obviously correct for this query, which is exactly how the bug shipped. Migration `055` documents the same class of stats-blind planner constraint at comparable length and is the closest precedent in the tree. Happy to cut further if you'd rather these read as terse.

**Query-plan assertions — loosened, partially.** Agreed the exact bound formatting is planner-output detail; it also caught me out once while writing the test, so the brittleness concern is real. Dropped it.

I kept `\(deleted_at=\? AND scheduled_at` rather than reducing to the index name alone. Matching only the name would pass if the planner seeked `deleted_at` and merely filtered `scheduled_at`, which is a materially worse plan than the one this PR is claiming. The retained fragment asserts both columns participate in the seek without pinning how SQLite renders the bounds.

Both assertions are additionally backed by the `not.toContain("USE TEMP B-TREE FOR ORDER BY")` check and by the pre/post migration test, which asserts the old plan before `migration074.up()` and the new one after.

Verification after these changes: `pnpm typecheck` clean, `pnpm lint:quick` clean, formatted, `git diff --check` clean; 602 passed across the database integration suite plus the scheduled-publish and registry unit suites. Postgres parity remains unverified locally for the reason given above.

### DIFF
--- a/.changeset/scheduled-sweep-index.md
+++ b/.changeset/scheduled-sweep-index.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes scheduled publishing so its recurring check no longer reads every content entry on each run. Sites running the scheduler on a frequent cron trigger, as the Cloudflare deployment guide recommends, previously saw database reads grow with the size of their content library rather than with the amount of scheduled work — a cost that is directly billable on D1 and was paid even when nothing was scheduled. Existing sites pick up the fix when migrations run on upgrade; no configuration or code changes are needed.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -175,7 +175,7 @@ Test representative upgrades from existing data, retry after partial completion,
 
 ## Indexes
 
-Every content table gets indexes on: `status`, `slug`, `created_at`, `deleted_at`, `scheduled_at` (partial, `WHERE scheduled_at IS NOT NULL`), `live_revision_id`, `draft_revision_id`, `author_id`, `primary_byline_id`, `updated_at`, `locale`, `translation_group`. Foreign key columns always get an index.
+Every content table gets indexes on: `status`, `slug`, `created_at`, `deleted_at`, `(deleted_at, scheduled_at)` (partial, `WHERE scheduled_at IS NOT NULL`), `live_revision_id`, `draft_revision_id`, `author_id`, `primary_byline_id`, `updated_at`, `locale`, `translation_group`. Foreign key columns always get an index.
 
 Naming: `idx_{table}_{column}` for single-column, `idx_{table}_{purpose}` for multi-column.
 
--- a/packages/core/src/database/migrations/074_content_deleted_scheduled_index.ts
+++ b/packages/core/src/database/migrations/074_content_deleted_scheduled_index.ts
@@ -1,0 +1,60 @@
+import type { Kysely } from "kysely";
+import { sql } from "kysely";
+
+import { listTablesLike } from "../dialect-helpers.js";
+
+/**
+ * Migration: replace the single-column partial `scheduled_at` index on content
+ * tables with a `(deleted_at, scheduled_at)` composite.
+ *
+ * The scheduled-publishing sweep reads `scheduled_at IS NOT NULL AND
+ * scheduled_at <= ? AND deleted_at IS NULL` ordered by `scheduled_at`. A
+ * stats-blind planner cannot seek the `scheduled_at`-only partial index and
+ * satisfy `deleted_at IS NULL` too, so it takes the `deleted_at=?` equality on
+ * whichever `deleted_at`-leading composite it finds first, reads every live row
+ * in the table, and sorts the result in a temp b-tree. The sweep runs per
+ * collection on every scheduler tick, so a site with no scheduled content still
+ * pays its entire content size in reads every tick.
+ *
+ * Leading with `deleted_at` lets the same index serve the equality and then walk
+ * `scheduled_at` in order, which satisfies the range and the ORDER BY together.
+ * The partial `WHERE scheduled_at IS NOT NULL` clause is kept so the index stays
+ * proportional to scheduled content rather than to the table. D1 never has
+ * `sqlite_stat1`, so the index shape is the only lever.
+ *
+ * Forward-only and idempotent (`IF NOT EXISTS`).
+ *
+ * The short `del_sched` suffix keeps the name inside Postgres's 63-byte
+ * identifier limit for long collection slugs. Keep it identical to the name in
+ * `schema/registry.ts`.
+ */
+export async function up(db: Kysely<unknown>): Promise<void> {
+	const tableNames = await listTablesLike(db, "ec_%");
+
+	for (const tableName of tableNames) {
+		// D1 DDL is non-transactional: create the replacement before dropping the
+		// old index so an interrupted migration always leaves a scheduled_at index
+		// in place.
+		await sql`
+			CREATE INDEX IF NOT EXISTS ${sql.ref(`idx_${tableName}_del_sched`)}
+			ON ${sql.ref(tableName)} (deleted_at, scheduled_at)
+			WHERE scheduled_at IS NOT NULL
+		`.execute(db);
+
+		await sql`DROP INDEX IF EXISTS ${sql.ref(`idx_${tableName}_scheduled`)}`.execute(db);
+	}
+}
+
+export async function down(db: Kysely<unknown>): Promise<void> {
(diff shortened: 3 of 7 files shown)


ascorbic (approved):
Thanks!

PR #2910, opened 2026-09-04
fix(core): bound the media-usage cleanup scan with a row-value cursor

### LINKED ISSUE #2901: Media-usage cleanup scan reads the whole table to return a 250-row page
### Description

**This is not a regression of #2844.** That fix is present and working: the lease is verified in its own statement and the scan's predicate is clean. The defect below sits in the same query but is reached on the normal, healthy-lease path — the scan returns its full page every single time and still reads the entire table to do it.

`MediaUsageRepository.findMediaUsageCleanupCandidates` paginates with a keyset cursor expressed as an `OR`:

```sql
select "u"."id", ..., "writer"."expires_at"
from "_emdash_media_usage" as "u"
left join "_emdash_media_usage_sources" as "s" on "s"."source_key" = "u"."source_key"
left join "_emdash_media_usage_generation_writes" as "writer"
  on "writer"."source_key" = "u"."source_key" and "writer"."generation" = "u"."generation"
where "u"."created_at" < ?
  and ("u"."created_at" > ? or ("u"."created_at" = ? and "u"."id" > ?))
order by "u"."created_at" asc, "u"."id" asc
limit ?
```

Under **bound parameters** SQLite cannot prove the two disjuncts reduce to a single range over the composite index, so it plans `MULTI-INDEX OR` — two index scans whose union is no longer in index order — and adds `USE TEMP B-TREE FOR ORDER BY` to restore the `ORDER BY`.

Sorting is a blocking operator. The entire `created_at` range must be read and joined through both `LEFT JOIN`s before the first of the `LIMIT`ed rows can be emitted. **The `LIMIT` and the keyset cursor provide no bound at all.**

Measured on a `_emdash_media_usage` table of 157,804 rows (49,859 sources), 142,870 rows between cursor and cutoff, varying only how the values are supplied:

| Binding | Rows read | Rows returned |
| --- | ---: | ---: |
| All values inlined as literals | **502** | 250 |
| All values bound as `?` | **427,020** | 250 |

**851×**, and `142,870 × 3 tables ≈ 428,610` confirms it is reading the whole window once per joined table.

Isolating which parameter matters (identical values throughout):

| Variant | Rows read | Returned |
| --- | ---: | ---: |
| All literals | 502 | 250 |
| Only `LIMIT` bound | 502 | 250 |
| Only cutoff bound | 502 | 250 |
| Only cursor bound | 15,434 | 250 |
| **All bound** | **427,020** | 250 |

The collapse is non-additive — it needs the cutoff and the cursor parameterised together, which is the signature of a planner decision rather than of any one value.

**Plans.** With literals — the intended ordered walk, terminates at `LIMIT`:

```
SEARCH u USING COVERING INDEX idx__emdash_media_usage_cleanup_scan (created_at>? AND created_at<?)
SEARCH s USING INDEX sqlite_autoindex__emdash_media_usage_sources_1 (source_key=?) LEFT-JOIN
SEARCH writer USING INDEX sqlite_autoindex__emdash_media_usage_generation_writes_2 (source_key=? AND generation=?) LEFT-JOIN
```

With bound parameters — what actually runs in production:

```
MULTI-INDEX OR
  INDEX 1
    SEARCH u USING COVERING INDEX idx__emdash_media_usage_cleanup_scan (created_at>? AND created_at<?)
  INDEX 2
    SEARCH u USING COVERING INDEX idx__emdash_media_usage_cleanup_scan (created_at=? AND id>?)
SEARCH s USING INDEX sqlite_autoindex__emdash_media_usage_sources_1 (source_key=?) LEFT-JOIN
SEARCH writer USING INDEX sqlite_autoindex__emdash_media_usage_generation_writes_2 (source_key=? AND generation=?) LEFT-JOIN
USE TEMP B-TREE FOR ORDER BY
```

Note the covering index is chosen correctly in both. This is not a missing-statistics problem and `ANALYZE` does not help — what changes is only whether the `OR` can be reduced to a range.

**This is invisible to any reproduction that inlines literals.** That is the most costly part of the defect: constant folding gives the cheap plan, so a hand-run repro of the exact production SQL exonerates it. It reproduces only through a prepared statement with bound parameters.

### Suggested fix

Use a row-value comparison, which states the keyset intent directly and matches the composite index without any `OR` reasoning. Verified on the same data — identical 250 rows, same first id:

```sql
where "u"."created_at" < ? and ("u"."created_at", "u"."id") > (?, ?)
```

| Rewrite | Rows read | Plan |
| --- | ---: | --- |
| **Row value** `(created_at, id) > (?, ?)` | **501** | `SEARCH u USING COVERING INDEX ... ((created_at,id)>(?,?) AND created_at<?)` |
| Keep the `OR`, add redundant `created_at >= ?` | 502 | single range, as literals |

The redundant-bound variant works by handing the planner the range it could not derive, but leaves the `OR` in place as a filter and depends on the planner continuing to prefer the range. The row-value form is the durable fix. SQLite has supported row values since 3.15.

Independently: this is the second defect in this one query after #2844, and both share a failure mode — **a `LIMIT`ed, keyset-paginated scan that silently degrades to a full-range scan**. It is worth auditing the `OR`-keyset shape wherever else it appears, and worth treating "the `LIMIT` cannot be reached early" as a property to test for rather than to assume.

### Steps to reproduce

1. Any SQLite/D1-backed install with a large `_emdash_media_usage` table (~150K occurrence rows).
2. Run the candidate scan as a **prepared statement with bound parameters**, using a cursor well before the cutoff.
3. Compare `rows_read` against the same SQL with the same values inlined as literals.
4. `EXPLAIN QUERY PLAN` both — the bound-parameter form shows `MULTI-INDEX OR` and `USE TEMP B-TREE FOR ORDER BY`.

### Environment

- emdash version: 0.37.0 (i.e. with the #2844 fix applied)
- `_emdash_media_usage`: 157,804 rows; `_emdash_media_usage_sources`: 49,859 rows
- Observed cost: ~450K rows read per execution, ~14.5M rows/hour, ~350M/day — six times the next-largest consumer on the database

### Logs / error output

```shell
# same SQL, same values, prepared statement with bound parameters
rows_read: 427020     rows_returned: 250

# same SQL, same values, literals inlined
rows_read: 502        rows_returned: 250
```

MA2153:
## What does this PR do?

`MediaUsageRepository.findMediaUsageCleanupCandidates` expressed its keyset cursor as an `OR` of a strict range and a tie-break equality. Under bound parameters SQLite cannot prove those disjuncts reduce to a single range over `idx__emdash_media_usage_cleanup_scan`, so it plans `MULTI-INDEX OR` and adds `USE TEMP B-TREE FOR ORDER BY`. Sorting is a blocking operator, so the `LIMIT` is unreachable until the entire cursor-to-cutoff range has been read and joined through both `LEFT JOIN`s — the sweep reads its whole backlog on every run to return one 250-row page.

This replaces the `OR` with a row-value comparison, `(u.created_at, u.id) > (?, ?)`, which states the same keyset intent as a single range the planner matches against the composite index directly. SQLite has supported row values since 3.15 and PostgreSQL supports them; both dialects are covered by tests.

Closes #2901

### The page is bounded, not merely unsorted

Removing the `TEMP B-TREE` is the visible half. The half that decides the cost is that SQLite compiles the row value into a **two-column vector seek** — `SeekGE` with a 2-element key — so the scan lands at the cursor's position inside the index rather than at the start of a `created_at` group and filtering forward.

That distinction is load-bearing here, because a bulk regeneration writes long runs of rows sharing one `created_at`. On a 50,000-row table where *every* row has an identical `created_at` and the cursor sits at row 49,990, the two forms behave as:

```
row-value:  SEARCH u USING COVERING INDEX ix ((created_at,id)>(?,?) AND created_at<?)
or-form:    MULTI-INDEX OR ... USE TEMP B-TREE FOR ORDER BY
```

Both return the same five rows; only the row-value form reaches them without walking the run. So the fix bounds the page cost by the page size in the degenerate shape too, not just when timestamps are distinct.

### The test

`packages/core/tests/integration/database/media-usage-cleanup-plan.test.ts` gains `walks the cleanup scan in index order from a bound cursor`. It compiles the statement the repository actually issues, runs `EXPLAIN QUERY PLAN` **with its real bound parameters**, and asserts the ordered index walk. This matters: the issue notes the defect is invisible to any repro that inlines literals, because constant folding gives the cheap plan.

Baseline against unfixed source (the test failing first):

```
AssertionError: expected 'MULTI-INDEX OR\nINDEX 1\nSEARCH u USI…' not to contain 'MULTI-INDEX OR'

  MULTI-INDEX OR
+ INDEX 1
+ SEARCH u USING COVERING INDEX idx__emdash_media_usage_cleanup_scan (created_at>? AND created_at<?)
+ INDEX 2
+ SEARCH u USING COVERING INDEX idx__emdash_media_usage_cleanup_scan (created_at=? AND id>?)
+ SEARCH s USING INDEX sqlite_autoindex__emdash_media_usage_sources_1 (source_key=?) LEFT-JOIN
+ SEARCH writer USING INDEX sqlite_autoindex__emdash_media_usage_generation_writes_2 (source_key=? AND generation=?) LEFT-JOIN
+ USE TEMP B-TREE FOR ORDER BY
```

After the fix the plan is the single ordered range that terminates at the `LIMIT`.

**What this test does and does not guard.** Every row it inserts shares one `created_at`, which is what makes it a worst case for the tie-break — but it also means the row assertion only pins the tie-break direction. It is a plan test, and it still passes against a keyset rewritten wrongly as `created_at >= ? AND id > ?`. Substituting that broken rewrite fails four other tests: the three sweep-state cases in this same file, and `advances a bounded cursor past old live rows to later garbage` in `media-usage-cleanup.test.ts`, which runs under `describeEachDialect` and covers resuming across a `created_at` boundary. So the cursor semantics are held by tests that already existed; what the new test adds is the plan.

### Verification

- **SQLite.** `media-usage-cleanup-plan.test.ts` and `media-usage-cleanup.test.ts` pass; the new test fails against unfixed source as quoted above.
- **PostgreSQL.** The existing plan test only exercised `cursor: null`, so the cursor path had no coverage in that dialect. It is now parameterised over both, asserting `EXPLAIN (COSTS OFF)` with real bound parameters shows the index scan and no `Sort`. Run against a real PostgreSQL 16: both cases pass, as do all 50 cases of the dialect-parametrised `media-usage-cleanup.test.ts` — which is where the cursor *semantics* are covered on that dialect, since the plan test only asserts the plan. **The plan case passes with and without the fix**: PostgreSQL plans the `OR` into an index scan without a `Sort`, so it is a portability guard on the new row-value SQL, not a second regression test. Flagging that explicitly so nobody reads it as evidence the bug reproduced there.
- **D1.** `EXPLAIN QUERY PLAN` against a production D1 carrying this schema returns the ordered covering-index range with both `LEFT JOIN`s as index searches and no `TEMP B-TREE`, confirming the syntax is accepted and the index is chosen there. Two honest limits on that check: the D1 CLI cannot bind parameters, so it cannot demonstrate the two-column seek the way a prepared statement does, and no automated test in this repo exercises `cleanupMediaUsage` on D1 — `tests/workerd/` covers activation, the maintenance engine, projection admission and collection deletion, but not the sweep. Worth noting alongside it that the D1 instance has no `sqlite_stat1`, so the local plan test running without `ANALYZE` is a faithful proxy for that planner rather than an accident.

### Scope

The issue also suggests auditing the `OR`-keyset shape wherever else it appears. Per the repository's scope-discipline rule I have not touched them here, but the deferral is worth characterising rather than leaving as a count.

Most sites carrying the shape pair it with a selective equality prefix ahead of the keyset — the media-usage work-queue listing filters by collection and state, `findCurrentEntryUsagePageByMediaId` filters by a single `media_id` — which bounds the range the planner is reasoning about and is exactly what this query lacked. The comparable one is `ContentRepository.findMany` (`content.ts`), which pages an unbounded range with a `LIMIT`, orders on arbitrary indexed fields, and unlike this background sweep can sit on a logged-out request. I have not measured it and am not claiming it is broken; it is the one I would point the same bound-parameter plan assertion at first. Worth a Discussion if maintainers want that audit; the property to test for is "the `LIMIT` can be reached early", which this PR's plan assertion demonstrates how to express.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable). Not applicable — no admin UI change.
- [x] New features link to an approved Discussion. Not applicable — bug fix.
- [x] I have included screenshots below if this PR changes the UI. Not applicable — no UI change.

Notes on the checks, so reviewers can see exactly what was and was not verified:

- `pnpm typecheck`: `packages/core` passes. `packages/cloudflare` fails on `MUTABLE_MEDIA_CACHE_CONTROL` not being exported from `emdash/media/image-endpoint` — that is a stale `packages/core/dist` in my working tree, pre-existing and untouched by this PR (the export is present in `packages/core/src/media/image-endpoint.ts:57`, and absent from the built `dist`).
- `pnpm lint`: `lint:quick` reports zero diagnostics. `lint --type-aware` reports one pre-existing diagnostic, `no-redundant-type-constituents` in `apps/release-service/src/ui/PublisherPage.tsx:283`, in a file this PR does not touch.
- `pnpm test`: targeted rather than full-suite. `media-usage-cleanup-plan.test.ts`, `media-usage-cleanup.test.ts` and `media-usage-migration.test.ts` pass on SQLite, and the first two also pass against a real PostgreSQL 16 with `EMDASH_TEST_PG` set. Without `EMDASH_TEST_PG` the two PostgreSQL plan cases skip, so a local run that reports them as skipped has not exercised that dialect — CI sets it.
- `pnpm format`: oxfmt reports both changed files already formatted.

## Screenshots / test output

Not applicable — no UI change. Test output is quoted above.

https://claude.ai/code/session_017WFL7b2QAwX1VZQ1jaj3d2


MA2153:
I have confirmed this fix in production:
image

khoinguyenpham04:
@MA2153 i'm reviewing it now, one of the browser tests was flaky so merged main into this branch to resolve that first.

.changeset/lucky-parrots-repeat.md
```diff
@@ -0,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes the media-usage cleanup sweep reading its entire pending backlog on every run once it resumes from a saved cursor. On SQLite and D1 the sweep now stops at its page limit instead of scanning and sorting every row between the cursor and the cutoff. A reported sweep over a 157,000-row `_emdash_media_usage` table read 427,020 rows to return a 250-row page; it now reads about 500. D1 sites with a large media-usage backlog will see rows read by the cleanup task fall accordingly, with no configuration change.
```
khoinguyenpham04: could you please simplify this, as this is a customer facing changeset?

khoinguyenpham04 (commented):
everything is good! thanks for tackling this issue. just a nit before merging

.changeset/lucky-parrots-repeat.md
MA2153: Done

khoinguyenpham04 (approved):
Thanks!

PR #2912, opened 2026-09-04
fix(core): require the revision token on MCP content writes

danielmlr:
## What does this PR do?

Requires `_rev` on the MCP `content_update`, `content_publish`,
`content_unpublish` and `content_discard_draft` tools, so an agent can no longer
write over changes it never read. The token was optional on those four while the
CLI has always required it on `content update`.

The status transitions inside `content_update` hand the token on to the publish
and unpublish handlers as well. A status-only call never checked the token it
carried, so a stale read could publish or unpublish the item while
`content_publish` and `content_unpublish` refused the same token.

The parameter descriptions named the token without saying what to do with it,
and now carry the protocol. A call that omits it fails with a message pointing
at `content_get` rather than "expected string, received undefined", since a
client holding a cached schema never re-reads a description. `content_update`
also states that a change to a published item is staged as a draft when
`status` is not set.

The MCP reference is updated to match. It listed `_rev` as optional on
`content_update` and left it out of the other three tools altogether, so a
client written from that page would have failed on all four.

Breaking for existing MCP clients, approved on Discord on 4 September along with
the decision to cover all four tools rather than `content_update` alone.

The unreleased changeset `atomic-expected-publish.md` said calls that omit
`_rev` keep the existing behaviour. Both changesets ship in the same release, so
that sentence is removed here.

Closes: N/A

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (or targeted tests for my change): `packages/core` is 6309 passing and `pnpm test:e2e` is 283 passing with 31 skipped; `packages/registry-verification` `test:workerd` fails identically on an unrelated branch, so it is not from this change
- [ ] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable). Do not include `messages.po` changes except in translation PRs — a workflow extracts catalogs on merge to `main`. (n/a: no admin UI strings)
- [ ] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/... (n/a: not a feature)
- [ ] I have included screenshots below if this PR changes the UI (n/a: no UI change)

- [x] This PR includes AI-generated code — model/tool: Claude Opus 5 (review pass: Claude Fable 5.1)

## Screenshots / test output

Not applicable for screenshots.

Tests carry the token the same two ways a caller does. `revOf` takes it from a
response for a chained write, `currentRev` reads it fresh where the subject is
not concurrency. Each of the four tools has a test that goes red when the
requirement is removed: reverting all four declarations to `.optional()` fails
exactly those four and leaves the rest of the suite green. Two more cover a
status-only `content_update`: a stale token is refused and the current one goes
through, and dropping the token from the publish and unpublish calls fails
exactly those two. Three existing tests changed meaning rather than shape: the one asserting that an update without
`_rev` still succeeds now asserts the rejection and the message, and the two
parallel-write tests guard state integrity and permission isolation, so a lost
race became an acceptable outcome there while corruption and a wrong refusal
did not. The missing-token message contains the word `required`, which the
field-validation suite's pattern accepts, so a further three `content_update`
calls kept passing without a token; they now carry it and are rejected by the
field check again.

Whether a required token confuses an agent was checked rather than assumed.
Five agents with no prior context were pointed at a running MCP server over
HTTP and given ordinary editorial tasks, one per tool. All five read the
requirement out of `tools/list`, called `content_get`, and wrote successfully on
the first attempt, including on publish, unpublish and discard, where the read
serves no purpose except obtaining the token.

Two limits of that exercise are worth stating. A sixth run under a writer
rewriting the same entry every few seconds never tripped a `CONFLICT`, because
the round trip won the race every time, so the retry half of the protocol is
covered by tests and not by observation. And two of the five agents
independently mistook a staged draft for a live change, which is why
`content_update` now says so.


danielmlr:
The fixes for the review are pushed.

The validation test file was the one MCP suite the first push skipped. Its
three `content_update` calls now carry the token, two from the create response
and one read fresh. Under a pattern that rejects the missing-token message all
seventeen tests in the file still pass, so the three are rejected by the field
check again rather than by the missing token.

Both changesets ship in the same release, so the older publish changeset's
sentence about calls that omit `_rev` is gone, and the body no longer flags
it as open.

The comment above the stub token in the authorization unit test now says why
any string works: the schema takes any string, and the mocked handlers
never compare it with a revision. The SDK validates the input before the scope
and role checks run, so the old comment had the order backwards.

ascorbic (approved):
Thanks!

PR #2913, opened 2026-09-04
fix(core): normalize scheduledAt to UTC in content.schedule()

### LINKED ISSUE #2896: content.schedule() stores scheduledAt without UTC normalization, breaking the lexicographic due-check for non-Z offsets
### Description

`ContentRepository.schedule()` (`packages/core/src/database/repositories/content.ts`, compiled as `content-BNfwh7xX.mjs:1046-1062` in `emdash@0.34.0`) writes the caller-supplied `scheduledAt` string to the `scheduled_at` column verbatim:

```js
async schedule(type, id, scheduledAt) {
  ...
  const scheduledDate = new Date(scheduledAt);
  if (isNaN(scheduledDate.getTime())) throw new EmDashValidationError("Invalid scheduled date");
  if (scheduledDate <= new Date()) throw new EmDashValidationError("Scheduled date must be in the future");
  ...
  await sql`UPDATE ... SET scheduled_at = ${scheduledAt}, ...`
```

It validates that the string parses to a future `Date`, but never normalizes it — it stores whatever offset the caller passed (`+09:00`, `-05:00`, etc.), not UTC.

The due-check in `findReadyToPublish` (same file, `1106-1118`) compares this raw column against `now` with a plain SQL `<=`, where `now` is always `new Date().toISOString()` (`Z`-suffixed UTC):

```js
const now = new Date().toISOString();
... WHERE scheduled_at <= ${now}
```

On SQLite/D1 this is a byte-for-byte lexicographic TEXT comparison, not a real datetime comparison. When `scheduled_at` carries a non-`Z` offset, the comparison is wrong by exactly that offset:

- **Positive offsets (e.g. `+09:00`, JST):** the item is judged "not due" until UTC wall-clock digits reach the same value, i.e. it publishes **~9h late**.
- **Negative offsets (e.g. `-05:00`, US Eastern):** the opposite happens — the item satisfies `scheduled_at <= now` **before** the real instant arrives, so it publishes **~5h early**. This is the more concerning direction, since content intended to stay unpublished can go live ahead of schedule.

### Steps to reproduce

1. Call `content_schedule` (or the admin API `handleContentSchedule`) with an ISO 8601 datetime that includes a non-`Z` offset, e.g. `"2026-09-03T20:39:00+09:00"`.
2. Wait past the real (UTC-equivalent) instant.
3. Observe the item is still `status: "scheduled"` — it does not flip to `published` until ~[offset] hours after the actual due time (or before it, for negative offsets).

### Expected behavior

Scheduling should be timezone-offset agnostic: any valid ISO 8601 input should publish at the correct UTC instant, regardless of what offset the caller used.

### Suggested fix

Normalize to UTC at write time in `ContentRepository.schedule()`:

```js
const scheduledDate = new Date(scheduledAt);
...
await sql`UPDATE ... SET scheduled_at = ${scheduledDate.toISOString()}, ...`
```

This keeps `scheduled_at` uniformly `Z`-suffixed, matching what `findReadyToPublish`'s `now` already assumes, so the lexicographic comparison is valid again — same fix shape as #917.

### Environment

- `emdash@0.34.0`, `@emdash-cms/cloudflare@0.34.0`
- Cloudflare Workers + D1
- Observed via the MCP admin API (`content_schedule`), but the bug is in the core write path and isn't specific to that caller.

htdtkshi:
## What does this PR do?

`content.schedule()` stores the caller-supplied `scheduledAt` string
verbatim, even though it already parses it into a `Date` for
validation a few lines above. `findReadyToPublish()` compares
`scheduled_at` against `new Date().toISOString()` (always UTC, "Z"
suffix) using plain string ordering. A non-UTC offset such as
`+09:00` sorts incorrectly against that "Z" string, so the item
publishes up to that many hours late (or early, for offsets ahead of
UTC).

Observed end-to-end on a production Cloudflare Workers deployment: a
post scheduled for `20:39 JST` via `content_schedule` (MCP tool) was
still `status: scheduled` at `21:31 JST`, ~52 minutes past due. The
cron trigger and `scheduled()` handler were both firing correctly
every minute (confirmed via the Cloudflare API schedules/logs
endpoints, zero errors) — the row just never compared as "due" until
the UTC clock caught up to the same digits, i.e. ~9 hours after the
intended time.

This PR reuses the already-validated `scheduledDate` and stores
`scheduledDate.toISOString()` instead of the raw input, so
`scheduled_at` is always UTC regardless of what offset the caller
passed.

Closes #2896

## Type of change

- [x] Bug fix

## Checklist

- [ ] `pnpm typecheck` passes — **not verified**, see note below
- [ ] `pnpm lint` passes — **not verified**, see note below
- [ ] `pnpm test` passes — **not verified**, see note below
- [ ] `pnpm format` has been run — **not verified**, see note below

**Note on unchecked items**: I wasn't able to run the toolchain in my
environment (`engines.node` requires >=22.16; I only had Node 18
available, no network access to fetch a newer runtime). The change
itself is a 4-line diff reusing a value the method already computes,
and the added test follows the existing suite's structure/fixtures
exactly, but I want to flag explicitly that `typecheck`/`lint`/`test`
have not been run against this branch by me. Happy to fix up per CI
results.

- [x] This PR includes AI-generated code — model/tool: Claude Opus 5,
  via Claude Code, at the direction of and reviewed by the repo owner
  (reported the underlying bug as #2896 from production use of
  `content_schedule` on a live tech.hatada.jp deployment)

## Screenshots / test output

Not applicable (no UI change). Test output not available locally —
see note above.


htdtkshi:

htdtkshi:
Thanks for the thorough review. Addressed both findings in 7f0b438b:

- Extracted a private `normalizeScheduledAt()` helper and applied it in both `update()` and `replaceDraftRevisionForUpdate()`, so every write path that persists `scheduled_at` is UTC-normalized, not just `schedule()`.
- Added regression tests for the `update()` path (non-UTC offset normalization, `null` still clears the field, invalid strings still rejected).
- Reworded the changeset to describe the user-facing behavior instead of leaning on internal mechanics (`findReadyToPublish()` / `scheduledDate.toISOString()`).

One gap I want to flag rather than silently skip: `replaceDraftRevisionForUpdate()` has no existing unit-test scaffold in this suite — it's only reachable via `updateDraftAware()`, which needs a revisions-enabled collection fixture that isn't set up anywhere in `content.test.ts` today. I didn't add one because I can't run the test suite in my environment to validate a new fixture actually exercises the path correctly, and I'd rather say so than guess. Happy to add it if a maintainer can point me at (or is fine merging without) coverage for that specific path.

ascorbic (changes requested):
Can you update the failing test's assertion. It's expecting the old format

ascorbic (approved):

Issue #2911, opened 2026-09-04
Non-translatable fields never sync to sibling locales on collections that support revisions


danielmlr:
## Description

Marking a field non-translatable has no effect on a collection that supports
revisions, which is the default. The edited value stays on the locale it was
entered in and the sibling rows keep what they had. The i18n guide says these
fields are "copied and kept in sync across all translations in the group", and
`Field.translatable` repeats it.

`syncNonTranslatableFields` has a single call site, inside
`handleContentUpdate`, guarded on the update carrying `data`. A save on a
revision-supporting collection stages a draft and leaves the content columns
alone, so the runtime passes `data: undefined` and the guard never opens.
Publication promotes the draft onto those columns without going through the
handler, so nothing syncs there either. The sync runs only where `supports`
excludes `revisions`.

Where it does run, it writes the siblings with raw SQL and advances neither
`version` nor `updated_at`. `_rev` is derived from those two, so a token read
before a sync still validates after it, and an editor holding it writes its
stale copy back with no conflict raised.

I don't think the fix is a one-liner. The natural place to sync is publication,
when the values reach the content columns, but that makes publishing one locale
mutate another locale's live row, and `updated_at` there is what sitemap
`<lastmod>` and `dateModified` read, the phantom modification #2143 was about.
Whether the sync is worth that cost is unresolved.

## Steps to reproduce

1. Configure i18n with two locales, `en` and `de`.
2. Create a collection with the default `supports`, one ordinary field and one
   field declared `translatable: false`.
3. Create an entry in `en`, then its `de` translation with `translationOf`.
4. Publish both.
5. Save a new value for the non-translatable field on the `de` entry, then
   publish it.
6. Read the `en` entry. The non-translatable field still holds the old value.

Creating the collection with `supports: []` instead makes step 6 pass, and
shows the second half: capture the `en` entry's `_rev` before step 5, and
afterwards its data has changed while that same token still validates.

I ran both as integration tests against real in-memory SQLite, not through the
admin UI.

## Environment

- emdash version: 0.36.0 (repository `main` at a213afabf)
- Node.js version: 26.8.1
- Runtime: Node
- OS: macOS 26.6.2

## Screenshots

Not applicable.

## Logs / error output

```shell
AssertionError: expected 'SKU-1' to be 'SKU-2' // Object.is equality

Expected: "SKU-2"
Received: "SKU-1"

  const en2 = await runtime.handleContentGet("posts", enId, "en");
[...]

PR #2919, opened 2026-09-05
feat(core): lock a content entry while someone is editing it

danielmlr:
## What does this PR do?

Prevents two editors from colliding on one entry, rather than reporting the collision once both have done the work. #2126 turns a stale save into a 409, but that arrives after the typing is done, and the only resolution left is to discard one side.

Opening an entry in the admin now takes a lease on it, one row per `(collection, entry_id)`, claimed in a single insert-or-steal statement so two editors racing on one entry cannot both win. The second editor is told who holds it and chooses read-only or take-over; the holder learns of a take-over on their next heartbeat or save, and their editor turns read-only with a banner naming who has it now. The save-conflict banner #2902 added sits inside that read-only region, so an entry held by someone else offers no "Save anyway" the server would refuse. `PUT`, `DELETE`, `/publish`, `/unpublish`, `/schedule` and `/discard-draft` refuse a write against someone else's live lease with `409 ENTRY_LOCKED` and name the holder; `overrideLock` opts out, and the CLI exposes `--override-lock`. The lease runs seven minutes; the admin re-asserts it every two minutes and on every save, and hands it back when the editor is left or the tab closes. Trashing an entry releases the lease server-side, since the client's own release is refused once the entry is out of its reach. Nothing is added to the logged-out path. Locking is on for every collection and switches off per collection.

The lock is documented where a caller meets it: the lock endpoints and `ENTRY_LOCKED` in the REST reference, `--override-lock` in the CLI reference and the CLI skill's editing flow, read-only or take-over in the content guide, and a note on the plugin capabilities page that a plugin content write is not stopped by an editor's lock.

Design and defaults are discussion #2819; ascorbic approved the split it proposes, `_rev` on the MCP write tools first and locking on top, in Discord.

Four things carry over from that thread:

- The refusal code is `ENTRY_LOCKED`, not the `WORK_LEASE_ACTIVE` the discussion named. `mediaUsageWorkRetryConflictSchema` pins that code's `details` to `{ leaseExpiresAt }` and is wired into the OpenAPI document at both media-usage retry paths, so reuse would make the published spec claim one code carries two incompatible shapes.
- The MCP content tools do not honour the lock. That half edits the same four tool definitions #2912 changed, and #2912 has merged, so it can follow as its own PR.
- `syncNonTranslatableFields` writes sibling locales with raw SQL and bypasses the lock exactly as it bypasses `_rev`. That is #2911, inherited rather than fixed here.
- Migration `075_entry_edit_locks` takes the next free number on `main`. Published migrations are immutable, so another 075 landing first means renumbering this one.

Two things stay outside the lock deliberately. `/restore` and `/permanent` take no lock check: a lease is only ever created by opening an entry, a trashed entry cannot be opened, and the trash boundary now clears the lease it inherits, so a check on either route could never refuse. Bulk actions still report a locked entry as a plain failure; naming the holder there means changing what the shared bulk helper returns for publish, unpublish and trash alike, and that belongs in its own PR.

## Type of change

- [ ] Bug fix
- [ ] Translation
- [x] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/2819

- [x] This PR includes AI-generated code — model/tool: Claude Opus 5 (review pass: Claude Fable 5.1)

## Screenshots / test output

Content editor with a modal dialog reading "This entry is open somewhere else", "Ada Lovelace is editing this entry. Open it read-only, or take over; they will be told the entry moved on.", above buttons Open read-only and Take over

The same dialog in the Arabic admin, with the sidebar, header and dialog buttons mirrored to right-to-left

Content editor in read-only mode with a banner reading "Read-only", "Ada Lovelace is editing this entry. Nothing you change here will be saved.", above a Take over button; the fields below are disabled

The same read-only banner in the Arabic admin, mirrored to right-to-left with the banner text still reading left-to-right

The refusal path is covered at the repository, at the handlers and at every refusing route, against real databases; the repository suite runs on Postgres too, in CI, where the integration job has one. The two-editor journey has no end-to-end test: the e2e fixture has a single identity, dev-bypass is hard-wired to `dev@emdash.local`, and a second live session needs the invite and virtual-authenticator flow. So the end-to-end tests cover the lease mechanics with that one identity, and the 409-to-banner seam is covered at the hook level.


danielmlr:
All four are answered; one is taken, three are not.

**Taken: the lock survived a soft delete.** `ad94d0794` releases it inside the delete transaction, and only when the delete happened, the way permanent delete already does. `a1f4a6c18` closes the second boundary: `ContentAccessWithWrite.delete`, the write API plugins get, soft-deletes through the repository and left the lease sitting there. Neither release sits in the route, because the MCP `content_delete` tool and a plugin never reach one. The holder could not clear the lease either: the release endpoint resolves through `handleContentGet` and 404s once the entry is in the trash. One integration test per boundary asserts the row is gone; both fail without the fix.

**Not taken: `/restore` and `/permanent`.** A lease is only ever created by opening an entry, `handleContentGet` does not return a trashed one, and both trash boundaries now clear the lease they inherit, including one held by someone else when the trash was forced with `overrideLock`. A check on either route could never refuse. The trash is a list rather than an editor, so viewing it holds nothing.

**Not taken: naming the holder in bulk actions.** `runBulkAction` keeps failed ids and discards the error that carries the holder, for bulk publish, unpublish and trash alike. Teaching it to carry refusal details changes every caller's result and belongs in its own PR.

ascorbic:
@danielmlr can you fix the conflict?

danielmlr:
The conflict is resolved. It was in `ContentEditor.tsx`: #2972 changed the `onPublishedAtChange` prop on the settings panel, this branch had wrapped that panel in the read-only fieldset, and both sides rewrote the same block. The merge keeps the fieldset and re-applies the prop inside it, so every other change #2972 made to the file came through untouched.

The checks on the current head do not carry this: the review job on the previous head failed rather than declining. It ran twice, dying first on a `400 Bad Request` from the AI binding and then, in the same `model_review` stage, on a review context of 1,055,775 bytes against the 1,048,576-byte budget. The failure shows only as a red mark on the previous commit, and the newest review on the PR is still the approval from 7 September.

ascorbic:
Thanks!

PR #2970, opened 2026-09-09
Keep the edit proxy on collection reads in edit mode

### LINKED ISSUE #2785: getEmDashCollection always returns a no-op edit proxy, so visual editing never works on list pages
## Summary

`getEmDashCollection` always returns entries whose `edit` is a no-op proxy, in every mode. Visual editing therefore cannot work on any list page, for any collection — spreading `{...entry.edit.title}` on a list renders no annotation and the toolbar never picks the element up.

`getEmDashEntry` is unaffected: it has an edit/preview branch that returns before the caching path. `getEmDashCollection` has no equivalent.

The type gives no hint of the difference — `ContentEntry.edit` is a required `EditProxy` on both.

**Affected version:** `emdash@0.30.0`. Verified against `main` at 1717d31; line numbers below are given for both.

## Mechanism

`getEmDashCollection` routes unconditionally through the request cache:

```
getEmDashCollection → requestCached → loadCollectionCached
  → cachedQuery(…)                    // entries stored via entrySnapshot()
  → snapshot.value.entries.map(reviveEntry)
```

and `reviveEntry` hardcodes the no-op:

```ts
// packages/core/src/query.ts
return { ...entry, data, edit: createNoop() } as ContentEntry<D>;
```

The comment above that line already states the intent — "rebuilt to the ContentEntry shape with a no-op edit proxy". An `EditProxy` cannot survive `entrySnapshot()`, so the real proxy built in `getEmDashCollectionUncached`:

```ts
edit: isEditMode ? createEditable(type, dbId, entryEditOptions(entry)) : createNoop(),
```

is unreachable through the public API. It is constructed, serialized away, and replaced.

`getEmDashEntry` builds the identical expression, but edit-mode requests never reach its caching path:

```ts
const serveDrafts = isPreviewMode || isEditMode;
…
if (serveDrafts) {
  …  // returns via wrapEntry() → createEditable()
}
```

Its own comment further down says so: *"Preview/edit requests took the `serveDrafts` branch above."*

| What | `main` (1717d31) `packages/core/src/query.ts` | built `emdash@0.30.0` `dist/query-DR73ZNfm.mjs` |
| --- | --- | --- |
| `getEmDashCollection` (no bypass) | :382 | :195 |
| `loadCollectionCached` | :424 | :207 |
| revive call on collection entries | :454 | :236 |
| `reviveEntry` → `edit: createNoop()` | :666 → :681 | :376 → :390 |
| unreachable `createEditable` (collection) | :750 | :430 |
| `serveDrafts` bypass (entry only) | :806 / :856 | :467 / :506 |
| revive call on the entry path | :980 | :617 |

## Reproduction

1. A collection with a routable detail page and a list page.
2. Detail page: `<h1 {...entry.edit.title}>` from `getEmDashEntry` — annotated, editable in the toolbar.
3. List page: `<h2 {...entry.edit.title}>` from `getEmDashCollection` — no annotation, not editable.

Same collection, same field, same edit-mode session.

## Impact

- Editors can only edit through detail pages. Anything shown exclusively on a list (a summary, a date, a card title) is not reachable by visual editing.
- A collection deliberately built without detail pages — an anchor-list page, for instance — has no editable surface at all, though nothing in the API or types says so.
- The failure is silent. `entry.edit.title` exists and spreads cleanly; it just yields nothing.

## Note on what this is not

This is not the object cache being stale. `loadCollectionCached` is on the path even for a cold request within a single page render — `requestCached` is per-request de-duplication, and the snapshot round-trip happens regardless of whether anything was cached. Disabling the object cache does not restore the proxy.

## Possible direction

The narrow fix mirrors what the entry path already does: skip the snapshot round-trip when `ctx.editMode` is set, and return `getEmDashCollectionUncached`'s result directly. Edit-mode requests are already exempt from route caching, so the lost de-duplication only affects sessions that are editing.

Happy to send a PR if that shape suits you.

jakevis:
## What does this PR do?

Fixes visual editing on list pages. Entries returned by `getEmDashCollection` always carried a no-op `edit` proxy, in every mode, so spreading `{...entry.edit.title}` on a list rendered no annotation and the toolbar never picked the element up. Only pages built from `getEmDashEntry` were click-to-edit.

`getEmDashCollection` routed every read through `loadCollectionCached`, which reduces entries to a JSON snapshot (`entrySnapshot`) and rebuilds them with `reviveEntry`. A `Proxy` cannot survive that round-trip, so `reviveEntry` re-attaches the no-op — discarding the `createEditable` proxy that `getEmDashCollectionUncached` built one line earlier.

The round-trip ran unconditionally, not only on a cache hit, so the annotation was lost even with no L2 backend configured. `shouldBypass()` governs whether a snapshot is *stored*, not whether entries pass through it, which left this invariant — stated above `entrySnapshot` — unenforced on the load path:

> L2 is only consulted for anonymous, non-preview, non-edit requests (see `shouldBypass` in object-cache), where `edit` is always the no-op variant — so dropping and recreating it is lossless.

This bypasses the snapshot for preview and edit-mode requests, mirroring the `serveDrafts` branch `getEmDashEntry` already uses for the same reason. The request-scoped cache still collapses duplicate queries within a render, so repeated widget reads are unaffected; only the distributed snapshot is skipped, and only for requests that must never be shared anyway.

Closes #2785

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable) — not applicable, no admin UI strings
- [ ] New features link to an approved Discussion — not applicable, bug fix
- [ ] I have included screenshots below if this PR changes the UI — not applicable, no UI change

## Screenshots / test output

No UI change.

`tests/unit/query-collection-edit-proxy.test.ts` reproduces the bug. Against `main` the first case fails:

```
 × annotates fields in edit mode, matching getEmDashEntry
   AssertionError: expected {} to have property "data-emdash-ref"
 Tests  1 failed | 1 passed (2)
```

With the fix, both pass. The second case pins the other half of the contract — that no annotations are emitted outside edit mode — so the bypass can't leak refs into shared HTML.

Full `packages/core` unit suite on this branch:

```
 Test Files  343 passed (343)
      Tests  4257 passed (4257)
```

`pnpm typecheck`, `pnpm lint` and `pnpm format` all clean.

### How this was found

Reported independently while building a site on `emdash@0.32.0`: FAQ, house-rules and supporters list pages were not editable, while the event detail page was. Confirmed at runtime in a single request with `editMode: true` — `getEmDashCollection` returned `Object.keys(entry.edit) === []` while `getEmDashEntry` returned a real annotation for the same session. Details in [#2785 (comment)](https://github.com/emdash-cms/emdash/issues/2785#issuecomment-5593509192).

### DIFF
--- a/.changeset/olive-crabs-repeat.md
+++ b/.changeset/olive-crabs-repeat.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes visual editing on list pages. Entries from `getEmDashCollection` now carry a working `edit` proxy in edit mode, so spreading `{...entry.edit.title}` renders the annotation and the toolbar makes the element editable. Previously every collection entry received a no-op proxy in every mode, so only pages built from `getEmDashEntry` were click-to-edit — fields shown exclusively in a list, and collections with no detail page, could not be edited on the page at all.
--- a/packages/core/src/query.ts
+++ b/packages/core/src/query.ts
@@ -398,8 +398,21 @@ export async function getEmDashCollection<T extends string, D = InferCollectionD
 	// Cursor-paginated calls are exempt: their limit is part of the
 	// pagination contract.
 	const bucketed = bucketFilter(filter);
+
+	// Preview and edit-mode requests skip `loadCollectionCached`. That path
+	// reduces every entry to a JSON snapshot (`entrySnapshot`) and rebuilds it
+	// with `reviveEntry`, which cannot carry the `edit` proxy and re-attaches
+	// the no-op — so annotations spread as `{...entry.edit.title}` would render
+	// nothing on list pages. `getEmDashEntry` has the same bypass for the same
+	// reason (see its `serveDrafts` branch). The request-scoped cache still
+	// collapses duplicate queries within the render.
+	const ctx = getRequestContext();
+	const serveDrafts = ctx?.editMode === true || ctx?.preview !== undefined;
+
 	const cached = await requestCached(collectionCacheKey(type, bucketed.fetchFilter), () =>
-		loadCollectionCached<T, D>(type, bucketed.fetchFilter),
+		serveDrafts
+			? getEmDashCollectionUncached<T, D>(type, bucketed.fetchFilter)
+			: loadCollectionCached<T, D>(type, bucketed.fetchFilter),
 	);
 	return bucketed.requestedLimit === undefined
 		? cached
--- a/packages/core/tests/unit/query-collection-edit-proxy.test.ts
+++ b/packages/core/tests/unit/query-collection-edit-proxy.test.ts
@@ -1,0 +1,97 @@
+import type { Kysely } from "kysely";
+import { describe, it, expect, beforeEach, afterEach, vi } from "vitest";
+
+import { handleContentCreate } from "../../src/api/index.js";
+import type { Database } from "../../src/database/types.js";
+import { getEmDashCollection, getEmDashEntry } from "../../src/query.js";
+import { runWithContext } from "../../src/request-context.js";
+import { setupTestDatabaseWithCollections, teardownTestDatabase } from "../utils/test-db.js";
+
+vi.mock("astro:content", () => ({
+	getLiveCollection: vi.fn(),
+	getLiveEntry: vi.fn(),
+}));
+
+import { getLiveCollection, getLiveEntry } from "astro:content";
+
+/**
+ * Visual editing on list pages.
+ *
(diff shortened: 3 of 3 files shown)


jakevis:

ascorbic (approved):
Thanks!

PR #3019, opened 2026-09-10
fix(core): give the JSON-LD nodes an @id so they can be referenced

marcin-misiewicz-pl:
## What does this PR do?

Core emits `BlogPosting`, its `publisher`, and `WebSite` without an `@id`. In JSON-LD that makes each an anonymous node: nothing can point at it, and a fuller description of the same thing published alongside it stays a separate entity instead of merging into one.

The visible consequence is on article pages. A site that publishes an Organization graph — from a plugin, or hand-written in a template — ends up with **two organisations**: the rich one, and core's `{ "@type": "Organization", name }` inside `publisher`. A consumer cannot merge them, so the article's publisher is the node carrying nothing but a name, and every property the site took care to state is attached to something the article does not reference.

```jsonc
// today, on one article page
{ "@type": "BlogPosting", "publisher": { "@type": "Organization", "name": "Example" } }
{ "@type": "Organization", "@id": "https://example.com/#organization",
  "logo": …, "address": …, "vatID": …, "sameAs": […] }   // ← unrelated, as far as a consumer can tell
```

`WebSite` has the same problem from the other side: with no `@id` there is nowhere to attach a `potentialAction`, so adding a sitelinks `SearchAction` means emitting a second `WebSite` node and claiming the site is two sites.

This adds three ids, chosen to be joinable:

| Node | `@id` |
| --- | --- |
| `BlogPosting` | `<canonical>#article` — a fragment, because `mainEntityOfPage` already identifies the WebPage by the bare canonical and reusing it would state that the article and its page are one thing |
| `publisher` | `<origin>/#organization` — the conventional form, and the one `url.replace(/\/$/, "") + "/#organization"` already produces |
| `WebSite` | `<origin>#website` |

**The slash before the fragment is load-bearing.** `https://example.com#organization` and `https://example.com/#organization` are different IRIs, and a mismatch does not error — it publishes two organisations instead of one.

`publisher` keeps its `@type` and `name`. A bare `{ "@id": … }` would leave a dangling reference on every site that publishes no Organization graph, which is worse than the anonymous node it replaces. **Nothing changes for those sites.**

`buildWebSiteJsonLd` had the origin-resolution chain inline and `buildBlogPostingJsonLd` now needs the same answer, so it moves to a shared helper — two copies would eventually disagree, and for a value used to build an `@id` that means silently publishing two entities. Deliberately **not** `resolveSiteOrigin()` from `absolute-url.ts`: that gives `SiteSettings.url` precedence over `page.siteUrl`, which would change which origin these graphs carry. Plausibly an improvement, but a behaviour change that does not belong in a fix about node identity.

Closes #

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

- [x] This PR includes AI-generated code — model/tool: Claude Opus 5 (via Claude Code)

## Screenshots / test output

Four tests added to `packages/core/tests/unit/plugins/page-seo.test.ts`. Three fail before the change; the fourth guards against the regression the obvious fix would introduce (a bare `@id` publisher losing its `@type`/`name`), and passes both before and after.

```
Test Files  1 failed (1)          # before
     Tests  3 failed | 11 passed (14)

Test Files  1 passed (1)          # after
     Tests  14 passed (14)
```

Full package suite:

```
Test Files  536 passed | 2 skipped (538)
     Tests  6442 passed | 10 skipped (6452)
```

### DIFF
--- a/.changeset/olive-cups-shine.md
+++ b/.changeset/olive-cups-shine.md
@@ -4,6 +4,6 @@
 
 Fixes the JSON-LD that core emits on public pages, which published nodes nothing could reference. `BlogPosting`, its `publisher`, and `WebSite` had no `@id`, so an Organization graph describing the site — from a plugin or written into a template — stood beside the article's publisher as a second, competing organisation instead of merging with it. Article pages carried two organisations, and the fuller one was not the article's publisher.
 
-Each of the three nodes now carries an `@id`: `<canonical>#article`, `<origin>/#organization`, and `<origin>#website`. The publisher keeps its `@type` and `name`, so a site publishing no Organization graph of its own is unaffected.
+Each of the three nodes now carries an `@id`: `<canonical>#article`, `<origin>/#organization`, and `<origin>/#website`. The publisher keeps its `@type` and `name`, so a site publishing no Organization graph of its own is unaffected.
 
 Sites already emitting an Organization graph should confirm its `@id` is `<origin>/#organization` — with the slash before the fragment. `https://example.com#organization` is a different IRI, and a mismatch produces two organisations rather than one.
--- a/.changeset/olive-cups-shine.md
+++ b/.changeset/olive-cups-shine.md
@@ -1,0 +1,9 @@
+---
+"emdash": patch
+---
+
+Fixes the JSON-LD that core emits on public pages, which published nodes nothing could reference. `BlogPosting`, its `publisher`, and `WebSite` had no `@id`, so an Organization graph describing the site — from a plugin or written into a template — stood beside the article's publisher as a second, competing organisation instead of merging with it. Article pages carried two organisations, and the fuller one was not the article's publisher.
+
+Each of the three nodes now carries an `@id`: `<canonical>#article`, `<origin>/#organization`, and `<origin>#website`. The publisher keeps its `@type` and `name`, so a site publishing no Organization graph of its own is unaffected.
+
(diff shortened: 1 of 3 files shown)


marcin-misiewicz-pl:
Thanks — both findings were right, and the second was a real defect.

**`siteOrigin` normalisation.** Confirmed and fixed. `page.siteUrl` is now parsed through `new URL(...).origin`, falling back to the raw string only when it does not parse. The failure was worse than cosmetic: a trailing slash produced `https://example.com//#organization`, which silently reintroduces the two-organisation problem this PR exists to fix — and only on sites that configure an origin, so it would not have shown up in a default setup. Added a test pinning the slashed form.

**Comments.** Removed. They were PR rationale in code: they justified the choice against `resolveSiteOrigin()` and narrated rejected alternatives, which AGENTS.md excludes. That reasoning is in the commit message and the PR description instead. The `siteOrigin` docstring now states the contract only, taking your suggested wording.

**`WebSite` fragment.** Taken — `<origin>/#website`, matching the organisation's form, so plugin authors extending either node follow one rule. Changeset updated to match.

`pnpm typecheck`, `pnpm lint`, `pnpm format` clean; `packages/core` is 6443 passed | 10 skipped.

The CLA is still outstanding and I will sign it separately.

marcin-misiewicz-pl:

PR #3031, opened 2026-09-10
fix(plugin-forms): give each embedded form its own element ids

### LINKED ISSUE #2909: plugin-forms: element ids collide when the same form is embedded more than once on a page
`src/astro/FormEmbed.astro` builds element ids from the form id and the field name (line 38):

```
function fieldId(name: string): string {
	return `${formId}-${name}`;
}
```

Those ids are used for every input (lines 96, 112, 121, 134, 169), for the honeypot (line 211), and as the `for` target of each `<label>` (line 75). They do not include anything that distinguishes one rendering of a form from another, so embedding the same form twice on a page produces duplicate ids.

This is not a hypothetical: our article template carries the newsletter form in the sidebar, again in the footer and again in a pop-up, so a single page has three elements with `id="newsletter-email"`. Clicking any of the three labels focuses the first one, and anything that resolves an id (a label, an `aria-describedby`, a test selector, a password manager) reaches the wrong instance.

**Suggested fix:** include a per-instance counter or a random suffix in `fieldId`, generated once per rendered form. Astro can do this at render time, so it costs nothing at runtime and the `for` and `id` pair stays consistent within an instance.

We work around it locally by generating our own instance-scoped ids for the error messages we associate, but we cannot fix the `label for` pairs from outside the component.

Environment: `@emdash-cms/plugin-forms` 0.2.5, EmDash 0.36, Astro 7.

eisenbruch:
## What does this PR do?

Fixes duplicate element IDs when the same form is embedded more than once on a page.

`fieldId()` built IDs from the form ID and the field name alone:

```ts
function fieldId(name: string): string {
	return `${formId}-${name}`;
}
```

Every rendering of a given form therefore produced the same IDs. Each rendering now carries a short per-instance suffix, so labels, inputs and the honeypot stay paired within their own copy.

The honeypot built its ID inline in two places (`` `${formId}-_hp` ``) and had the same problem, so it goes through `fieldId()` now too.

Closes #2909

### Verified on a live site

A site embedding its newsletter form three times per page — a rail panel, an inline module and a pop-up — serves two IDs three times each on every page:

```
duplicate ids: newsletter-email x3, newsletter-interests x3
```

Clicking the second or third label focuses the first input, which is the reported symptom.

### Why this is safe to change

Element IDs are not part of the plugin's public API, and nothing outside the component reads them: `fieldId()` is used only inside `FormEmbed.astro`, and the client script scopes its lookups to the form element (`form.querySelector(...)`, `[data-error-for="name"]`). The one document-wide lookup, in `handlePopState`, matches on `[data-ec-form][data-form-id=...]` rather than an element ID, so multi-page navigation is unaffected. No tests or e2e selectors reference the ID shape.

`data-form-id` deliberately still repeats across copies of the same form — that is the form's identity, not an element identity, and changing it is out of scope here.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] I have added/updated tests for my changes (if applicable) — see note
- [ ] I have included screenshots below if this PR changes the UI — described instead, see below

Honest notes on the ticks:

- **No test added.** `packages/plugins/forms/tests` covers schemas, the public definition and the client submit response; there is no harness that renders `FormEmbed.astro`, and adding one for this felt like more surface than the fix warrants. If you would like the component rendered under `experimental_AstroContainer` — asserting that two renderings of the same form share no element IDs — say so and I will add it.
- `pnpm --filter @emdash-cms/plugin-forms test` is 19/19, `pnpm lint:json` reports nothing for the touched file, and `pnpm format` is clean.
- No new strings, so nothing to wrap; no `messages.po` changes included.

## Screenshots / test output

**No screenshot.** The change is invisible — identical markup with different `id` attribute values — and the only site I can reproduce it on is a private pre-launch client site. The observable difference is in the HTML: three copies of a form previously emitted `id="newsletter-email"` three times, and now emit three distinct IDs, with each `<label for>` matching the input beside it.

I have not rendered the patched component. The package ships the `.astro` file as source with no build step, and I could not run it against a realistic page locally. The change is a string suffix and the existing suite passes, but I would rather say that than imply I watched it render.

### DIFF
--- a/.changeset/hungry-pianos-sing.md
+++ b/.changeset/hungry-pianos-sing.md
@@ -1,0 +1,5 @@
+---
+"@emdash-cms/plugin-forms": patch
+---
+
+Fixes duplicate element IDs when the same form is embedded more than once on a page. Field IDs were built from the form ID and the field name alone, so a form appearing in, say, a sidebar and a pop-up produced several elements sharing an ID: clicking a label focused the first copy rather than the one beside it, and anything resolving an ID — `aria-describedby`, a password manager, a test selector — reached the wrong instance. Each rendering now suffixes its IDs with a per-instance value, so labels, inputs and the honeypot stay paired within their own copy. Element IDs are not part of the plugin's API and nothing else references them; the client script scopes its lookups to the form element.
--- a/packages/plugins/forms/src/astro/FormEmbed.astro
+++ b/packages/plugins/forms/src/astro/FormEmbed.astro
@@ -34,9 +34,14 @@ const hasFiles = form.pages.some((p: FormPage) =>
 	p.fields.some((f: FormField) => f.type === "file")
 );
 
-/** Generate an element ID for a field */
+/** Suffix that distinguishes this rendering from any other on the same page. The same form is often embedded more
+ * than once - a sidebar, a footer and a pop-up - and without it every copy repeats the same element ids, so a
+ * label, an aria-describedby or a password manager resolves to the first copy rather than the one being used. */
+const instance = Math.random().toString(36).slice(2, 8);
+
+/** Generate an element ID for a field, unique to this rendering of the form */
 function fieldId(name: string): string {
-	return `${formId}-${name}`;
+	return `${formId}-${name}-${instance}`;
 }
 ---
 
@@ -205,10 +210,10 @@ function fieldId(name: string): string {
 				style="position:absolute;left:-9999px;"
 				aria-hidden="true"
 			>
-				<label for={`${formId}-_hp`}>Leave blank</label>
+				<label for={fieldId("_hp")}>Leave blank</label>
 				<input
 					type="text"
-					id={`${formId}-_hp`}
+					id={fieldId("_hp")}
 					name="_hp"
 					tabindex="-1"
 					autocomplete="off"
--- a/packages/plugins/forms/tests/form-embed-ids.render.test.ts
+++ b/packages/plugins/forms/tests/form-embed-ids.render.test.ts
@@ -1,0 +1,80 @@
+/**
+ * Renders FormEmbed twice, the way a page embedding the same form in a rail and a pop-up does, and pins that the
+ * two renderings share no element ids.
+ */
+import { experimental_AstroContainer as AstroContainer } from "astro/container";
+import { beforeAll, describe, expect, test, vi } from "vitest";
+
+import type { PublicFormDefinition } from "../src/public-definition.js";
+
+const definition: PublicFormDefinition = {
+	name: "Newsletter",
+	slug: "newsletter",
+	pages: [
+		{
+			fields: [
+				{
+					id: "email",
+					type: "email",
+					label: "Email",
+					name: "email",
+					required: true,
+					width: "full",
+				},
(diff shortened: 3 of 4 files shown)


eisenbruch:
Added the regression test you asked for.

`tests/form-embed-ids.render.test.ts` renders `FormEmbed` twice through `experimental_AstroContainer` and asserts three things: the two renderings share no element ids, ids are unique within a rendering, and every `<label for>` resolves to an input in its own copy.

It fails on the previous `fieldId()` and passes on this branch:

```
× two renderings of the same form share no element ids
  AssertionError: expected [ 'newsletter-email', …(2) ] to deeply equal []
```

One thing worth flagging for review: the package's `vitest.config.ts` now goes through astro's `getViteConfig` so a `.astro` component can be imported at all. That follows `packages/core`'s repro render tests, and the existing `include` glob and node environment are unchanged — the three existing suites still run and pass (22/22 across 4 files). Happy to move the test to a separate config and script instead if you would rather keep the default one plain.

eisenbruch:
This PR's diff has now been run on a real site, which it did not have before. Posting the result because the PR has been open a while with no verification beyond CI.

We applied this exact diff as a local patch over `@emdash-cms/plugin-forms@0.2.6` (fetched with `gh pr diff 3031`, applied unchanged) and deployed it to our staging environment on 0.38.0. On a page carrying two embeds of the same form, a newsletter module and a subscribe popup:

- 2 forms rendered, **0 duplicate ids**
- every `<label for>` resolves to an input inside its own form
- clicking the popup's "Email" label keeps focus in the popup

Before the patch, on the same page, clicking the popup's "Email" label moved focus to the form behind the modal. That is the user-visible shape of #2909 and the reason I care about it: the duplicate ids are the defect, but the symptom is focus landing in a form the user cannot see, behind an open modal, which is an accessibility problem rather than a cosmetic one.

To be exact about what that is and is not: this is our own patched build of 0.2.6 with this diff, on staging, checked in a browser. It is not a test of the branch as CI builds it, and the verification was done by a colleague on our side rather than by me personally. No code changes are proposed here; the PR stands as it is.

Production has not had it yet. If anything about the diff changes before merge, treat the above as applying to the current revision only.

eisenbruch:
Update to my comment above: this is now on our production site, and the earlier note saying production had not had it is out of date.

Read from the served HTML of the live homepage rather than from the build, and checked by me directly this time. The same newsletter form is embedded twice on that page, and the ids are now:

```
newsletter-email-h0flzc      newsletter-email-6jifcq
newsletter-interests-h0flzc  newsletter-interests-6jifcq
```

Across the whole page: 5 forms, 18 ids, **0 duplicates**, and all 4 `<label for>` targets resolve to an existing id. Without this diff both embeds produce `newsletter-email`, and the popup label moves focus into the form behind the modal.

Still the same qualification as before on what this is: our own patched build of `@emdash-cms/plugin-forms@0.2.6` carrying this PR's diff unchanged, not the branch as CI builds it. But it is now a live site rather than staging, serving real traffic, with the fix behaving as intended.

PR #3032, opened 2026-09-10
fix(plugin-forms): let the plugin's validation run and its error spans fill

### LINKED ISSUE #2908: plugin-forms: the form has no novalidate, so the plugin's own validation never runs and its error spans stay empty
`@emdash-cms/plugin-forms` 0.2.5 renders `<form class="ec-form" method="POST" ...>` in `src/astro/FormEmbed.astro` (line 43) with no `novalidate`, while its own validation lives in a delegated submit handler in `src/client/index.ts` (line 62, `document.addEventListener("submit", handleSubmit)`).

On any browser that runs constraint validation, the browser blocks the submit event before that handler is reached. The consequence is that `handleSubmit` never runs on an invalid form, `showErrors` (line 385) is never called, and the `[data-error-for]` spans the template renders (line 191) stay empty for the whole life of the page. The reader sees the browser's own bubble instead, and any styling or wiring built around those spans appears dead.

We noticed because we were adding ARIA associations to those spans and could not make them fill, on any field, with any input. Setting `noValidate` on the form from JavaScript makes the plugin's validation run exactly as documented.

**Suggested fix:** set `novalidate` on the form and let the plugin own validation, or drop the plugin's client-side validation and lean on the browser's. Setting it from JavaScript rather than in the markup keeps native validation for readers without JavaScript, which is the behaviour we settled on locally.

**Second, related problem in the same area.** When the plugin does show an error, nothing ties the message to the field it belongs to: no `aria-invalid` on the input, no `id` on the message, no `aria-describedby` from the input, and focus is not moved to the first failure. A screen-reader user is told the form failed and has no way to learn which field or why. `showFieldError` (line 381) already has both the field name and the message element in hand, so the association is a few lines there.

Environment: `@emdash-cms/plugin-forms` 0.2.5, EmDash 0.36, Astro 7, Cloudflare Workers. Reproduced on a live site with three embedded forms.

eisenbruch:
## What does this PR do?

Fixes the plugin's own field errors never rendering.

`FormEmbed.astro` emits `<form class="ec-form" method="POST" ...>` with no `novalidate`, while validation lives in a delegated submit handler. On any browser running constraint validation, the submit event is blocked before that handler is reached, so `validateVisibleFields()` never runs, `showFieldError()` is never called, and the `[data-error-for]` spans the template renders stay empty for the life of the page. The reader sees the browser's own bubble instead, and anything built around those spans — styling, ARIA associations — appears dead.

The client script now turns native validation off as it initialises each form:

```ts
form.dataset.ecInitialized = "1";
form.noValidate = true;
```

Closes #2908

### Why from JavaScript rather than `novalidate` in the markup

Adding `novalidate` to the template would switch validation off for everyone, including readers whose JavaScript never loads — and for them the plugin's handler never runs either, so they would submit an invalid form to the server with no client-side check at all. Setting it during init keeps native validation as the fallback and hands over only once the script that replaces it is actually running.

This also matches the pattern the file already documents at the top: *"Following the same progressive enhancement pattern as Astro's `<ClientRouter />`"*.

### Nothing changes about which inputs are valid

`validatePage()` and `validateVisibleFields()` already call `input.checkValidity()` and read `input.validationMessage` — the same constraint validation API the browser was using. The rules and the message text are identical; only the presentation moves, from a browser bubble to the plugin's own error span, which is what the template was built for.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] I have added/updated tests for my changes (if applicable) — see note
- [x] User-visible strings in the admin UI are wrapped for translation (if applicable)
- [ ] I have included screenshots below if this PR changes the UI — described instead, see below

Honest notes on the ticks:

- **No test added.** The behaviour needs a real browser: it depends on the browser blocking a submit event, which is precisely what a jsdom-style harness does not do. `packages/plugins/forms/tests` has no browser harness today. If you would like this covered in the Playwright suite — submit an invalid embedded form and assert the `[data-error-for]` span fills — I am happy to add it, but that is a new e2e fixture and felt like a separate decision.
- `pnpm --filter @emdash-cms/plugin-forms test` is 19/19, `pnpm lint:json` reports nothing for the touched file, `pnpm format` is clean.
- No new strings; no `messages.po` changes included.

## Screenshots / test output

**No screenshot.** The visible difference is a browser's native validation bubble being replaced by the plugin's own error text inside the form, and the only site I can reproduce it on is a private pre-launch client site.

I have not watched the patched script run. The change is one assignment on a form element, the reporter of #2908 confirmed that setting `noValidate` from JavaScript makes the plugin's validation run as documented, and the existing suite passes — but I would rather state that than imply I verified it in a browser.

### DIFF
--- a/.changeset/tidy-donkeys-repeat.md
+++ b/.changeset/tidy-donkeys-repeat.md
@@ -1,0 +1,5 @@
+---
+"@emdash-cms/plugin-forms": patch
+---
+
+Fixes the plugin's own field errors never appearing. Native constraint validation blocked the submit event before the plugin's handler could run, so an invalid form showed the browser's default bubble and the `[data-error-for]` spans the template renders stayed empty for the life of the page — leaving any styling or ARIA wiring built around them apparently dead. The client script now turns native validation off as it initialises each form, so its own messages render in those spans. It is set from JavaScript rather than in the markup, so a reader without JavaScript keeps native validation; the plugin applies the same `checkValidity()` rules and the same `validationMessage` text, so nothing about which inputs are considered valid changes.
--- a/packages/plugins/forms/src/client/index.ts
+++ b/packages/plugins/forms/src/client/index.ts
@@ -47,6 +47,11 @@ export function initForms() {
 		document.querySelectorAll<HTMLFormElement>("[data-ec-form]").forEach((form) => {
 			if (form.dataset.ecInitialized) return;
 			form.dataset.ecInitialized = "1";
+			// Hand validation to this script now that it is running. Native constraint validation blocks the submit
+			// event before handleSubmit can reach it, so the field errors below never render and the [data-error-for]
+			// spans stay empty. Set here rather than in the markup so a reader without JavaScript keeps the browser's
+			// own validation; validateVisibleFields() then applies the same checkValidity() rules and messages.
+			form.noValidate = true;
 			restoreState(form);
 			initMultiPage(form);
 			initConditions(form);
--- a/packages/plugins/forms/src/client/index.ts
+++ b/packages/plugins/forms/src/client/index.ts
@@ -47,10 +47,7 @@ export function initForms() {
 		document.querySelectorAll<HTMLFormElement>("[data-ec-form]").forEach((form) => {
 			if (form.dataset.ecInitialized) return;
 			form.dataset.ecInitialized = "1";
-			// Hand validation to this script now that it is running. Native constraint validation blocks the submit
-			// event before handleSubmit can reach it, so the field errors below never render and the [data-error-for]
-			// spans stay empty. Set here rather than in the markup so a reader without JavaScript keeps the browser's
-			// own validation; validateVisibleFields() then applies the same checkValidity() rules and messages.
+			// Set here, not in the markup, so a reader without JavaScript keeps native validation.
 			form.noValidate = true;
 			restoreState(form);
 			initMultiPage(form);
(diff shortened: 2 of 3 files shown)


eisenbruch:
Both points addressed.

**Comment trimmed** to the one thing that is not obvious from the code:

```ts
// Set here, not in the markup, so a reader without JavaScript keeps native validation.
form.noValidate = true;
```

**Test added**, `tests/client-init.test.ts`, in jsdom. It fails without the fix:

```
× turns native validation off so the plugin's own errors can render
  AssertionError: expected false to be true
× does it for every embedded form, not just the first
  AssertionError: expected [ false, false ] to deeply equal [ true, true ]
```

So the assignment cannot be dropped again without a red test, which was the concern.

**On Playwright specifically:** I did not add one, and the reason is worth your call rather than mine. The behaviour that needs a real browser is the submit event no longer being blocked, and `e2e/fixture` does not install this plugin — so a spec means adding the plugin to the fixture, seeding a form and adding a public page that embeds it. That is a new fixture surface rather than a test, and I would rather you decided whether you want it than have me grow the fixture unasked. Happy to do it if you do.

What the jsdom test pins is narrower but real: that the flag is set, for every embedded form, and that the markup does not set it — so the no-JavaScript fallback stays intact.

Issue #3043, opened 2026-09-10
Content list cannot be filtered by taxonomy term: fieldFilters reaches indexed columns but never content_taxonomies


eisenbruch:
## Description

A collection's content list can be filtered by status, author, byline, date range, free text, and any **indexed custom field**. It cannot be filtered by a **taxonomy term**, in the admin or over the API, even on a collection where a taxonomy is attached and every entry carries a term.

So a taxonomy can organize a site's content, drive its URLs and its navigation, and still be invisible in the one screen an editor works in all day.

## Measured

EmDash core 0.37.0, `@emdash-cms/admin` 0.36.0, on a site with several collections that each have a taxonomy attached (one flat, one hierarchical) and a term on every entry, from hundreds to thousands of entries per collection. None of those lists offers a term filter. Every one offers exactly: search, `All statuses`, `All authors`, `All bylines`, `Date range`.

Over the API:

```
GET /content/{collection}?limit=1                                    total N
GET /content/{collection}?limit=1&fieldFilters={"<indexed field>":"x"}  filtered   ← works
GET /content/{collection}?limit=1&fieldFilters={"<taxonomy>":"x"}       error      ← not a field
GET /content/{collection}?limit=1&taxonomy=<taxonomy>&term=x            total N    ← ignored
```

The last line is the one that costs time: an unknown query parameter is silently dropped, so a caller filtering by term gets a complete, plausible, unfiltered list back with a 200.

## Why this is a small gap rather than a missing feature

Both halves already exist and neither needs inventing.

**The filter mechanism is there.** `contentListQuery` (`src/api/schemas/content.ts:139`) already accepts `fieldFilters`, documented as "JSON-encoded indexed custom-field filters, combined with AND semantics". `ContentFieldFilters` (`src/content-list-query.ts:29`) supports exact match, `null`, `{ in: [...] }` and range bounds. `handleContentList` (`src/api/handlers/content.ts:538`) assembles a `where` from it, and `ContentRepository.findMany` resolves it at `content.ts:703`. What it filters is columns on the `ec_*` table. Terms are not columns; they live in `content_taxonomies`, and nothing in that query path joins it.

**The UI pattern is there.** The list already renders four filter controls, each driving a query parameter. A collection already declares which taxonomies apply to it, so the admin can render one control per applied taxonomy with no per-site configuration, and collections without a taxonomy get nothing new.

## Why it matters

[...]

PR #2799, opened 2026-08-30
feat: make the reference field type merge-ready

saariuslystoned:
## What does this PR do?

This is a tested, author-preserving continuation of [#1928](https://github.com/emdash-cms/emdash/pull/1928), opened directly against upstream `main` so the result can run through EmDash's authoritative CI, bot, preview, visual, query-count, security, and review rails.

MA2153's work remains the foundation:

- The merge commit keeps the exact #1928 head, `808cde87a34fa818a2f29ddafacf5053d3859398`, as its first parent and preserves the original commits and authors.
- It merges upstream `main` at `1717d31b351164a5f78e95fe004ee582c7c50f40` and resolves the three conflicts in the content editor, transaction helper, and route tests.
- Focused continuation commits enforce required and single-reference constraints across create, update, direct edge writes, and seed application, including translation-create inheritance.
- Storage-less reference fields stay outside `body.data` validation while the existing top-level `references` payload remains accepted.
- No legacy-data migration, public-query hydration, MCP behavior, Block Kit work, Dinkus work, or unrelated cleanup is included.

This follows the maintainer direction on #1928 to resolve the current-main conflicts without adding migration scope. It does not automatically close #1928; maintainers can coordinate which PR remains canonical.

Approved feature discussion: [#386](https://github.com/emdash-cms/emdash/discussions/386)

## Type of change

- [x] Bug fix
- [x] Feature (requires [maintainer-approved Discussion](https://github.com/emdash-cms/emdash/discussions/categories/ideas))
- [ ] Refactor (no behavior change)
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests
- [ ] Chore (dependencies, CI, tooling)

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md)
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (or targeted tests for my change)
- [x] `pnpm format` has been run
- [x] I have added/updated tests for my changes (if applicable)
- [x] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (the continuation adds no new user-visible strings)
- [x] I have added and reviewed the user-facing [changeset](https://github.com/emdash-cms/emdash/blob/main/.changeset/README.md)
[...]

PR #1928, opened 2026-07-10
feat: relations, and reference fields as entry pickers

MA2153:
## What does this PR do?

Adds **relations** as first-class schema objects, and turns `reference` fields into entry pickers that link through them — core storage, the site query API, generated types, seeds and export, two migrations, and the admin surface for all of it.

A relation joins two collections under a site-unique slug, with a label for each side and an optional limit on how many entries each side may link. A reference field binds to a relation and views its links from one end, so one relation can back a field on either collection. Links live in `_emdash_content_references` keyed by translation group, not in a column on the collection table, so a selection is shared across an entry's translations.

Feature Discussion: https://github.com/emdash-cms/emdash/discussions/386

Closes #2403

### Core (`emdash`)

**Schema.** `_emdash_relations` becomes one row per relation keyed by a unique `slug` (migration `083_relations_structural`, replacing the per-locale rows migration 043 created), and `_emdash_content_references.relation_group` becomes `relation_id`. Relation IDs are preserved, so existing links stay valid. A relation carries a label and an optional singular form for each role, plus a per-side limit. Deleting a reference field no longer deletes its relation unless the caller passes `deleteRelation=true`; deleting a collection removes every relation it is an end of, along with the fields on the far collection that view them. Reading a relation reports what deleting it would take: the fields bound to it and how many links it holds.

**Storage.** A bound reference field adds no column. `SchemaRegistry` skips the DDL for it, `createContentTable` applies the same guard for seeded fields, and a selection is written as edges in the same transaction as the entry. `relations` joins the reserved collection slugs, since the admin serves its screen at that path.

**Writes.** Entry create and update bodies carry `references`, keyed by field slug. Required and selection-limit constraints are enforced on both ends of the relation, so a field on the parent side that would hand a selected entry more parents than the relation allows is refused, not only one that selects too many children itself. Large replacements are chunked for D1's bind-parameter limit.

**Drafts.** On a collection that keeps revisions, a selection changed on a published entry is staged in the draft and becomes live when the entry is published — through the publish action, a scheduled publish, or a revision restore — and is discarded with the draft. Publishing re-checks the whole selection against the relation's limits, so a draft cannot carry a selection past a schema change that would now reject it. Revision compare reports `_references` on both sides, filled in from the published selection for the fields a draft did not stage, so an unchanged reference field does not read as one the draft removed.

[...]

PR #3046, opened 2026-09-10
feat: filter the content list by taxonomy term

eisenbruch:
## What does this PR do?

Adds `termFilters` to the content list and a dropdown per applied taxonomy in the admin, so a collection organized by a taxonomy can be filtered by it, in the screen editors work in, and over the API.

Related: #3043 (closed in favor of Discussion #3055)

### Why this shape

`fieldFilters` already ANDs its keys over indexed columns. Terms are not columns, so this is a sibling parameter rather than an extension of that one: it keeps "column" and "term" distinguishable, and avoids a collision when a field and a taxonomy share a name.

```
?termFilters={"topics":["rodeo","polo"],"places":["kentucky"]}
```

OR within a taxonomy, AND across taxonomies, which is how the controls read: two terms of one taxonomy widen a search, two taxonomies narrow it.

### Core

One `EXISTS` subquery per key, applied in both `findMany` and `countWithResolvedFilters` so totals and pages agree.

Two details the junction imposes, both of which **match nothing rather than erroring** when got wrong, so they are worth stating:

- `content_taxonomies.entry_id` stores the ec_* row's `translation_group`, not its id, unlike `_emdash_content_bylines.content_id` alongside it, which stores the row id.
- it points at `taxonomies.translation_group`, so a term matches through any of its locale variants unless the list is locale-scoped.

### Two deliberate failure choices

Both are the same principle: **a filter that silently matches everything returns a complete, plausible, unfiltered list with a 200**, which is the hardest kind of wrong answer to notice.

- A taxonomy **not attached to the collection** is a `VALIDATION_ERROR`, not an empty result and not a no-op.
- A key that resolves to **no slugs matches nothing**, rather than degrading to "no filter". This mirrors the existing byline filter, which already does this for an empty id set.

At the API-client and UI layers the same idea appears as: "no selection" is the *absence of a key*, never an empty array.

### Admin

Nothing is configured per site. A collection already declares which taxonomies apply to it, so the controls derive from that: a collection with no taxonomy renders nothing new, and one that gains a taxonomy gains its filter. Terms load per taxonomy and only for those on screen. A hierarchical taxonomy is flattened depth-first and indented, matching the entry editor's picker.

### Not included

Sorting by a term, and multi-select within a taxonomy. Both fit the shape (the parameter already takes an array), but neither is needed for the filtering case and I would rather not guess at the UI for them.

### Discussion

Discussion: https://github.com/emdash-cms/emdash/discussions/3055. I opened this PR before it, which was the wrong order; the Discussion sets out the shape, the two failure choices, how this relates to #1928 and #2799, and a separate question about unknown query parameters being stripped.

## Type of change

- [ ] Bug fix
- [x] Feature (requires [maintainer-approved Discussion](https://github.com/emdash-cms/emdash/discussions/categories/ideas)): see the Discussion above
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes: `packages/core` and `packages/admin` both clean. `packages/admin` has one pre-existing error in `src/components/editor/ordered-list.ts`, present identically on a clean `origin/main` in my environment.
- [x] `pnpm lint` passes: `oxlint --type-aware` clean on every changed file
- [x] `pnpm test` passes (or targeted tests for my change): see the note below
- [x] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n): the two new strings use `useLingui`; no `messages.po` changes included
- [x] New features link to a Discussion: https://github.com/emdash-cms/emdash/discussions/3055

- [x] This PR includes AI-generated code — model/tool: Claude Opus 4.6 (Claude Code)

## Screenshots / test output

**The new control, on a collection with the `category` taxonomy applied**, and opened to show that taxonomy's terms. Captured on `pnpm ux:site:start` with the `editorial-small` profile.

Posts list toolbar showing, left to right: a search box, All statuses, All bylines, All categories, Created, and Date range. All categories is the new control and matches the styling of the others.

The All categories dropdown opened over the Posts list, listing All categories with a check beside it, then News, Tutorials and Opinion, the terms of the category taxonomy seeded by the editorial-small profile.

**Verified against a running site** (`pnpm ux:site:start`, `editorial-small`, patched core and admin), which is the more useful evidence anyway:

```
GET /content/posts?limit=1
  → {"success":true,"data":{"items":[…],"total":3}}

GET /content/posts?limit=1&termFilters={"category":["opinion"]}
  → {"success":true,"data":{"items":[],"total":0}}
    (correct: the profile seeds terms but assigns none)

GET /content/posts?limit=1&termFilters={"nosuchtaxonomy":["x"]}
  → {"success":false,"error":{"code":"VALIDATION_ERROR",
     "message":"Taxonomy not applied to posts: nosuchtaxonomy"}}
```

The admin request carries it correctly: `…&termFilters=%7B%22category%22%3A%5B%22opinion%22%5D%7D` → 200.

Worth recording that the middle case is exactly the bug being fixed: **before rebuilding core, that same request returned all three rows**, because the unknown parameter was dropped silently.

**Unit tests:** seven added, covering OR within a taxonomy, AND across taxonomies, an empty filter matching nothing, an unknown slug matching nothing, totals staying stable across cursor pages, and combining with an indexed field filter. They seed `content_taxonomies` by hand precisely to pin the two junction details above.

```
tests/database/repositories/content.test.ts   77 passed
tests/unit/api/schemas.test.ts + content       128 passed (both files)
```

**On the wider suite, stated plainly rather than hidden:** `tests/unit tests/database` reports 36 failed files / 458 failed tests on this branch, and **exactly the same 36 / 458 on a clean `origin/main`** in my environment, from what looks like a local `better-sqlite3` native module problem. Same run: 3,954 passing here against 3,947 on main, i.e. the seven added. I could not get a green baseline to compare against, so I am flagging it rather than claiming green.

`tests/database/repositories/content.test.ts` also passes on Node 22.23.2 (77 passed), after a follow-up commit that moves `applyTermFilters` below `applyBylineFilter` so the byline method keeps its own doc comment.

### DIFF
--- a/.changeset/great-eyes-brake.md
+++ b/.changeset/great-eyes-brake.md
@@ -5,7 +5,7 @@
 
 Adds taxonomy-term filtering to the content list, in the admin and over the API.
 
-A collection's list could be narrowed by status, author, byline, date range, free text and any indexed custom field, but not by a taxonomy term — so a site organised by categories or tags could not be browsed by them in the screen editors work in.
+A collection's list could be narrowed by status, author, byline, date range, free text and any indexed custom field, but not by a taxonomy term, so a site organized by categories or tags could not be browsed by them in the screen editors work in.
 
 The admin now shows one dropdown per taxonomy applied to the collection, beside the existing filters. Nothing is configured per site: a collection already declares which taxonomies apply to it, so a collection with no taxonomy is unchanged and one that gains a taxonomy gains its filter. Hierarchical taxonomies are indented.
 
@@ -15,6 +15,6 @@ Over the API, `GET /content/{collection}` accepts `termFilters`, a JSON object k
 ?termFilters={"topics":["rodeo","polo"],"places":["kentucky"]}
 ```
 
-An entry matches **any** of a taxonomy's slugs and **every** taxonomy named — OR within a taxonomy, AND across taxonomies. Term slugs resolve through their translation group, so a term matches across locales unless the request is locale-scoped. The filter composes with `fieldFilters` and every existing filter, and `total` reflects it, so cursor pages stay consistent.
+An entry matches **any** of a taxonomy's slugs and **every** taxonomy named: OR within a taxonomy, AND across taxonomies. Term slugs resolve through their translation group, so a term matches across locales unless the request is locale-scoped. The filter composes with `fieldFilters` and every existing filter, and `total` reflects it, so cursor pages stay consistent.
 
 Two cases fail rather than returning a misleading list. A taxonomy that is not applied to the collection is rejected with `VALIDATION_ERROR` instead of being ignored, and a taxonomy given an empty array matches nothing instead of being treated as no filter. Callers that previously passed an unrecognised filter parameter and received an unfiltered list will now see an error where the parameter names a real but unapplied taxonomy.
--- a/.changeset/great-eyes-brake.md
+++ b/.changeset/great-eyes-brake.md
@@ -1,0 +1,20 @@
+---
+"emdash": minor
+"@emdash-cms/admin": minor
+---
+
+Adds taxonomy-term filtering to the content list, in the admin and over the API.
+
+A collection's list could be narrowed by status, author, byline, date range, free text and any indexed custom field, but not by a taxonomy term — so a site organised by categories or tags could not be browsed by them in the screen editors work in.
+
+The admin now shows one dropdown per taxonomy applied to the collection, beside the existing filters. Nothing is configured per site: a collection already declares which taxonomies apply to it, so a collection with no taxonomy is unchanged and one that gains a taxonomy gains its filter. Hierarchical taxonomies are indented.
+
+Over the API, `GET /content/{collection}` accepts `termFilters`, a JSON object keyed by taxonomy name:
+
+```
+?termFilters={"topics":["rodeo","polo"],"places":["kentucky"]}
+```
+
+An entry matches **any** of a taxonomy's slugs and **every** taxonomy named — OR within a taxonomy, AND across taxonomies. Term slugs resolve through their translation group, so a term matches across locales unless the request is locale-scoped. The filter composes with `fieldFilters` and every existing filter, and `total` reflects it, so cursor pages stay consistent.
+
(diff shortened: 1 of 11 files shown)


eisenbruch:
Two notes on the CI run, to save a reviewer the digging:

- The 10 visual-regression diffs are all the new "All categories" control in the content-list filter bar; the fixture's `posts` collection has the `category` taxonomy applied, so it now renders there. Nothing else on those screens moved.
- The D1 Tests failure is `migrations-d1.test.ts > replays every migration written since the guarded-DDL pattern` timing out at 30s. This PR adds no migrations, and the same job failed on `main` yesterday (22:52 UTC) and passed on later runs.

eisenbruch:
Rebased onto `main` after #3364 (bulk tagging), and the review findings addressed in `dca23a12`:

- **Clear filters and term filters.** `FilterBar` now counts a term selection as an active filter (`isTermFilterActive`, exported from `TermFilters.tsx`) and `handleClear` resets it. New test: with only a term selected, Clear filters is offered and resets it to `{}`; with none selected, it is not offered. The first test fails on the previous commit and passes on this one.
- **Labels.** The taxonomy label goes to Lingui as written (`All ${label}`, `Filter by ${label}`), no `toLowerCase()`.
- **Indentation.** Child terms are indented with `paddingInlineStart` in rem, the same pattern `TaxonomySidebar` uses, instead of literal non-breaking spaces.
- The handler comment on the unattached-taxonomy rejection is cut to one line.

The rebase conflicted only on the imports in `repositories/content.ts` (the datetime normalization import added on `main`). Checked: `ContentList.test.tsx` and `TaxonomySidebar.test.tsx` (108 tests), core `repositories/content.test.ts` and `api/schemas.test.ts` (133), admin `tsgo --noEmit`, core `tsc --noEmit`, `oxfmt --ignore-path .gitignore --check`, `oxlint --type-aware --deny-warnings`.

eisenbruch:
CI on `dca23a12`, two failures, neither in the change itself:

- **Smoke Isolated Install / Tests**: `isolated-install.test.ts:383` "The symbol `fetchWithServerOutput` has already been declared". That duplicate was on `main` when I rebased and was removed by #3414. Rebased again (`85202179`).
- **Visual Regression, 10 content-list screens**: the only difference is the new "All Categories" term filter between the bylines and date-field filters, in both LTR and RTL. I checked the expected, actual and diff images. The label is untranslated in the RTL pseudo-locale because it is a new string and the catalogs are extracted on merge. These are intended, so a maintainer's `/accept-baselines` is what's needed there.

Issue #3049, opened 2026-09-10
Media delete reports success when the storage delete failed, and the public file route serves objects with no DB row, so a "deleted" file can stay downloadable forever


marks-zyz:
### Description

`DELETE /_emdash/api/media/:id` reports `{"deleted": true}` even when the object was never removed from storage, and the public file route serves objects with no database lookup. Together these mean a "deleted" media file can stay publicly downloadable forever, while the admin shows it as gone.

**Two halves of the problem, both on 0.37.0:**

**1. The storage delete failure is swallowed.** In `packages/core/src/astro/routes/api/media/[id].ts` the DB row is deleted first, then:

```ts
if (emdash.storage) {
  const repo = new MediaRepository(emdash.db);
  await removeUploadAttempt(emdash.storage, repo, result.data.storageKey, {
    allowUntracked: true,
  });          // line 160 — return value ignored
}
return apiSuccess({ deleted: true });   // line 165 — unconditional
```

`removeUploadAttempt` (`packages/core/src/media/upload-attempts.ts:20-25`) catches every storage error, `console.error`s it and returns `false`:

```ts
try {
  await storage.delete(storageKey);
} catch (error) {
  console.error("[media] upload cleanup failed:", error);
  return false;
}
```

The route never reads that boolean, so any storage-side failure (R2 incident, bucket lock, credentials, S3 endpoint error) produces a success response and a deleted DB row.

**2. The public file route needs no DB row.** `packages/core/src/astro/routes/api/media/file/[...key].ts` (90 lines) resolves the key straight against storage. It has zero references to `emdash.db` or `MediaRepository`; the only key-based rejection is the `backups/` prefix. So an object with no `media` row is still served at `/_emdash/api/media/file/<storageKey>` with `200`.

**Why this is worse than an orphan-storage-cost issue:** nothing ever reclaims these objects. `MediaRepository.cleanupPendingUploads` (`packages/core/src/database/repositories/media.ts:617-623`) starts *from* the database (`where status = 'pending'`), so an object whose row is gone is unreachable by the cleanup. The file is invisible in the admin, unreferenced by the DB, permanently public, and only its URL is needed to fetch it. For a site that deletes a file precisely because it should not have been published (personal data, a wrong document, a takedown request), "delete" that silently does nothing is a privacy/GDPR problem, not a housekeeping one.

**Suggested fix:**

[...]

PR #3067, opened 2026-09-11
fix(media): delete the stored file on every delete path and surface storage failures

swissky:
## What does this PR do?

Fixes the first half of #3049: a media delete could report `{ deleted: true }` while the object was never removed from storage, and because the row was already gone nothing could ever find the object again.

- `handleMediaDelete(db, id, storage)` now owns storage deletion. `EmDashRuntime.handleMediaDelete` passes its storage adapter, so the MCP `media_delete` tool removes the file too (its description already promised that; it only deleted the row). The REST route just forwards the result.
- Before the row is deleted, the storage key is registered in `_emdash_media_upload_attempts` with status `cleanup` (`MediaRepository.trackStorageKeyForCleanup`, an upsert on the primary key). The existing cleanup sweep (`runSystemCleanup` → `findUploadAttemptsForCleanup` → `removeUploadAttempt`) already retries rows in that state once the media row is gone, so a failed storage delete is retried on every cron tick instead of being lost. `deleteCompletedUploadAttempts` skips `cleanup` rows so a cron tick landing between the upsert and the row delete cannot reap the marker.
- The delete order stays DB row first, then storage — the ordering #2273 introduced so a delete racing an in-flight upload cannot strand the freshly uploaded object. That existing race test still passes.
- `DELETE /_emdash/api/media/:id` responds `{ deleted: true, storageDeleted: boolean }` (OpenAPI `MediaDeleteResponse`). Additive; the admin client only checks the envelope.

Not changed: the public file route still serves any key that exists in storage (second half of the issue). Gating it on a `media` row adds a query to a logged-out route; with deletes now retried until the object is gone, I'd leave that for a maintainer decision.

Tests (all fail before the fix): route reports `storageDeleted: false` when storage rejects, the object stays reachable and the next `runSystemCleanup` removes it; MCP `media_delete` removes the stored object; a `cleanup`-marked key survives the sweep while its media row still exists.

Refs #3049 (leaves the file-route question open, see above)

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

- [x] This PR includes AI-generated code — model/tool: Cursor + Claude Fable 5.1

## Screenshots / test output

Not applicable (no UI change). i18n and Discussion items are n/a: no admin strings, bug fix.


swissky:
Addressed the review: `ctx.media.delete()` now delegates to `handleMediaDelete` (same row-first order, storage delete, cleanup marker), with a test that fails without the change. Changeset names the plugin API too. (693448187)

packages/core/src/api/handlers/media.ts
```diff
@@ -270,29 +272,38 @@ function isForeignKeyViolation(error: unknown): boolean {
 }
 
 /**
- * Delete media item
+ * Delete a media item and its stored object.
+ *
+ * The object is registered for cleanup before the row is removed, so when
+ * the storage delete fails the cleanup sweep retries it; `storageDeleted`
+ * tells the caller whether the object is already gone.
```
khoinguyenpham04, line 279: could we trim + simplify this comment?

packages/core/src/plugins/context.ts
```diff
@@ -689,16 +690,16 @@ export function createMediaAccessWithWrite(
 		},
 
 		async delete(id: string): Promise<boolean> {
-			const deleted = await mediaRepo.delete(id);
+			const result = await handleMediaDelete(db, id, storage);
```
khoinguyenpham04, line 693: Could we route `local-runtime.delete()` through `handleMediaDelete()` and test storage failure? It currently leaves no cleanup marker, so #3049 still affects the local provider.

packages/core/src/database/repositories/media.ts
```diff
@@ -196,6 +196,28 @@ export class MediaRepository {
...
+	 * leaving the object unreferenced and unreachable.
+	 */
+	async trackStorageKeyForCleanup(mediaId: string, storageKey: string): Promise<void> {
+		const now = new Date().toISOString();
+		await this.db
+			.insertInto("_emdash_media_upload_attempts")
+			.values({
+				media_id: mediaId,
+				storage_key: storageKey,
+				status: "cleanup",
```
khoinguyenpham04, line 211: plese make the cleanup batch advance past repeated failures. right now 100 permanently failing rows can prevent newer deletions from ever being retried.

khoinguyenpham04 (commented):
Thanks! Just a few things.

swissky:
Rebased onto main and addressed all three findings:

- Trimmed the comment on `handleMediaDelete` down to a single line.
- The local media provider's `delete()` now delegates to `handleMediaDelete`, so it gets the same storage cleanup and retry-queue behavior as the other delete paths. A missing id resolves silently (idempotent); an actual deletion failure rejects.
- Cleanup retry fairness: a failed storage deletion now bumps the attempt's `updated_at` (`deferUploadAttemptCleanup`), and `findUploadAttemptsForCleanup` orders by `updated_at`, so a persistently failing key moves behind the rest of the batch instead of blocking it. Covered by a new unit test.

Also aligned `ctx.media.delete()` and the local provider on the same contract: `NOT_FOUND` is reported as `false`/silent success, operational failures throw.

khoinguyenpham04 (approved):
Thanks!

ascorbic (approved):

PR #3071, opened 2026-09-11
perf: add local render CPU benchmark

ascorbic:
## What does this PR do?

Adds a repeatable local CPU benchmark for server-rendered pages in the production-built Cloudflare performance fixture.

`pnpm render-cpu` keeps three distinct measurements instead of averaging them together: module evaluation from `wrangler check startup`, repeated first requests in fresh workerd processes, and batches after route warm-up. It captures V8 CPU profiles through the inspector protocol, drains every streamed response, waits briefly for request-owned continuation work, and reports median active CPU, median absolute deviation, garbage collection, response wall time, and the corresponding cold/warm query count. Raw `.cpuprofile` files and a JSON summary are written outside the repository for flamegraph analysis.

The query-count harness now sends browser-shaped `Accept` headers. This exercises the HTML layout-prefetch path that real navigations use and updates the D1 query snapshots accordingly. It also keeps Astro's agent-detected dev server attached to the harness lifecycle and fails promptly if the seed server exits before becoming ready.

The resulting D1 snapshot increase (`+67` queries across cold and warm routes) is newly measured existing behavior, not runtime work introduced by this PR. The previous `*/*` requests skipped the HTML prefetch branch entirely; browser navigations already take it in production. Making that cost visible gives follow-up optimization work a truthful baseline.

The fixture accepts an alternate Wrangler config path so the same build can be grounded against an isolated deployed Worker without committing account resource IDs.

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests
- [x] Chore (dependencies, CI, tooling)

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] I have added and reviewed the user-facing [changeset](https://github.com/emdash-cms/emdash/blob/main/.changeset/README.md) (if this PR changes a published package)

The localization, changeset, Discussion, and screenshot items are not applicable: this changes repository-only performance tooling and a private fixture, with no published package or UI change.

## Screenshots / test output

Not applicable — no UI changes.

Validated on the merged current `origin/main` head with:

- `pnpm --config.verifyDepsBeforeRun=warn build`
- `pnpm --config.verifyDepsBeforeRun=warn typecheck`
- `pnpm --config.verifyDepsBeforeRun=warn lint`
- `pnpm --config.verifyDepsBeforeRun=warn format:check`
- `pnpm --config.verifyDepsBeforeRun=warn query-counts --target sqlite`
- `pnpm --config.verifyDepsBeforeRun=warn render-cpu -- --route home --startup-runs 1 --cold-runs 2 --requests 2 --batches 1 --warmups 1` (includes D1 query snapshot verification)
- `pnpm --config.verifyDepsBeforeRun=warn render-cpu -- --skip-prepare` (full startup, first-request, and warm baseline across ten routes)

The same fixture was deployed to an isolated benchmark Worker for grounding. Across ten page types, local and deployed median CPU had `r = 0.948` correlation and `0.75 ms` mean absolute error. Query count explained 82% of local CPU variance and 73% of deployed CPU variance.

The cold-path split is material: on the merged head, Worker module evaluation was 23.5 ms median, first requests were 43–80 ms by route before adding startup, and warm renders were 4.2–12.1 ms. A warm-only profile therefore misses most of the CPU cost paid by a fresh isolate.


MA2153 (commented):
I used this harness to investigate warm/cold CPU on the perf fixture, then cross-checked it against OS-accounted CPU. The structure is right — separating module eval, first request and warm batches is exactly the split that matters, and the query-count integration is genuinely useful. But the headline metric it reports is not CPU, and I think that needs to be either fixed or labelled before anyone builds on the numbers.

### 1. `cpuPerResponseMs` overstates CPU by ~2× on I/O-bound routes

`summarizeProfile` defines active CPU as `sampled − idle` (`scripts/render-cpu.mjs`). On routes that await a binding, that is substantially more than the CPU actually consumed. Reading `utime+stime` from `/proc/<workerd>/stat` across 200 warmed sequential requests, same build, same machine:

| Route | queries | harness (sampled) | OS CPU |
| --- | --- | --- | --- |
| `/rss.xml` | 2 | 12.09 ms | 6.80 ms |
| `/posts` | 10 | 26.60 ms | 15.45 ms |
| `/` | 10 | 32.70 ms | 16.60 ms |
| `/posts/[slug]` | 16 | 39.46 ms | 24.45 ms |

The OS figure is repeatable to ±0.3 ms/request (two independent baselines on `/`: 16.85 and 16.30). The printed regression inherits the factor: 1.97 ms/query sampled vs 1.25 ms/query OS (R² 0.946 vs 0.991).

### 2. In local dev the figure includes the D1 emulator's own execution

Bare Worker, 10 sequential queries, identical client code, identical one-row results — only the SQL does more work inside SQLite:

```
trivial SELECT            profiled active 8.19 ms/req
recursive CTE (20k rows)  profiled active 31.50 ms/req
```

The extra time lands on `setAttributes`, `enterSpan`, `URL`, `toJson` and `addD1MetaToSpan` — i.e. on the D1 *client's* frames. In production that work runs on D1's servers and is not billed to the Worker. Anything derived from this harness that looks like "cost of the D1 client" is partly the local database.

### 3. Per-function self-time in the emitted profiles doesn't survive ablation

The PR writes `.cpuprofile` files "for flamegraph analysis", so people will read them that way. Two checks against the fixture's own code:

- `mapRowToData`: profile assigns it 0.58 ms/request on `/`. Amplified to run **10× per row**, OS CPU was 16.55 ms/req against baselines of 16.30 and 16.85 — no measurable change, where the profile predicts +5.2 ms.
- `formatDate`: profile assigns it 2.07 ms/request (6.6% of warm CPU). It is called twice per request; true cost is 33 µs/call measured in situ (+198 calls/request → +6.45 ms OS CPU), matching a bare-Worker sweep at 29.4 µs/call. Off by ~30×.

### 4. The reported MAD is within-process and reads as more precision than exists

Warm `/` batches inside one run: 26.81–33.42 ms, reported as MAD 0.54 ms. Across independent runs of the same unmodified build I saw 27.3–32.7 ms. `wrangler check startup` over 5 runs: 27.1, 27.5, 31.4, 35.1, 40.5. A sub-millisecond MAD next to a 5–13 ms spread will be read as an error bar and isn't one. The first startup run is also a consistent outlier (79.3 ms in one run) — worth discarding rather than relying on the median to absorb it.

### Suggested change

Add OS CPU accounting alongside the sampled figure — roughly 15 lines, and it makes the harness's own claims checkable:

```js
const pids = /* workerd pids spawned by this dev server */;
const ticks = (pid) => { const p = readFileSync(`/proc/${pid}/stat`, "utf8").split(" "); return Number(p[13]) + Number(p[14]); };
// before/after each batch; cpuMs = (after - before) / 100 * 1000
```

I verified it detects injected work exactly (+198 `toLocaleDateString` calls/request → +6.45 ms, 32.6 µs/call, matching the bare-Worker sweep) and that it agrees with the profiler on a pure busy-loop (12.45 vs 12.95 ms). It's Linux-only, so gate it and keep the sampled number as the portable fallback.

Short of that, I'd rename the field to `sampledActiveMs`, note in the docstring that local D1 work is included, and say in the PR body that the "warm renders 4.2–12.1 ms" range is sampled rather than billed CPU.

eisenbruch:
A measurement your harness would surface immediately, and a question it raises.

**`ANALYZE` had never been run on our D1**, and until it was, the planner was ignoring an index that already existed. EmDash's own single-listing-by-slug query went from **9,676 rows read and 7.5 ms to 3 rows and 0.2 ms** from that one statement, at 5,798 runs an hour on a 9,676-entry collection. Nothing was missing: the correct index was there the whole time, and without statistics SQLite had no idea `slug` was selective.

That is the part worth your attention rather than our numbers. **An unused index looks exactly like a missing one**, and the usual response is to add another index, which we did more than once before finding this. Any EmDash site past its seed data is in this state silently, because nothing in the framework or the docs runs `ANALYZE` or mentions it.

So, as a question rather than a request: should EmDash run `ANALYZE` after a seed or a bulk import, or at least document it as a step? It seems to fit this PR's purpose, since a benchmark that reports query counts against a fixture will read very differently before and after that statement, and a baseline taken without it may be measuring the absence of statistics rather than the cost of the code.

Happy to re-run any of this against a branch if it would help, and to open a docs PR if you would rather have it written down than built in. For completeness, the same site produced #3186, a content list filtered by an explicit id list that still gets an `ORDER BY` and is answered with a table scan; that one is a plan problem rather than a statistics problem, and `ANALYZE` does not fix it.

Context on the environment: EmDash 0.38.0 on Cloudflare Workers with D1, measured on production September 17, 2026.

PR #3074, opened 2026-09-11
fix(core): don't save the inline Portable Text editor when nothing changed

### LINKED ISSUE #2877: Visual Portable Text publish leaves key-only pending changes
### Description

Publishing an inline Portable Text edit from visual editing updates the public page correctly, but a new draft is saved after the publish. When the author returns to admin, the entry still shows the blue Publish button and `Pending changes`.

The live and draft content are semantically identical. In D1 they differ only in generated `_key` values.

On current `main`, `pmToPortableText()` creates new random `_key` values on each `getBlocks()` call. `save()` compares raw JSON, and the `pagehide` keepalive flush can therefore treat unchanged Portable Text as changed after the publish reload. The Portable Text save promise is also separate from the toolbar `pendingSavePromise`, so the floating Publish action does not wait for it.

Expected: Publish waits for the inline Portable Text save, and the subsequent reload does not create a semantically unchanged draft. The entry remains Published with no pending changes.

Related to #2123, but not a duplicate: #2123 and #2860 cover the admin ContentEditor autosave/publish race, while this reproduction uses InlinePortableTextEditor and the floating visual-editing toolbar. Also related to #446: its toolbar guard tracks scalar `saveField()` calls, but not InlinePortableTextEditor saves.

### Steps to reproduce

1. Open a published entry with a Portable Text body in visual editing mode.
2. Edit the body.
3. Move focus to the floating toolbar, letting the blur save begin.
4. Click Publish.
5. Confirm that the public page renders the new content.
6. Return to the admin editor or content list.
7. Observe a blue Publish button and `Pending changes`.

### Environment

- emdash version: 0.36.0
- Astro version: 7.2.10
- Runtime: Cloudflare Workers with D1
- Deployment: production
- OS: macOS

### Screenshots

No screenshot is attached. The visible result is a blue Publish button in the editor and `Pending changes` in the content-list status column after the new content is already live.

### Logs / error output

```shell
{
  "live_revision_created_at": "2026-09-02 21:03:51",
  "draft_revision_created_at": "2026-09-02 21:04:01",
  "raw_content_equal": false,
  "equal_after_recursively_removing__key": true
}
```

eisenbruch:
## What does this PR do?

With visual editing on, the inline Portable Text editor saves the body every time focus leaves it and every time the page is left, even when nobody typed. Each save creates a draft that differs from the live version only in `_key` values, so entries show "Pending changes" that no one made.

Closes #2877

### Cause

- `pmToPortableText()` mints a new `_key` for every block, span and link markDef on every call (`k()` is `Math.random()`), and spans refer to link markDefs by those keys.
- `save()` compared `JSON.stringify(getBlocks())` against `initialRef`, which starts as the raw stored value and, after a save, becomes another fresh serialization. The two never match.
- `handleBlur` and the `pagehide` keepalive flush both call `save()`, so an untouched body is saved on every blur and on every page leave.

### Fix

One file, `InlinePortableTextEditor.tsx`. Change detection moves from serialized Portable Text to the editor's own document:

- The document the loaded content produces is recorded as the saved document once the editor exists.
- `save()` returns early when `editor.state.doc.eq(savedDoc)`, ProseMirror's structural equality over node types, attributes, marks and text. ProseMirror nodes have no `_key`, so the random keys never enter the comparison.
- After a successful save, the saved document becomes the document whose blocks were sent. The blocks are serialized once and that same array is sent, rather than calling `getBlocks()` again.

Two side effects, both improvements:

- An edit undone before focus leaves is no longer saved.
- An edit typed while a save is in flight is no longer absorbed into the baseline. Previously `initialRef = getBlocks()` after the response picked it up, so that edit could never be saved.

The stored shape does not change: the PUT still sends the editor's blocks with their keys. Making keys deterministic, so they stop churning on real saves, is a separate change and is not in this PR.

### Can this hide a real edit?

No. `Node.eq` compares the whole document tree: node types and order, attributes (heading level, alignment, list fields, image and link attributes), marks and text. Reordering, splitting or merging blocks, typing, or changing a link all produce a document that is not `eq`. The only thing it cannot see is `_key`, which lives only in the serialized Portable Text, and ignoring `_key` is the point.

I first tried a canonical-JSON comparison that dropped `_key` and remapped link-mark references to their markDef position. It worked, but it needed about 40 lines of normalization, plus care for every place Portable Text carries a key reference. Comparing the editor document avoids that class of mistake.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes: no errors in the changed files. `tsgo --noEmit` in `packages/core` reports errors only in untouched files (`src/api/handlers/registry.ts`, `src/registry/*`, `src/plugins/types.ts`, `src/utils/slugify.ts`) in my environment.
- [x] `pnpm lint` passes: `oxlint --type-aware` clean on the changed files
- [x] `pnpm test` passes (or targeted tests for my change): all 15 files in `tests/unit/components` pass (89 tests)

- [x] This PR includes AI-generated code — model/tool: Claude (Claude Code)

## Screenshots / test output

Not applicable (no UI change). New file `tests/unit/components/inline-portable-text-unchanged-save.test.ts`; the stored body includes a link paragraph. Against current `main` without the fix:

```
× does not save when focus leaves an unedited body
    expected [ …3 PUTs ] to have a length of +0 but got 3
× does not save on page leave when the body was never touched
    expected [ …1 PUT ] to have a length of +0 but got 1
× saves a real edit once, and not again on the next blur
    expected [ …2 PUTs ] to have a length of 1 but got 2
× does not save an edit that was undone before focus left
    expected [ …1 PUT ] to have a length of +0 but got 1
```

With the fix, all four pass.

### DIFF
--- a/.changeset/inline-editor-unchanged-saves.md
+++ b/.changeset/inline-editor-unchanged-saves.md
@@ -1,0 +1,5 @@
+---
+"emdash": patch
+---
+
+Fixes visual editing saving a new draft when nothing changed. With edit mode on, the inline Portable Text editor saved the body every time focus left it and every time the page was left, even if no one had typed, because each save check compared freshly generated block keys against the stored ones. Entries picked up drafts that differed only in `_key` and showed "Pending changes" in the admin. The editor now compares its document with the one last saved, so an unedited body is never saved and a real edit is saved once.
--- a/packages/core/src/components/InlinePortableTextEditor.tsx
+++ b/packages/core/src/components/InlinePortableTextEditor.tsx
@@ -2096,6 +2096,12 @@ export function InlinePortableTextEditor({
 	tablePlaceholder = TableBlockNode.options.placeholder,
 }: InlinePortableTextEditorProps) {
 	const initialRef = React.useRef(value);
+	// The editor document as last known to be stored: the one the loaded
+	// content produced, then whatever each successful save sent. Saves compare
+	// against it structurally. Comparing serialized Portable Text against the
+	// raw `value` never matched, because serialization mints new `_key`s on
+	// every call, so every blur and every page leave saved a new draft.
+	const savedDocRef = React.useRef<Editor["state"]["doc"] | null>(null);
 	const savingRef = React.useRef(false);
 	const editorRef = React.useRef<ReturnType<typeof useEditor>>(null);
 
@@ -2159,9 +2165,9 @@ export function InlinePortableTextEditor({
 			// one that can still land.
 			if (savingRef.current && !options?.keepalive) return;
 
-			const current = JSON.stringify(getBlocks());
-			const initial = JSON.stringify(initialRef.current);
-			if (current === initial) return;
+			const doc = editorRef.current?.state.doc;
+			if (!doc || !savedDocRef.current || doc.eq(savedDocRef.current)) return;
+			const blocks = getBlocks();
 
 			savingRef.current = true;
 			try {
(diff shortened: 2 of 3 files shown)


ascorbic (approved):
Thanks!

PR #3095, opened 2026-09-13
fix(cloudflare): resolve D1 databases by name without total_pages

### LINKED ISSUE #2840: `emdash migrate --d1 <name>` always fails: Cloudflare's D1 list API omits `result_info.total_pages`
Selecting a D1 database by name is documented in both the [CLI reference](https://docs.emdashcms.com/reference/cli/) (`--d1 <uuid-or-name>`) and [Manage Core Database Migrations](https://docs.emdashcms.com/deployment/core-migrations/) ("You can instead provide `--account-id` with `--d1 <database-uuid-or-name>`. Name lookup must resolve to exactly one database.").

It cannot succeed. Every name lookup fails, regardless of account, token, or how many databases match.

### Reproduction

```bash
export CLOUDFLARE_ACCOUNT_ID="..."
export CLOUDFLARE_API_TOKEN="..."   # D1 Edit
pnpm emdash migrate --status --account-id "$CLOUDFLARE_ACCOUNT_ID" --d1 my-site-production
```

**Expected:** the named database resolves and a status report prints.

**Actual:** exit code 1, and

```
Cloudflare D1 database list total_pages is invalid.
```

Passing the same database's UUID instead (`--d1 <uuid>`) works, as does `--wrangler-config wrangler.jsonc`. Only the name path is affected.

### Cause

`databaseByName` validates pagination through `listResultInfo`, which requires `total_pages`:

`packages/cloudflare/src/db/d1-migration-target.ts:248`

```ts
totalPages: listInteger(value.total_pages, "total_pages", 0),
```

`listInteger` rejects anything that isn't a number, so an absent key throws.

Cloudflare's `GET /accounts/{account_id}/d1/database` does not return `total_pages`. Live response `result_info`, both with and without a `name` filter:

```json
{ "count": 5, "page": 1, "per_page": 100, "total_count": 5 }
```

`total_pages` is simply not part of that envelope, so `listInteger(undefined, ...)` throws before any match is considered. Because `totalPages` also drives the `while (page <= totalPages)` loop and the cross-page consistency check, the whole paging routine is built on a field the endpoint never sends.

### Why tests don't catch it

The fetch stub in `packages/cloudflare/tests/db/d1-migration-target.test.ts:25` synthesizes the envelope by hand and includes `total_pages`:

```ts
result_info: {
  page,
  per_page: 100,
  count: result.length,
  total_count: totalCount,
  total_pages: totalPages,
},
```

The fixture asserts a contract the real endpoint doesn't honor, so the name path passes in CI and fails universally in practice.

### Suggested direction

Derive the page count from `total_count` / `per_page` rather than requiring `total_pages`, or treat the field as optional and fall back to "stop when a short page arrives". Either way it would be worth aligning the test fixture with a recorded real response so this class of drift is caught.

### Secondary note

Cloudflare's `name` parameter is a **substring** filter, not an exact match — `?name=my-site` returns `my-site-production`, `my-site-staging`, and so on. `databaseByName` handles this correctly at the end (it filters on `database.name === name`), but it calls `metadataDatabase` on *every* row of the page first, and that helper throws `Preview D1 databases cannot be used for deployment migrations.` on any row whose `version !== "production"`. So once the `total_pages` issue is fixed, an unrelated non-production database whose name merely contains the search string can still fail the lookup. Filtering by exact name before validating each row would avoid that.

### Environment

- `emdash` / `@emdash-cms/cloudflare` 0.35.0
- Node 26.1.0, pnpm, Linux
- Reproduced against the live Cloudflare API with a D1 Edit token

dchaudhari7177:
## What does this PR do?

`emdash migrate --d1 <name>` could never resolve a name. `databaseByName` validated pagination with `listInteger(value.total_pages, …)`, but the D1 list endpoint's `result_info` is `{ count, page, per_page, total_count }` with no `total_pages`, so every lookup threw `Cloudflare D1 database list total_pages is invalid.` before looking at a single row. UUIDs and `--wrangler-config` were unaffected.

- **Page count.** When `total_pages` is absent, it is derived as `ceil(total_count / per_page)`. When present it is validated as before. Everything downstream (the page loop, the cross-page consistency check, `MAX_DATABASE_LIST_PAGES`, the final `seenCount === totalCount` check) is unchanged and now runs on a value the endpoint actually implies.
- **Substring matches (the issue's secondary note).** `name` is a substring filter, and every row went through `metadataDatabase` before the exact-name filter. That throws for any non-production row, so once paging worked, a preview `site-db-staging` on the same page would still have failed the lookup for `site-db`. Rows are now filtered by exact name first, and only the matches are validated as deployment targets.

Closes #2840

## Type of change

- [x] Bug fix

## Checklist

- [ ] `pnpm typecheck` passes: `tsc --noEmit -p packages/cloudflare` reports the same 7 errors on this branch and on `main` (unbuilt `emdash` workspace types), none in the touched file
- [x] `pnpm lint` passes: `oxlint --deny-warnings` on the touched files
- [x] `pnpm test` passes (or targeted tests for my change): see below
- [x] `pnpm format` has been run: `oxfmt`
- [ ] User-visible strings in the admin UI are wrapped for translation: n/a
- [ ] New features link to an approved Discussion: n/a, bug fix
- [ ] I have included screenshots below if this PR changes the UI: n/a

## Screenshots / test output

Not applicable (CLI). The test stub `listApiResponse` synthesized `total_pages`, which is why CI never saw this. The new cases in `packages/cloudflare/tests/db/d1-migration-target.test.ts` use a `liveListApiResponse` shaped like the endpoint the issue recorded:

- resolves a name from a single page;
- walks two pages derived from `total_count: 2, per_page: 1`;
- ignores a preview `site-db-staging` listed next to `site-db`;
- still reports `No D1 database named site-db` for an empty result.

```
main:  5 failed | 10 passed   # the 4 new cases ("total_pages is invalid") + 1 pre-existing
head:  1 failed | 14 passed   # only the pre-existing one
```

The one failure on both trees is `uses the selected Wrangler environment and its own binding array`, which asserts `/project/wrangler.jsonc` and gets a Windows path separator on this machine. CI on Linux should pass it. I did not run this against the live Cloudflare API.

### DIFF
--- a/.changeset/d1-name-lookup-paging.md
+++ b/.changeset/d1-name-lookup-paging.md
@@ -1,0 +1,5 @@
+---
+"@emdash-cms/cloudflare": patch
+---
+
+Fixes `emdash migrate --d1 <name>` failing for every database name with "Cloudflare D1 database list total_pages is invalid". The D1 list endpoint does not return `total_pages`, so the page count is now derived from `total_count` and `per_page` when it is absent. A preview database whose name only contains the requested name (the `name` filter matches substrings) no longer fails the lookup either.
--- a/packages/cloudflare/src/db/d1-migration-target.ts
+++ b/packages/cloudflare/src/db/d1-migration-target.ts
@@ -213,8 +213,13 @@ async function databaseByName(
 			throw new Error("Cloudflare D1 database list pagination is invalid.");
 		}
 		seenCount += resultInfo.count;
+		// `name` is a substring filter, so a page can hold other databases whose names merely
+		// contain it. Only an exact match is validated as a deployment target; a preview database
+		// named `site-db-staging` must not fail the lookup for `site-db`.
 		matches.push(
-			...envelope.result.map(metadataDatabase).filter((database) => database.name === name),
+			...envelope.result
+				.filter((database) => isRecord(database) && database.name === name)
+				.map(metadataDatabase),
 		);
 		page += 1;
 	}
@@ -245,8 +250,14 @@ function listResultInfo(value: unknown, expectedPage: number, resultCount: numbe
 		perPage: listInteger(value.per_page, "per_page", 1),
 		count: listInteger(value.count, "count", 0),
 		totalCount: listInteger(value.total_count, "total_count", 0),
-		totalPages: listInteger(value.total_pages, "total_pages", 0),
+		totalPages: 0,
 	};
+	// The D1 database list returns `page`, `per_page`, `count` and `total_count` but no
+	// `total_pages`, so derive it when it is absent rather than rejecting every name lookup.
+	result.totalPages =
+		value.total_pages === undefined
+			? Math.ceil(result.totalCount / result.perPage)
+			: listInteger(value.total_pages, "total_pages", 0);
 	if (
 		result.page !== expectedPage ||
 		result.perPage > DATABASE_LIST_PAGE_SIZE ||
--- a/packages/cloudflare/tests/db/d1-migration-target.test.ts
+++ b/packages/cloudflare/tests/db/d1-migration-target.test.ts
@@ -37,6 +37,22 @@ function listApiResponse(
 	});
 }
 
+/** The envelope the live D1 list endpoint returns: `result_info` has no `total_pages` (#2840). */
+function liveListApiResponse(
+	result: unknown[],
+	page: number,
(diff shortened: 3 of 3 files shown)


ascorbic (approved):
Thanks!

PR #3103, opened 2026-09-14
fix(core): size gallery images to their grid cell, not the viewport

### LINKED ISSUE #2930: Gallery renderer uses full viewport in sizes for multi-column images
## Problem

The built-in Portable Text `Gallery.astro` generates responsive `srcset` values for provider images, but its `sizes` value does not account for the gallery's column count:

```ts
sizes = width ? `(min-width: ${width}px) ${width}px, 100vw` : "100vw";
```

The gallery itself renders `repeat(var(--columns, 3), 1fr)`. On viewports below the source image's intrinsic width, each image may occupy only one third of the content width while `sizes` tells the browser it occupies `100vw`. The browser can therefore choose a substantially larger `srcset` candidate than the rendered cell needs.

## Reproduction

1. Add a Portable Text gallery with three columns and provider-backed images that expose `getSrc`.
2. Render it inside a normal article content column.
3. Inspect an image's generated `sizes` and the selected `currentSrc` at a viewport narrower than the image's intrinsic width.
4. The generated fallback is `100vw`, despite the image rendering in one of three columns.

The issue is more visible on DPR 2 mobile and tablet screens, where the oversized slot estimate can select an unnecessarily large derivative for every gallery item.

## Expected

The generated slot estimate should reflect the configured number of columns and responsive layout, or the renderer should expose a way for the consuming site to provide an accurate `sizes` value.

Possible directions:

- derive a default from `node.columns`, with the existing two-column mobile breakpoint reflected in `sizes`, or
- add an optional `sizes` prop / per-gallery field that consumers can override when the gallery is rendered in a narrower container.

A default could approximate the current CSS, including gaps, while an override would cover custom content widths and custom gallery renderers.

## Environment

Observed in EmDash 0.36.0. The relevant code is `packages/core/src/components/Gallery.astro`.

dchaudhari7177:
## What does this PR do?

`Gallery.astro` told browsers each image fills the viewport. Provider images got `sizes="(min-width: Wpx) Wpx, 100vw"`, and locally stored images with dimensions go through `<Image layout="constrained">`, whose default is the same estimate. But the gallery lays images out in `repeat(var(--columns, 3), 1fr)` with a `1rem` gap, and two columns at 640px and below. So the browser sizes the slot at roughly three times the rendered cell and picks a correspondingly large `srcset` candidate. The local path's srcset runs up to 3200w for a 1600px image, so on DPR 2 screens this means downloading far more than each cell shows.

- `gallerySizes(columns)` in `media/responsive.ts` builds the cell estimate from the gallery's own CSS:
  `(max-width: 640px) calc((100vw - 1rem) / 2), calc((100vw - 2rem) / 3)` for three columns. One column gives `100vw` above the breakpoint. A missing, zero or non-integer count falls back to three, the count `Gallery.astro` already defaults to.
- `Gallery.astro` uses it for both the provider `` and the `AstroImage` path.
- A new optional `sizes` prop overrides it. This is the issue's second direction, for galleries rendered in a container narrower than the viewport, where a viewport-based estimate is still too large. Both directions from the issue are covered: an accurate default and a consumer override.

Closes #2930

## Type of change

- [x] Bug fix

## Checklist

- [ ] `pnpm typecheck` passes: not run as `tsgo` (not installed here). `tsc --noEmit -p packages/core` reports no errors.
- [x] `pnpm lint` passes: `oxlint` on the touched files
- [x] `pnpm format` has been run: `oxfmt` (via stdin on the LF content) and `prettier` for the `.astro` file, no changes
- [ ] User-visible strings in the admin UI are wrapped for translation: n/a, no admin strings
- [ ] New features link to an approved Discussion: n/a, bug fix. The `sizes` prop is the override the issue proposes alongside it.
- [ ] I have included screenshots below if this PR changes the UI: n/a, the change is an attribute value, shown below

## Screenshots / test output

`tests/repro/gallery-sizes.render.test.ts` (new) renders `Gallery` through the Astro container, with a mock provider injected via `__emdashTestMediaProviders`:
- provider images in a default gallery get the three-column cell estimate, and their srcset is still provider-generated
- `columns: 4` gives `calc((100vw - 3rem) / 4)`
- a local image with width/height (`data-astro-image="constrained"`) gets the same estimate
- a `sizes` prop wins

With `main`'s `Gallery.astro` and `responsive.ts`:
```
× provider images are sized to one of three default columns
× the slot follows the block's column count
× locally stored images with dimensions get the same cell estimate
× a consumer-provided sizes wins
Tests  4 failed (4)
```
With this branch: 4 passed. `tests/unit/media/responsive.test.ts` gains `gallerySizes` cases (column counts, one column, the fallback for `undefined`/`0`/`2.5`/`NaN`): 21 passed. The existing `image-render` and `focal-point-render` repro suites still pass (23 tests across the three files).


ascorbic (approved):
Thanks!

PR #3128, opened 2026-09-15
fix(core): don't throw when a request has no route-cache handle

eisenbruch:
## What does this PR do?

EmDash's request-context middleware calls `optOutOfRouteCache(context.cache)` for `_preview` requests, `_edit` requests in client toolbar mode, and responses it injects the toolbar into. The function assumed `context.cache` is always present. It is not, and when it is missing the middleware throws `TypeError: Cannot read properties of undefined (reading 'set')` and the whole response fails instead of rendering.

No existing issue; found on a production site.

### When `context.cache` is undefined

In Astro 7.3.2 (the version in this repo's lockfile):

- `FetchState.defineProviderGetters()` (`core/fetch/fetch-state.js`) adds a getter only for keys a handler registered with `provide()`. `cache` is registered by `provideCache()` in `core/cache/handler.js`.
- The routing handler (`core/routing/handler.js`) returns `renderErrorFromState(...)` for a URL with no matching route **before** `render()` calls `provideCache(state)`. The 404 page rendered on that path runs middleware on a context with no `cache` property.

On our site (Astro 7.2, `@astrojs/cloudflare`) it took pages down in three separate ways before we traced it: a signed-in editor opening a retired URL (the 404 page), a mistyped image path from inside the admin (the 404 page again), and a middleware rewrite. Each time the page returned an error instead of a 404 or the rewritten content.

### Fix

`optOutOfRouteCache()` now uses `cache?.set(false)`, and `RouteCache` includes `undefined` so the type says what Astro actually provides. The function is already documented as a no-op when no cache provider is configured; it is now also a no-op when the request has no cache handle. Nothing else changes: `Cache-Control: private, no-store` is still set on toolbar and preview responses, and requests that have a cache handle opt out exactly as before.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes (`packages/core`, after `pnpm build` at the root)
- [x] `pnpm test` passes (or targeted tests for my change): the three `tests/unit/astro/request-context-*` files, 18 tests
- [ ] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable). Not applicable.
- [ ] I have included screenshots below if this PR changes the UI. Not applicable.

## Screenshots / test output

Not applicable for UI. Two tests added to `tests/unit/astro/request-context-route-cache.test.ts`, each running the middleware with no `cache` on the context: a `_preview` request, and a signed-in editor request that gets the toolbar (which also asserts `Cache-Control: private, no-store` is still set).

Against `main` without the `request-context.ts` change:

```
 × serves a preview request instead of throwing
 × serves an editor request with the toolbar instead of throwing
TypeError: Cannot read properties of undefined (reading 'set')
      Tests  2 failed | 4 passed (6)
```

With it:

```
 Test Files  3 passed (3)
      Tests  18 passed (18)
```


ascorbic (approved):
Thanks!

Issue #2389, opened 2026-08-09
Production admin builds render raw ICU for messages missing from the compiled catalog


MA2153:
### Description

In a production build of the admin, any message that isn't present in the compiled catalog renders as its raw ICU source instead of formatted text. Plain strings survive (the message id *is* the English source), but anything with a placeholder or an ICU construct leaks:

- `` t`Fields: ${list}` `` renders as `Fields: {0}`
- `plural(n, { one: "…", other: "…" })` renders as `{0, plural, one {…} other {…}}`

Dev builds are unaffected, so this never shows up while working on a feature — only in a built admin.

**Why it happens.** `@lingui/core` (5.9.5) installs its runtime message compiler only outside production:

```js
// I18n constructor
if (process.env.NODE_ENV !== "production") {
  this.setMessagesCompiler(compileMessage);
}

// I18n._()
let translation = messageForId || message || id;
if (isString(translation)) {
  if (this._messageCompiler) {
    translation = this._messageCompiler(translation);
  } else {
    console.warn(`Uncompiled message detected! … ICU features such as interpolation and plurals will not work properly for that message.`);
  }
}
```

`packages/admin/src/App.tsx` calls `i18n.loadAndActivate({ locale, messages })` without ever setting a compiler, so in production the fallback string is returned verbatim and the warning is the only signal.

**Why the catalog is missing the message.** Catalogs are regenerated by the `auto-extract` workflow on merge to `main`, and feature branches are asked not to include `messages.po` changes. That is a good policy, but it means a string is absent from *every* catalog — English included — for the entire life of the branch that introduces it. `lingui compile` can't emit a source fallback for a message it has never seen, so the compiled catalogs shipped by that branch's build simply don't contain the id.

**Practical consequence: `pkg.pr.new` previews are unusable for reviewing any change that introduces strings.** A preview package is built with the branch's catalogs and with `NODE_ENV=production`, which is exactly the combination above. Anyone installing a preview to try out new UI sees raw ICU in place of every new plural or interpolated string, in every locale. The problem disappears on merge once extraction runs, so it has never been visible on a release.

Two things worth deciding separately:

1. Whether the runtime should degrade gracefully — `i18n.setMessagesCompiler(compileMessage)` from `@lingui/message-utils/compileMessage` makes an uncompiled fallback format correctly instead of printing its source. It costs ~12KB raw (~4KB gzipped) in the admin bundle and also covers untranslated locales, not just unextracted ones.
[...]

PR #3130, opened 2026-09-15
fix(admin): don't pass untranslated plugin labels to Lingui

eisenbruch:
## What does this PR do?

The admin console fills with `Uncompiled message detected!` warnings, one per plugin page label per render (for example "SEO", "Forms", "Submissions", "Audit History"). The labels render correctly; the warnings are noise that buries real errors.

No existing issue. Related to #2389, which covers raw ICU for messages missing from the compiled catalog.

### Cause

`resolvePluginPageLabel()` runs every declared plugin label through the shared Lingui instance (`i18n._(label)` in `Sidebar.tsx`, `translateDynamic` in `AdminCommandPalette.tsx`). Most plugin labels have no catalog entry. For a missing id, `I18n._()` falls back to the id string, and when no message compiler is installed, which is the case when `NODE_ENV` is `production`, it logs the warning for that string.

### Fix

`declaredLabelTranslator(i18n, translate)` (exported from `Sidebar.tsx`) wraps the translate function: a label goes to `translate` only when `i18n.messages[label]` is defined, the same lookup `I18n._()` uses to decide a message is missing; otherwise the label is returned as declared. The sidebar and the command palette both use it. Labels a plugin catalog translates still translate, and untranslated labels render exactly as before.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes (`packages/admin`, after `pnpm build` at the root)
- [x] `pnpm test` passes (or targeted tests for my change): `tests/components/Sidebar.test.tsx` (27 tests) and the `AdminCommandPalette` tests
- [ ] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable). Not applicable: no new strings.
- [x] I have added and reviewed the user-facing [changeset](https://github.com/emdash-cms/emdash/blob/main/.changeset/README.md) (`@emdash-cms/admin`: patch)
- [ ] I have included screenshots below if this PR changes the UI. Not applicable: no visual change, console output only.

## Screenshots / test output

Two tests in `tests/components/Sidebar.test.tsx`: a label the catalog has still translates, and a label it lacks comes back as declared with no `console.warn`. The second test removes the message compiler (`setMessagesCompiler(undefined)`), since outside production Lingui installs one and would not warn.

With the wrapper reduced to calling `translate` for every label (the previous behavior):

```
 × returns a label the catalog lacks as declared, without a missing-message warning
AssertionError: expected "Mock" to not be called at all, but actually been called 1 times
      Tests  1 failed | 26 passed (27)
```

With the change: `Tests  27 passed (27)`.


eisenbruch:
Both stale comments fixed in 0bfc3612, taking your suggested text as written.

The JSDoc on `resolvePluginPageLabel` and the inline comment in `AdminCommandPalette` were left describing the unguarded behavior, which is exactly the kind of comment that outlives the code it describes. Nothing else changed, so the tests and the behavior are as reviewed.

eisenbruch:
Rebased onto main (`b578e380`). Two conflicts, both from `isSafePluginPagePath` landing in the meantime; I kept both sides in each case.

- `Sidebar.tsx`: main's `isSafePluginPagePath` guard and import are kept, and the label lookup still goes through `declaredLabelTranslator`.
- `AdminCommandPalette.tsx`: same guard kept, and the comment is the rewritten one from the last review round, describing the guarded behavior rather than the old unguarded lookup.

No behavior change from the rebase itself. `oxfmt --check` is clean across the tree.

PR #3142, opened 2026-09-16
fix(core): let an entry shed a deleted field's value instead of stranding it

### LINKED ISSUE #3141: Deleting a field strands every entry that has a pending draft: the draft keeps the value, and no save is accepted afterwards
### Description

Deleting a field from a collection leaves any entry that has a **pending draft** unable to be saved again, through the API or the admin.

The draft revision stores the entry's whole `data` as JSON, so the deleted field's value stays in it. `handleContentGet` returns the draft, so a client reads the stale key back. On the way in, `validateContentData` (`packages/core/src/api/handlers/validation.ts`) rejects any key that is not a current field:

```
hits: unknown field on collection 'posts'
```

There is no request body that gets the entry out of this state:

- **Send the data you just read** (what the admin editor does): rejected, `unknown field`.
- **Send the key as `null` or `""`** to clear it: rejected the same way.
- **Omit the key**: accepted, but `handleContentUpdate` merges the body into the stored data (`{ ...baseData, ...processedData }`, where `baseData` is the draft revision), so the key survives and the next full save fails again.

A published entry with no pending draft is fine: its values live in columns that the delete drops.

On our site this stranded two entries, an article draft and a listing draft, after two long-retired fields were removed. They can only be repaired by recreating them.

Either of these would fix it, and they are not exclusive:

1. **`deleteField` strips the field from stored entry data and from revisions** as part of the delete, so nothing is left behind.
2. **Validation tolerates a key that is not in the collection**, or accepts an explicit `null` as "remove this key". This is the smaller change, and it is the only one of the two that rescues entries stranded by a delete that has already happened.

I am happy to open a PR for either once one of my open PRs clears; say which you would prefer.

### Steps to reproduce

Integration test against `main` (79c69b2), in `packages/core/tests/integration/content/`. All four cases pass, which is to say the bug is present:

```ts
// setup: collection `posts` with `title` and `hits`; an entry published with hits: 42,
// then updated (not published) to hits: 43, so it has a pending draft; then:
await registry.deleteField("posts", "hits");
```

```ts
it("still reads back the deleted field's value, so a client sends it again", async () => {
  const got = await runtime.handleContentGet("posts", id);
  expect((got.data.item.data as Record<string, unknown>).hits).toBe(43);
});

it("refuses the save when the client sends the data it just read", async () => {
  const got = await runtime.handleContentGet("posts", id);
  const data = { ...(got.data.item.data as Record<string, unknown>), title: "edited" };
  const saved = await runtime.handleContentUpdate("posts", id, { data });
  expect(saved.success).toBe(false);
  expect(saved.error.message).toContain("unknown field on collection");
});

it("refuses an explicit null or empty string for the deleted field", async () => {
  for (const value of [null, ""]) {
    const saved = await runtime.handleContentUpdate("posts", id, { data: { title: "edited", hits: value } });
    expect(saved.success).toBe(false);
  }
});

it("accepts a save that omits the key, but the value survives the merge", async () => {
  const saved = await runtime.handleContentUpdate("posts", id, { data: { title: "edited" } });
  expect(saved.success).toBe(true);
  expect((saved.data.item.data as Record<string, unknown>).hits).toBe(43);
});
```

```
 Test Files  1 passed (1)
      Tests  4 passed (4)
```

Printing the rows in the same scenario shows where the value sits after the delete:

```
revision 01M2M1VD7HEAS84SPNNYP4R2VM {"title":"p1 draft","hits":43}
get data: {"title":"p1 draft","hits":43}
save with stale key: hits: unknown field on collection 'posts'
```

### Environment

- emdash 0.38.0 on Cloudflare Workers with D1 (where we hit it), and `main` at 79c69b2 with the sqlite test dialect (where the test above runs).
- Node 22.

### Screenshots

Not applicable.

eisenbruch:
## What does this PR do?

Fixes #3141: an entry with a pending draft cannot be saved again once one of its fields is deleted.

A draft revision stores the whole `data` as JSON, so a deleted field's value stays in it and every read hands the key back. On the way in, `validateContentData` refuses any key that is not a current field, and there is no request body that gets the entry out:

- send the data you just read (what the admin editor does): `hits: unknown field on collection 'posts'`;
- send the key as `null` or `""`: refused the same way;
- omit the key: accepted, but the update merges into the draft's data, so the key survives and the next full save fails again.

This takes the validation side of the two fixes offered in the issue, because it also rescues entries that a delete has already stranded. Two changes in `EmDashRuntime.handleContentUpdate`:

- **`dropUnknownKeysAlreadyStored()`** removes a key the collection has no field for from the incoming data **when the entry already stores it** (checking the draft revision when there is one, otherwise the entry). A key the entry does not already store is left alone, so an unknown field is still reported as one and a typo still fails.
- The merged data is written through **`keepKnownFields()`**, so a saved entry sheds keys the collection has no field for instead of carrying them through every revision that follows. Reserved `_`-prefixed keys are kept, and a collection whose fields cannot be read is left alone rather than emptied.

I have not touched `handleContentCreate`: a new entry has nothing stored to shed, so an unknown key there is exactly what the existing error is for.

The alternative in the issue, stripping the field from stored data inside `schema_delete_field`, is worth doing as well, and it is not in this PR. It prevents the next delete from stranding anything, but it cannot help an entry stranded by a delete that has already happened.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes (`packages/core`)
- [x] `pnpm test` passes (or targeted tests for my change): `tests/integration/content` and `tests/unit/fields`, 26 files, 220 tests
- [ ] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable). Not applicable, no admin strings.
- [ ] I have included screenshots below if this PR changes the UI. Not applicable.

## Screenshots / test output

Not applicable for UI.

`packages/core/tests/integration/content/deleted-field-stranded-entry.test.ts` sets up a published entry with a pending draft, deletes the field, and then covers the four cases. Against `main` the first three fail:

```
 × saves the data it just read, and sheds the deleted field
 × accepts an explicit null for the deleted field, and refuses it once the entry is clean
 × drops the value from the merged data even when the key is omitted
 ✓ still refuses a key the entry does not already store
```

With the change:

```
 Test Files  1 passed (1)
      Tests  4 passed (4)
```

And the suites around it:

```
 Test Files  26 passed (26)
      Tests  220 passed (220)
```

Opened as a draft: I have six PRs waiting for review, which I understand to be the cap. Mark it ready whenever suits you, or say if you would rather I close it and reopen it later.


eisenbruch:
Thanks, the `updateDraftAware` gap is real and I had missed it. Pushed in 5129a8a4.

**`ContentRepository.updateDraftAware` (needs fixing).** It now applies the same two rules as the runtime handler: a key with no matching field is dropped rather than rejected when the entry already stores it, reading the draft revision when there is one, and the merged data is written through `keepKnownFields`. The unknown-field check runs after that drop, so a key the entry has never stored still throws `Unknown field`. That covers `plugins/context.ts` and both sandbox bridges.

Rather than duplicate the behavior I moved it into `packages/core/src/content/known-fields.ts`, exporting `keepKnownFields` and `staleStoredKeys`, both taking a plain `ReadonlySet<string>` of field slugs. The runtime and the repository read their field lists differently (schema registry versus the joined `_emdash_fields` query), so a set keeps the helpers free of either shape.

Two new cases in the integration test cover the repository path: a read-then-write save that sheds the key, and a typo key that is still refused. The first fails against this branch without the `content.ts` change.

**Field-list loads (suggestion).** You were right about the count, and `SchemaRegistry.getCollectionWithFields` does not cache, so each one was two queries. `handleContentUpdate` now loads it once up front and threads it through: `normalizeMediaFields` takes it as an optional preloaded argument, `dropUnknownKeysAlreadyStored` takes the slug set, and the draft merge reuses it. That removes three of the four loads, six queries per save. The fourth is inside `validateContentData`, which builds its own registry; leaving that alone kept the change contained.

**Comments and test header (suggestions).** Taken as written, all three.

`pnpm typecheck` and `oxlint --type-aware --deny-warnings` are clean, `oxfmt` has been run, and `tests/integration/content` plus `tests/unit/fields` pass at 26 files and 222 tests.

PR #704, opened 2026-04-21
feat(admin): optional link on portable-text image blocks

drudge:
## What does this PR do?

Adds an optional `link: { href; blank? }` on portable-text image blocks, round-tripped through the PT ↔ PM converters. The existing editor Link buttons (main toolbar and bubble menu) now operate on image selection: selecting an image and clicking Link opens the URL popover pre-populated with the image's current link, and apply/remove updates the image node's attrs. `Image.astro` wraps the rendered `` in an `<a>` using `sanitizeHref()` when `link.href` is set. `ImageDetailPanel` gains a Link URL input and an "Open in new tab" checkbox as a secondary editing surface.

Split from #679 per @ascorbic's review ("it'd be worth opening those as separate PRs … the image link is definitely a separate concern"). The image-link portion and its Copilot follow-ups land here; the `repeater` BlockKit element will follow on the original PR #679 branch (force-pushed), and `media_picker` in a third PR.

Relative paths like `/page` are accepted — the Link URL input is `type="text"` (not `type="url"`) so it matches how text-link inputs behave elsewhere in the editor. `sanitizeHref()` (already tested) enforces the allow-list at render time.

Copilot comments addressed in scope:

- `types.ts` comment now describes real behavior ("the image is rendered inside an `<a>` using `sanitizeHref`-validated `href`")
- `ImageDetailPanel.tsx` Link URL input is `type="text"` (was flagged when it was `type="url"` and rejected relative paths)
- `prosemirror-to-portable-text.ts` drops half-populated or whitespace-only `link` objects on the way out — new round-trip test covers it

Closes #

## Type of change

- [ ] Bug fix
- [x] Feature (requires [maintainer-approved Discussion](https://github.com/emdash-cms/emdash/discussions/categories/ideas))
- [ ] Refactor (no behavior change)
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests
- [ ] Chore (dependencies, CI, tooling)

Maintainer sign-off for this scope is @ascorbic's review comment on #679 ("Thanks for this. It's a good addition.") combined with the instruction to split that PR into focused replacements.

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md)
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (or targeted tests for my change)
- [x] `pnpm format` has been run
[...]

PR #3167, opened 2026-09-16
feat(admin): optional link on portable-text image blocks

### LINKED ISSUE #2497: Image hyperlink applied in the editor is silently dropped on save (Portable Text image block has no link field)
### Summary

A hyperlink applied to an image in the admin editor is silently dropped when the entry is saved. The link appears to work while editing, but it never reaches the stored Portable Text, so it is absent from the public render — and it is also gone if the entry is reopened in the editor.

Version: 0.32.0 (also inspected the 0.32.0 published source; all references below are to that tarball).

### Steps to reproduce

1. In a `portableText` field, insert an image into the body.
2. Select the image and apply a link via the editor's link control. The editor session shows it applied.
3. Save/publish the entry.
4. View the public page rendered via `<PortableText>` — the image renders with no `<a>` wrapper.
5. Reopen the entry in the editor — the link is gone there too.

### Root cause (from the 0.32.0 source)

The Portable Text image block cannot carry a link, and both converters drop it:

- `src/content/converters/types.ts` — `PortableTextImageBlock` has fields for `asset`, `alt`, `caption`, `width`/`height`, `displayWidth`/`displayHeight`, but **no link/href field**.
- `src/content/converters/prosemirror-to-portable-text.ts` — `convertImage()` maps only `provider`, `mediaId`, `src`, `alt`, `title`, and the four dimension attrs. Any link mark/attr on the image node is discarded on save.
- `src/content/converters/portable-text-to-prosemirror.ts` — the reverse `convertImage()` likewise restores no link, so the loss is invisible until you look at the published page (or reload the editor).
- `src/components/Image.astro` (public renderer) has no link handling either, so even a hand-authored block with an extra field would not render one.

### Expected

Either round-trip the link (add e.g. `link?: { href: string; blank?: boolean }` to `PortableTextImageBlock`, carry it through both converters, and wrap the rendered `<figure>`/`` in an anchor in `Image.astro`), or disable the link control for image nodes in the editor so the UI doesn't accept input it cannot persist.

### Impact

Content editors reasonably use linked images (e.g. a promotional image linking to a signup page). Today the editor accepts the link and everything downstream loses it silently — the failure is invisible at authoring time.

kwmr:
## What does this PR do?

Supersedes #704 (@drudge's implementation, re-ported onto current `main` per the handoff in that PR). Approved in Discussion #2741.

Adds an optional `link: { href: string; blank?: boolean }` to Portable Text image blocks, round-tripped through the PT ↔ ProseMirror converters.

- **Editor**: the existing Link buttons (toolbar and bubble menu) now act on an image selection — the popover is pre-filled with the image's current link; apply/remove updates the image node's attrs. `ImageDetailPanel` gains a "Link URL" input and an "Open in new tab" checkbox as a secondary surface.
- **Renderer**: `Image.astro` wraps the image in `<a>` when `link.href` is set, via `sanitizeHref()`; `blank` adds `target="_blank" rel="noopener noreferrer"`.
- **Converters**: half-populated links (empty/whitespace `href`) are dropped on save so `{ blank: true }` never leaks into Portable Text.

Relative to #704: re-ported onto current `main` (the original no longer applied cleanly — `identityAttrs`/`attrsWithPortableTextKey`, image `alignment` and Kumo imports had landed on the same lines); new UI strings are wrapped with Lingui per the review; and the case flagged on #704 where editing an image link's URL from the toolbar/bubble menu dropped an existing `blank: true` is fixed, with tests.

Closes #2497

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] User-visible strings in the admin UI are wrapped for translation. No `messages.po` changes included.
- [x] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/2741
- [x] I have included screenshots below

- [x] This PR includes AI-generated code — model/tool: Claude Code (Claude Fable 5.1 (this PR); an earlier Claude Code session did the Aug 27 port）

## Screenshots / test output

Image settings panel with the new Link URL field and Open in new tab checkbox

Targeted tests run locally: `packages/core` converter suites (45 passing) and `packages/admin` browser tests for `ImageDetailPanel`, toolbar, bubble menu, image selection and alignment (169 passing).


kwmr:

kwmr:
Thanks for the review. All three are addressed in 5aa743b.

1. Legacy link shape — added `normalizeImageLink()` in core and routed
   the PT type, `Image.astro`, the PT→PM converter and the admin editor
   through it, so `link: "url"` from Gutenberg imports renders correctly
   and is upgraded to `{ href, blank? }` on the next round-trip.
   `rewrite-url-helpers` now handles both shapes too. Admin has no
   dependency on core, so it carries a small mirror of the helper.
   I left `gutenberg-to-portable-text` emitting the string shape for
   now to keep this PR contained; happy to switch it to the object
   shape here or in a follow-up, whichever you prefer.

2. `ImageDetailPanel` resync — `linkHref`/`linkBlank` are now reset
   in the node-switch effect, with the resync test extended to cover
   them.

3. Same-page anchors — `Image.astro` now mirrors `marks/Link.astro`
   and never opens `#section` links in a new tab.

Tests added for the normaliser, the legacy-string round-trip, URL
rewriting for both shapes, and the three render cases.

---
_Generated by [Claude Code](https://claude.ai/code)_

kwmr:
Thanks for the second pass. Addressed in 1665fe5.

`shouldShow` now also admits a `NodeSelection` whose node is an image,
so the image branches of the link handling are reachable. When the
selection is an image the menu renders only the link control; the
text-mark buttons are meaningless there and are hidden rather than
shown as no-ops.

Added bubble-menu tests for a selected image: the menu appears with
the link control only, applying a URL sets `link`, editing the URL
keeps `blank`, and Remove link clears it.

---
_Generated by [Claude Code](https://claude.ai/code)_

kwmr:
Thanks for the re-review.

On the `shouldShow` point, I don't think the guard blocks images: a
`NodeSelection` spans the node (`to = from + node.nodeSize`), so
`selection.empty` is `false` for a selected image. I checked this in
the real editor (the actual `ImageExtension`, not the test mock) in a
browser: selecting the image by click or with the arrow keys gives a
`NodeSelection` from 8 to 9 with `empty: false`, the bubble menu
appears with the link control, and applying a URL from it sets the
image's `link`. The existing bubble-menu tests for a selected image
cover the same path and pass, so I've left `shouldShow` as it is.
Happy to add a test against the real `ImageExtension` if that would
help.

The `aria-label` point was right: the tooltip already said "Image
link" but the accessible name stayed "Insert Link". Fixed in 3fe197c,
and the tests that open the link control with an image selected now
find the button by the exact name "Image link", so they'll catch the
two drifting apart.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #3169, opened 2026-09-17
feat(plugins): add sandboxed taxonomy writes

ascorbic:
## What does this PR do?

Adds the `taxonomies:write` sandboxed-plugin capability as a complete vertical slice. Plugins can create taxonomy terms and apply idempotent add/remove assignment deltas through `ctx.taxonomies`; write implies read, uses term row or translation-group IDs, validates collection attachment, entry and term ownership, configured locales, translation identity, and hierarchy, and invalidates taxonomy caches after successful mutations.

The mutation path is owned by core and is shared by native execution, Cloudflare Worker Loader, and Node/workerd. Registry metadata, manifest reconciliation, marketplace validation, installation/update consent, generated CLI artifacts, plugin-test fixtures and inspectors, public docs, both creating-plugins skills, and package changesets carry the same authority.

A forward migration makes `(translation_group, locale)` unique for taxonomy terms. Existing invalid duplicate rows are split into independent groups after their assignments are copied, and the migration is restart-safe.

Taxonomy-definition management, attachment changes, assignment replacement, term updates, and term deletion remain out of scope. This adds no logged-out route or query.

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] New features link to an approved Discussion: No matching public Discussion covers this term/assignment capability; implementation was requested directly as PR 9 of the maintainer-authored sandboxed capability program.
- [ ] I have included screenshots below if this PR changes the UI — not applicable; this adds localized capability consent copy to the existing dialog without changing its rendered structure or interaction.

- [x] This PR includes AI-generated code — model/tool: GPT-5 Codex

## Screenshots / test output

Screenshots are not applicable. The existing consent dialog renders one new Lingui-wrapped capability label; there is no component, layout, or interaction change.

Validated at `725d983bb7a25e9448f085f6898ebc2901e1dc64`:

- `pnpm build`
- `pnpm typecheck`
- `pnpm lint:json | jq '.diagnostics | length'` (`0`)
- `pnpm format:check`
- `pnpm --dir docs build`
- native taxonomy locale, hierarchy, translation race, assignment delta, migration upgrade/retry, manifest reconciliation, consent/update, registry lexicon, plugin CLI, Cloudflare bridge, workerd bridge, runtime-backed Worker Loader, and real Node/workerd isolate tests
- independent adversarial review and material-fix re-check: passed with no remaining findings


ascorbic:
Thanks for the re-review. I’m not applying the suggested flat-taxonomy guard because it breaks the existing taxonomy contract rather than completing this capability.

I implemented that exact guard on the earlier head. CI then reproduced four regressions in `tests/integration/mcp/taxonomy.test.ts`: the established parent/child, maximum-depth, cycle, and child-protected-delete journeys intentionally exercise the built-in flat `tags` taxonomy. Removing only the blanket `hierarchical` check restored all 140 MCP and plugin-capability tests.

The write capability still validates every structural invariant in `validateParentTerm()`: parent existence, same-taxonomy ownership, translation-group self-parenting, ancestor cycles, and the 100-level depth bound. `hierarchical` remains the existing presentation/tree flag; this PR does not redefine it or change trusted/MCP behavior. The docs and changeset therefore promise hierarchy validation, not a new prohibition on parent IDs for flat definitions.

The independent adversarial re-check also reviewed this exact compatibility decision at `2511a906c` and found no remaining issue. All current-head CI checks are green.

~ 🤖 Codex

ascorbic:
Follow-up on the flat-taxonomy finding: the earlier compatibility conclusion has been superseded by the explicit decision to make hierarchical=false reject parent assignment consistently.

Head 1adaa8422 moves the guard into handleTermCreate and handleTermUpdate, resolves the effective definition by exact locale, configured default, then deterministic fallback, and returns stable VALIDATION_ERROR for a non-null parent on a flat taxonomy. The MCP hierarchy journeys now use hierarchical fixtures, and handler-level SQLite/PostgreSQL-gated reproductions cover both create and update without mutating the rejected term. The changeset, MCP reference, canonical plugin skill, generated skill, and plugin capability docs now state the rejection contract.

Local validation passed: root build and typecheck, type-aware lint (0 diagnostics), focused core tests (163), plugin-cli tests (481), and docs build (85 pages). The exact origin/main...1adaa8422 diff passed an independent adversarial re-review with no remaining findings.

~ 🤖 Codex

ascorbic:

```js
```

**For this PR:** `080_taxonomy_translation_locale_unique` sorts last of the three, so it is the safest to renumber — but it must be renumbered to `081`/`082` if either #3182 or #3184 merges first, which on current ordering is the likely outcome.

Otherwise this PR reviewed clean: capability gating in both bridges, no plugin-supplied identifier reaching SQL, bounded hierarchy walk with no cycle on create, per-locale slug scoping, and runtime-backed tests including a real workerd isolate.

~ 🤖 Codex

ascorbic:
Confirmed and fixed at 99de35d51. I reproduced the strict-ordering failure with Kysely 0.29.2: after applying 080_content_translation_locale_unique and 081_redirect_write_guards, introducing the old 080 taxonomy key fails before pending migrations run with the corrupted-prefix error.

This PR now uses 082_taxonomy_translation_locale_unique everywhere: filename, static import, provider key, migration test, and changeset. The runner integration suite also enforces alphabetically ordered registration with unique, increasing three-digit sequence numbers, so combining two 080 migrations or appending a lower-number registration fails CI.

The coordinated landing order is #3182 as 080, #3184 as 081, then this PR as 082. #3169 must not merge or deploy before those dependencies. Once both land, I will update this branch from main and advance the taxonomy upgrade test baseline from 079 to 081 while keeping the exact assertion that only 082 applies.

Validation passed: focused migration tests (27/27), root build, full package typecheck, type-aware lint (0 diagnostics), docs build, and an independent adversarial review of origin/main...99de35d51 with no findings.

~ 🤖 Codex (gpt-5.6-sol)

PR #3173, opened 2026-09-17
feat(plugins): encrypt sandboxed plugin settings

ascorbic:
## What does this PR do?

Adds a plugin-scoped `ctx.settings` API and encrypts fields declared as `type: "secret"` before persistence. Native plugins, Cloudflare Worker Loader plugins, and Node/workerd plugins use the same versioned AES-GCM envelope, with plugin ID and setting key as authenticated data.

The first configured key encrypts new values and `kid` selects older keys during rotation. Existing plaintext settings remain readable and migrate when saved again. Missing, wrong, and tampered keys fail closed at plugin reads, while the admin can still replace an unreadable credential. Exact current and previous secret values are redacted from host plugin logs with bounded state.

Generated settings, the CLI descriptor, runtime-backed plugin-test helpers, Cloudflare templates, public docs, backup guidance, and the canonical/generated creating-plugins skill now describe the same contract. The existing `ctx.kv` `settings:` namespace remains a compatibility alias throughout EmDash 0.x.

This does not add a capability or `declaredAccess` facet: settings already use a host-fixed per-plugin namespace, and an administrator grants the plugin a credential by entering it into that plugin's settings form.

Related planning: maintainer-directed encrypted-settings slice from the sandboxed plugin capability expansion plan.

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] User-visible strings in the admin UI are [wrapped for translation](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md#internationalization-i18n) (if applicable). No admin strings or `messages.po` files changed.
- [ ] New features link to an approved Discussion: no public Discussion URL; this is the maintainer-directed PR 13 from the supplied capability plan.
- [x] I have included screenshots below if this PR changes the UI. Not applicable: no rendered UI changed.

- [x] This PR includes AI-generated code — model/tool: GPT-5 Codex

## Screenshots / test output

Screenshots: Not applicable; this changes host/runtime behavior, tests, templates, and documentation rather than rendered UI.

Validation at current head `3d6600536d4b645e15cd1dbe50e34c26640f3d2f`:

- `pnpm build`
- `pnpm typecheck`
- `pnpm lint:json | jq '.diagnostics | length'` → `0`
- package matrix: core 6,906 tests, Cloudflare 437 tests, workerd 149 tests, plugin CLI 481 tests, plugin-test 14 tests, and all other reached package suites passed
- `pnpm --dir docs build`
- mandatory independent adversarial review passed at the exact current head


ascorbic:
Addressed the stale encryption-key documentation in 9ef10a716. The startup validator now documents its current non-throwing behavior and the settings-boundary fail-closed behavior; I also updated the adjacent runtime comment and `resolveSecrets` note so they no longer describe plugin encryption as future work.

Validated with `pnpm format`, `pnpm lint:quick`, `pnpm build`, `pnpm typecheck`, type-aware lint, and `git diff --check`.

~ 🤖 Codex

ascorbic (commented):
Reviewed at 9ef10a716. The envelope design holds up: AAD bound to plugin ID and key, `kid` selection, one `createSettingsAccess` boundary shared by the in-process context and both bridges, and the admin handler able to replace an unreadable secret. Four things, in order of how much they'd hurt after release.

### 1. A missing key now fails saves, and on Node the scaffolded key is never loaded

`encryptPluginSetting` throws `PLUGIN_SETTING_ENCRYPTION_KEY_MISSING` when no key resolves. That is a behavior change for every existing site with a `type: "secret"` field and no key: saving it worked before this PR and errors after. Failing closed is a reasonable call, but two things make it rough in practice.

**The operator can't tell why.** `handlePluginSettingsUpdate` calls `encodePluginSettingValue` inside its `try`, and the bare `catch` maps everything to `PLUGIN_SETTINGS_UPDATE_ERROR` / "Failed to update plugin settings". The admin toast shows that string. The specific code never leaves the handler. Suggest catching `PluginSettingEncryptionError` and returning its code with a message that names `EMDASH_ENCRYPTION_KEY`, so the admin can map it to something actionable.

**On Node, `.env` does not reach `process.env`.** `readDefaultEnv` reads `process.env` only, which is right. But Astro copies `.env` into `process.env` only during `astro build` (`astro/dist/env/vite-plugin-env.js`, `buildStart` returns early unless `isBuild`), and nothing loads it for `node ./dist/server/entry.mjs`. I checked this on `demos/simple` with astro 7.0.0: an invalid `EMDASH_ENCRYPTION_KEY` in `.env` produced no `[emdash] EMDASH_ENCRYPTION_KEY is invalid` line after the runtime initialised under `astro dev`; the same value exported in the shell produced it on the first request.

`create-emdash` writes the key into `.env` for Node templates, and `docs/reference/cli.mdx:710-711` still says "The same `.env` file is read by Node" at this head. So a fresh Node site, following the docs, will have a key on disk, no key at runtime, and a generic failure the first time someone saves the forms plugin's Turnstile secret. Before this PR that gap was harmless because nothing consumed the key.

Minimum fix here: correct the `cli.mdx` sentence, say in `deployment/nodejs.mdx` and the changeset that the variable must be in the real process environment, and surface the error code per the point above. Whether the Node templates should run with `--env-file=.env` is a separate decision.

### 2. The `nodejs_compat_populate_process_env` requirement looks overstated

The changeset says existing Cloudflare sites "must add" the flag, and all four templates gain it. Cloudflare's docs say it is on by default when `nodejs_compat` is enabled and `compatibility_date` is 2025-04-01 or later ([environment variables](https://developers.cloudflare.com/workers/configuration/environment-variables/), [changelog](https://developers.cloudflare.com/changelog/post/2025-03-11-process-env-support/)). The templates are on `2026-02-24`, so the flag should be redundant there, and the upgrade note only applies to sites pinned before April 2025.

If the flag was added because the admin route actually saw an empty `process.env` on the templates' date, that's worth a comment in the PR, since it would mean something else is off. If not, scope the changeset sentence to older compatibility dates and drop the template change.

### 3. The bundled plugins that hold secrets are not covered

Encryption keys off `settingsSchema`, and no file under `packages/plugins` changes. At this head:

- webhook-notifier still writes `settings:secretToken` through `ctx.kv.set` (`plugin.ts:516`) with no schema, so it stays plaintext.
- atproto does the same for `settings:appPassword` (`plugin.ts:730`).
- atproto also stores `state:accessJwt` and `state:refreshJwt` (`atproto.ts:299`). Those are live credentials and the settings API has no path for them.

The new docs say the right thing ("A `secret_input` in Block Kit only masks browser input"), but our own plugins don't follow it yet. One wrinkle for the fix: declaring `settingsSchema` also sets `hasSettings` (`api/handlers/plugins.ts:91,139`), which turns on the generated settings form and gear icon. A Block Kit plugin that declares a schema only to get encryption ends up with two settings UIs. Worth deciding whether that is intended before converting the two plugins, and whether `state:` credentials need an answer in this slice or a follow-up.

Fine as a follow-up PR, but the changeset reads as though plugin secrets are now encrypted, and for two of the three bundled plugins that store one, they aren't.

### 4. Rotation asks operators to do something they can't verify

The docs say to keep old keys "until every plugin secret has been saved again". Nothing reports which stored envelopes still reference which `kid`, or which declared secrets are still plaintext, so there is no way to know when dropping a key is safe. Getting it wrong is unrecoverable for that value. Not blocking, but a status surface (kids in use, plaintext count per plugin, no values) should follow closely, and the rotation docs could say plainly that there is currently no way to check.

~ 🤖 Claude Fable 5.1

ascorbic:
Review of `9ef10a716`. One blocker. The crypto itself is solid — the problem is a read path that was left behind.

## Blocker: the public `getPluginSetting()` / `getPluginSettings()` helpers return the raw ciphertext envelope

The write path now encrypts (`api/handlers/plugin-settings.ts:193-198`, via `encodePluginSettingValue`), but the public read helpers were not updated:

```ts
export async function getPluginSettingWithDb<T = unknown>(
	pluginId: string, key: string, db: Kysely<Database>,
): Promise<T | undefined> {
	const options = new OptionsRepository(db);
	const value = await options.get<T>(`plugin:${pluginId}:settings:${key}`);
	return value ?? undefined;
}
```
<sub>`packages/core/src/settings/index.ts:357-365`; `getPluginSettingsWithDb` has the same gap at `:382-399`</sub>

These read the exact rows (`plugin:<id>:settings:<key>`) the PR starts writing as `{v:1,kid,iv,ciphertext}`, but never call `decodePluginSettingValue`/`decryptPluginSetting` and never detect an envelope. Both are exported from the package root (`packages/core/src/index.ts:474-475`), and their own docstrings say "Use `getPluginSetting()` in templates and plugin rendering code."

**Failure scenario:** an existing install upgrades. A plugin declares `apiKey: { type: "secret" }`. Template or native-plugin code that isn't holding a `PluginContext` calls `await getPluginSetting<string>("mail-plugin", "apiKey")`. Before the upgrade that returned the plaintext string; after the operator next saves that setting in the admin UI it returns an object — still typed `string` by the generic — so the call site sends `Bearer [object Object]` upstream. No throw, no log, no type error. `getPluginSettings()` does the same for every secret in the map. Nothing in the changeset or docs mentions it.

**Fix** — both AAD inputs (`pluginId`, key) are already in scope, so no schema change is needed:

```ts
const value = await options.get(`plugin:${pluginId}:settings:${key}`);
return isEncryptedPluginSetting(value)
	? ((await decryptPluginSetting(pluginId, key, value)) as T)
	: (value ?? undefined);
```

…and the same envelope check per entry in `getPluginSettingsWithDb`. If reading secrets outside the plugin runtime is meant to be unsupported, these should throw `PluginSettingEncryptionError` rather than hand back ciphertext — but either way it can't silently return an envelope, and a deliberate break needs the changeset to say so.

---

Checked and clean, so the crypto isn't the issue: key material comes only from `process.env` via `readDefaultEnv()` or the Worker secret binding `this.env.EMDASH_ENCRYPTION_KEY` — no `import.meta.env` anywhere in the key path, and the sandbox env passes only `PLUGIN_ID`/`PLUGIN_VERSION`/`BRIDGE`, so the key never reaches plugin isolates. AES-GCM, fresh 12-byte `crypto.getRandomValues` IV per encryption, 32-byte raw key, AAD binding `["emdash-plugin-setting", 1, pluginId, key]` so cross-plugin and cross-key envelope reuse fails (with a test), `kid` for rotation, fatal UTF-8 decode, no `Math.random`. Legacy plaintext still reads and is re-encrypted on save, and there's no backfill migration, so nothing can be left unreadable. Missing/unknown/tampered keys fail closed without leaking plaintext into errors, `buildSettingsResponse` still returns only `secretsSet` booleans, both routes gate on `requirePerm(user, "plugins:manage")`, and the plugin-controlled `list` prefix can't escape the `plugin:<id>:settings:` namespace. Tests use real WebCrypto and real SQLite/D1/Worker-Loader/workerd round-trips including rotation and tamper cases — no mocked crypto.

~ 🤖 Codex

ascorbic:
Addressed the latest reviews in fa2d07df7:

- `getPluginSetting*` and `getPluginSettings*` now decrypt stored secret envelopes instead of returning ciphertext. New envelopes carry the exact `$emdash: "plugin-setting"` brand, so malformed branded records fail closed while arbitrary legacy JSON continues to round-trip.
- Missing and malformed `EMDASH_ENCRYPTION_KEY` errors are actionable and mixed updates still write nothing. Malformed-key guidance prioritizes restoring the backed-up key.
- Node documentation now states that standalone Node does not load `.env` automatically.
- Cloudflare guidance now scopes `nodejs_compat_populate_process_env` to compatibility dates before `2025-04-01`; redundant modern-date template and test-host flags were removed.
- The changeset and canonical secrets documentation now state that undeclared KV/state credentials, including the current AT Protocol and webhook notifier values, are outside this encryption path.
- Rotation documentation now states that EmDash cannot yet report key IDs in use, so operators must track resaves and verify integrations before removing an old key. The stale `create-emdash` bump was removed.

Validated with the root build and typecheck, type-aware lint, docs build, 63 focused core tests, 14 runtime-backed plugin tests, formatting, and `git diff --check`. The independent adversarial gate passed on the exact working diff before push.

~ 🤖 Codex (gpt-5.6-sol)

PR #3176, opened 2026-09-17
fix(cloudflare): fall back to process.env for media-provider credentials

### LINKED ISSUE #2453: Cloudflare media-provider *EnvVar options ignore process.env under Node adapters
## Summary

The Cloudflare Images and Cloudflare Stream media-provider runtimes document `*EnvVar` options as runtime environment-variable names, but they resolve those names only from `cloudflare:workers` `env`. Under a Node adapter, exported `process.env` values are ignored and provider operations fail as missing credentials.

This is both an implementation defect for an advertised adapter-neutral config shape and a documentation ambiguity: the error says to set an environment variable, and the provider examples say the default/custom names are read at runtime, without saying that only Workers bindings are consulted.

## Minimal reproduction (Node adapter)

```js
// astro.config.mjs
import node from '@astrojs/node';
import { cloudflareImages } from '@emdash-cms/cloudflare';
import { defineConfig } from 'astro/config';
import emdash from 'emdash/astro';

export default defineConfig({
  adapter: node({ mode: 'standalone' }),
  integrations: [
    emdash({
      mediaProviders: [
        cloudflareImages({
          accountIdEnvVar: 'CF_ACCOUNT_ID',
          accountHashEnvVar: 'CF_IMAGES_ACCOUNT_HASH',
          apiTokenEnvVar: 'CF_IMAGES_TOKEN',
        }),
      ],
    }),
  ],
});
```

Export non-secret test values before starting Astro (real credentials are needed only to get past resolution):

```sh
CF_ACCOUNT_ID=test-account \
CF_IMAGES_ACCOUNT_HASH=test-hash \
CF_IMAGES_TOKEN=test-token \
astro dev
```

Open the media library or invoke a provider operation. Expected: the provider resolves the exported Node environment variables and reaches the API (which may reject these test values). Actual: resolution stops locally with `Cloudflare Images: Missing CF_ACCOUNT_ID`.

Cloudflare Stream has the same behavior for `accountIdEnvVar` and `apiTokenEnvVar`.

## Current implementation

Both runtimes import `env` directly from `cloudflare:workers` and `resolveEnvValue()` reads only:

```ts
(env as Record<string, string | undefined>)[envVar]
```

- `packages/cloudflare/src/media/images-runtime.ts`
- `packages/cloudflare/src/media/stream-runtime.ts`

Direct config values work because they are checked first, but inlining an API token into the generated `virtual:emdash/media-providers` module is not an appropriate general workaround for secrets.

## Expected behavior / candidate shape

Please preserve the existing precedence and Workers behavior:

1. A direct config value remains authoritative when explicitly supplied.
2. If a `cloudflare:workers` binding exists for the requested name, use it.
3. Only when that Workers binding is absent, fall back to `process.env[envVar]` in runtimes where `process` is available.
4. Otherwise throw the existing missing-variable error.

Conceptually:

```ts
const workersValue = (env as Record<string, string | undefined>)[envVar];
const nodeValue =
  typeof process !== 'undefined' ? process.env?.[envVar] : undefined;
const value = workersValue ?? nodeValue;
```

An adapter-generated env virtual module would also be reasonable if it avoids a direct Node-global reference. The key contract is that Workers bindings retain precedence and secrets remain runtime values rather than being serialized into generated source.

Documentation should explicitly distinguish Workers bindings from Node `process.env` and state the precedence.

connorblack:
## What does this PR do?

Fixes `cloudflareImages()` and `cloudflareStream()` (`@emdash-cms/cloudflare`) so their `accountIdEnvVar`, `accountHashEnvVar`, and `apiTokenEnvVar` options fall back to `process.env` when no Cloudflare Workers binding exists for that name. `resolveEnvValue()` previously read only `cloudflare:workers`'s `env`, so a Node-adapter site with credentials exported as environment variables failed with `Missing CF_ACCOUNT_ID` (etc.) even though the variable was set. A Workers binding of the same name still takes precedence when one exists, and a direct config value still takes precedence over both; the missing-variable error is unchanged when neither source has it.

Docs now state that precedence once, above both provider option tables in the configuration reference.

@emdashbot started on this but didn't finish before its work window ended (checkpoint in the issue thread). This is a fresh implementation following the precedence @ascorbic specified in the issue body: direct config value → Workers binding → `process.env`.

Closes #2453

## Type of change

- [x] Bug fix

## Checklist

- [x] `pnpm typecheck` passes (`@emdash-cms/cloudflare`, after `pnpm build` at the root)
- [x] `pnpm test` passes — full `@emdash-cms/cloudflare` suite: 444 passed, 2 skipped
- [x] User-visible strings... Not applicable, no admin UI changed.
- [x] I have added and reviewed the user-facing changeset (`@emdash-cms/cloudflare`: patch)
- [x] New features link to an approved Discussion — Not applicable, this is a bug fix.
- [x] Screenshots — Not applicable, no UI change.

## Screenshots / test output

Not applicable for UI. `tests/media/images-runtime.test.ts` (new) and `tests/media/stream-runtime.test.ts` (extended) each cover four cases: Workers-binding-only, process.env fallback, direct-value override, and neither-present.

Against `main` without the runtime changes (source reverted, tests kept):

```
 × falls back to process.env when no Workers binding is set
   → Cloudflare Images: Missing CF_IMAGES_ACCOUNT_HASH...
 × falls back to process.env when no Workers binding is set
   → Cloudflare Stream: Missing CF_ACCOUNT_ID...
 × prefers the Workers binding over process.env
   → Cloudflare Stream: Missing CF_STREAM_TOKEN...
 Test Files  2 failed (2)
      Tests  3 failed | 5 passed (8)
```

With the fix:

```
 Test Files  2 passed (2)
      Tests  10 passed (10)
```


connorblack:
@ascorbic one thing I noticed while in there but left alone: `resolveEnvValue()` is duplicated verbatim between `images-runtime.ts` and `stream-runtime.ts` — pre-existing, not something this PR introduced, I just applied the same fix to both copies. Want me to sweep it into a shared helper as part of this PR, or leave it as-is? Didn't want to fold an unrequested refactor into a bugfix PR per the scope-discipline note in CONTRIBUTING.md, so I left it duplicated for now.

connorblack:

connorblack:
recheck

PR #3181, opened 2026-09-17
fix(plugin-forms): defer the submission webhook and log a failed response

### LINKED ISSUE #3136: plugin-forms: webhook call is not awaited or passed to waitUntil, and a non-2xx response is never logged
### Description

The webhook step in `packages/plugins/forms/src/handlers/submit.ts` (step 9, "fire and forget") looks like this:

```ts
ctx.http
  .fetch(settings.webhookUrl, { method: "POST", ... })
  .catch((err) => { ctx.log.error("Webhook failed", ...) });
```

This causes two problems:

1. **It isn't awaited, and it isn't handed to `waitUntil`.** The handler returns straight after. On Cloudflare Workers, work that isn't registered with `waitUntil` isn't guaranteed to finish once the response has been sent ([docs](https://developers.cloudflare.com/workers/runtime-apis/context/#waituntil)). EmDash already exports `after()` for exactly this.
2. **Only a rejected promise gets logged.** `fetch` resolves on a 4xx or 5xx, so those responses are ignored. `createHttpAccess` follows redirects, and the plugin has `allowedHosts: ["*"]`, so a webhook behind an auth wall redirects to the login page and resolves with that page's 200. In every one of these cases the submission looks successful, the webhook did nothing, and nothing appears in the logs.

We ran into this with a webhook pointed at a route on the same site, which turns a submission into a draft entry. Submissions were stored and emails went out, but no draft was ever created and nothing was logged. On Workers there's a third wrinkle for same-origin webhooks: the call goes out over the public hostname, so it passes through whatever sits in front of the site (Cloudflare Access, in our case). A service binding to the Worker itself avoids that, but that's probably a docs note rather than something the plugin should do.

Suggested fix for 1 and 2:

```ts
after(async () => {
  try {
    const res = await ctx.http!.fetch(url, init);
    if (!res.ok) ctx.log.error("Webhook failed", { status: res.status, url });
  } catch (err) {
    ctx.log.error("Webhook failed", { error: String(err), url });
  }
});
```

The redirect-to-login case would still return 200. Treating a redirected response as a failure, or at least logging the final URL when it differs from the webhook URL, would catch it.

### Steps to reproduce

1. Create a form and set its `webhookUrl` to an endpoint that returns 500 (or one that redirects to a login page).
2. Submit the form.
3. The submission is stored and the visitor sees the confirmation, but "Webhook failed" is never logged.

### Environment

- emdash 0.38.0, `@emdash-cms/plugin-forms` 0.2.6. The same code is on main at f57974f.
- Runtime: Cloudflare Workers

### Screenshots

Not applicable.

eisenbruch:
## What does this PR do?

Closes #3136.

The submission webhook was a floating promise, and only a rejected one was logged. Two consequences, both silent:

1. **Nothing awaited it and nothing handed it to the host.** `submitHandler` returned immediately after firing it, so on Cloudflare Workers the isolate could be torn down before the request went out. It now runs inside `after()`, which registers the promise with `waitUntil` where the host provides one.
2. **`fetch` only rejects on a transport error.** A 4xx and a 5xx both resolve, and so does the sign-in page an authenticated endpoint redirects to, because `createHttpAccess` follows redirects and the plugin declares `allowedHosts: ["*"]`. Every one of those looked like a delivered webhook and wrote nothing to the log. The response is now inspected, and a 2xx that arrived by redirect is logged as well, since the body came from somewhere other than the configured URL.

We hit this on a live site with a webhook pointed at a route on the same host: submissions stored, notification emails sent, no webhook effect, nothing in the logs to say so.

The emails in steps 7 and 8 are already awaited, so they are left alone.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes (`@emdash-cms/plugin-forms`)
- [x] `pnpm lint` passes (`oxlint --type-aware --deny-warnings`, 0 diagnostics)
- [x] `pnpm test` passes (or targeted tests for my change): `packages/plugins/forms`, 5 files, 26 tests
- [x] I have added and reviewed the user-facing [changeset](https://github.com/emdash-cms/emdash/blob/main/.changeset/README.md) (`@emdash-cms/plugin-forms`: patch)

## Screenshots / test output

Not applicable, no UI change.

`packages/plugins/forms/tests/webhook.test.ts` covers four cases: a 5xx, a redirect to a different URL, a transport error, and a clean success that logs nothing. The first two fail against `main` and pass with this change; the other two pass on both.

There is one thing I did not do, and it is worth a maintainer's opinion. A same-origin webhook on Workers still goes out over the public hostname, so it passes through whatever sits in front of the site — Cloudflare Access, in our case — and gets the sign-in page. A service binding back to the Worker avoids that, but it looks like a docs note rather than something the plugin should reach for, so this PR only makes the failure visible rather than trying to solve it.


eisenbruch:
You are right on all three, and the test deserved the criticism more than the code did. Fixed in e69a5df3.

**The redirect guard.** I confirmed it in `createHttpAccess`: the wrapper loops with `globalThis.fetch(currentUrl, { redirect: "manual" })` and returns the last response, so `redirected` is always false however many hops it took. Guarding on it meant the auth-wall case never fired in production, which is the failure the PR exists to surface. It now compares the final URL, which the wrapper does carry, through a small `sameTarget` helper that ignores a trailing-slash difference so a webhook answered at its own address with a slash is not reported as a redirect. It also requires `response.url` to be non-empty, so a runtime that does not populate it stays quiet rather than warning on every delivery.

**The test.** This is the part I got wrong rather than merely incomplete. Setting `redirected: true` on a constructed `Response` meant it exercised a shape that does not exist in production and would have passed with the guard deleted, which is worse than no test. It now reproduces the wrapper's own manual-redirect loop over a stubbed `globalThis.fetch`, so the handler sees the response shape it really gets, and there is a fifth case covering the trailing-slash redirect that should stay quiet.

I did try driving the real `createHttpAccess` directly, which is what you suggested first. It resolves the hostname over DoH before every request, so the 5xx case failed with `Could not resolve hostname: DoH lookup failed` against a test host. Reproducing the loop is the alternative you offered; if you would rather the wrapper exposed redirect metadata so plugins need not reimplement this, say so and I will do that instead, since it would remove the duplication and let the test use the real thing.

**Verified the way it should have been the first time:** I put the old `response.redirected` guard back and confirmed the auth-wall case now fails, then restored the fix. Full forms suite green at 5 files and 27 tests, `oxlint --type-aware --deny-warnings` clean, typecheck clean, `oxfmt` run.

**The changeset** no longer claims more than the code does; it names the URL comparison and why `Response.redirected` cannot be used here.

ascorbic (approved):

PR #3182, opened 2026-09-17
feat(plugins): support translation-aware content creation

ascorbic:
## What does this PR do?

Adds translation-aware content creation for sandboxed plugins as the next slice stacked on #3171.

Plugins can pass `{ locale, translationOf }` to `ctx.content.create()`. EmDash validates the active source in the same collection, atomically joins its translation group, preserves non-translatable fields, byline credits, and taxonomy assignments, and runs the normal content validation and save-hook path across native, Cloudflare Worker Loader, and Node.js workerd execution.

The database permits one active row per case-insensitive locale in a translation group. The forward-only migration repairs historical duplicates before adding the unique index, and stable `CONFLICT`, `NOT_FOUND`, `VALIDATION_ERROR`, and `SAVE_REJECTED` codes cross both sandbox transports. Save-hook-originated creates do not re-enter save hooks.

Closes: n/a

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] New features link to an approved Discussion: n/a — this is the next slice of the maintainer-provided sandboxed capability expansion plan used by #3171.

## Screenshots / test output

Not applicable. This PR has no admin UI changes.

Validated with:

- `pnpm build`
- `pnpm typecheck`
- `pnpm format:check`
- `pnpm lint:json` (zero diagnostics)
- `pnpm --dir docs build`
- focused native, SQLite migration/concurrency, Cloudflare Worker Loader, Node.js workerd, plugin-test, and plugin CLI suites
- full `@emdash-cms/plugin-test` runtime-host suite (15/15)
- independent adversarial review of the exact stacked diff, including material-fix rechecks and a final clean pass

One full workerd run encountered a transient local port collision; the affected wall-time test passed immediately in isolation, and the complete bridge suite plus the translation/restart real-workerd journey passed.


ascorbic:
Checked this before changing the migration. PostgreSQL truncates identifiers written in commands to `NAMEDATALEN - 1` bytes; that applies when the identifier is referenced by `DROP INDEX` as well as when it is created, including quoted identifiers. The existing full generated name in `down()` therefore resolves to the same stored 63-byte identifier. SQLite stores and resolves the full name consistently.

I also drafted the suggested truncated-name helper and a long-slug rollback assertion. Independent adversarial review found that the assertion passes against the existing implementation too, so it does not reproduce a regression. I discarded that non-fix and left the current green head unchanged.

PostgreSQL reference: https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS

~ 🤖 Codex

ascorbic:
Review of `f262c2e45`, scoped to this PR's own diff (`origin/feat/plugin-content-discovery...origin/feat/plugin-translation-create`). One blocker. #3171 underneath it came back clean.

## Blocker: on Postgres the restore conflict mapping never matches, so 409 becomes 500

```ts
function isTranslationLocaleConflict(error: unknown): boolean {
	if (!(error instanceof Error)) return false;
	const message = error.message.toLowerCase();
	return (
		message.includes("active_tg_locale") ||
		(message.includes("unique constraint failed") &&
			message.includes("translation_group") &&
			message.includes("locale"))
	);
}
```
<sub>`packages/core/src/api/handlers/content.ts:67-76`</sub>

Postgres raises `duplicate key value violates unique constraint "<name>"`, which contains neither `unique constraint failed` nor the column names — so on Postgres the entire check rests on the index name containing `active_tg_locale`.

That name gets truncated. The migration itself already accounts for this:

```ts
const indexName = `uidx_${tableName}_active_tg_locale`;
const storedIndexName = isPostgres(db) ? indexName.slice(0, 63) : indexName;
```
<sub>`080_content_translation_locale_unique.ts:70-71`, and again at `:84-85`</sub>

`uidx_ec_{slug}_active_tg_locale` is 25 characters plus the slug, and Postgres truncates identifiers at 63 bytes. Any collection slug of 39+ characters (well inside the 63-char slug limit) stores a name ending `_active_tg_local`, `_active_t`, and so on — the substring test fails.

**Failure scenario:** Postgres site, collection slug ≥ 39 chars. Trash the `fr` translation, create a replacement `fr` translation, then restore the trashed one. `handleContentRestore` falls through to `CONTENT_RESTORE_ERROR` → HTTP 500 with a `console.error`, instead of the documented 409 `CONFLICT` ("An active translation already exists in this locale"). The operator gets an unexplained server error from an ordinary Restore click.

The branch's own `content-translation-index-name.test.ts` asserts the truncated names via `expectedName()`, so the truncation is already understood here — it just isn't reflected in the matcher.

**Fix:** match the name the migration actually computes instead of a suffix. `isTranslationLocaleConflict` already receives the collection at both call sites:

```ts
const expected = `uidx_ec_${collection}_active_tg_locale`.slice(0, 63).toLowerCase();
// ... message.includes(expected) || (existing SQLite clause)
```

---

Checked and clean: `translation_group` is taken atomically via `COALESCE(translation_source.translation_group, translation_source.id)` inside `INSERT … SELECT … WHERE id = ? AND deleted_at IS NULL`, so the group is reused and a source trashed mid-save yields zero rows → `NOT_FOUND`. Cross-collection sources are impossible (the SELECT is scoped to the same `ec_*` table). `(translation_group, lower(locale))` uniqueness is enforced by migration 080 *and* mirrored in `SchemaRegistry.createContentTable`, closing the read-then-insert race. Slug uniqueness stays per-locale via `generateUniqueSlug(collection, source, effectiveLocale)`. Locale is re-validated host-side by `resolveContentCreateLocale` even though both bridges validate first, `content:write` is re-checked in each bridge, and `originHook` comes from the host invocation path rather than plugin data, so save-hook suppression isn't spoofable. Migration 080's split loop provably terminates and doesn't collide with `main` (at 079).

One non-blocking note, since this changes a shared host path: `runtime.handleContentCreate` also serves admin `POST /_emdash/api/content/[collection]`, so admin translation creation now also rejects unconfigured locales with `VALIDATION_ERROR`, force-inherits non-translatable fields from the source, and can return `CONFLICT`. Consistent with the group-shared-field model, but the `emdash` changeset frames these as plugin-only.

~ 🤖 Codex

ascorbic:

```js
```

**For this PR:** `080_content_translation_locale_unique` sorts first of the three, so it's the only one that is safe to merge at 080 without renumbering — but only if it merges *before* #3184 and #3169 ship. If either of those lands first, this one must be renumbered.

~ 🤖 Codex

ascorbic:
Addressed both review points at `77d7ab559`.

- `isTranslationLocaleConflict()` now computes the collection-specific stored index name, including PostgreSQL's 63-byte truncation, while retaining the existing SQLite column-based clause.
- Added a 39-character collection-slug restore regression to the dialect integration suite. The focused suite passes locally (8 tests); lint, full build, and typecheck also pass. An independent adversarial review found no further bugs.
- Kept `080_content_translation_locale_unique` as migration 080. This PR must land before #3184's planned 081 and #3169's planned 082; no independent renumbering was made here.

~ 🤖 Codex (gpt-5.6-sol)

PR #3184, opened 2026-09-17
feat(plugins): add sandboxed redirect access

ascorbic:
## What does this PR do?

Adds capability-gated, versioned redirect CRUD for sandboxed plugins across native, Worker Loader, and Node/workerd runtimes. It reuses host validation and cache invalidation, adds serialized database guards, and extends the runtime-backed test host. Related foundation: #3162.

Merge order: #3182 is merged as migration 080; this PR uses 081. #3169 follows with 082.

## Type of change

- [x] Feature

This is a maintainer-directed capability slice following #3162; there is no separate Discussion URL.

## Checklist

- [x] I have read `CONTRIBUTING.md`
- [x] Typecheck passes
- [x] Lint passes
- [x] Tests pass
- [x] Formatting run
- [x] Tests added or updated
- [x] Admin consent copy uses Lingui
- [x] User-facing changeset added and reviewed
- [ ] Approved Discussion link (not applicable; maintainer-directed plan)
- [x] Screenshots (not applicable; no component or layout change)

- [x] AI-generated code — model/tool: gpt-5.6-sol / Codex

## Screenshots / test output

Not applicable; no interface component or layout change.

Validated at `83ab759b3`: full build, format check, typecheck, type-aware lint, 144 focused SQLite/PostgreSQL migration and redirect tests, 13 D1/workerd migration tests, both sandbox bridges and the combined Stack A authoring/manifest/test-host consumers, Worker startup closure (18 modules / 1,728,969 bytes), changeset validation, `git diff --check`, and independent adversarial conflict review.


ascorbic:
Review of `d93801621`. Two blockers; everything else I checked came back clean.

## 1. The new loop trigger aborts legitimate slug renames, rolling back the whole content save

`createAutoRedirect` collapses chains *before* deleting the rules whose source is the new URL:

```ts
await repository.collapseChains(oldUrl, newUrl, fence);
// The new URL serves live content again — any redirect from it would shadow the page.
... .deleteFrom("_emdash_redirects").where("source", "=", newUrl)
```
<sub>`packages/core/src/database/repositories/redirect.ts:539`</sub>

`emdash_redirect_loop_update` (migration `080_redirect_write_guards.ts:221`, and the Postgres branch at `:99-115`) evaluates the intermediate state and rejects it.

Repro against SQLite 3.51 with the trigger body copied verbatim from the migration:

```sql
INSERT INTO _emdash_redirects VALUES ('r1','/blog/b','/promo',1);
INSERT INTO _emdash_redirects VALUES ('r2','/promo','/blog/a',1);
-- editor renames a post's slug a -> b; step 1 of createAutoRedirect:
UPDATE _emdash_redirects SET destination='/blog/b' WHERE destination='/blog/a' AND enabled=1;
```
```
Runtime error near line 25: redirect loop (19)
```

Two ordinary manual redirects that coexist fine today (`/blog/b → /promo`, `/promo → /blog/a`) make the rename impossible. The transient loop involves `r1` — the row the *very next* statement deletes.

This is not confined to the redirects admin. `createSlugChangeRedirect` runs inside the content transaction (`api/handlers/content.ts:1108`, and the publish path at `:1694`, both passing `trx`) and has no try/catch, so the `RAISE(ABORT)` propagates out and rolls back the entire update. The editor gets `CONTENT_UPDATE_ERROR`, and the rename stays impossible until someone deletes the redirect by hand.

**Fix:** run the `deleteFrom(_emdash_redirects).where("source","=",newUrl)` statement *before* `collapseChains(oldUrl, newUrl, fence)`. I verified the end state is identical (those rows are deleted either way) and the transient loop never forms:

```
r2|/promo|/blog/b
```

## 2. Migration 080 fails closed against previous-version code, breaking content saves mid-deploy

```sql
WHEN NEW.write_generation = 0 AND (
  NEW.is_pattern = 1 OR EXISTS (SELECT 1 FROM _emdash_redirects WHERE enabled = 1 AND is_pattern = 1))
BEGIN SELECT RAISE(ABORT, 'pattern redirect writes require the current runtime'); END
```
<sub>`080_redirect_write_guards.ts:177-198`, Postgres equivalent at `:93-99`</sub>

`write_generation` is `notNull().defaultTo(0)`, so every write from the previously released binary aborts once the site has one enabled pattern rule.

The code comment frames this as protecting *pattern* writes, but the `EXISTS` branch catches non-pattern writes too — including the auto-redirect inside the content-update transaction. So on a site with any enabled pattern redirect, slug renames and publishes 500 on every still-running old instance (in-flight Workers isolates after a deploy, rolling Node replicas), and permanently after a code rollback, since migrations are forward-only and `down()` is a no-op at `:258`.

That's the case AGENTS.md § Migrations rules out: "old application code must tolerate the expanded schema during a rolling deploy". The migration test asserts the abort as intended behaviour and the changeset doesn't mention it.

**Fix:** drop `emdash_redirect_legacy_pattern_insert` / `emdash_redirect_legacy_pattern_update` and the `write_generation = 0` branch in `emdash_redirect_validate_write`. Unfenced writes already lose the race safely, so old binaries are no worse off than they are today. If the guard stays, it must at minimum be unreachable from the content-save path and called out in the changeset as an upgrade-ordering requirement.

---

Checked and clean: `source`/`destination` both go through `urlPath` (leading `/`, rejects `//`, CRLF, traversal) on the plugin path too, and the middleware re-checks `//` and `/\` and skips `/_emdash*` — no open redirect, no admin/API hijack, no `javascript:` or protocol-relative escape. Status codes are constrained in zod, the workerd bridge validator, and `RedirectStatus`. No new queries on the logged-out path. Both bridges gate every op on `redirects:read`/`redirects:write`, `auto` is rejected via `assertNoAutomaticRedirectMarker`, `_rev` is enforced atomically in the `WHERE`, `findMany` clamps to 100 and returns `{items, nextCursor}`, all SQL is parameterised, and the changeset covers all eight packages.

~ 🤖 Codex

ascorbic:

```js
```

**For this PR:** `080_redirect_write_guards` sorts after #3182's migration. If #3182 ships first, this one is fine as-is; if this one ships first, #3182 must renumber. Coordinate before merging either.

This is separate from, and in addition to, the two blockers in my earlier comment.

~ 🤖 Codex

ascorbic:
Re-reviewed at `0c96a431b`. Blocker 1 and the migration renumber are genuinely fixed. Blocker 2 is fixed for the case I reported, but the same rolling-deploy rule is still violated by a different path — so this isn't quite done.

## Fixed

**Slug-rename abort.** The delete now precedes `collapseChains` in `createAutoRedirect` (`redirect.ts:538-555`), and it's a real reorder rather than a relaxed trigger — `emdash_redirect_loop_update`/`_insert` are byte-identical to the previous head, and no try/catch was added around the call. My original repro now completes, leaving `/blog/a → /blog/b` and `/promo → /blog/b`. The fence `EXISTS` guard survives on the delete, `idx_redirects_managed_source` isn't newly exposed (the delete targets `newUrl`, the insert uses `oldUrl`), and the revert case A → B → A leaves no self-redirect.

**Pattern guards.** `emdash_redirect_legacy_pattern_insert`/`_update` and the `write_generation = 0` branch of the plpgsql `emdash_redirect_validate_write` are gone — `git grep 'legacy_pattern\|pattern redirect writes require'` has no hits. An old binary's unfenced insert/update now succeeds.

**Migration numbering.** `081_redirect_write_guards.ts`, with the static import and `getMigrations()` key updated and no leftover `080` anywhere on the branch. That matches the agreed split (#3182 = 080, this = 081, #3169 = 082 — both of the others have landed their side).

## Still open: the loop trigger aborts the *previous release's* rename sequence

Blocker 2 was about old binaries failing against the new schema. Removing the pattern guards fixed the path I named, but the loop trigger reaches the same outcome, because the reorder that fixes blocker 1 only exists in the new code.

`origin/main` still does collapse-then-delete:

```ts
// Collapse chains: update any existing redirects pointing to the old URL
await this.collapseChains(oldUrl, newUrl);
...
await this.db.deleteFrom("_emdash_redirects").where("source", "=", newUrl).execute();
```
<sub>`origin/main:packages/core/src/database/repositories/redirect.ts:385-390`</sub>

That is exactly the order `emdash_redirect_loop_update` rejects, and an old binary has no `write_generation` to opt out of the loop triggers the way it can now opt out of the fence triggers. Same two-redirect seed as before, replayed with main's statement order against the trigger text as `081` creates it:

```
--- OLD BINARY (main's order) against migration 081 triggers ---
Runtime error near line 22: redirect loop (19)
```

`createSlugChangeRedirect` is un-caught inside the content transaction on `main` too, so a still-running old isolate 500s and rolls back the content save for that rename shape — during the rolling window, and permanently after a rollback, since `down()` is still a no-op. Postgres behaves the same: the `RAISE EXCEPTION 'redirect loop'` fires on any `destination` update regardless of `write_generation`.

The fence triggers just gained a generation-aware opt-out; the loop triggers need the equivalent, or they need to not fire on a write that an old binary can produce.

## Changeset

`.changeset/bright-rivers-redirect.md` now says "The expanded redirect schema remains compatible with writes from previous host processes during rolling deployments." That's true for plain inserts and updates, but overstated while the above holds. Worth not shipping the unqualified claim until it does.

For completeness, I checked the fix commit for new regressions and found none beyond this: the `NEW.write_generation <> OLD.write_generation` fence relaxation doesn't open a hole (every current write still carries the application-level fence predicate and `withLease()` bumps `generation`), and the new `emdash_redirect_revision_update` AFTER trigger doesn't re-enter the fence/loop/guard triggers or disturb `recordHit()`.

~ 🤖 Codex (claude-opus-5)

PR #3185, opened 2026-09-17
feat(plugins): add publication policy hooks

ascorbic:
## What does this PR do?

Adds capability-gated publication policy hooks for native and sandboxed plugins without granting content read, write, or lifecycle authority. Events identify every action origin and authenticated human actor.

Cancellation is validated and revision-fenced. Scheduler rejections become bounded dashboard records with revision-safe cleanup. Visual-editor actions use renewable editor-bound tokens.

Includes registry consent, both runners, runtime-backed tests, docs, authoring guidance, and a changeset. Builds on #3162.

## Type of change

- [ ] Bug fix
- [x] Feature (maintainer-directed follow-on to #3162; no separate Discussion URL)
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] Targeted behavior tests pass
- [x] Tests are added/updated
- [x] Admin strings use Lingui; no catalogs included
- [x] Changeset added and reviewed
- [ ] Approved Discussion link (not applicable; maintainer-directed capability plan following #3162)
- [x] Rendered UI screenshot included

- [x] This PR includes AI-generated code — model/tool: gpt-5.6-sol

## Screenshots / test output

Admin dashboard showing scheduled entries blocked by publication policy

Validated with root build/typecheck, lint, docs, focused tests, query counts, and adversarial review.


packages/core/src/emdash-runtime.ts
```diff
@@ -3333,6 +3368,127 @@ export class EmDashRuntime {
...
+	private async checkContentPolicy(
+		name: ContentPolicyHookName,
+		collection: string,
+		id: string,
+		options: ContentPolicyMutationOptions,
+		rejectionCode: "PUBLISH_REJECTED" | "SCHEDULE_REJECTED" | "UNPUBLISH_REJECTED",
+		failureCode: "CONTENT_PUBLISH_ERROR" | "CONTENT_SCHEDULE_ERROR" | "CONTENT_UNPUBLISH_ERROR",
+		scheduledAt?: string,
+	): Promise<ContentPolicyCheck> {
+		const current = await handleContentGet(this.db, collection, id);
```
khoinguyenpham04: Could we run publish and schedule checks against the effective draft, since checking the published version can allow an unapproved draft to go live?

packages/core/src/api/handlers/content.ts
```diff
@@ -1623,6 +1644,56 @@ export async function handleContentUnschedule(
...
+	id: string,
+	options: { _rev: string; rejection: ScheduledPolicyRejection },
+): Promise<ApiResult<ContentResponse>> {
+	try {
+		const expectedRevision = decodeRevisionPrecondition(options._rev);
+		const item = await withTransaction(db, async (trx) => {
+			const repo = new ContentRepository(trx);
+			const resolvedId = (await resolveId(repo, collection, id)) ?? id;
+			const updated = await repo.unschedule(collection, resolvedId, expectedRevision);
+			await new OptionsRepository(trx).set(scheduledPolicyRejectionKey(collection, resolvedId), {
```
khoinguyenpham04: On D1, unscheduling can succeed even if saving the rejection fails, leaving no retry or dashboard notice. I think we should make these writes atomic.

packages/core/src/scheduled-publish.ts
```diff
@@ -131,7 +131,7 @@ export async function publishDueContent(
 				});
 				if (result.success) {
 					batch.push({ collection: collection.slug, id: item.id });
-				} else if (result.error?.code === "NOT_DUE") {
+				} else if (result.error?.code === "NOT_DUE" || result.error?.code === "PUBLISH_REJECTED") {
```
khoinguyenpham04, line 174: If the oldest 100 entries keep failing their hooks, every sweep selects the same rows, so later scheduled content is never processed.

docs/src/content/docs/plugins/creating-plugins/hooks.mdx
```diff
@@ -241,6 +242,36 @@ Runs after content is successfully deleted.
...
+
+### `content:beforePublish`
+
+The following hook requires an approval marker before content can become live:
+
+```typescript title="src/plugin.ts"
+"content:beforePublish": async (event) => {
+	if (event.content.approvalStatus !== "approved") {
+		return { cancel: true, reason: "Approve this entry before publishing." };
+	}
```
khoinguyenpham04, line 266: This reads the wrong field, so it rejects every publish. I think it should use `event.content.data.approval_status` here and in the matching skill example?

khoinguyenpham04 (commented):
I found a few bugs and nits, please have a look!

docs/src/content/docs/plugins/creating-plugins/hooks.mdx
ascorbic: Fixed in `453aed518`. Both the public docs and canonical plugin-authoring skill now narrow `event.content.data` and read `approval_status` from that custom-field object. They also document that publish/schedule receive the effective draft while unpublish receives live content.

~ 🤖 Codex

packages/core/src/api/handlers/content.ts
ascorbic: Fixed fail-closed in `453aed518`. D1 cannot transact generic Kysely statements, so the rejection is persisted before unscheduling: an option-write failure leaves the due row untouched, an unschedule failure leaves both the due row and actionable notice, and a stale `_rev` compare-deletes only the rejection revision written by that attempt. Transactional dialects still commit or roll back both writes together. Added no-transaction coverage for all three outcomes.

~ 🤖 Codex

packages/core/src/scheduled-publish.ts
ascorbic: Fixed in `453aed518`. Bounded scheduled sweeps now persist a deterministic `(scheduled_at, id)` keyset cursor, wrap after the forward window is exhausted, and advance it with versioned compare-and-set so an older overlapping sweep cannot move progress backward. Added sequential starvation and controlled out-of-order concurrency regressions.

~ 🤖 Codex

packages/core/src/emdash-runtime.ts
ascorbic: Fixed in `453aed518`. Publish and schedule policy checks now strictly hydrate the effective draft data and staged slug, but only after `_rev` validation and only when a policy hook exists. Draft-read failures fail closed; unpublish intentionally continues to inspect live content. Added regressions for divergent live/draft values, read failure, no-hook behavior, and stale revisions.

~ 🤖 Codex

ascorbic:
Correction to my earlier review of this PR — I cleared it, and I missed one thing. Found while re-reviewing the stacked #3194.

## Blocker: `hooks.content-policy:register` can't be published through the marketplace

The capability is registered in core (`packages/core/src/plugins/manifest-schema.ts:31`) and gated correctly at runtime via `HOOK_REQUIRED_CAPABILITY`, but it's absent from the marketplace worker's hardcoded `VALID_CAPABILITIES`:

```
$ git show origin/codex/plugin-publication-policy:packages/marketplace/src/routes/author.ts | grep content-policy:register
(no output)
```

That array backs both `createPluginSchema.capabilities` (registration) and `manifestSchema.capabilities` (`:829`), the latter being what validates an uploaded manifest on the version-publish path, plus the `/dev` upload routes that reuse it. So a plugin declaring `hooks.content-policy:register` is rejected with `400 Invalid manifest: capabilities.N: Invalid enum value` — the policy-hook feature is undistributable through that channel even though the runtime side is sound.

It fails closed, so there's no privilege concern; the feature just can't ship to marketplace users.

This turned out to be a pattern rather than a one-off — the same gap exists on #3170 (`comments:*`), #3172 (`media:*`), and #3194 (`content:publish`/`content:restore`). #3170 has since reworked `author.ts` to source the list from `PLUGIN_CAPABILITIES` in `@emdash-cms/plugin-types` rather than keep a copy, which fixes the whole class. If #3170 merges first, this PR needs no edit; otherwise add `"hooks.content-policy:register"` to `VALID_CAPABILITIES` here.

Everything else in my original review stands at the current head `453aed518`.

~ 🤖 Codex (claude-opus-5)

packages/core/src/astro/routes/api/visual-editing/content/[collection]/[id]/publish.ts
```diff
@@ -0,0 +1,31 @@
...
+	VISUAL_ACTION_TOKEN_INVALID,
+	verifyVisualEditingActionToken,
+} from "../../../../../../../visual-editing/action-token.js";
+import { publishContent } from "../../../../content/[collection]/[id]/publish.js";
+
+export const prerender = false;
+
+export const POST: APIRoute = async (context) => {
+	const { emdash, user } = context.locals;
+	if (!emdash?.db || !user) return apiError("NOT_CONFIGURED", "EmDash is not initialized", 500);
```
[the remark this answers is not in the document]
Comment (ascorbic) [reply packages/core/src/astro/routes/api/visual-editing/content/[collection]/[id]/publish.ts]: Fixed in fabefe329. The route now distinguishes an uninitialized runtime from a missing user and returns UNAUTHORIZED with HTTP 401. The regression verifies the response code/body and that publication is never invoked.\n\n~ 🤖 Codex (gpt-5.6-sol)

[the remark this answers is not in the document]
Comment (ascorbic) [reply packages/core/src/astro/routes/api/visual-editing/content/[collection]/[id]/publish.ts]: Fixed in fabefe329 with a logged-out visual-publish regression: HTTP 401, UNAUTHORIZED, and no publication call.\n\n~ 🤖 Codex (gpt-5.6-sol)

packages/core/src/visual-editing/toolbar.ts
```diff
@@ -549,6 +552,59 @@ export function renderToolbar(config: ToolbarConfig): string {
...
 
   var isEditMode = toolbar.getAttribute("data-edit-mode") === "true";
+  var visualActionToken = ${JSON.stringify(actionToken)};
+  var visualActionTokenErrorCode = ${JSON.stringify(VISUAL_ACTION_TOKEN_INVALID)};
+  var visualActionRefreshTimer = null;
+
+  function showVisualActionRecovery() {
+    if (visualActionRefreshTimer !== null) clearTimeout(visualActionRefreshTimer);
+    visualActionRefreshTimer = null;
+    statusEl.innerHTML = '<span class="emdash-tb-badge emdash-tb-badge--error">Editing session expired. Refresh the page to continue.</span>';
```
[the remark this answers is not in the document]
ascorbic: Fixed in fabefe329. The publication/recovery labels now come from the active Lingui catalog, are passed into the generated toolbar, and are safely escaped for HTML and inline script contexts. Focused localization and toolbar behavior tests cover translated states.\n\n~ 🤖 Codex (gpt-5.6-sol)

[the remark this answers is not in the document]
ascorbic: Fixed in fabefe329. The recovery, refresh, failure, publish, and publishing labels now resolve through Lingui and are supplied by request locale rather than embedded English.\n\n~ 🤖 Codex (gpt-5.6-sol)

packages/core/src/visual-editing/toolbar.ts
```diff
@@ -549,6 +552,59 @@ export function renderToolbar(config: ToolbarConfig): string {
...
+    publishBtn.disabled = true;
+    publishBtn.textContent = "Refresh page";
+  }
+
+  function showPublishError(message) {
+    statusEl.textContent = message || "Publish failed. Check your permissions and try again.";
```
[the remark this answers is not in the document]
ascorbic: Fixed in fabefe329. The fallback publish error is now a Lingui-resolved label supplied to the toolbar script, with translated-label behavior covered by tests.\n\n~ 🤖 Codex (gpt-5.6-sol)

ascorbic:
Addressed the marketplace validation blocker in fabefe329: registration and uploaded manifests now use the canonical @emdash-cms/plugin-types capability list, so hooks.content-policy:register is accepted while unknown capabilities still fail. Added focused schema coverage.\n\n~ 🤖 Codex (gpt-5.6-sol)

packages/core/src/astro/routes/api/visual-editing/toolbar-labels.ts
```diff
@@ -0,0 +1,17 @@
+export const prerender = false;
+
+import { loadVisualEditingToolbarLabels } from "@emdash-cms/admin/locales/server";
+import type { APIRoute } from "astro";
+
+export const GET: APIRoute = async ({ request }) => {
+	const labels = await loadVisualEditingToolbarLabels(request);
+	return Response.json(
```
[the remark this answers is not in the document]
ascorbic: Fixed in cc676a844. The route now uses apiSuccess, preserves private no-store caching, and the regression expects the standard success envelope. ~ 🤖 Codex (gpt-5.6-sol)

PR #3199, opened 2026-09-18
perf(cloudflare): reduce Worker startup module closure

ascorbic:
## What does this PR do?

Reduces the module graph evaluated when an EmDash Cloudflare Worker starts. The Worker entry exports `PluginBridge` directly, loads scheduled maintenance only when a Cron Trigger runs, and loads the bridge's repository runtime on the first sandbox RPC through a narrow `emdash/plugins/host` entry. Existing `emdash` and `@emdash-cms/cloudflare/worker` exports remain compatible.

Adds a production-build guard that walks static imports from the generated Worker entry. CI now rejects more than 30 eager modules, more than 2 MB of eager JavaScript, or eager WXR, Tiptap/ProseMirror, plugin-registry, and media-maintenance sources.

The query-count harness now disables Astro 7.3's separate dev and preview background modes so local agent runs capture the same query events as CI.

Discussion: https://github.com/emdash-cms/emdash/discussions/3143

Measured on the production-built Cloudflare starter with Wrangler 4.125.0, using one discarded warm-up and nine startup samples:

| Metric | Before | After | Change |
| --- | ---: | ---: | ---: |
| Eager modules | 144 | 18 | -87.5% |
| Eager bytes | 5,721,716 | 1,734,286 | -69.7% |
| Startup active CPU median | 44.0 ms | 19.0 ms | -56.8% |
| Startup CPU range | 41.0–45.3 ms | 18.7–20.1 ms | |

The narrow bridge runtime adds 928 KB on the first sandbox RPC, keeping startup plus first plugin use at approximately 2.66 MB instead of loading the 5.72 MB root barrel. Testing Cloudflare's `new_module_registry` flag increased the optimized median to 35.4 ms, so this PR does not enable it.

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [x] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [ ] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/3143

The localization, new-feature, and screenshot items are not applicable: this changes Worker startup timing without adding admin UI or a new product capability.

## Screenshots / test output

Not applicable — no UI changes.

Validated on the exact merged `origin/main` head with:

- `pnpm install --frozen-lockfile`
- `pnpm build`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm format:check`
- `pnpm --dir packages/cloudflare test` — 436 passed
- focused core content-access tests — 102 passed
- production workerd bridge tests — 31 passed
- `pnpm query-counts --target d1` — count and SQL snapshots match
- `pnpm --dir templates/starter-cloudflare build`
- `pnpm startup:check` — 18 modules, 1,734,286 bytes
- `wrangler check startup` — 19.0 ms median across nine measured runs

The package API linter (`publint`) passes for both changed packages. The secondary `attw --pack` process crashes after packing with `Cannot read properties of undefined (reading 'filename')` for both packages and does not report a package diagnostic.


eisenbruch:
Tested on our real Worker rather than the starter, since this is the site that motivated the thread. Short version: **it works, and it removes the workaround we had been carrying.** Details, with the caveat that our own entry was already lazy, so our headline numbers move less than yours and that is expected.

## What we are

`equineinfoexchange.com`, launched yesterday: Astro 7 + EmDash 0.38.0 on Workers/D1/R2/KV, ~9,700 directory listings, ~3,400 HTML pages. We had already worked around the eager closure on our side, by making our build inject the `scheduled` handler with a dynamic `import()` rather than a static one. So we start from 21 eager modules, not 144.

## The measurement that matters

Because our entry was already lazy, the honest test is **what a static import costs**, which is what an ordinary site does and what we had to stop doing:

| our Worker, `scheduled` imported **statically** | eager modules | eager bytes |
| --- | ---: | ---: |
| 0.38.0 | 137 | 5,712,000 |
| #3199 preview | **22** | **1,500,000** |

Same entry, same build, only the packages differ. **That is the whole of our workaround made unnecessary** — one extra module and 0.02 MB over our hand-lazied build. We can delete the injection hack and go back to a plain static import.

With our existing lazy entry, before and after are unchanged, as expected: 21 modules / 1.48 MB → 21 / 1.50 MB, startup active CPU median **29.5 ms → 31.0 ms** over nine runs each (min 26.6/27.6, max 34.3/52.4). No regression, no gain, because we were already at the floor. Wrangler 4.x, `wrangler check startup`, one warm-up discarded.

The `PluginBridge` change is the part I had concluded was unfixable from outside — a Durable Object class has to be a module-scope export, so no amount of deferring on our side could help. Good to be wrong about that.

## Two things you may want to know

**1. `emdash/dist/astro/middleware.mjs` goes from 161 KB to 43 KB**, because it stops inlining the runtime. That is a nice second-order confirmation the split does what it says. It also **moves `mergedData` / `validateContentData` / `getCollectionWithFields` out of that module into `emdash-runtime-*.mjs`**. Only relevant to people patching built dist files, which is us: we carry local patches for #3129 and #3142 that target `astro/middleware.mjs`, and they will need retargeting when this lands. Not a blocker and not a request — the patches are ours and temporary — but worth knowing that the entry's contents change shape, not just its import graph.

**2. `@emdash-cms/plugin-forms` is unaffected.** It does not reference `plugins/host` or the sandbox, and it stays where it was in the eager closure (1 of 21 modules).

## Caveats

Our four remaining local patches could not be applied to the preview build, because they target content-hashed dist filenames (`api-BvxJmnFY.mjs` and friends) that are rebuilt under new hashes. I checked each one and the upstream code they anchor to is still present under the new names, so this is a packaging artifact of preview builds rather than a conflict with this PR. The numbers above are therefore from an **unpatched** preview install.

Tested on our staging environment, not production.

The CI guard is the part I would have asked for if you had not included it. The regression we hit was invisible precisely because nothing failed — the build passed, the deploy passed, every page served, and the only symptom was cold pages taking 1–3 seconds. A threshold that fails the build is what stops that recurring.

Happy to re-run any of this, or to test again once it is on a release we can install normally.

ascorbic:
Thanks for checking. I'll merge this and it'll be in the next release

PR #3202, opened 2026-09-18
fix(core): search finds nothing for a query with an operator word and an apostrophe

### LINKED ISSUE #3193: Search silently returns nothing when the query contains both an English "and"/"or"/"not"/"near" and an apostrophe
## Summary

`escapeQuery()` treats the ordinary English words **and**, **or**, **not** and **near** as FTS5 operators, case-insensitively, and responds by handing the user's raw string to FTS5 **unquoted**. If that string also contains an apostrophe — which any possessive does — FTS5 raises `fts5: syntax error near "'"`, `isFts5SyntaxError()` catches it, and the search returns `[]`.

The result is a query that looks completely reasonable and finds nothing, with no error and no signal to the user. Typing a business's own full name into a site search is the case that hits it.

We found this on a directory of 9,676 listings. Searching for `Tennessee Walking Horse Breeder's and Exhibitor's Association` — the exact stored title of a published entry — returns 0 results. Dropping either the `and` or the apostrophes finds it immediately.

**Affected version:** 0.38.0. The code is unchanged on `main`.

## Reproduction

Minimal, no EmDash needed. `escapeQuery` is transcribed verbatim from `packages/core/src/search/query.ts`:

```js
import { DatabaseSync } from "node:sqlite";               // node --experimental-sqlite
const db = new DatabaseSync(":memory:");
db.exec(`CREATE VIRTUAL TABLE f USING fts5(title, content, tokenize='porter unicode61')`);
db.prepare(`INSERT INTO f VALUES(?,?)`).run("Tennessee Walking Horse Breeder's and Exhibitor's Association", "");

const FTS_OPERATORS = /\b(AND|OR|NOT|NEAR)\b/i;
function escapeQuery(q) {
  q = (q || "").trim();
  if (!q) return "";
  if (q.startsWith('"') && q.endsWith('"') && q.length >= 2) return `"${q.slice(1, -1).replace(/"/g, '""')}"`;
  const escaped = q.replace(/"/g, '""');
  if (FTS_OPERATORS.test(q)) return escaped;              // <-- raw passthrough
  const terms = escaped.split(/\s+/).filter((t) => t.length > 0);
  if (!terms.length) return "";
  return terms.map((t) => `"${t}"*`).join(" ");
}

for (const q of [
  "Tennessee Walking Horse Breeder's and Exhibitor's Association",
  "Tennessee Walking Horse Breeder's Exhibitor's Association",
  "Breeder's and Exhibitor",
  "O'Brien not here",
  "Breeders and Exhibitors",
]) {
  const m = escapeQuery(q);
  let out;
  try { out = db.prepare(`SELECT count(*) c FROM f WHERE f MATCH ?`).get(m).c; }
  catch (e) { out = "THROWS: " + e.message; }
  console.log(JSON.stringify(q), "->", JSON.stringify(m), out);
}
```

| query | what `escapeQuery` produces | FTS5 |
| --- | --- | --- |
| `Tennessee Walking Horse Breeder's and Exhibitor's Association` | the string, raw | **`fts5: syntax error near "'"`** |
| `Tennessee Walking Horse Breeder's Exhibitor's Association` | `"Tennessee"* "Walking"* … "Breeder's"* …` | 1 |
| `Breeder's and Exhibitor` | the string, raw | **`fts5: syntax error near "'"`** |
| `O'Brien not here` | the string, raw | **`fts5: syntax error near "'"`** |
| `Breeders and Exhibitors` | the string, raw | 1 |

Only the two ingredients together fail. An apostrophe alone takes the quoted-term path and is fine; an operator word alone parses as barewords and is fine.

Reproduced end to end against a live D1 deployment through `search()`, and through the site's own search page.

## Root cause

`packages/core/src/search/query.ts`:

```ts
const FTS_OPERATORS_PATTERN = /\b(AND|OR|NOT|NEAR)\b/i;
…
const escaped = query.replace(DOUBLE_QUOTE_PATTERN, '""');
if (FTS_OPERATORS_PATTERN.test(query)) return escaped;   // user is "writing FTS5 syntax"
const terms = escaped.split(WHITESPACE_SPLIT_PATTERN).filter((t) => t.length > 0);
return terms.map((t) => `"${t}"*`).join(" ");
```

Two decisions combine:

1. **The operator test is case-insensitive.** FTS5's own operators are uppercase-only — `and` is a bareword to FTS5, `AND` is the operator. Matching `/i` means every sentence containing the English word "and" is classified as a syntax query. That is a large fraction of natural search input: "boarding and training", "tack and feed", "Smith and Sons".
2. **The escape hatch escapes only double quotes.** Once on the raw path, nothing protects `'`, `(`, `)`, `*`, `:`, `^` or `-`, all of which are meaningful or illegal to FTS5 in a bareword.

The failure is silent because `isFts5SyntaxError()` correctly converts the resulting throw into `[]`. That is the right behaviour for genuinely malformed syntax; it is the wrong outcome for a user who never intended to write syntax.

## Impact

It is not limited to apostrophes. Anything that reaches the raw path and is not valid FTS5 fails the same way, for example `Smith & Sons (Tack and Feed)`, or a hyphenated name plus an "and". And when the raw string *is* valid FTS5, the behaviour is different again but still surprising: prefix matching is silently dropped, so `boarding and train` no longer matches "training", while `boarding train` does.

There is no way for a site to opt out. `escapeQuery` is not configurable and the site search page passes user input straight through.

## Possible directions

Offered as starting points.

- **Make the operator test case-sensitive**, `/\b(AND|OR|NOT|NEAR)\b/`. This alone fixes the reported case and matches FTS5's own rule: a user typing uppercase AND plausibly means the operator, a user typing "and" does not. It is a one-character change and the least surprising semantics.
- **Escape the raw path properly** rather than only its double quotes, so a query that opts into syntax still cannot produce a hard error from an apostrophe in a name.
- **Fall back instead of returning empty.** When `isFts5SyntaxError` fires, retry once with the fully quoted term construction before giving up. A search that quietly returns nothing is worse than one that ignores the user's operators.
- Either way, a regression test over a title containing both an apostrophe and "and" would pin it.

Happy to send a PR for the case-sensitivity change plus a test if that is the direction you would take.

## Environment

- emdash 0.38.0, Cloudflare D1, `listings` collection with `search` enabled
- Also reproduced on stock SQLite FTS5 via `node:sqlite`

eisenbruch:
## What does this PR do?

Closes #3193.

FTS5's operators are uppercase only: `AND` is the operator, `and` is an ordinary bareword. `escapeQuery()` in `packages/core/src/search/query.ts` tested for them case-**in**sensitively:

```ts
const FTS_OPERATORS_PATTERN = /\b(AND|OR|NOT|NEAR)\b/i;
```

So any query containing the English word *and*, *or*, *not* or *near* was classified as a syntax query and sent down the raw path, where only double quotes are escaped. An apostrophe then reached SQLite unescaped, raised `fts5: syntax error near "'"`, and `isFts5SyntaxError()` turned that into an empty result. A perfectly ordinary search silently found nothing, with no error and nothing to tell the reader why.

The fix is the `i` flag:

```diff
-const FTS_OPERATORS_PATTERN = /\b(AND|OR|NOT|NEAR)\b/i;
+const FTS_OPERATORS_PATTERN = /\b(AND|OR|NOT|NEAR)\b/;
```

Uppercase operators keep working, because those are the ones FTS5 honours. Lowercase prose now takes the safe path and is quoted.

We hit this on a live directory of 9,676 listings: searching for a listing's own exact stored title, `Tennessee Walking Horse Breeder's and Exhibitor's Association`, returned zero results. Typing a business's full name into a site search is not an exotic query.

### The two alternatives, and why they are worse

- **Escape the apostrophe on the raw path.** It fixes the crash but keeps the cause: ordinary prose is still read as operator syntax, so `and`, `or` and `not` silently change the meaning of a search.
- **Strip punctuation before searching.** It breaks tokenization. `O'Brien` becomes the single token `obrien`, which matches no document holding `O'Brien`.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes (`emdash`; the `apps/aggregator` failure on `PackageView` is pre-existing and reproduces on a clean `main`)
- [x] `pnpm lint` passes (`oxlint --type-aware --deny-warnings`)
- [x] `pnpm test` passes (or targeted tests for my change): the search suites, 13 files and 54 tests

## Screenshots / test output

Not applicable, no UI change.

`packages/core/tests/integration/search/operator-words.test.ts` covers five cases. Three fail against `main` and pass with the change:

```
× finds a title containing both an apostrophe and the word and
× finds a title containing an apostrophe and any of the operator words
× keeps prefix matching for a query containing a lowercase operator word
```

I verified that by reverting the one character on this branch and re-running, rather than by reasoning about it.

One detail worth keeping if the tests are edited: the prefix-matching case deliberately uses a partial word rather than a stem. `boarding and train` would pass either way, because Porter reduces `training` to `train` on both sides; it uses `boarding and wil` against `Willow`, which only matches with the prefix operator.


eisenbruch:
All three taken in 698953a4, as written.

- `query.ts`: the block comment is now the two-line invariant. The bug narrative belongs in this PR and the issue, not in the source.
- `operator-words.test.ts`: same, a present-tense note above the suite.
- The changeset now leads with the verb and names the user-facing behavior.

No logic change; `oxfmt --check` is clean across the tree.

PR #1838, opened 2026-07-06
fix(admin): surface field-level validation messages instead of a generic error

marcusbellamyshaw-cell:
## What does this PR do?

`throwResponseError()` only read `error.message`, which the API always sets to the generic "Invalid request data" for `VALIDATION_ERROR` responses — the actual per-field problems live in `error.details.issues` and were silently discarded. This surfaces the real field-level messages to the user instead of the generic error, fixed once in the shared helper so every caller routed through `parseApiResponse` benefits (createTaxonomy and 20+ others), not just the reported taxonomy-creation case.

Closes #255

## Type of change

- [x] Bug fix
- [ ] Feature (requires [maintainer-approved Discussion](https://github.com/emdash-cms/emdash/discussions/categories/ideas))
- [ ] Refactor (no behavior change)
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests
- [ ] Chore (dependencies, CI, tooling)

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/emdash-cms/emdash/blob/main/CONTRIBUTING.md)
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test` passes (or targeted tests for my change)
- [x] `pnpm format` has been run
- [x] I have added/updated tests for my changes (if applicable)
- [ ] User-visible strings in the admin UI are wrapped for translation (n/a — no new user-visible strings, this surfaces existing API-provided messages)
- [x] I have added a changeset (if this PR changes a published package)
- [ ] New features link to an approved Discussion (n/a — bug fix)

## AI-generated code disclosure

- [x] This PR includes AI-generated code — model/tool: Claude Sonnet 5

## Screenshots / test output

`packages/admin/tests/lib/api-client.test.ts` — 1050/1050 admin tests pass; `pnpm lint:quick` clean; `pnpm typecheck` clean.

PR #3205, opened 2026-09-18
fix(admin): name the field by its label when a save fails validation

### LINKED ISSUE #2426: Admin's per-field validation messages never fire for content saves — core's content validator cannot supply error.details.issues
## Summary

`formatValidationIssues` in the admin client renders per-field validation messages only when the response carries `error.details.issues`. Core supplies that array for request-shape failures, but **cannot** supply it for content-data validation — the validator's result type has no `details` member, and it collapses the per-field strings it has already built into a single joined `message`.

The result is that the admin's per-field error rendering never fires for the validation failure authors actually hit: saving a content entry with an invalid field. It falls back to one concatenated string, and the field-level detail the server already computed is discarded on the way out.

Versions: `@emdash-cms/admin@0.32.0`, `emdash@0.32.0`.

### The consumer requires `details.issues`

`@emdash-cms/admin`, `src/lib/api/client.ts` (shipped in `dist/plugins-CQAV8Ap0.js`, sourcemap `sources[0]`):

```ts
/**
 * Extract per-field validation issue messages from a `VALIDATION_ERROR`
 * response's `error.details.issues` array (see `packages/core/src/api/parse.ts`).
 * Returns undefined when the shape doesn't match, so callers can fall back
 * to the generic top-level message.
 */
function formatValidationIssues(error: Record<string, unknown>): string | undefined {
	if (error.code !== "VALIDATION_ERROR") return undefined;
	if (!isRecord(error.details)) return undefined;
	const issues = error.details.issues;
	if (!Array.isArray(issues) || issues.length === 0) return undefined;
	…
}
```

and `throwResponseError` in the same file falls back when it returns `undefined`:

```ts
message = formatValidationIssues(error);
if (!message && typeof error.message === "string") message = error.message;
```

The docblock names the producer it was written against: `packages/core/src/api/parse.ts`.

### Core supplies it — on one path only

`emdash`, `src/api/parse.ts:111`:

```ts
const issues = result.error.issues.map((issue: z.ZodIssue) => ({
	path: issue.path.join("."),
	message: issue.message,
}));

return apiError("VALIDATION_ERROR", "Invalid request data", 400, { issues });
```

So request-shape failures satisfy the consumer, and the admin renders per-field messages. That path works as designed.

### Core cannot supply it for content data — by type

`emdash`, `src/api/handlers/validation.ts:26`:

```ts
type ValidationResult =
	| { ok: true }
	| { ok: false; error: { code: "VALIDATION_ERROR" | "COLLECTION_NOT_FOUND"; message: string } };
```

No `details` member exists. The validator builds per-field strings throughout — `:99`, `:111`, `:124`, `:155`, `:193`, `:199` — each already carrying its field:

```ts
issues.push(`${key}: unknown field on collection '${collection}'`);
issues.push(`${formatIssuePath(issue.path)}: ${issue.message}`);
issues.push(`${field.slug}: required (empty value not allowed)`);
issues.push(`${ref.field}: target '${ref.id}' not found in collection '${target}'`);
```

then collapses them at `:204`:

```ts
if (issues.length === 0) return { ok: true };
return {
	ok: false,
	error: {
		code: "VALIDATION_ERROR",
		message: issues.join("; "),
	},
};
```

The structured array exists one line before it is discarded.

`src/emdash-runtime.ts:2725` passes that error straight through, so nothing downstream re-expands it:

```ts
const validation = await validateContentData(this.db, collection, processedData, { partial: false });
if (!validation.ok) {
	return { success: false as const, error: validation.error };
}
```

### Net

Same `VALIDATION_ERROR` code, two shapes. `formatValidationIssues` returns `undefined` for every content-save failure and `throwResponseError` falls back to `error.message` — the joined string.

## Reproduction

1. Save a content entry with two invalid fields (for example a required field left empty and a reference pointing at a missing target).
2. Observe the admin error.

Expected, given `formatValidationIssues` exists: two field-scoped messages.
Actual: one string, `"<field-a>: required (empty value not allowed); <field-b>: target 'x' not found in collection 'y'"`, surfaced by the generic fallback.

A consumer wanting the fields back has to split on `"; "` and then on the first `": "` — guesswork the moment a field slug or a Zod message contains either separator.

## Net effect

The per-field improvement is present in the admin and unreachable on the path that matters. For a non-technical author the practical result is a save that fails with one undifferentiated blob of text; in our case an editor read a repeated save failure as a permissions problem and stopped, because nothing in the response distinguished "this field is invalid" from "you may not do this". The server knew which field was wrong and had already formatted the message.

## Suggested fix

Add the structured payload alongside the existing message in `validation.ts`, matching the shape `parse.ts` already emits:

```ts
return {
	ok: false,
	error: {
		code: "VALIDATION_ERROR",
		message: issues.join("; "),   // unchanged, back-compatible
		details: { issues },          // new — satisfies formatValidationIssues
	},
};
```

Carrying `issues` as `{ path, message }` objects rather than pre-joined strings would make the two paths identical for consumers, and the field name is already isolated at each `issues.push` site. `ValidationResult` at `:26` needs the corresponding `details?` member.

No admin change is required — `formatValidationIssues` starts working the moment the payload arrives.

danielmlr:
## What does this PR do?

Names each rejected field by its editor label and says what it needs, so `excerpt: Too big: expected string to have <=160 characters` becomes "Summary can have at most 160 characters."

The content validator joined per-field failures into one message, so #1838's per-field formatting never ran for content saves. It now returns them in `error.details.issues` with path, code and any failed bound; `error.message` is unchanged. The admin's save toasts turn each issue into a translatable sentence; other errors keep the server message. Like `SAVE_REJECTED`'s details, the issues serve the admin, so the changeset and OpenAPI leave them out; the core test pins each code the admin words, so a Zod upgrade that renames one fails there.

Closes #2426

Part of #2417 (item 2, [requested here](https://github.com/emdash-cms/emdash/issues/2417#issuecomment-5509662441)).

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

## Screenshots / test output

A Summary field (`minLength: 20`) and a Reading time field (`min: 1`, `max: 60`) rejected by one autosave.

Post editor in English: Summary holds "Too short" and Reading time holds 99, both outlined red. The Autosave failed toast reads "Summary needs at least 20 characters. Reading time must be at most 60."

The same editor in Arabic; the new sentences stay English until the catalog is translated.

The same editor in Arabic with Reading time 0: the toast title is Arabic, its two sentences English.

- Failing on `main`: three of the four new core cases (the fourth pins the unchanged `error.message`) and both new cases in `router.test.tsx`; `content-validation-errors.test.ts` covers the wording.
- `pnpm exec playwright test autosave` → 3 passed, one of them new; `pnpm typecheck` clean; `pnpm lint:json` → 0 diagnostics
- `pnpm --filter @emdash-cms/admin test` → 2367 passed; `pnpm --filter emdash test` → 6897 passed, 1 failed: `cron-schedule.test.ts` fails in a Europe/Berlin shell and passes with `TZ=UTC`, outside this diff


ascorbic (approved):
Thanks

PR #3235, opened 2026-09-21
Add content search to the editor link input

swissky:
## What does this PR do?

Adds an internal link picker to the rich text editor, as proposed in Discussion #3177 (Layer 1). The link popovers (toolbar and bubble menu) replace the plain URL input with a combined destination field, modeled on WordPress's link control:

- Typing plain text searches existing entries by title across all searchable, routable collections and lists them below the field. Typing anything URL-like (scheme, path, anchor, domain) behaves exactly as before.
- Picking a result inserts the entry's public URL resolved from the collection's `urlPattern`, including locale prefixes and date tokens (the entry is fetched on demand only for date-token patterns). Entries whose URL needs a publish date can't be picked until they are published — the picker explains why instead of baking a broken `{year}` href into the content.
- Editors with `content:read_drafts` also find drafts, marked with the standard status badge.

Two additive core changes support this:

- The search API and the `search()`/`searchCollection()` helpers gain a `scope=title` option that restricts the FTS5 match to the collection's title column, so body-text matches don't surprise authors picking a link target. Queries containing FTS5 operators are always term-quoted under this scope so no clause can escape the column filter (covered by a regression test).
- The auth middleware resolves the session on `GET /_emdash/api/search`, so `status=draft` honors the `content:read_drafts` permission instead of silently falling back to published results. Anonymous requests skip the session lookup entirely and stay on the zero-query hot path (covered by a unit test asserting no user lookup happens without a session cookie).

## Type of change

- [ ] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] New features link to an approved Discussion: https://github.com/emdash-cms/emdash/discussions/3177

## Screenshots / test output

Searching "team" in the link popover lists matching entries by title, with the first result pre-highlighted for Enter:

Link popover in the editor showing a search input with "team" and one result, the Team page, highlighted below it

Drafts are listed for editors with draft access and marked with the standard status badge:

Link popover showing the search "upcoming" with one result, the draft post Upcoming Announcement, carrying a Draft badge

Picking a result inserts the entry's resolved public URL as the link:

Editor content with the selected text now underlined as a link to /pages/team, and the bubble menu showing the active link state


swissky:
Both findings addressed in 96d8c6b29:

1. **Date-token drafts with a publish date**: `resolveHref` now requires `entry.status === "published"` in addition to a publish date before resolving a dated URL, so scheduled entries and unpublished-after-release drafts are refused like undated drafts. Added the suggested regression case (draft carrying a publish date).

2. **Middleware comment**: reworded as suggested — anonymous callers skip the user DB lookup, not the session read.

On the `DOMAIN_LIKE` suggestion: I'd like to keep the heuristic as is. It only applies to single-token dotted input (anything with a space is always treated as a search), and any shape that still recognizes `example.com` also matches `vue.js` — the right side of the dot is indistinguishable from a TLD. Searching for a dotted title still works by typing any other part of it, while failing to recognize a pasted bare domain would insert a broken relative href with no recovery hint. Happy to revisit if authors report friction.

swissky:
Took the `DOMAIN_LIKE` suggestion after all, with a dual-affordance middle ground in 55a4aefbe rather than narrowing the regex (any shape that still accepts `example.com` also matches `vue.js`): bare domain-like input now searches too, but no result is pre-highlighted there — Enter keeps applying the typed text as the link target, while arrow keys or a click select a result explicitly. Scheme- and path-like input still disables search entirely. Covered by two new tests (Enter stays on the raw text; arrowing into the list picks the result).

ascorbic (approved):
Thanks!

Issue #3147, opened 2026-09-16
404 logging and active URL redirects fail due to ambiguous column reference "hits" on PostgreSQL


palockocz:
### Description

In the EmDash admin dashboard (Redirects section), both existing URL redirects and the 404 error log tab are non-functional in a Node.js + PostgreSQL production environment.

When a user accesses a URL that should trigger a redirect (e.g. /stitek/domácí-zvířata -> /stitek/domaci-zvirata) or any non-existent page, the system attempts to log the hit in the _emdash_404_log table. However, the SQL UPSERT query fails due to an ambiguous column reference error (hits). Because this database operation throws an unhandled error during page lookup, the redirect middleware fails before performing the actual HTTP redirect (and the Hits counter in the UI remains at 0).

**Root Cause:**
In the SQL query ON CONFLICT ("path") DO UPDATE SET "hits" = hits + 1, ..., PostgreSQL cannot determine whether hits refers to the existing row in _emdash_404_log or the EXCLUDED table.

### Steps to reproduce

1. Run EmDash CMS connected to a PostgreSQL database.
2. Create a new redirect rule in the admin panel (e.g., Source: /stitek/domácí-zvířata -> Destination: /stitek/domaci-zvirata).
3. Access the source URL /stitek/domácí-zvířata in your browser.
4. Notice that the redirect does not execute and a 500/404 error occurs instead.
5. Check the PostgreSQL database logs for the ambiguous hits column error.
6. Return to Admin UI -> Redirects: Observe that the redirect's Hits counter stays at 0 and the 404 Errors tab remains empty.

### Environment

- EmDash CMS v0.38.0
- Runtime Node.js v22
- Database PostgreSQL v18
- ENV production

### Screenshots

<img width="1683" height="958" alt="Image" src="https://github.com/user-attachments/assets/0f615cd7-5631-4c51-ae71-0db5ab461f24" />

<img width="1687" height="971" alt="Image" src="https://github.com/user-attachments/assets/202e8059-f5aa-4254-b045-65e216790a98" />

### Logs / error output

```shell
2026-09-16 10:06:03.312 UTC [66659] ERROR:  column reference "hits" is ambiguous at character 200

[...]

PR #3238, opened 2026-09-21
fix(core): qualify hits column in 404-log upsert for PostgreSQL

swissky:
## What does this PR do?

On PostgreSQL, every 404 hit failed to be recorded: the `log404` upsert's `ON CONFLICT ("path") DO UPDATE SET "hits" = hits + 1` uses an unqualified `hits`, which Postgres rejects at parse time as ambiguous between the existing row and the `EXCLUDED` pseudo-row (`column reference "hits" is ambiguous`). SQLite resolves the same reference leniently, so the SQLite-only test suite never caught it. As a result, the 404 Errors tab stayed empty and 404 hit counts never incremented on Postgres installs.

The fix qualifies the reference as `_emdash_404_log.hits + 1` (via `sql.ref`), matching the idiom already used for the same situation in `entry-locks.ts` and `media/usage/activation.ts`. The two tests that exercise the upsert's conflict branch (dedup by path, concurrent inserts) now run on both dialects via `describeEachDialect`, so `EMDASH_TEST_PG` CI catches this class of regression. I checked all other `doUpdateSet` call sites in the repo: this was the only remaining unqualified raw-SQL self-reference.

Verified red→green against a real PostgreSQL 18 container: before the fix, both dialect tests fail with the exact error from the issue report; after the fix, all 83 tests in `tests/integration/redirects/` pass on SQLite and Postgres.

Note on scope: the issue reports a second symptom — the redirect itself (`/stitek/domácí-zvířata` → …) not executing. That cannot be caused by this bug: `log404` runs after `next()` and is fire-and-forget with `.catch(() => {})` (identical in v0.38.0), and matched redirects never reach `log404` at all. The redirect failure is a separate defect (the exact-match lookup compares the percent-encoded `context.url.pathname` against the stored raw-Unicode source), tracked separately so this issue isn't considered fully resolved by this PR alone.

Fixes the 404-logging half of #3147.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

## Screenshots / test output

Not applicable (no UI change).

Postgres 18, before the fix (both new dialect tests fail with the reported error):

```
FAIL  tests/integration/redirects/log404-bounded.test.ts > RedirectRepository.log404 — path upsert [postgres] > dedups repeat hits by path instead of inserting new rows
error: column reference "hits" is ambiguous
 ❯ RedirectRepository.log404 src/database/repositories/redirect.ts:649:18
Tests  2 failed | 7 passed (9)
```

After the fix:

```
✓ tests/integration/redirects/ (5 files)
Test Files  5 passed (5)
     Tests  83 passed (83)
```


swissky:
Done in 2602e6a — moved the two capacity/eviction tests into the `describeEachDialect` block as well, so the conflict branch on seeded paths and `enforce404Cap`'s delete-with-subquery now run on Postgres too. Verified green on both dialects against Postgres 18 locally (11 tests).

swissky:
Fixed the comment wording ("so Promise.all doesn't produce real parallelism there") — the intended word was "there" (on SQLite), so I restructured the sentence instead of switching to "their".

ascorbic (approved):
Thanks

PR #3265, opened 2026-09-22
fix(core): warn on a where object with no range bounds, and chunk long IN lists for D1

### LINKED ISSUE #3256: Collection `where`: a non-range object drops the filter with no warning, and an array over 100 values returns no rows on D1
**Version:** 0.38.0, `@emdash-cms/cloudflare`, D1. `loader.ts` is unchanged on `main` for both code paths below.

Two silent failures in `buildFieldConditions` (`packages/core/src/loader.ts`). Both return a successful result that is wrong, with nothing logged.

### 1. Any non-array object is treated as a range, and a range with no bounds adds no condition

```ts
function isWhereRange(value) {
  return value !== null && typeof value === "object" && !Array.isArray(value);
}
```

`buildFieldConditions` then reads only `gt`, `gte`, `lt`, `lte`. An object with none of them pushes zero conditions, so the key disappears from the query.

```ts
// intended: slug IN (...)
await getEmDashCollection("articles", {
  status: "published",
  limit: 120,
  where: { slug: { in: slugs } },
});
```

**Expected:** the filter applied, or a warning that the shape is not supported.
**Actual:** no `slug` condition at all. The call returns the 120 most recently created published articles. The result is well formed, so the caller has no signal.

The supported spelling is `where: { slug: slugs }`. The two neighbouring mistakes in the same function (an invalid field name, a range on a taxonomy key) both `console.warn`; this one does not.

**Suggested fix:** in the range branch, if none of `gt/gte/lt/lte` is present, `console.warn` and skip, matching the invalid-field-name case. Tightening `isWhereRange` to require at least one of the four keys would do the same.

### 2. An array value binds one parameter per element, with no chunking

```ts
} else if (Array.isArray(value)) {
  if (value.length > 0) conditions.push(sql`${ref} IN (${sql.join(value.map((v) => sql`${v}`))})`);
}
```

D1 allows 100 bound parameters per statement. With 120 slugs:

```ts
const r = await getEmDashCollection("articles", { status: "published", limit: 120, where: { slug: slugs } });
// r.entries.length === 0, no error surfaced to the caller
```

With the same slugs in batches of 40, the expected entries come back.

**Expected:** an error, or the list chunked.
**Actual:** zero entries and no error reached the caller, which reads as "nothing matched".

**Suggested fix:** throw (or warn and chunk) when an array filter would exceed the adapter's parameter limit. #219 fixed the same limit in `hydrateEntryBylines`.

### Related

- #219, the same D1 variable limit on another path.
- #3186, an explicit id list still gets `ORDER BY` and walks the ordering index: same function family, different defect.

eisenbruch:
## What does this PR do?

Two silent failures in `buildFieldConditions` (`packages/core/src/loader.ts`), both measured on a Cloudflare D1 site and described in #3256:

- An object `where` value with none of `gt`, `gte`, `lt`, `lte` (for example `{ slug: { in: [...] } }`) is classified as a range, contributes no condition, and the query returns unfiltered rows with nothing logged. It now warns and is skipped, the same as the invalid-field-name case beside it. The supported spelling stays the bare array.
- An array value binds one parameter per element. D1 allows 100 per statement, so `where: { slug: <120 slugs> }` returned `entries: []` with an error object and nothing thrown to the caller. Arrays are now split into `IN` clauses of `SQL_BATCH_SIZE` (50) joined with `OR`, the same approach `hydrateEntryBylines` took for #219.

Tests: two new warning tests fail on `main` and pass here; a 130-value `IN` test passes on both (the limit is D1's, not SQLite's) and guards the chunking.

Supersedes the bot's draft #3263, which was opened on a reply of mine that was not an acceptance; its body says the reporter confirmed the preview build, which did not happen (the preview needs `main`'s `@emdash-cms/plugin-types`).

Closes #3256.

## Type of change

- [x] Bug fix

## Checklist

- [x] `pnpm lint` passes (oxlint on the two files)
- [x] `pnpm test` passes (`loader-field-filters.test.ts`, 17 of 17)
- [x] `pnpm format` has been run (prettier clean)

- [x] This PR includes AI-generated code — model/tool: Claude Fable 5.1 (Claude Code)

## Screenshots / test output

```
 ✓ tests/unit/loader-field-filters.test.ts (17 tests)
 Test Files  1 passed (1)
      Tests  17 passed (17)
```
Without the loader change: 2 failed | 15 passed (the two warning tests).


eisenbruch:
You are right, and the array chunking in this PR does not do what the changeset says. I have checked it rather than argue it.

`chunks()` exists for exactly this and its own docstring says so: "keep SQL `IN (?, ?, …)` clauses within Cloudflare D1's bound-parameter limit (~100 per statement)". Per statement. This PR imports that helper and then rejoins the chunks into a single statement with `OR`, so a 130-element array still binds 130 parameters in one statement and the reported `entries: []` failure would still happen. Splitting the clause changes the SQL text and nothing about the parameter count.

It is also out of step with every existing call site. `relation.ts:565`, and the same shape in `byline.ts` and `content.ts`, iterate the chunks as separate awaited queries and merge in JS. That is the pattern that actually works.

One thing to add against my own PR, which your review was generous about: the new test cannot fail for the reason it claims to guard. It runs against the unit-test SQLite, where the variable limit is 32766, so a 130-parameter statement is fine there and the test would pass with or without the chunking. It is not a regression test for the D1 bug, and the changeset's claim that long arrays are handled is unverified. That is my error and the misleading part is mine, not the reviewer's reading of it.

The empty-range-object warning half stands on its own and is unaffected.

Converting this back to draft rather than leaving a wrong claim sitting in the queue. Doing the array half properly means the loader running one query per chunk and merging, which is a change above `buildFieldConditions` rather than inside it, and it needs to be measured against a real D1 before I say it is fixed. I will either split the warning half into its own PR or come back with the chunked-query version.

eisenbruch:
Reviewing my own PR after the rewrite. The mechanism is right now: one statement per chunk, `buildFieldConditions` back to a single `IN`, and a test that asserts the statement count rather than a row count the unit harness cannot fail. Two defects in the merge, one confirmed.

**1. NULL ordering diverges from the single-statement path on `asc`.** The comparator checks null before applying the direction:

```js
if (x == null) return 1;
if (y == null) return -1;
return (x < y ? -1 : 1) * (d === "asc" ? 1 : -1);
```

so nulls sort last whichever way the direction points. SQLite treats NULL as lower than any value, so `ORDER BY v ASC` puts them first. Checked, not assumed:

```
sqlite> select v from t order by v asc;   -- NULL, a, b
sqlite> select v from t order by v desc;  -- b, a, NULL
```

Failure case: `where: { slug: [130 slugs] }, orderBy: { published_at: "asc" }` with any entry whose `published_at` is null. Under 51 slugs that entry comes first, over 50 it comes last, and the page contents change with the length of the filter. The null branch needs to respect `d`.

**2. The chunk statements drop `LIMIT`, which is unbounded when the array column is not unique.** Fine for `id` or `slug`, where the result is bounded by the list. But the test's own column, `series`, is not unique, and `where: { series: [130 values] }` where each value matches many rows now reads every matching row of all 130 into memory to sort in JS, where the single statement stopped at `limit + 1`. Worth either bounding each chunk with `fetchLimit` (valid because each chunk is ordered on the same keys, so no chunk can contribute a row past its own first `fetchLimit`) or saying in the comment that the shape is assumed near-unique.

Two smaller things. The changeset still describes the old approach, "split into several `IN` clauses", where it is now several statements, and it says an array over 50 is split when only the first long array in a `where` is. And the `id` tiebreak takes its direction from `sortKeys[0]`, which will not match the appended `id` direction when `orderBy` mixes directions across keys.

Leaving this in draft until 1 and 2 are addressed.

eisenbruch:
Both defects are fixed correctly, and my fourth point was wrong. Checked in the source rather than taken on trust.

NULLs: `if (x == null) return -sign` puts a null first when ascending and last when descending, which is what SQLite does. Right fix.

The chunk bound, `LIMIT fetchLimit + offset`, is correct and worth stating why, since it looks like it could drop rows. The merge needs global rows `[offset, offset + fetchLimit)`. Every chunk is ordered on the same keys, so any row that lands in that window globally must be within its own chunk's first `offset + fetchLimit` rows: a row at position N in its chunk has at least N rows ahead of it globally. So nothing reachable is cut. And with `fetchLimit` null the bound is empty, which is right, because the caller asked for everything.

On the `id` tiebreak you are right and I was wrong. `getPrimarySort` at loader.ts:665 returns the **first** pattern-valid key, so `sortKeys[0]` matches the direction `buildOrderByClause` appends exactly. I had assumed it derived from something else without reading it. Ignore that part of my earlier comment.

Also checked the thing neither of us raised: `buildOrderByClause` sorts by every valid key in object key order and then `id`, and the merge comparator walks the same keys in the same order, so multi-key `orderBy` agrees between the two paths.

One residual limit, not worth changing, just worth knowing: the merge compares strings with JS `<`, which is UTF-16 code-unit order, while SQLite's default `BINARY` collation is UTF-8 byte order. Those agree across the BMP and diverge only for astral-plane characters, and only for an array filter over 50 ordered by a text column containing them.

Reads correct to me now. Leaving it in draft until a slot frees rather than swapping one of the six out again.

packages/core/src/loader.ts
```diff
@@ -1358,9 +1372,22 @@ export function emdashLoader(): LiveLoader<EntryData, EntryFilter, CollectionFil
...
 					const localeFilter = locale ? sql`AND locale = ${locale}` : sql``;
 					const cursorCond = cursorCondition ? sql`AND ${cursorCondition}` : sql``;
-					const fieldConds = buildFieldConditions(fieldFilters);
-					const fieldCondsSQL =
-						fieldConds.length > 0 ? sql`${sql.join(fieldConds, sql` AND `)}` : null;
+					// A long array filter binds one parameter per value and D1 allows 100 per statement, so a
+					// list longer than `SQL_BATCH_SIZE` is run as one statement per chunk of that field, with
+					// the same order and no LIMIT/OFFSET, and merged below. Only the first long array is split;
+					// two long arrays in one `where` would multiply statements and are left to bind as they are.
+					const longEntry = Object.entries(fieldFilters).find(
```
ascorbic: This splits only the first long array. Every generated statement still includes every other array unchanged, so two long field filters can still exceed D1's per-statement parameter limit. For example, chunks of 50 combined with a second 60-value array already bind more than 100 values before status, locale, cursor, taxonomy, limit, and offset parameters. The public `where` shape permits multiple fields, so the fix needs to bound every long list rather than documenting that only one is handled.

packages/core/src/loader.ts
```diff
@@ -1427,8 +1462,52 @@ export function emdashLoader(): LiveLoader<EntryData, EntryFilter, CollectionFil
...
-						${limitOffsetClause}
+						${paged ? limitOffsetClause : chunkLimit}
 					`.execute(db);
+					};
+					if (fieldFilterSets.length === 1) {
+						result = await runOne(fieldFilterSets[0], true);
+					} else {
+						// Each chunk is ordered on its own; the merged set is re-sorted on the same keys, then
+						// paged as the single statement would have been.
+						const parts = await Promise.all(fieldFilterSets.map((f) => runOne(f, false)));
```
ascorbic: Can you keep the merge, ordering, and pagination in the database. This path adds `ceil(values / 50)` logged-out round trips and reimplements SQL ordering in JavaScript: the comparator deliberately follows SQLite NULL placement, which is opposite PostgreSQL's default placement, and JavaScript string comparison can differ from database collation. That means crossing 50 values can change page contents by dialect. The existing `jsonTextValues(db, values)` helper expands a JSON array with one bind parameter on both SQLite/D1 and PostgreSQL; using it in the `IN` subquery would preserve the single-query ordering and pagination contract and naturally support multiple long arrays.

packages/core/tests/unit/loader-field-filters.test.ts
```diff
@@ -149,6 +150,108 @@ describe("Loader field filters", () => {
...
+			const nullsAt = merged
+				.map((id, i) => [id, i] as const)
+				.filter(([id]) => /post-(0|4|8|12|16|20|24|28|32|36|40|44|48|52|56)$/.test(id))
+				.map(([, i]) => i);
+			if (dir === "asc") expect(Math.max(...nullsAt)).toBeLessThan(15);
+			else expect(Math.min(...nullsAt)).toBeGreaterThanOrEqual(merged.length - 15);
+		}
+	});
+
+	it("runs a long IN list as several statements and merges them in order, with limit and offset", async () => {
```
ascorbic: This test doesn't do what it needs to do. The unit SQLite build accepts thousands of parameters, so the test proves only the chosen multi-statement mechanism, not that a 120/130-value collection query works on D1. Add a workerd-backed D1 loader regression with more than 100 values, assert the returned entries, and pin the total compiled parameter count. Please also check the indexed `slug` plan requested by the reporter so the correctness fix does not introduce a full scan.

ascorbic (changes requested):
Thanks. This is a good fix, and the rewrite is the right approahc. 

The large-array path still needs another revision before it is safe to merge, though: it handles only the first long array, moves database ordering and pagination into a dialect-specific JavaScript merge, adds multiple logged-out round trips, and has no regression at the D1 boundary that originally failed. The existing portable jsonTextValues(db, values) helper appears to avoid all of that by representing each long list with one bind parameter on D1/SQLite and PostgreSQL while leaving ordering, cursor handling, limit, and offset in the database. Please use that single-query approach, add real-D1 coverage, and update the PR body and changeset to describe the final behaviour.

Issue #3259, opened 2026-09-21
Taxonomy pivot query drives from the first `where` key rather than the most selective, and is skipped entirely when a field filter is present


eisenbruch:
**Version:** 0.38.0, D1. `loader.ts` is unchanged on `main` for both points. This builds on #1834 / #1962 (`buildTaxonomyPivotQuery`), which fixed the single-taxonomy case; these are the two ways a real query still ends up reading thousands of rows.

### 1. With two taxonomies, the driving side is whichever key was written first

```ts
const firstGroups = groupSets[0] ?? [];
const restGroups = groupSets.slice(1);   // each becomes AND EXISTS (... ct2 ...)
```

`groupSets` follows the key order of the caller's `where` object. The first taxonomy is walked through the pivot and every other one is a correlated `EXISTS` per candidate row, so the rows read depend on which key the caller happened to type first.

Measured with D1's `meta.rows_read`, same data, same result set:

| Query | `where` order | Rows read |
|---|---|---|
| listings in one place, of given business types | `{ business_type, places }` | 3,131 |
| same | `{ places, business_type }` | 539 |
| articles for a topic in a place (about 13 results) | `{ topics, places }` | 7,896 to 15,906 |

In that dataset 1,890 articles carry 3,312 topic tags and only 278 place tags, so driving from `places` starts from a few dozen pivot rows and driving from `topics` starts from hundreds to thousands.

**Expected:** the plan does not depend on object key order.
**Suggested fix:** order `groupSets` by estimated selectivity before building the query. `resolveTermGroups` has already run for each filter at that point; a `COUNT(*)` per group set on `content_taxonomies (collection, taxonomy_id)` is an index-only read, or the denormalized term counts from #2210 could be used. Even "fewest resolved term groups first" would be a better default than insertion order.

### 2. Any field filter abandons the pivot path (from reading the source; not separately measured)

```ts
if (taxonomyFilters.length > 0 && Object.keys(fieldFilters).length === 0) {
  // buildTaxonomyPivotQuery
} else {
  // scan ec_<collection>, one AND EXISTS (...) per taxonomy filter, correlated per row
}
```

`where: { topics: [...], type: "guide" }` therefore takes the pre-#1962 shape and walks the collection table. Field conditions could be applied to `r` inside the pivot query the same way `status`, `locale` and `deleted_at` already are.

### Related

- #1834 / #1962, the fix these two cases fall outside.
[...]

PR #3266, opened 2026-09-22
fix(core): drive the taxonomy pivot query from the most selective taxonomy, not the first where key

eisenbruch:
## What does this PR do?

`buildTaxonomyPivotQuery` walks `groupSets[0]` through `content_taxonomies` and checks every other set with a correlated `EXISTS` per candidate row. `groupSets` followed the caller's `where` key order, so the rows read depended on how the object was written: the same listings query read 3,131 rows as `{ business_type, places }` and 539 as `{ places, business_type }` (D1 `meta.rows_read`, #3259).

With two or more taxonomy filters, the resolved term groups are now counted in `content_taxonomies (collection, taxonomy_id)` (an index-only read per set) and sorted ascending, so the most selective taxonomy drives. One taxonomy is untouched. Results are identical; only the plan changes.

The second half of #3259 (any field filter abandons the pivot path) is the documented limitation in the comment above this branch and is not changed here.

Tests: a new key-order test (many posts in the category, one with the tag) asserts both orders return the same ids; the existing two-taxonomy tests still pass.

Refs #3259.

## Type of change

- [x] Bug fix

## Checklist

- [x] `pnpm lint` passes
- [x] `pnpm test` passes (`loader-taxonomy-filter.test.ts` and `loader-field-filters.test.ts`, 23 of 23)

- [x] This PR includes AI-generated code — model/tool: Claude Fable 5.1 (Claude Code)

## Screenshots / test output

```
 Test Files  2 passed (2)
      Tests  23 passed (23)
```


eisenbruch:
Agreed on "one query beats two", and done in `050cbb9d`: the probe is now a single `SELECT taxonomy_id, COUNT(*) ... WHERE taxonomy_id IN (every set's ids) GROUP BY taxonomy_id`, summed per set in memory, so a query with N taxonomy filters costs one extra round trip rather than N. The inline comment is cut to what the code needs (the smallest set drives; one grouped count covers every set); the numbers stay in the changeset and the issue.

On the test: it asserts result equality across key orders, which is the contract, and I have not added a rows-read assertion because the unit harness has no statement or row counter to assert against (the figures in #3259 are D1's own `rows_read` from production). If there is a preferred way to count statements in these tests, point me at it and I will add one. Typecheck, type-aware lint and format are clean; the taxonomy-filter suite passes.

eisenbruch:
Rebased onto `main` after #3300 (`d7afc6cc`), no conflict. #3300 changed the pivot join this PR's ordering feeds, so I re-measured whether the ordering still matters under both joins. It does.

The statement `buildTaxonomyPivotQuery` emits, on D1 with bound parameters, `meta.rows_read`. Each figure was run twice with the same result, and both orders returned the same ids:

| query | order | `CROSS JOIN` | `JOIN` (#3300) |
|---|---|---:|---:|
| listings, business type + place (CA) | business type first | 4,692 | 13,218 |
| same | place first (what this PR picks) | 2,540 | 11,784 |
| listings, business type + place (TX) | business type first | 4,337 | 13,051 |
| same | place first | 2,694 | 11,957 |
| articles, topic + place, `published_at` | topic first | 1,054 | 1,054 |
| same | place first | 470 | 470 |

The `JOIN` column for listings is a separate regression from #3300 in the temp-sort branch, filed as #3408 with a fix in #3409. With both merged, listings run as the `CROSS JOIN` column and articles as the `JOIN` column, and this PR's ordering still takes each about half.

PR #3275, opened 2026-09-22
fix(core): match redirect sources containing non-ASCII characters

### LINKED ISSUE #3239: Redirects with non-ASCII source paths never match (percent-encoding mismatch)
### Description

Redirects whose source path contains non-ASCII characters (diacritics, umlauts, any character a browser percent-encodes) never match, so the redirect does not execute and the visitor gets a 404.

The exact-match lookup in the redirect middleware compares the request's `context.url.pathname` — which the WHATWG URL parser always returns percent-encoded (`/stitek/domácí-zvířata` becomes `/stitek/dom%C3%A1c%C3%AD-zv%C3%AD%C5%99ata`) — against the redirect source exactly as stored. The admin stores whatever the author typed, typically the raw Unicode path. Neither side is normalized:

- `packages/core/src/redirects/cache.ts:94` keys the exact-match map with the stored `source` verbatim.
- `packages/core/src/astro/middleware/redirect.ts:70` looks up the percent-encoded `pathname` in that map.
- No `decodeURI`/`encodeURI`/normalization exists anywhere in the write path (`api/schemas/redirects.ts`) or the match path.

Pattern redirects have the same mismatch: the compiled regex runs against the encoded pathname.

Expected: a redirect created for `/stitek/domácí-zvířata` fires when a visitor opens that URL, regardless of whether the stored source is raw Unicode or percent-encoded.

Actual: the lookup misses, the request falls through to a 404 (and is logged as a 404 miss under the encoded path).

This is the second, still-unfixed half of #3147: the `ambiguous column "hits"` error reported there breaks 404 logging on PostgreSQL (fix: #3238), but it cannot break redirect execution — `log404` runs after `next()` and is fire-and-forget with `.catch(() => {})`. The reporter's non-executing redirect `/stitek/domácí-zvířata` → `/stitek/domaci-zvirata` is explained by this encoding mismatch instead.

A workaround exists (enter the source already percent-encoded in the admin), but nothing tells the author that, and a copied browser-URL pastes as decoded Unicode in most browsers.

Likely fix direction — needs a maintainer call on where to normalize: canonicalize sources to their percent-encoded form on write (plus a data migration for existing rows), or normalize both sides at match time (encode the stored source when building the cache, so existing data keeps working without migration). The second option is non-breaking and confined to `cache.ts`.

### Steps to reproduce

1. Run EmDash (any database).
2. In the admin, create a redirect with source `/stitek/domácí-zvířata` and destination `/stitek/domaci-zvirata`.
3. Open `/stitek/domácí-zvířata` in a browser (the browser sends it percent-encoded).
4. Observe a 404 instead of the redirect; the redirect's hit counter stays at 0 and the path shows up in the 404 log (encoded) once #3238 lands.

### Environment

- emdash version: 0.38.0 (verified unchanged on current `main`)
- Node.js version: 22
- Runtime: Node
- OS: any

### Screenshots

Not applicable.

### Logs / error output

```
$ node -e 'console.log(new URL("http://x/stitek/domácí-zvířata").pathname)'
/stitek/dom%C3%A1c%C3%AD-zv%C3%AD%C5%99ata
```

linropx:
## What does this PR do?

Redirect rules whose source path contains non-ASCII characters never matched. The admin stores a source exactly as the author typed it — normally raw Unicode — while the middleware compares it against `context.url.pathname`, which the URL parser always returns percent-encoded. A rule for `/stitek/domácí-zvířata` therefore missed the lookup and the visitor got a 404.

This normalizes stored sources to their percent-encoded form when the redirect rule cache is built (`packages/core/src/redirects/cache.ts`), so exact-match and pattern rules both compare like for like.

Normalization is idempotent: a source that is already percent-encoded (the previously undocumented workaround) is preserved rather than double-encoded, Astro pattern syntax (`[param]`, `[...rest]`) and trailing slashes are unaffected, and existing rules keep working without a data migration.

Closes #3239

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [ ] `pnpm typecheck` passes
- [ ] `pnpm lint` passes

**Why four boxes are unchecked:** my local `pnpm install` stopped on `better-sqlite3`, which needs a native build toolchain this Windows machine does not have (the prebuilt binary download failed too), so `oxlint`, `oxfmt` and `prettier` are not installed locally and `pnpm typecheck` needs a prior `pnpm build`. I ran the affected test file directly with vitest instead. No admin UI strings are touched by this change.

- [x] This PR includes AI-generated code — model/tool: WorkBuddy agent (DeepSeek V4.1 Flash)

## Screenshots / test output

Targeted test run (`packages/core`, vitest, in-memory SQLite):

```
 ✓ tests/unit/astro/middleware-redirect.test.ts (20 tests) 1300ms
     Tests  20 passed (20)
```

The new tests were verified to fail before the fix — this is the regression they lock in:

```
 × fires for a redirect whose stored source is raw Unicode
 × matches a pattern whose literal segment contains non-ASCII characters
```

`still fires when the stored source is already percent-encoded` passes both before and after the fix; it guards against double-encoding an already-encoded source.


linropx:

linropx:
recheck

ascorbic (approved):
Thanks!

Issue #3249, opened 2026-09-21
Flaky E2E: content-actions.spec.ts "keeps publishing choices usable in a 320-pixel viewport" (Cloudflare target)


MA2153:
## Description

The E2E test `e2e/tests/content-actions.spec.ts:233` — **"Schedule content › keeps publishing choices usable in a 320-pixel viewport"** — flakes on the Cloudflare target. It measures the "Publishing summary" group against the "Schedule" button and asserts they share an x coordinate, but intermittently reads them mid-relayout:

```
Error: expect(received).toBeLessThanOrEqual(expected)

Expected: <= 1
Received:    9.759288787841797
```

Both the initial attempt **and** retry #1 failed in the same run, with a *different* offset each time (9.76px, then 5.92px). The shrinking, sub-pixel, non-repeating value is the signature of a measurement taken while the layout is still settling rather than of a real misalignment.

The failure screenshot, captured moments after the assertion, shows the panel in its **correct** state — the summary card and the Schedule button flush at the same x and the same width. The layout is fine; the measurement is early.

### Likely cause

The test switches the viewport and then measures without waiting for the breakpoint-driven re-render to commit:

```ts
await page.setViewportSize({ width: 320, height: 576 });
// ...
await page.getByRole("button", { name: "Settings" }).click();
const summaryBox = await publishingSummary.boundingBox();
const scheduleBox = await scheduleButton.boundingBox();
expect(Math.abs(summaryBox!.x - scheduleBox!.x)).toBeLessThanOrEqual(1);
```

`ContentEditor` tracks the `lg` breakpoint in React state fed by a `matchMedia` listener (`packages/admin/src/components/ContentEditor.tsx:445-453`):

```ts
const [isBelowLg, setIsBelowLg] = React.useState(
    () => typeof window !== "undefined" && window.matchMedia("(max-width: 1023px)").matches,
);
React.useEffect(() => {
    const mq = window.matchMedia("(max-width: 1023px)");
    const onChange = () => setIsBelowLg(mq.matches);
    mq.addEventListener("change", onChange);
    return () => mq.removeEventListener("change", onChange);
}, []);
```

[...]

PR #3279, opened 2026-09-22
fix(core): give the toolbar's admin link one window per entry

### LINKED ISSUE #3278: Visual editing toolbar: "Open in admin" uses one named window for the whole site, so a second entry steals the first one's admin tab
**Version 0.38.0**, confirmed still present on `main`.

The toolbar's admin link targets a single named window shared by every entry on the site, so opening the admin for a second entry navigates the tab the first one is already in.

### Where

Two places, both using the same constant name:

- `packages/core/src/visual-editing/toolbar.ts:64` — `<a class="emdash-tb-admin" id="emdash-tb-admin" href="#" target="emdash-admin" ...>`, with `updateStatus` setting only `adminLink.href` at :732 and leaving `target` as rendered.
- `packages/core/src/visual-editing/toolbar.ts:957` — `openAdmin(annotation)` calls `window.open(url, "emdash-admin")`.

Because the name is static, the browser reuses one browsing context for all of them.

### Reproduction

1. Sign in as an editor and enable the visual editing toolbar.
2. Open two different articles in two tabs.
3. Click "Open in admin" on the first. An admin tab opens.
4. Click "Open in admin" on the second.

Expected: the second entry opens its own admin tab, or at least does not disrupt the first.
Actual: the second entry replaces the first in that same tab. Any unsaved state in the first entry's admin view is gone, and the editor is silently moved to a different record.

It is worst in the workflow the toolbar exists for: going through several entries on a page and opening each one to fix something.

### Suggested fix

Make the window name per entry rather than global, which keeps the intent of a named target. A repeat click on the *same* entry still reuses that entry's tab instead of piling up duplicates, and two entries no longer fight over one window:

```js
adminLink.target = "emdash-admin-" + (ref.collection + "-" + ref.id).replace(/[^A-Za-z0-9_-]/g, "_");
```

set in `updateStatus` alongside the existing `href`, and the matching change at the `window.open` call. The sanitising replace matters because a window name is not allowed to contain whitespace and ids and collection names are not constrained to a safe alphabet here.

`_blank` is the other option and would also fix the collision, but it opens a fresh tab on every click and loses the de-duplication the named target was presumably chosen for. Per entry seems the better behaviour.

Happy to open a PR for this.

eisenbruch:
## What does this PR do?

The visual editing toolbar names the admin window with a single site-wide constant, so every entry shares one browsing context. Clicking "Open in admin" on a second entry navigates the tab the first one is already in, discarding whatever state was there and silently moving the editor to a different record.

Two call sites, both using the same name:

- `packages/core/src/visual-editing/toolbar.ts:64` renders `target="emdash-admin"` on the anchor, and `updateStatus` sets only `href`, leaving that target in place.
- `openAdmin()` calls `window.open(url, "emdash-admin")`.

This names the window after the entry instead, through one helper used by both:

```js
function adminWindowName(collection, id) {
  return "emdash-admin-" + (collection + "-" + id).replace(/[^A-Za-z0-9_-]/g, "_");
}
```

Keeping a name rather than switching to `_blank` preserves what the named target was presumably for: clicking the same entry twice reuses that entry's window instead of piling up duplicate tabs. The sanitising replace is load-bearing rather than defensive, because a window name may not contain whitespace and neither collection names nor ids are constrained to a safe alphabet at this point.

Two things a reviewer may want to check: `adminWindowName` is a function declaration at the same scope as `updateStatus`, so hoisting covers the call that appears earlier in the emitted script; and the changed code lives inside the client-script template literal, so typecheck and type-aware lint do not analyse it.

Closes #3278

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

**Correcting myself on tests.** An earlier version of this description said I could not find a harness over the emitted toolbar script. That was wrong: `packages/core/tests/unit/visual-editing/toolbar.test.ts` already exists and already extracts the script and runs slices of it under `node:vm`. Four cases now follow that pattern, in `620363ba`:

- two entries get two different window names, which is the bug
- one entry gets the same name every time, so a repeat click still reuses that entry's window
- characters a browsing context name cannot carry are replaced
- both the link target and the `openAdmin` fallback are wired to the helper, with explicit assertions that neither falls back to the old `"emdash-admin"` constant

All four fail against `main` with "adminWindowName was not rendered" and pass with the fix; the file goes 19 tests to 23.

**Screenshots** is the remaining unchecked box: the change alters which window a link targets and produces no visual difference. No user-visible strings are added or changed.

## Screenshots / test output

Not applicable for screenshots: no rendered output changes, only the target window of an existing link.

On testing, stated precisely, because an earlier version of this description wrongly said the behaviour was "verified by hand":

- **The bug is a real user report on 0.38.0, not a browser test.** Our site owner hit it in ordinary use as an admin on production: two article tabs open, "Open in admin" on one and then the other, and the second replaced the first in the same tab.
- **The code reasoning is verified.** Both call sites were confirmed in the installed 0.38.0 source and in the built `dist` output, and the patched output was confirmed after applying the change.
- **Nobody here has clicked through the fix in a browser.** The original behaviour was not reproduced by me and the fix has not been confirmed visually, so please treat it as unverified at the UI level.

CI on this branch: Typecheck, Format, Lint, Integration Tests and the unit shards pass.

`E2E tests (2/8)` failed, and it is [#3249](https://github.com/emdash-cms/emdash/issues/3249), the tracked flake, not this change. Read from the job log rather than assumed: the single failure is `content-actions.spec.ts:233` "Schedule content > keeps publishing choices usable in a 320-pixel viewport", the same pixel-alignment assertion named in that issue, failing `toBeLessThanOrEqual(1)` with 9.76 and then 7.78 on retry, with 42 other tests passing in the shard. That test exercises the content schedule panel's responsive layout; this PR only changes the window name on the toolbar's admin link in `visual-editing/toolbar.ts`, so there is no shared path.


eisenbruch:
Editing my own PR body to correct a claim I should not have made: the Testing section originally said the behaviour was "verified by hand on 0.38.0". Nobody here has clicked through it. The bug is a real user report from our site owner in ordinary admin use on production, and the code reasoning is verified at both call sites in source and in the built output, but the fix itself is unverified at the UI level. The section now says that.

Two points that may come up in review:

- `adminWindowName` is a function declaration at the same scope as `updateStatus`, so hoisting covers the call that appears earlier in the emitted script.
- The sanitising replace is load-bearing rather than defensive: a collection name or id containing whitespace would otherwise produce an invalid window name.

eisenbruch:
CI note on `620363ba`, so the red marks are not mistaken for something in this change.

`E2E shard (8/8)` failed at harness setup, not on an assertion. The test server never came up:

```
Error: http://localhost:4444/_emdash/api/setup/status did not return ok within 120000ms (last 0):
  fetch failed
[cause]: Error: http://localhost:4444/_emdash/api/setup/status did not return ok within 120000ms (last 0):
```

15 tests passed in that shard before the timeout, and no test reported a failure. `E2E Tests` and `Tests` are the aggregate gates reporting the same shard, so all three red marks are one event.

I cannot re-run it: `gh run rerun --failed` returns "Must have admin rights to Repository". I have not pushed an empty commit to force a fresh matrix either, since this is a draft and that would spend the full E2E matrix on a harness timeout. Marking it ready for review will re-trigger CI when a slot frees, which is the natural point to get a clean run.

The commit itself is test-only: four cases added to the existing `packages/core/tests/unit/visual-editing/toolbar.test.ts`, which passes locally at 23 tests and passed `Test Shard` and `Typecheck` in this same run. Nothing in it touches the E2E harness or anything shard 8 exercises.

eisenbruch:
Both points taken, in `a7770575`:

1. The test that quoted the two call sites' source text is gone. What remains is behavioural: two entries get two names, one entry gets the same name every time, unsafe characters are replaced, and neither call site falls back to a shared `"emdash-admin"` (the negative check you said was fine). 23 passing in `toolbar.test.ts`.
2. The comment above `adminWindowName` now says only what shapes the function: a window name may not contain whitespace, and ids are not constrained to a safe alphabet. The bug narrative lives in the PR description and the changeset.

PR #3292, opened 2026-09-22
fix(admin): localize dates and keep them whole in RTL layouts

itaides:
## What does this PR do?

Fixes three related bugs in how the admin prints dates and times, all of them visible in any non-English admin and worst in a right-to-left one (`ar`, `fa`, and `he` in the PR that follows this one).

**1. Relative times were hardcoded English.** `formatRelativeTime()` in `packages/admin/src/lib/utils.ts` built its phrase in English (`3 mins ago`) and was not wrapped for translation, so no catalog could reach it — the dashboard and the revision list read English in every locale. In an RTL layout the English phrase is then reordered by the bidirectional algorithm and the number appears at the wrong end, reading as `mins ago 3`.

It now writes the phrase with `Intl.RelativeTimeFormat` in the active locale (`לפני 3 דקות`, `vor 3 Minuten`), which needs no catalog entry and gets each language's plural rules for free. This is exactly what the admin already did one directory away, in `settings/PasskeyItem.tsx` — that near-identical private copy is deleted and the shared helper used instead.

**2. Dates followed the browser's language, not the admin's.** Fifteen call sites formatted with `toLocaleDateString()` / `toLocaleString()` and no locale argument, so an admin switched to Hebrew still printed English months when the browser was English. Every one of them now passes the admin's locale through a shared `formatDate(value, locale, options?)`, which also parses SQLite's `YYYY-MM-DD HH:MM:SS` as UTC the way `parseTimestamp` does.

**3. Formatted dates were not bidi-isolated.** A formatted date is digits and punctuation, whose direction the bidirectional algorithm takes from the text around it, so `22 בספט׳ 2026, 22:48` rendered in the publish panel with its comma against the wrong number. Dates are now wrapped in `<bdi>` where they are rendered on their own, and in Unicode isolates (a small `isolate()` helper, U+2068/U+2069) where they are interpolated into a translated sentence and there is no element to wrap.

No `msgid` changes: the interpolations were already numbered placeholders, so no catalog churn.

## Type of change

- [x] Bug fix
- [ ] Translation
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

Two existing assertions change with the behaviour, both in `tests/components/RevisionHistory.test.tsx`: `Intl` writes `5 minutes ago` rather than `5 mins ago`, and `yesterday` rather than `1 day ago` (`numeric: "auto"`, as in the passkey list).

- [x] This PR includes AI-generated code — model/tool: Claude Opus 5 (Claude Code). Every hunk was reviewed by the author before committing, and the bugs were found by using the admin in Hebrew rather than by reading the code.

## Screenshots / test output

The bug as it appears in an RTL admin, from the Hebrew catalog in the PR that follows: the dashboard's "Recent activity" column reads `mins ago 3` — English, and reordered — and the publish panel prints `22 בספט׳ ,2026 22:48`, with the comma bound to the wrong number. After this change the same two places read `לפני 3 דקות` and `22 בספט׳ 2026, 22:48`.

```
$ pnpm --filter @emdash-cms/admin test
 Test Files  181 passed (181)
      Tests  2454 passed (2454)

$ pnpm typecheck   # clean
$ pnpm lint        # clean
$ pnpm format:check # clean
```


itaides:
Thanks — all five addressed in 20ca6b0.

- **MediaDetailPanel**: you're right that it was a regression, not just a duplicate. `item.createdAt` is a naive SQLite timestamp and `new Date()` read it as local time, so media timestamps were shifted by the browser's offset. The private helper is gone; the call sites use the shared `formatDate` with a named `MEDIA_DATE_FORMAT` for the date-with-time display they want.
- **RegistryPluginDetail**: the copy is gone too. The `try`/`catch` stays as `formatRegistryDate`, because those timestamps come from a third-party registry and a malformed one should show as-is rather than throw, but parsing now happens in one place.
- **`.js` extensions**: added on every internal import this PR touches, and `Redirects.tsx` imports `lib/utils` once.
- **UserList**: the date cell is wrapped in `<bdi>` like the others.
- **`dateStyle: "medium"` default**: good catch — that would have restyled every plain date. `formatDate` now defaults to the locale's own format, which is exactly what `toLocaleDateString()` produced, so no call site changes how it looks. The one deliberate change left is the comment detail panel (`9/22/2026 10:48:00 PM` → one medium/short pair), and the changeset now says so.

On the failing-test requirement: the new tests in `tests/lib/utils.test.ts` fail against the current code — `formatRelativeTime(ts, "he")` returns `3 minutes ago` there instead of `לפני 3 דקות` — so they reproduce the bug rather than only cover the fix.

`pnpm typecheck`, `pnpm lint`, `oxfmt` and the admin suite (2,453 tests) are green locally. Since this changes how dates render, the RTL visual-regression baselines may need a maintainer's `/accept-baselines`.

itaides:

